PR #258, opened 2014-10-02
Fix bug fetching out unset timestamp values.

muirmanders:
If your row exists but has an empty timestamp column, your time.Time
bind variable was being initialized to epoch time 0. Now it returns
early and does not modify the bind variable. I considered having it
instead set the bind variable to the go zero time.Time value, but
since that is techincally a legitimate time value as well it still
doesn't solve the semipredicate problem.

I tweaked the timestamp marshal code to do "val.UnixNano() / 1000000"
instead of "val.UnixNano() / time.Millisecond" because it doesn't make
sense to do operations directly between integers and time.Duration values.
The latter only works because the base unit for time.Duration happens to
be nanoseconds.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -19,6 +19,7 @@ import (
 	"testing"
 	"time"
 	"unicode"
+
 	"speter.net/go/exp/math/dec/inf"
 )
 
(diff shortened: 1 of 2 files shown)


serbaut:
Scanning should either yield a value or give an error imo. I could not find anything in the docs about scanning null values but silently ignoring null is not what you would expect.

I would expect this code to always set foo.time or give an error:

    for iter.Scan(&foo.id, &foo.time) {
    // foo.time gets the previous foo.time if time is null!
    }

jtbandes:
I agree with @serbaut; I recently ran into the same issue. The particularly subtle/problematic case is `for iter.Scan(&time)`, whose behavior is completely wrong when encountering null time values in some (but not all) results.

Just as scanning a null value can produce an empty string, it seems it should be able to produce an empty (zero) time, a.k.a. `time.Time{}`.

Zariel:
@jtbandes feel free to open a ticket with current and expected behaviour

PR #282, opened 2014-11-08
Use godep to vendor imports, and stay go gettable

Zariel:


0x6e6562:
@Zariel I'm just trying to get some traction on this pull request and work out whether it should stay open as a PR or whether it needs to go off to the discussion list for further consideration?

As for vendoring with godep (or any other package manager), I think it could be a hard sell to introduce the package manager as a dependency, if you get what I mean. In other languages there is a dominant package manager, so this kind of question is a moot point, but there is nothing official for go, so we might be being a little bit too opinionated by introducing this.

I'm wondering if we can achieve some kind of half way house by changing the build such that there is a core with no dependencies whatsoever and then we have optional providers for each of the concerns that introduces a 3rd party library?

PR #287, opened 2014-11-25
Improved marshaling of interface "Marshaler"

c0va23:
The Marshal did not work for pointers to interface Marshaller. I added the test cases, which had experienced this problem.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -39,4 +39,3 @@ Dean Elbaz <elbaz.dean@gmail.com>
 Mike Berman <evencode@gmail.com>
 Dmitriy Fedorenko <c0va23@gmail.com>
 Zach Marcantel <zmarcantel@gmail.com>
-James Maloney <jamessagan@gmail.com>
--- a/marshal.go
+++ b/marshal.go
@@ -54,6 +54,10 @@ func Marshal(info *TypeInfo, value interface{}) ([]byte, error) {
 		}
 	}
 
+	if v, ok := value.(Marshaler); ok {
+		return v.MarshalCQL(info)
+	}
+
 	switch info.Type {
 	case TypeVarchar, TypeAscii, TypeBlob:
 		return marshalVarchar(info, value)
@@ -156,8 +160,6 @@ func unmarshalNullable(info *TypeInfo, data []byte, value interface{}) error {
 
 func marshalVarchar(info *TypeInfo, value interface{}) ([]byte, error) {
 	switch v := value.(type) {
-	case Marshaler:
-		return v.MarshalCQL(info)
 	case string:
 		return []byte(v), nil
 	case []byte:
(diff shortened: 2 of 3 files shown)


c0va23:
I do rabase of this PR.

AUTHORS
```diff
@@ -39,4 +39,3 @@ Dean Elbaz <elbaz.dean@gmail.com>
 Mike Berman <evencode@gmail.com>
 Dmitriy Fedorenko <c0va23@gmail.com>
 Zach Marcantel <zmarcantel@gmail.com>
-James Maloney <jamessagan@gmail.com>
```
0x6e6562: Why is this author being removed from the list?

0x6e6562:
I've just kicked off the build again for [571](https://travis-ci.org/gocql/gocql/builds/47229311), to see if it is an intermittent failure.

0x6e6562:
OK, so the build passes now and it looks like this patch is in a mergable state - the only thing remaining is to clarify the change to the `AUTHORS` file.

AUTHORS
c0va23: It revert-commit.

0x6e6562: Sorry, I don't understand what you mean. @james-maloney submitted #291, so I'm not sure that his name should be removed from the `AUTHORS` list.

Also, in the context of #291, this #287 is effectively undoing the change from #291 - what is the intention behind moving the

```go
if v, ok := value.(Marshaler); ok {
    return v.MarshalCQL(info)
}
```

below the 

```go
if valueRef := reflect.ValueOf(value); valueRef.Kind() == reflect.Ptr {
    if valueRef.IsNil() {
        return nil, nil
    } else if _, ok := value.(Marshaler);  !ok {
        return Marshal(info, valueRef.Elem().Interface())
    }
}
```

block?

c0va23: Without this rollback tests fall. See: https://travis-ci.org/gocql/gocql/builds/47227393

0x6e6562: Right, so you're talking about this test failure:

```
=== RUN TestMarshal
--- FAIL: TestMarshal (0.00 seconds)
panic: value method gocql.CustomString.MarshalCQL called using nil *CustomString pointer [recovered]
```

I can see what you are trying to achieve and in your favor you have written a covering test case, but considering that this is very critical to the library, I think this change requires a second opinion (@phillipCouto).

We need to be sure what the best behavior is.

mykaul:
v2: Added a guard in `DowngradingConsistencyRetryPolicy.Attempt()` that prevents downgrading from SERIAL/LOCAL_SERIAL to a non-serial consistency level (e.g. ONE, QUORUM), which would break serial read (Paxos) guarantees. Also added a unit test verifying this behavior.

PR #294, opened 2014-12-11
keepalive - idle, interval and count

curlup:
We use tcp keepalive a lot in our env. So I though that it would be nice to have all keepalive options configurable. 
Previously one could only specify KeepAlivePeriod. In that case other two settings (idle and count) will have default values. And for my desktop ubuntu it's 2 hours and 9 tries respectfully. These timings are huge and we want to be able to set them appropriate to the needs.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -40,6 +40,12 @@ type DiscoveryConfig struct {
 	Sleep time.Duration
 }
 
+type KeepaliveConfig struct {
+	Interval 	time.Duration
+	Idle 		time.Duration
+	Count 		int
+}
+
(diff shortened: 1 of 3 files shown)


Zariel:
The idea seems like a good one but I dont agree with the idea of adding another library to gocql. Instead, how about having a post connection create hook where users can set tcp options or what ever on the raw connection?

0x6e6562:
@curlup Is this pull request still alive?

0x6e6562:
@curlup Since this pull request has not progressed in a while, I'm putting it in a dormant state - feel free to contribute further if the motivation for this patch is still alive.

PR #300, opened 2014-12-30
Add func (*Query)MapScan(map[string]interface{})

matope:
Iter.MapScan(map[string]interface{}) Method is an useful interface to imprement data bindings.
It would be nice if *Query also has this method.

Could you review this PR?

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -605,66 +605,28 @@ func TestSliceMap(t *testing.T) {
 	if returned, retErr := session.Query(`SELECT * FROM slice_map_table`).Iter().SliceMap(); retErr != nil {
 		t.Fatal("select:", retErr)
 	} else {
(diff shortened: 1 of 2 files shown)


phillipCouto:
I +1 the addition of the MapScan to the Query type. As for the testing it looks like you reduced boiler plate code for the test and used an existing function. 

To me this PR looks good.

0x6e6562:
OK, I've reviewed the outright version for `cassandra_test.go` and it is now clear that the changes to this file a good simplification and that the test coverage has indeed not been reduced.

0x6e6562:
Since I was indifferent about the change and @phillipCouto thinks it's a good idea, I think the best thing to do is to get on and merge this.

@matope Thank you for your contribution and apologies for the time taken to merge this pull request.

matope:
Thanks a lot!

PR #302, opened 2015-01-16
Fix unmarshal list and set

c0va23:
This PR fix repeated unmarshaling of list into slice for case where second list less that first list.

For Example lists

- [valueA]
- [valueA valueB]
- [valueB]

unmarshaled to

- [valueA]
- [valueB valueB] <- bug here
- [valueB]

Complete example see into TestMarshalList(https://github.com/probkiizokna/gocql/compare/unmarshal_list_fix?expand=1#diff-e0406c3c675ff397bbcb989f434b7aacR619)


0x6e6562:
If I see this correctly, then should we not land #287 first and then merge that into this patch?

c0va23:
It PR should be merged after 287.

PR #295, opened 2014-12-11
tls connetion timeout

curlup:
Proper connection timeout for SSL connections

PR #304, opened 2015-01-16
Support for Go 1.4

0x6e6562:
This patch investigates whether Go 1.4 can be supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
 go:
   - 1.2
   - 1.3
+  - 1.4
 
 before_install:
   - sudo apt-get update
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ before_install:
 install:
   - sudo apt-get install -y libjna-java python-pip
   - sudo pip install cql PyYAML six
-  - go get code.google.com/p/go.tools/cmd/vet
-  - go get code.google.com/p/go.tools/cmd/cover
+  - go get golang.org/x/tools/cmd/vet
+  - go get golang.org/x/tools/cmd/cover
   - git clone https://github.com/pcmanus/ccm.git
   - pushd ccm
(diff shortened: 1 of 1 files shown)


Zariel:
Should we also remove 1.2 as an official supported version to maintain the latest 2 supported major versions?

0x6e6562:
@Zariel Good question - this patch started as an investigation as to whether 1.4 can be supported, but ultimately I think you've got a point - we can't just increase the maintenance scope ad infinitum. Maybe we should come to some of consensus before ploughing ahead with this change. @phillipCouto ?

Zariel:
I think when we added 1.3 we removed 1.1 and agreed to support the last 2, let me see if I can find it.

Zariel:
Also given that supporting 1.2 as a minimum blocks issues such as #295

phillipCouto:
I think supporting the latest version and one previous is a good model. We should execute the sunsetting of 1.2 similar to 1.1. Create a tag signaling the latest version of 1.2 support and then discontinue it.

0x6e6562:
OK, agreed - I'll add some notes to the README about the sunset model.

0x6e6562:
@Zariel I think there is some confusion here - I don't think this pull request was quite ready to go. I had left it alone for a couple of weeks because I have been so snowed under with other tasks. I'm not sure that all of the components in the Travis matrix pass (I've just kicked a couple of failing ones off again). Also we have support for 1.2 still, that was supposed to get sunset. And this change in the maintenance model wasn't documented yet.

0x6e6562:
I've kicked the build off again - 594.5 seems to have issues getting past TestRingDiscovery though.

0x6e6562:
OK, so the build has finally passed, so the jump to 1.4 is OK. Once I get the head space, I'll remove 1.2 and document the sunset model accordingly.

Issue #292, opened 2014-12-09
Marshal string to int


purohit:
Hi,

I'm loading rows of hundreds of columns through a spreadsheet in Go. Since encoding/csv automatically parses rows as string slices, it would be nice if Marshal supported string -> int or string-> decimal columns.

I could write this into the marshaller, but would you accept such a pull request?

Thanks!

0x6e6562:
@purohit Sorry for not responding to this issue for such a long time, I completely overlooked it in my review stack.

So if I understand you correctly, as an alternative to creating a custom Marshaler, you would like to extend `marshalInt(info *TypeInfo, value interface{})` so that in your application, you can represent integer values as strings, but when you come to insert them into Cassandra, the marshalling functionality will then be able to automatically say, "hey, you've got a string, but you're trying to stick into into an integer column, so I'll convert this for you"?

If so, then how do you delineate between 32 and 64 bit integers?

purohit:
Exactly right. I'd switch on the cassandra type -- if it's a `TypeCounter` or a `TypeBigInt`, I'd use 64 bit per the Cassandra spec (http://www.datastax.com/documentation/cql/3.0/cql/cql_reference/cql_data_types_c.html) and 32-bit if it's a `TypeInt`. What do you think?

https://github.com/gocql/gocql/blob/6ba061b4e491647fcbd2839b7e65d982c95d4104/helpers.go#L27

corydoras:
Marshalling and unmarshlling string to integer and integer to string conversion would certainly make some of my code using the gocql client simpler and easier to read. 

That said, I'm not sure if I am for or against it. Code readability is great, but it also feels a bit lazy and potentially less iniffecient.

purohit:
I don't understand - what would be less efficient? I'm approaching this from the angle that someone already has data typed as a string, so they have no choice but to marshal it anyway.

In the codebase, there are a lot of pairs of [CQL data type -> Go data type] conversions. I propose to include more for completeness; can't think of a reason to exclude these in particular. 

Thoughts?

corydoras:
What I meant by "potentially less efficient" was simply that it could encourage people to not bother storing data types in the correct data types. i.e. Storing a number in a string field so all operations to and from mysql require a conversion. 

Just to be clear, I am not agreeing or disagreeing if this feature should be in there, like I said, it certainly would be handy for some of my code.

purohit:
I understand, good point. I'd argue that the burden of storing data sanely
[...]

PR #305, opened 2015-01-20
Marshal and unmarshal from [big]int <-> string

purohit:
This is re: #292


Zariel:
LGTM, could you add yourself to the AUTHORS file so we can merge?

purohit:
Done. Thanks!

Zariel:
Thanks for the contribution!

PR #306, opened 2015-01-20
Add a callback function for when connection states change.

Zariel:
Following on from #294, this will allow users to implement what ever logic they like for their connections.

### DIFF
--- a/connectionpool.go
+++ b/connectionpool.go
@@ -171,6 +171,10 @@ func (c *SimplePool) connect(addr string) error {
 		return err
 	}
 
+	if fn := c.cfg.ConnState; fn != nil {
+		fn(conn, StateNew)
+	}
+
 	return c.addConn(conn)
 }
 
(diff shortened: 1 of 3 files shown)


connectionpool.go
```diff
@@ -274,6 +278,11 @@ func (c *SimplePool) removeConnLocked(conn *Conn) {
 		delete(c.connPool, conn.addr)
 	}
 	delete(c.conns, conn)
+
+	// im not so keen on executing this with c.mu held
```
0x6e6562: I can see why you're not keen on this firing whilst the lock is being held - what are the pros and contras of this approach?

0x6e6562:
Apologies for leaving this unattended for so long - a bug in the iOS Github client I use lead me to be believe the change was not mergable, so I hadn't looked at it until now :-(

Anyway, the idea is sound, we just need to bottom out the implications of holding the lock during the execution of the callback.

Zariel:
Basically it means that the call may block the connection pool, which might
be bad. But that would be the users fault, could mention it in the docs.

0x6e6562:
OK, that's a fair point. The question then remains how do you document this in a way that, over time, this potential lock remains documented, so that even after refactoring, it is clear that any custom code is in a critical section?

In addition, I don't think any of the code needs to change, but I think the documentation needs to get tightened up, so that it is clearer what is actually going on. But I will comment on each bit individually.

Maybe there's some mileage in having a test case as a documenting artifact, or just an running example?

Zariel:
Ill update this to have an example and improve the docs, im not so happy this is tied to the connection pool right now, and in the future might make sense to move it to the parent Session

0x6e6562:
Ah, ok - maybe it would help to elaborate on the practical use cases for this change so that we can judge where this is best housed?

Zariel:
So this is mainly to allow users to track connections inside the pool, so they can track hosts going up and down or so, main use case was an issue I cant find right now. Ill close this and if it becomes a feature request ill update it.

PR #307, opened 2015-02-02
Added PageSize as a property for ClusterConfig

dankinder:
This is intended to make it easy to set a default PageSize for all CQL sessions that share the same configuration.

Looked like a pretty trivial change. Hopefully doesn't violate project philosophy or anything :smirk:


phillipCouto:
LGTM :) Thanks for the contribution.

PR #308, opened 2015-02-05
Token aware query routing

justinretailnext:
Adds a new connection pool implementation and supporting functionality to optionally route queries using a token aware policy. 

The new connection pool implementation (policyConnPool) is a factoring of SimplePool which splits the code for managing a pool of connections for a single host into an independent struct (hostConnPool) from the higher level multi-host connection pool. It also introduces two policy interfaces HostSelectionPolicy and ConnSelectionPolicy for selecting a prioritized list of hosts for a query and for selecting a connection for a host respectively.

I have added a ccm based test script connectionpool_systems_test.sh and a go test connectionpool_systems_test.go which test the behaviors of SimplePool and policyConnPool independently when nodes are stopped and started. This test demonstrates some situations where SimplePool will return a nil *Conn from Pick when nodes are stopped.

Note that I have left SimplePool in place as-is, though the NewRoundRobinConnPool function returns an equivalent connection pool.

This pull request also introduces a new dependency on github.com/spaolacci/murmur3 (BSD-like licence) for the Murmur3 Partitioner, though if desired this dependency could be replaced by simply including the Murmur3 128-bit algorithm directly in Murmur3Partitioner which is what the DataStacks Java driver does.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -42,3 +42,5 @@ Zach Marcantel <zmarcantel@gmail.com>
 James Maloney <jamessagan@gmail.com>
 Ashwin Purohit <purohit@gmail.com>
 Dan Kinder <dkinder.is.me@gmail.com>
+Justin Corpron <justin@retailnext.com>
+
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -135,7 +135,7 @@ func TestRingDiscovery(t *testing.T) {
 		time.Sleep(*flagAutoWait)
 	}
 
-	size := len(session.Pool.(*SimplePool).connPool)
+	size := session.Pool.Size()
 
 	if *clusterSize != size {
 		t.Logf("WARN: Expected a cluster size of %d, but actual size was %d", *clusterSize, size)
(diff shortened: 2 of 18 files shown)


0x6e6562:
First of all, thank you very much for submitting this patch, it looks like some quite interesting functionality that will certainly interest other members of the community.

For now, I've only given the patch a cursory look over in order to give you some initial feedback.

This is mainly because the patch is quite big and it will take some time to digest it. Because of this, the obvious question that is going to arise is whether the patch can be broken done into a series of meaningful smaller patches. This is just about being able to review changes more thoroughly and trying to avoid introducing bugs by having smaller change sets to eyeball. 

One big question springs to mind: what purpose is the parser implementation serving?

Also, is this approach taking any inspiration from other drivers?

I'm wondering whether a bullet point list of the strategy of the patch might help us review the change, seeing as it is so large? This would be just to try to give the discussion some structure.

As for the dependency on spaolacci/murmur3, I'm not seeing this as a deal breaker. I've used this particular implementation before on other projects, so I am familiar with it. That said, we'll need to get buy from @phillipCouto or @Zariel, they might see this differently.

As for the test run, I've noticed that build 587 failed on one of the runs, but it looked like the VM ran out of resources to run ccm, so I kicked it off again and that test now passes.

Following on from that, there are a couple of failures with Cassandra 1.2.x:

```
=== RUN TestKeyspaceMetadata
--- FAIL: TestKeyspaceMetadata (1.72 seconds)
	cassandra_test.go:1445: failed to query keyspace metadata with err: Undefined name type in selection clause
=== RUN TestRoutingKey
--- FAIL: TestRoutingKey (1.47 seconds)
	cassandra_test.go:1506: Expected routing key info, but was nil
```

I'm not sure what these are due to, eventually they may be touching on something that can't really be supported in 1.2.x, but we'll need to find out what the cause is and then take a judgement call from there.

0x6e6562:
I also take it that since I began to write my initial comment,  a13cde9 has been pushed into the patch - should we assume for now that this is still work in progress?

justinretailnext:
I'm trying to resolve the travis CI issues and I am rebasing the changes into the 4 commits that make up this pull request so that everything is tidy.

I agree that this is a big change which is why I've kept it split into the 4 commits as each represents a component/step that I took to make this all happen. If it makes sense I'm fine with doing a series of PRs and potentially making whatever changes make sense.

0x6e6562:
Obviously the smaller, the better. There's nothing wrong with aiming for the big picture, but breaking things down into smaller logical changes will help everybody and will get the overall patch reviewed quicker. It will also give us a chance to structure the discussion better.

justinretailnext:
Sure thing, I'll pull a branch off of the first commit and double check that I pass the integration tests and then get a PR setup.

0x6e6562:
Just to be clear, I was referring to a series of PRs - individual commits are rarely reviewed in isolation - the unit of review is the PR. So keeping them small is viewed favorably by all concerned :-)

justinretailnext:
I think we're on the same page, unless there's a PR option I'm not aware off?

0x6e6562:
I also think we are on the same page - I was writing my last comment just as you posted your previous comment, so our thoughts crossed in mid air, but I think they are aligned now.

Zariel:
Very nice, will take a look in more detail when I have the chance

PR #309, opened 2015-02-05
Adds a function to retrieve schema metadata of the session's keyspace

justinretailnext:
Adds a function to get the schema metadata from a Session for the session's current keyspace. Also includes a couple typo fixes in the TypeInfo code.

The implementation is based on the [DataStax Java Driver](https://github.com/datastax/java-driver), specifically the code found in these Java classes:
* [ControlConnection.java](https://github.com/datastax/java-driver/blob/2.1/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java)
* [Metadata.java](https://github.com/datastax/java-driver/blob/2.1/driver-core/src/main/java/com/datastax/driver/core/Metadata.java)
* [KeyspaceMetadata.java](https://github.com/datastax/java-driver/blob/2.1/driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java)
* [TableMetadata.java](https://github.com/datastax/java-driver/blob/2.1/driver-core/src/main/java/com/datastax/driver/core/TableMetadata.java)
* [ColumnMetadata.java](https://github.com/datastax/java-driver/blob/2.1/driver-core/src/main/java/com/datastax/driver/core/ColumnMetadata.java)
* [CassandraTypeParser.java](https://github.com/datastax/java-driver/blob/2.1/driver-core/src/main/java/com/datastax/driver/core/CassandraTypeParser.java)

I've tried to simplify and consolidate the ideas from the Java code into distinct steps which are:
1. Get the keyspace schema row
2. Get the table schema rows for the keyspace
3. Get the column schema rows for the keyspace
4. Compile/reason the previous information into a single result (the KeyspaceMetadata) and determine the PartitionKey and ClusteringColumns information

The [type parser code](https://github.com/retailnext/gocql/commit/4951424501dd747c1215d45709d715add95568b1#diff-86d6a6e453f6c762fa29b1272cc0faedR475) used to parse the key_validator, default_validator, and comparator column values from the table rows is basically a from-scratch implementation as the Java code in CassandraTypeParser was difficult to understand IMHO. The type parser is used to determine the TypeInfo's for each of the columns in a table.

There are a couple instances where the code for handling schema information from the V1 protocol differs from the V2+ protocol as indicated with comments. Specifically, these differences are because V1 protocol does not support the "type" field in the system.schema_columns table and that table only includes "regular" columns.

One limitation of note is that the typeParser code currently treats UDT(User), Tuple, and DynamicComposite types as Custom (specifically it treats everything it can't parse as custom); however, additional conditional branches can be added to detect these types and handle them appropriately once the Type enums and TypeInfo changes support these various types.

For testing there is a unit test for the type parser using various key validator, default validator and comparator values found in my development cassandra schema. There is also an integration test which tests the KeyspaceMetadata() function of Session, ensuring that PartionKey and ClusteringColumns information is filled out as expected.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -1427,3 +1427,280 @@ func TestEmptyTimestamp(t *testing.T) {
 		t.Errorf("time.Time bind variable should still be empty (was %s)", timeVal)
 	}
 }
+
+func TestGetKeyspaceMetadata(t *testing.T) {
+	session := createSession(t)
+	defer session.Close()
+
+	keyspaceMetadata, err := getKeyspaceMetadata(session)
+	if err != nil {
+		t.Fatalf("failed to query the keyspace metadata with err: %v", err)
+	}
+	if keyspaceMetadata == nil {
+		t.Fatal("failed to query the keyspace metadata, nil returned")
+	}
+	if keyspaceMetadata.Name != "gocql_test" {
+		t.Errorf("Expected keyspace name to be 'gocql' but was '%s'", keyspaceMetadata.Name)
+	}
+	if keyspaceMetadata.StrategyClass != "org.apache.cassandra.locator.SimpleStrategy" {
+		t.Errorf("Expected replication strategy class to be 'org.apache.cassandra.locator.SimpleStrategy' but was '%s'", keyspaceMetadata.StrategyClass)
+	}
+	if keyspaceMetadata.StrategyOptions == nil {
+		t.Error("Expected replication strategy options map but was nil")
+	}
+	rfStr, ok := keyspaceMetadata.StrategyOptions["replication_factor"]
+	if !ok {
+		t.Fatalf("Expected strategy option 'replication_factor' but was not found in %v", keyspaceMetadata.StrategyOptions)
+	}
+	rfInt, err := strconv.Atoi(rfStr.(string))
+	if err != nil {
+		t.Fatalf("Error converting string to int with err: %v", err)
+	}
+	if rfInt != *flagRF {
+		t.Errorf("Expected replication factor to be %d but was %d", *flagRF, rfInt)
+	}
+}
+
+func TestGetTableMetadata(t *testing.T) {
+	session := createSession(t)
+	defer session.Close()
+
+	if err := createTable(session, "CREATE TABLE test_table_metadata (first_id int, second_id int, third_id int, PRIMARY KEY (first_id, second_id))"); err != nil {
+		t.Fatalf("failed to create table with error '%v'", err)
+	}
+
+	tables, err := getTableMetadata(session)
+	if err != nil {
+		t.Fatalf("failed to query the table metadata with err: %v", err)
+	}
+	if tables == nil {
+		t.Fatal("failed to query the table metadata, nil returned")
+	}
+
+	var testTable *TableMetadata
+
+	// verify all tables have minimum expected data
+	for i := range tables {
+		table := &tables[i]
+
+		if table.Name == "" {
+			t.Errorf("Expected table name to be set, but it was empty: index=%d metadata=%+v", i, table)
+		}
+		if table.Keyspace != "gocql_test" {
+			t.Errorf("Expected keyspace for '%d' table metadata to be 'gocql_test' but was '%s'", table.Name, table.Keyspace)
+		}
+		if table.KeyValidator == "" {
+			t.Errorf("Expected key validator to be set for table %s", table.Name)
+		}
+		if table.Comparator == "" {
+			t.Errorf("Expected comparator to be set for table %s", table.Name)
+		}
+		if table.DefaultValidator == "" {
+			t.Errorf("Expected default validator to be set for table %s", table.Name)
+		}
+
+		// these fields are not set until the metadata is compiled
+		if table.PartitionKey != nil {
+			t.Errorf("Did not expect partition key for table %s", table.Name)
+		}
+		if table.ClusteringColumns != nil {
+			t.Errorf("Did not expect clustering columns for table %s", table.Name)
+		}
+		if table.Columns != nil {
+			t.Errorf("Did not expect columns for table %s", table.Name)
+		}
+
+		// for the next part of the test after this loop, find the metadata for the test table
+		if table.Name == "test_table_metadata" {
+			testTable = table
+		}
+	}
+
+	// verify actual values on the test tables
+	if testTable == nil {
+		t.Fatal("Expected table metadata for name 'test_table_metadata'")
+	}
+	if testTable.KeyValidator != "org.apache.cassandra.db.marshal.Int32Type" {
+		t.Errorf("Expected test_table_metadata key validator to be 'org.apache.cassandra.db.marshal.Int32Type' but was '%s'", testTable.KeyValidator)
+	}
+	if testTable.Comparator != "org.apache.cassandra.db.marshal.CompositeType(org.apache.cassandra.db.marshal.Int32Type,org.apache.cassandra.db.marshal.UTF8Type)" {
+		t.Errorf("Expected test_table_metadata key validator to be 'org.apache.cassandra.db.marshal.CompositeType(org.apache.cassandra.db.marshal.Int32Type,org.apache.cassandra.db.marshal.UTF8Type)' but was '%s'", testTable.Comparator)
+	}
+	if testTable.DefaultValidator != "org.apache.cassandra.db.marshal.BytesType" {
+		t.Errorf("Expected test_table_metadata key validator to be 'org.apache.cassandra.db.marshal.BytesType' but was '%s'", testTable.DefaultValidator)
+	}
+	expectedKeyAliases := []string{"first_id"}
+	if !reflect.DeepEqual(testTable.KeyAliases, expectedKeyAliases) {
+		t.Errorf("Expected key aliases %v but was %v", expectedKeyAliases, testTable.KeyAliases)
+	}
+	expectedColumnAliases := []string{"second_id"}
+	if !reflect.DeepEqual(testTable.ColumnAliases, expectedColumnAliases) {
+		t.Errorf("Expected key aliases %v but was %v", expectedColumnAliases, testTable.ColumnAliases)
+	}
+	if testTable.ValueAlias != "" {
+		t.Errorf("Expected value alias '' but was '%s'", testTable.ValueAlias)
+	}
+}
+
+func TestGetColumnMetadata(t *testing.T) {
+	session := createSession(t)
+	defer session.Close()
+
+	if err := createTable(session, "CREATE TABLE test_column_metadata (first_id int, second_id int, third_id int, PRIMARY KEY (first_id, second_id))"); err != nil {
+		t.Fatalf("failed to create table with error '%v'", err)
+	}
+
+	if err := session.Query("CREATE INDEX index_column_metadata ON test_column_metadata ( third_id )").Exec(); err != nil {
+		t.Fatalf("failed to create index with err: %v", err)
+	}
+
+	columns, err := getColumnMetadata(session)
+	if err != nil {
+		t.Fatalf("failed to query column metadata with err: %v", err)
+	}
+	if columns == nil {
+		t.Fatal("failed to query column metadata, nil returned")
+	}
+
+	testColumns := map[string]*ColumnMetadata{}
+
+	// verify actual values on the test columns
+	for i := range columns {
+		column := &columns[i]
+
+		if column.Name == "" {
+			t.Errorf("Expected column name to be set, but it was empty: index=%d metadata=%+v", i, column)
+		}
+		if column.Table == "" {
+			t.Errorf("Expected column %s table name to be set, but it was empty", column.Name)
+		}
+		if column.Keyspace != "gocql_test" {
+			t.Errorf("Expected column %s keyspace name to be 'gocql_test', but it was '%s'", column.Name, column.Keyspace)
+		}
+		if column.Kind == "" {
+			t.Errorf("Expected column %s kind to be set, but it was empty", column.Name)
+		}
+		if session.cfg.ProtoVersion == 1 && column.Kind != "regular" {
+			t.Errorf("Expected column %s kind to be set to 'regular' for proto V1 but it was '%s'", column.Name, column.Kind)
+		}
+		if column.Validator == "" {
+			t.Errorf("Expected column %s validator to be set, but it was empty", column.Name)
+		}
+
+		// find the test table columns for the next step after this loop
+		if column.Table == "test_column_metadata" {
+			testColumns[column.Name] = column
+		}
+	}
+
+	if *flagProto == 1 {
+		// V1 proto only returns "regular columns"
+		if len(testColumns) != 1 {
+			t.Errorf("Expected 1 test columns but there were %d", len(testColumns))
+		}
+		thirdID, found := testColumns["third_id"]
+		if !found {
+			t.Fatalf("Expected to find column 'third_id' metadata but there was only %v", testColumns)
+		}
+
+		if thirdID.Kind != REGULAR {
+			t.Errorf("Expected %s column kind to be '%s' but it was '%s'", thirdID.Name, REGULAR, thirdID.Kind)
+		}
+
+		if thirdID.Index.Name != "index_column_metadata" {
+			t.Errorf("Expected %s column index name to be 'index_column_metadata' but it was '%s'", thirdID.Name, thirdID.Index.Name)
+		}
+	} else {
+		if len(testColumns) != 3 {
+			t.Errorf("Expected 3 test columns but there were %d", len(testColumns))
+		}
+		firstID, found := testColumns["first_id"]
+		if !found {
+			t.Fatalf("Expected to find column 'first_id' metadata but there was only %v", testColumns)
+		}
+		secondID, found := testColumns["second_id"]
+		if !found {
+			t.Fatalf("Expected to find column 'second_id' metadata but there was only %v", testColumns)
+		}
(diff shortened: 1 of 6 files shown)


justinretailnext:
This is the first PR of a series of PRs which split up the original PR #308 into more easily reviewable portions.

0x6e6562:
It is good that this patch breaks down #308 into smaller portions. One thing is not quite clear to me, but I've probably just overlooked it - is token aware query routing actually depending on getting this schema metadata? If so, does it rely on all of the metadata or just parts of it.

Functionally I am fine with this patch, as it is quite similar to the schema compiler in cqlc: https://github.com/relops/cqlc/blob/master/generator/schema.go

But I think we need to clarify the procedure before going to much further. It is one thing to break #308 down into more reviewable units, but it is another thing to get acceptance of the overall thrust of #308. 

Neither I nor @phillipCouto or @Zariel have come to an agreement on whether we this approach is good for gocql.

Functionally I would say yes, without doubt, especially as there are other people interested in the functionality.

But the big question is how.

If the goal is to land something like #308, albeit in small steps, there is still the decision to make as to whether we want to increase the maintenance scope of the core library to include of the constituent parts of this patch.

As [I pointed out in a related issue](https://github.com/gocql/gocql/issues/175#issuecomment-72766352), gocql has thus far taken a minimalistic approach to the layering of the library. There are good reasons outlined there why that is currently the case. It also points out that this is not necessarily set in stone, but we would need to get some kind of consensus on how we would want to change the maintenance focus of the library.

Up until now, the maintenance team has taken the approach to prioritize stability and robustness.

Specifically relating to this patch, these are my primary maintenance concerns:

* I'd like a bullet point plan of how this patch fits in with the #308 mini-roadmap before we start doing stuff.
* The schema parsing is a big thing to land and will probably take a few iterations to get rid of all of the bugs. At the same time, there is no apparent consumer of the schema metadata to be testing all of the functionality in live. So this means we would have a period where there is a lot of unused code on the master branch.
* The schema parsing is not a full proof of concept for UDTs, because it does not implement them. Therefore it is difficult to know whether this particular change will be sufficient for on ongoing UDT binding work.
* We will effectively be introducing a big building site into the master branch, which is likely to affect stability. This will take a bit of time to mature out.
* We still haven't agreed on whether this kind of thing belongs in the core or at a higher layer.

I'm not vetoing this change on those grounds, but it would be inconsistent of me to suddenly just accept a change in direction without consultation.

So to keep the ball rolling, it would good if @justinretailnext could outline his plan and also if @phillipCouto and @Zariel could help out with the guiding the approach.

0x6e6562:
Following on from my last point, I think we need to be clear that the DataStax Java driver will be heavily influencing this change, which may or may not be a good thing. But we would effectively be cloning their approach.

justinretailnext:
Based on feedback and a desire to address your maintenance scope concerns more directly, I have pushed an update to the commit in this PR. I have factored the code to allow each of the core functions (getKeyspaceMetadata, getTableMetadata, getColumnMetadata, and compileMetadata) to be tested independently. I have made the V1 and V2+ variations into distinct branches in the code where needed. There is now a unit test for both V1 and V2+ variations of the compileMetadata function as well as an integration test for each of getKeyspaceMetadata, getTableMetadata, and getColumnMetadata.

In response to your questions, yes the metadata is required for constructing the routing key for a particular query. Specifically the routing key construction to be added in the next PR uses the PartitionKey field of TableMetadata to determine which query values make up the partition key if at all. Also the token aware policy must take into account the value of StrategyClass from KeyspaceMetadata. There are indirect dependencies on some of the other metadata, particularly when constructing metadata for V1 protocol, however fields such as ClusteringColumns are provided for feature completeness, not out of any immediate need.

In terms of PR roadmap for this token aware connection pool change set:
1. Add keyspace metadata
2. Add routing key computation
3. Add token partitioners
4. Add token aware policy and policy based connection pool

I am currently reworking the final code change (4) to remove some of the API changes I had introduced and to improve the quality and completeness of the feature. I also would note that due mostly to a merge conflict with AUTHORS, that this PR is not currently ready to merge cleanly which I will address once you all are prepared to move forward with the PR with regards to the scope of the feature.

0x6e6562:
Thanks for the heads up. Sorry about the delay in responding to this patch, but I need to hear back from the other maintainers before proceeding with this. I'll get in contact with them directly and then hopefully we can move this forwards quickly.

0x6e6562:
Before going into any low level review on this my high level take is that we should move ahead with this. Whilst I do have maintenance concerns, on balance the benefits outweigh the costs.

In doing so I'd like to make sure that we have sufficient integration test coverage (i.e. a representative catalogue of relevant schema test cases that we will support, but done in way that extra test schema scenarios could be added in the future should other users find legitimate bugs). I'd also like this schema API to exported in such a way that it can be reused by other layers. Replacing the schema compiler in cqlc with this would help validate this new API.

Zariel:
I personally think this should be in gocql as it provides a solid grounding to implement more core driver features, such as token aware routing.

cassandra_test.go
```diff
@@ -1427,3 +1427,261 @@ func TestEmptyTimestamp(t *testing.T) {
...
+		column := &columns[i]
+
+		if column.Name == "" {
+			t.Errorf("Expected column name to be set, but it was empty: index=%d metadata=%+v", i, column)
+		}
+		if column.Table == "" {
+			t.Errorf("Expected column %s table name to be set, but it was empty", column.Name)
+		}
+		if column.Keyspace != "gocql_test" {
+
```
Zariel: Looks like this should cause Errorf?

helpers.go
```diff
@@ -97,9 +97,9 @@ func getApacheCassandraType(class string) Type {
 		case "MapType":
 			return TypeMap
 		case "ListType":
-			return TypeInet
```
Zariel: Good spot :+1:

metadata.go
```diff
@@ -0,0 +1,833 @@
...
+
+	current *KeyspaceMetadata
+	err     error
+}
+
+func (s *schemaDescriber) init() {
+	s.err = s.RefreshSchema()
+}
+
+func (s *schemaDescriber) GetSchema() (*KeyspaceMetadata, error) {
```
Zariel: `GetSchema` and `RefreshSchema` methods should be documented as to what the difference is, or not be exported

Zariel:
For this I would be in favour of merging mostly as-is, the external interface looks fine, the test coverage looks good, and gives a good building block for improving the driver and does nothing by itself so it wont impact other users.

Needs a rebase though

justinretailnext:
New rebase of the same code with changes per comments from @Zariel and fixes for integration tests. 

Looks like Travis is running out of memory on two of the test runs, I'm not sure if that's due to my changes however.

Zariel:
:+1: from me, @0x6e6562 thoughts?

0x6e6562:
There is a semi-known issue with the OOM errors with Travis. We thought we had gotten to the bottom of these issues, but obviously that is not the case. However, we can always re-trigger the specific failing legs, that usually does lead to a successful build.

Seeing as this patch is now ready for final review, I'll try to get through this ASAP to keep the ball rolling.

0x6e6562:
From an API point of view, I've patched cqlc to be the first external consumer of this new API (see https://github.com/relops/cqlc/pull/19). Right now, it knowlingly fails, because of the following issue.

What I'd like to do is to validate this patch using the patch to cqlc.

But right now, I don't think I've got the usage right:

```go
md, err := session.KeyspaceMetadata()
```

This returns the error:

```
Key may not be empty
```

Does this mean I need to set the keyspace field on the session?

justinretailnext:
Yes the code currently does expect that Keyspace in ClusterConfig is set to a non-empty keyspace name. 

The KeyspaceMetadata() function could be modified to take the keyspace as a parameter, though it would also need a way to signal to the function that the keyspace in ClusterConfig should be used. This is because the user of the Session may not have access to the Session's keyspace since cfg is private.

    // KeyspaceMetadata returns the schema metadata for the keyspace specified or if
    // keyspace is "" returns the metadata for the keyspace specified in the 
    // ClusterConfig for this session.
    func (s *Session) KeyspaceMetadata(keyspace string) (*KeyspaceMetadata, error)

0x6e6562:
It's a good question, but I think it would be clearer to mandate the keyspace name as an argument to `KeyspaceMetadata()` and return an error if is an empty string. That means the `Keyspace` field in the cfg instance would become irrelevant. This is more explicit. It is also more consistent IMHO since we use the `Keyspace` field to route to the correct key space, but in the case of metadata we are always querying the system key space, and the target schema key space is a query argument, and hence is on a different level.

Zariel:
Yeah I think that makes sense, otherwise you would be needing to change the cfg before querying the keyspace as a sort of side effect.

cassandra_test.go
```diff
@@ -1427,3 +1427,280 @@ func TestEmptyTimestamp(t *testing.T) {
...
+	for i := range tables {
+		table := &tables[i]
+
+		if table.Name == "" {
+			t.Errorf("Expected table name to be set, but it was empty: index=%d metadata=%+v", i, table)
+		}
+		if table.Keyspace != "gocql_test" {
+			t.Errorf("Expected keyspace for '%d' table metadata to be 'gocql_test' but was '%s'", table.Name, table.Keyspace)
+		}
+		if table.KeyValidator == "" {
```
0x6e6562: It's great that we've got test coverage for all of these attributes, but wouldn't it be clearer and more robust in the face of refactoring, if the assertion were a comparison to a concrete string, as opposed to just asserting that the fields are not empty?

cassandra_test.go
```diff
@@ -1427,3 +1427,280 @@ func TestEmptyTimestamp(t *testing.T) {
...
+			t.Errorf("Expected key validator to be set for table %s", table.Name)
+		}
+		if table.Comparator == "" {
+			t.Errorf("Expected comparator to be set for table %s", table.Name)
+		}
+		if table.DefaultValidator == "" {
+			t.Errorf("Expected default validator to be set for table %s", table.Name)
+		}
+
+		// these fields are not set until the metadata is compiled
```
0x6e6562: In the scope of this test case, at what point will the metadata be compiled?

metadata.go
```diff
@@ -0,0 +1,836 @@
...
+)
+
+type ColumnIndexMetadata struct {
+	Name    string
+	Type    string
+	Options map[string]interface{}
+}
+
+// Column kind values
+const (
```
0x6e6562: Is there a reason why these constants can't be enums instead of strings?

metadata.go
```diff
@@ -0,0 +1,836 @@
...
+type ColumnMetadata struct {
+	Keyspace       string
+	Table          string
+	Name           string
+	ComponentIndex int
+	Kind           string
+	Validator      string
+	Type           TypeInfo
+	Order          ColumnOrder
+	Index          ColumnIndexMetadata
```
0x6e6562: I think`ColumnMetadata` should have an extra field to denote whether the column is the last component of a partitioned key and therefore is eligible for use in

```
WHERE column IN (1,2,3)
```

clauses.

This would mean that you would need to re-process each `ColumnMetadata` element in order to assess the relative position of that column using the `ComponentIndex` field.

The advantage of doing this in the compiler is that it can be done once, rather than having downstream consumers having having to compute this. It also makes the metadata API more expressive.

0x6e6562:
I've been able to get a basic version of cqlc compiling against this patch, so the API looks good in a round about way. There's still a bit of work to do, but it looks as if we are going in the right direction.

In the meantime, I've created #314, which I noticed propagates an already existing bug through to cqlc, which didn't previously exist, because cqlc was doing it's own schema compilation. So we've been able to flush out some more bugs in the library by doing this exercise.

I've also created #315, which should make the rest run on Travis easier.

So as and when #314 and #315 land, it would be a good idea to merge them into this patch.

metadata.go
justinretailnext: The last column in a partition key would be returned by:
    pk := session.KeyspaceMetadata("keyspace_name").Tables["table_name"].PartitionKey
    last := pk[len(pk)-1]

It seems any further enhancement of the column metadata would be beyond the immediate need of the goal of token aware routing and could be explored as an independent effort.

metadata.go
justinretailnext: The reasoning is that this value is exposed as a string by cassandra and could be any number of possible values beyond what I need for token-aware routing and the string type work suitably. 

I'd also note that leaving this as a string is generally forward compatible regardless of what values may become valid in the future, where as coercing this value to a enum (e.g. int) would require limiting the expected values.

cassandra_test.go
justinretailnext: the compileMetadata function is not called in this function, the expectation is that getTableMetadata will never attempt to provide these values (possibly overly thorough).

cassandra_test.go
justinretailnext: The primary test is on the metadata returned for the explicitly created table "test_table_metadata", and this loop is primarily meant to filter for that table. As a secondary test, I have added assertions that enforce the greatest common expectation of the getTableMetadata call which is simply that, regardless of what tables might be in this keyspace, these fields of the metadata will be filled (non-empty strings, and non-nil). Attempting to expect specific values would make this test brittle to changes in test function ordering, and what tables those tests create.

justinretailnext:
I've pushed a rebased version of the code including #314 and #315 as well as the addition of the keyspace name parameter to KeyspaceMetadata.

0x6e6562:
I've now been able to get the cqlc integration test suite to [pass against this new metadata API](https://travis-ci.org/relops/cqlc/builds/53937027), so this is a good validation of the shape of the API and its current implementation.

metadata.go
0x6e6562: That's a reasonable point - it would be inconsistent of me to require the scope of this pull request get expanded - after all, we want functionality to land in bite sized pieces.

metadata.go
0x6e6562: OK, fair point - let's leave them as strings then.

cassandra_test.go
0x6e6562: `func getTableMetadata()` is documented thusly:

```
// query for only the table metadata in the specified keyspace
```

Are you just saying here that this function only returns the table level metadata, and any further metadata would be need to harvested by introspecting the individual columns of the table?

cassandra_test.go
0x6e6562: I take your point about having to be concrete about the table names. To implement a test in a non-brittle way would require some level of test sophistication, and given that we already have an extensive integration test suite for the new API (the cqlc integration test suite now uses this API), it would silly to divert engineering effort into creating a fantastic test suite at the cost of adding real features.

cassandra_test.go
```diff
@@ -1427,3 +1427,280 @@ func TestEmptyTimestamp(t *testing.T) {
 		t.Errorf("time.Time bind variable should still be empty (was %s)", timeVal)
 	}
 }
+
+func TestGetKeyspaceMetadata(t *testing.T) {
```
0x6e6562: Can we add a one line note to indicate what this test is verifying?

cassandra_test.go
```diff
@@ -1427,3 +1427,280 @@ func TestEmptyTimestamp(t *testing.T) {
...
+			t.Errorf("Expected %s column kind to be '%s' but it was '%s'", thirdID.Name, REGULAR, thirdID.Kind)
+		}
+
+		if thirdID.Index.Name != "index_column_metadata" {
+			t.Errorf("Expected %s column index name to be 'index_column_metadata' but it was '%s'", thirdID.Name, thirdID.Index.Name)
+		}
+	}
+}
+
+func TestKeyspaceMetadata(t *testing.T) {
```
0x6e6562: This test case should a have a 1-line note to say what it is testing.

metadata_test.go
```diff
@@ -0,0 +1,670 @@
...
+							t.Errorf("Expected %s.Tables[%s].Columns[%s].Kind to be '%v' but was '%v'", expected.Name, keyT, keyC, ec.Kind, ac.Kind)
+						}
+					}
+				}
+			}
+		}
+	}
+}
+
+func TestTypeParser(t *testing.T) {
```
0x6e6562: This test requires a 1-line explanation.

0x6e6562:
We are very nearly there :-)

Code wise, everything looks fine and the fact that cqlc integrates so well with this is a great indication that this patch is good.

There are just some minor documentation issues to fix and the reason behind one comment to clarify.

Travis-wise we are just getting the intermittent OOM issues, which has nothing to do with this patch and will rectify themselves when the build gets re-triggered.

So once these documentation issues have been sorted, this pull request is good to merge.

justinretailnext:
The test functions are named for the function they test. Adding "Test of functionName" to the test function named "TestFunctionName" is pedantic. The naming is also consistent with other test suites in gocql.

Please keep in mind that I am maintaining 3 other branches / future PRs for the token aware routing functionality; every change means a interactive rebase and full integration tests for each. I would appreciate if we could move this process forward and take note of these "polish" issues as independent future work.

Zariel:
+1, we should merge this and fix nits, slight improvements later.

0x6e6562:
Please bear in mind that I spend a lot of effort keeping the gocql in a good working state, so when I'm asking for certain things to be done, I'm not doing it to make people jump through unnecessary hoops.

The point about documenting the test cases is inline with an agreement between @Zariel and @phillipCouto and myself to document test cases going forwards, so I am just sticking to a previous agreement.

But I'm not going to stand here and split hairs about things - especially when two other people are calling me pedantic :-)

I'm happy with the code and I've tested it personally with a non-trivial use case and fundamentally I really appreciate the effort that you've put into it, so I don't want to put you off in the future.

Thank you very much for your hard work on this pull request.

PR #311, opened 2015-02-06
Fix a panic establishing sessions using NewSession

obeattie:
Previously, the only way to establish the *first* connection was to use `ClusterConfig.CreateSession()`. This is due to the global prepared statement cache only being initialised here.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -366,7 +366,7 @@ func (c *Conn) ping() error {
 func (c *Conn) prepareStatement(stmt string, trace Tracer) (*QueryInfo, error) {
 	stmtsLRU.Lock()
 	if stmtsLRU.lru == nil {
-		initStmtsLRU(1000)
+		initStmtsLRU(defaultMaxPreparedStmts)
 	}
 
 	stmtCacheKey := c.addr + c.currentKeyspace + stmt
--- a/conn.go
+++ b/conn.go
@@ -364,13 +364,16 @@ func (c *Conn) ping() error {
 }
 
 func (c *Conn) prepareStatement(stmt string, trace Tracer) (*QueryInfo, error) {
-	stmtsLRU.mu.Lock()
+	stmtsLRU.Lock()
+	if stmtsLRU.lru == nil {
+		initStmtsLRU(1000)
+	}
 
 	stmtCacheKey := c.addr + c.currentKeyspace + stmt
 
 	if val, ok := stmtsLRU.lru.Get(stmtCacheKey); ok {
 		flight := val.(*inflightPrepare)
-		stmtsLRU.mu.Unlock()
+		stmtsLRU.Unlock()
 		flight.wg.Wait()
 		return flight.info, flight.err
 	}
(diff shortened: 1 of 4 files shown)


conn.go
```diff
@@ -364,13 +364,16 @@ func (c *Conn) ping() error {
 }
 
 func (c *Conn) prepareStatement(stmt string, trace Tracer) (*QueryInfo, error) {
-	stmtsLRU.mu.Lock()
+	stmtsLRU.Lock()
+	if stmtsLRU.lru == nil {
+		initStmtsLRU(1000)
```
0x6e6562: It would be nicer if this value referred to a constant that is defined somewhere.

obeattie: Done.

0x6e6562:
Thank you very much for your contribution. On the whole it looks pretty sane, apart from the comment about the magic number.

Could you outline in the summary why you may not choose to use `ClusterConfig.CreateSession()` to establish a connection? Not only would that help me understand the motivation behind the change (because I'm interested in how and why people use gocql), but it would also serve as documentation for other people who find themselves in a similar situation to yourself.

0x6e6562:
Also, as per `CONTRIBUTING.md`, could you add yourself to the `AUTHORS` file please?

obeattie:
The reason is that I want to use a custom connection pool.

Specifically, if I use `ClusterConfig.CreateSession`, [a new connection pool is constructed each time](https://github.com/gocql/gocql/blob/master/cluster.go#L9). I could, in retrospect, supply a constructor that always vends the same pool. There are also other limitations with using the `ClusterConfig.CreateSession` (specifically that I need to "check-out" and "check-in" connections, and there is no way to get the underlying `Conn` from a `Session`).

It seems to me that the APIs of `ConnectionPool`, `Session`, and `Conn` are rather tightly coupled, but I didn't want to embark on a wholesale refactor or anything like that.

Thanks for your review. I've added myself to `AUTHORS` as requested.

0x6e6562:
The default implementation of 

```
type NewPoolFunc func(*ClusterConfig) ConnectionPool
```

is `NewSimplePool (defined in `connectionpool.go`) as specified by default on the `ClusterConfig` instance.

So whilst not devaluing your patch, strictly speaking you could have supplied your own `ConnPoolType` callback.

I think it is fair to say that the relationships between the various pool, connection and session objects can always evolve and become more flexible and customizable, as they have already done so in the lifetime of this library. So I agree with your sentiment about them tightly coupled, even though there actually is already some form of abstraction in place (my observation above supports this point - you didn't necessarily need to patch the library to achieve the specific goal of this pull request).

Also, I can see that you have other requirements (check-in/check-out connections, getting the connection from the session) - is landing this particular patch a requirement for achieving those requirements?

On reflection, this patch adds value because it makes the library a little bit more flexible, so I am inclined to merge it. But I'd still like to discuss non-obvious changes before applying them to the core library.

obeattie:
I mentioned above in my comment that I could supply a constructor that always vends the same pool. Is this what you mean?

And yes, I think even if I did that, I still could not do checking-in/out without doing what I described above (because of the restrictions of not being able to get the underlying `Conn` from a `Session`).

I do think regardless of what I may be trying to do with it, and the correctness of that, a user of this library would expect to be able to call `NewSession` without first calling `ClusterConfig.CreateSession`. Panics in any form are pretty unexpected and terrible.

0x6e6562:
In general we treat panics as bugs and that is not the expected behavior of the library, so this is why the patch should land.

We are also open to evolving the other abstractions, so it might be worth kicking off a discussion about your specific requirements.

But for now, thanks a lot for this contribution, it is much appreciated.

PR #312, opened 2015-02-24
Fix memory leakage  when running bulk insertions

criloz:
it was using 950 MB after running by  5 minutes

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -457,6 +457,12 @@ func (c *Conn) executeQuery(qry *Query) *Iter {
 	}
 	switch x := resp.(type) {
 	case resultVoidFrame:
+		stmtsLRU.Lock()
+		stmtCacheKey := c.addr + c.currentKeyspace + qry.stmt
+		if _, ok := stmtsLRU.lru.Get(stmtCacheKey); ok {
+			stmtsLRU.lru.Remove(stmtCacheKey)
+		}
+		stmtsLRU.Unlock()
 		return &Iter{}
 	case resultRowsFrame:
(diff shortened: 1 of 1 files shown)


phillipCouto:
Can we see an example of the batch statements you are executing?

criloz:
BATCH with 2000 insertions. now all its running ok, the app barely go over 18 mb

phillipCouto:
Yes but you are introducing an inefficiency by purging the LRU which defeats the purpose of it to start with. Can you paste an example of one of your INSERT statements from your program. Preferably the Go code that adding the insert statement to the batch would be great.

criloz:
```go
css.buffer[worker] = "BEGIN BATCH\n" + css.buffer[worker] + "\nAPPLY BATCH;"
q := css.session.Query(css.buffer[worker])
```

is there a way to disable this cached?

phillipCouto:
What version of Cassandra are you using?

criloz:
Ahh  sorry I just notice that there is a Batch type struct, sorry my error. I will close the pull request.

phillipCouto:
no problem. Also try to use parameters in your statements. Like:
```
batch.Query("INSERT INTO Table (Col1,Col2,Col3) VALUES (?,?,?)",1,2,3)
```

This will only cache one copy of your query and keep your memory usage to a minimum and improve write performance in Cassandra.

criloz:
Thanks!!

0x6e6562:
@criloz Does this pull request need to remain open?

criloz:
@0x6e6562  tried, but the github ui don't let me  close it.

PR #316, opened 2015-03-05
Authentication integration test

0x6e6562:
This patch adds an authentication test to the integration test suite.

The main motivation behind this patch is to have some test coverage for authentication, without going overboard. Authentication in protocol v1 is not supported in gocql in any case. There also seems to be an issue with 2.0.11 requiring that the default user is actively created, whereas in 2.1.x this default user seems to exist. So what I've done is to only run the integration tests against 2.1.x.


0x6e6562:
Right now this is stalled because the `go test` invocation just seems to hang and it is not at all clear why.

0x6e6562:
OK, so I've settled for proto 2 only and Cassandra 2.1+ (in 2.0.x it looks like you need to actively create the user).

So now I need to put the main tests back and clean up.

Zariel:
Was that test pass a reproducible pass?

0x6e6562:
I think it was reproduceable. The reason why it is now broken is because I've re-introduced the original tests and in doing so I've broken something with a TLS setting, but I can't see what that is right now.

0x6e6562:
The multiline quoting style of `ccm updateconf` seems to be an issue, so I reverted the old style of putting everything on 1 line.

Zariel:
yeah, I think we should change it to be like this

```bash
	local conf=(
		"client_encryption_options.enabled: true"
		"client_encryption_options.keystore: testdata/pki/.keystore"
		"client_encryption_options.keystore_password: cassandra"
		"client_encryption_options.require_client_auth: true"
		"client_encryption_options.truststore: testdata/pki/.truststore"
		"client_encryption_options.truststore_password: cassandra"
		"concurrent_reads: 2"
		"concurrent_writes: 2"
		"rpc_server_type: hsha"
		"rpc_min_threads: 1"
		"rpc_max_threads: 1"
		"write_request_timeout_in_ms: 5000"
		"read_request_timeout_in_ms: 5000"
		"compaction_throughput_mb_per_sec: 0"
	)

	for f in "${conf[@]}"; do
		ccm updateconf "$f"
	done
```

0x6e6562:
I think you suggestion is good, but seeing as I'm fire fighting other issues and that I seem to have a running `ccm updateconf`, I would prefer to defer this improvement to a separate patch.

Zariel:
indeed, looks like your getting closer though.

0x6e6562:
It is amazing the amount of effort we have stuck into this ccm business, but I guess that's what it takes if you want to automate testing. On a separate note, we still have an OOM issue, I think.

Zariel:
try removing the heap constraints, we have like 4gb of ram in travis which should be plenty for 3 nodes

Zariel:
The nodes may also be doing a compaction after the tests are ran so it might be worth either disabling compaction all together or adding sleeps

0x6e6562:
Thanks for the tip, but I'm going to have to leave things for now, so I'll try that later.

My current resting place is this issue:

```
+ccm status
Cluster: 'test'
---------------
node1: UP
node3: UP
node2: UP
+ccm node1 nodetool status
Datacenter: datacenter1
=======================
Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address    Load       Tokens  Owns    Host ID                               Rack
UN  127.0.0.1  344.24 KB  256     ?       2105861c-ae88-4af7-9908-a60ceb89923f  rack1
UN  127.0.0.2  277.74 KB  256     ?       2c7ed228-d5b4-45dd-b87b-3c11b7834efc  rack1
UN  127.0.0.3  263.66 KB  256     ?       2153bc9f-de83-4793-b422-3f1a24516554  rack1

Note: Non-system keyspaces don't have the same replication settings, effective ownership information is meaningless
++ccm liveset
+go test -v . -timeout 15s -run=TestAuthentication -tags integration -runauth -proto=2 -cluster=127.0.0.1,127.0.0.3,127.0.0.2 -clusterSize=3 -autowait=5000ms
=== RUN TestAuthentication
2015/03/07 10:25:08 connectionpool.go:170: connect: failed to connect to "127.0.0.1:9042": EOF
--- FAIL: TestAuthentication (0.07 seconds)
	cassandra_test.go:136: Authentication error: no connections were made when creating the session
FAIL
exit status 1
FAIL	github.com/gocql/gocql	0.079s
```

Somehow the node ownership is an issue, but I can't see why.

0x6e6562:
How do disable compaction with ccm?

Zariel:
probably nodetool on each node

Zariel:
Are we logging everything that can cause a dial failure in the ssl case?

0x6e6562:
Don't think we're doing anything special for logging with dial issues.

0x6e6562:
I'm thinking it is not worth pursing this approach - maybe if the auth config were in its own separate build matrix, we would not be battling a ccm restart?

Zariel:
i wonder if this is related to not setting write timeouts? (the lack of log output and hang)

0x6e6562:
Which write timeout are you referring to?

Zariel:
If you follow the Write() path in conn.go we never set a WriteDeadline, I tried adding it in #328 but I don't think it actually causes an error.

On another note I've realised we could handle query timeouts better, the java-driver will send the query, start a timer then once that expires it will return and throw a timeout up the call stack. We instead rely on ReadDeadlines on the TCP connection which is not entirely indicative of a query timing out as we have potentially many queries on a single connection. We can simulate similar behaviour by having a select with a timeout on the read from the channel waiting for the response, but would need to be careful that we line up sending a receiving correctly.

0x6e6562:
Very interesting. So basically we need to define the semantics of what we think causes a timeout. This could be due to aggregate I/O latency in the server cluster or due to network latency or a bit of both. But assuming that network latency is a good measure for end to end latency might be a bad assumption?

Zariel:
Yeah, I think we should have a RW timeout on the conn itself but consider a query timedout at a specific time as well.

We also need to check that we do the right thing in the recv() loop, we essentially want to keep reading headers, then the body, but we might not always be receiving a header within the timeout due to many connections or other reasons.

Zariel:
@0x6e6562 can you rerun this against the current master?

0x6e6562:
@Zariel I've just merged the latest master into this patch and I get the same EOF issue.

Zariel:
Do these tests pass locally?

0x6e6562:
Yes, they are running locally against a local ccm instance.

Zariel:
When I run this in my vagrant box I get this error,

`io.netty.handler.ssl.NotSslRecordException: not an SSL/TLS record: 02000001000000160001000b43514c5f56455253494f4e0005332e302e30
        at io.netty.handler.ssl.SslHandler.decode(SslHandler.java:860) ~[netty-all-4.0.23.Final.jar:4.0.23.Final]
        at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:249) ~[netty-all-4.0.23.Final.jar:4.0.23.Final]
        at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:149) ~[netty-all-4.0.23.Final.jar:4.0.23.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:333) ~[netty-all-4.0.23.Final.jar:4.0.23.Final]
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:319) ~[netty-all-4.0.23.Final.jar:4.0.23.Final]
        at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:787) ~[netty-all-4.0.23.Final.jar:4.0.23.Final]
        at io.netty.channel.epoll.EpollSocketChannel$EpollSocketUnsafe.epollInReady(EpollSocketChannel.java:722) ~[netty-all-4.0.23.Final.jar:4.0.23.Final]
        at io.netty.channel.epoll.EpollEventLoop.processReady(EpollEventLoop.java:326) ~[netty-all-4.0.23.Final.jar:4.0.23.Final]
        at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:264) ~[netty-all-4.0.23.Final.jar:4.0.23.Final]
        at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:116) ~[netty-all-4.0.23.Final.jar:4.0.23.Final]
        at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:137) ~[netty-all-4.0.23.Final.jar:4.0.23.Final]
        at java.lang.Thread.run(Thread.java:745) [na:1.7.0_65]`

0x6e6562:
Interesting - so maybe I've somehow managed to mix in TLS into the ccm setup?

Zariel:
I think the setting for SSL is incorrect as well,

`ERROR [main] 2015-04-07 12:50:44,342 CassandraDaemon.java:550 - Exception encountered during startup
java.lang.RuntimeException: Failed to setup secure pipeline
        at org.apache.cassandra.transport.Server$SecureInitializer.<init>(Server.java:315) ~[apache-cassandra-2.1.4.jar:2.1.4]
        at org.apache.cassandra.transport.Server.run(Server.java:171) ~[apache-cassandra-2.1.4.jar:2.1.4]
        at org.apache.cassandra.transport.Server.start(Server.java:114) ~[apache-cassandra-2.1.4.jar:2.1.4]
        at org.apache.cassandra.service.CassandraDaemon.start(CassandraDaemon.java:463) [apache-cassandra-2.1.4.jar:2.1.4]
        at org.apache.cassandra.service.CassandraDaemon.activate(CassandraDaemon.java:546) [apache-cassandra-2.1.4.jar:2.1.4]
        at org.apache.cassandra.service.CassandraDaemon.main(CassandraDaemon.java:622) [apache-cassandra-2.1.4.jar:2.1.4]
Caused by: java.io.IOException: Error creating the initializing the SSL Context
        at org.apache.cassandra.security.SSLFactory.createSSLContext(SSLFactory.java:147) ~[apache-cassandra-2.1.4.jar:2.1.4]
        at org.apache.cassandra.transport.Server$SecureInitializer.<init>(Server.java:311) ~[apache-cassandra-2.1.4.jar:2.1.4]
        ... 5 common frames omitted
Caused by: java.io.FileNotFoundException: testdata/pki/.truststore (No such file or directory)
        at java.io.FileInputStream.open(Native Method) ~[na:1.7.0_65]
        at java.io.FileInputStream.<init>(FileInputStream.java:146) ~[na:1.7.0_65]
        at java.io.FileInputStream.<init>(FileInputStream.java:101) ~[na:1.7.0_65]
        at org.apache.cassandra.security.SSLFactory.createSSLContext(SSLFactory.java:114) ~[apache-cassandra-2.1.4.jar:2.1.4]
        ... 6 common frames omitted`

Zariel:
It looks like SSLOpts are not being enabled, `2015/04/07 12:57:48 conn.go:121: cfg={ProtoVersion:2 CQLVersion:3.0.0 Timeout:600ms NumStreams:128 Compressor:<nil> Authenticator:{Username:cassandra Password:cassandra} Keepalive:0 SslOpts:<nil>}`

Zariel:
I got this to pass but using `createCluster`. Seems to me the problem is that we are setting force client encryption, but then some of the tests are not using the ssl options.

I don't know why the other tests pass, they probably shouldnt. Cassandra seems pretty bad at handling the error, its basically waiting for TLS handshake but we send a startup frame, then it gets an exception from Netty and closes the connections, giving us the EOF. Ill create some seperate issues and try to resolve this.

Zariel:
created #346 and #345

Zariel:
I think we should also enable this for all version, and have the testsuite create the user as part of createSession after setting up the keyspace.

0x6e6562:
So you're suggesting that #346 and #345 need to land before this patch can land (and transitively #339)?

Zariel:
I have a fix for #345 coming up

Zariel:
I would suggest adding this to `createKeyspace` then removing the $AUTH matrix entry

```go
	err = session.Query("CREATE USER IF NOT EXISTS gocql WITH PASSWORD 'gocql'").Consistency(All).Exec()
	if err != nil {
		tb.Fatalf("error creating user: %v", err)
	}
```

0x6e6562:
The suggestion about eagerly testing the auth settings is good, but does that not then mean we have no test coverage for non-authenticated sessions?

Zariel:
This should have ran with `-runssl`

integration.sh
```diff
@@ -41,25 +51,31 @@ function run_tests() {
 		proto=3
 	fi
 
-	go test -timeout 5m -tags integration -cover -v -runssl -proto=$proto -rf=3 -cluster=$(ccm liveset) -clusterSize=$clusterSize -autowait=2000ms -compressor=snappy ./... | tee results.txt
+	if [ "$auth" = true ]
+    then
+    	go test -v . -timeout 15s -run=TestAuthentication -tags integration -runauth -proto=$proto -cluster=$(ccm liveset) -clusterSize=$clusterSize -autowait=1000ms
```
Zariel: need to add `-runssl` to this

Zariel:
Does it make sense to only run a single test though with the authentication? The only difference in the startup is that instead of getting OP_READY you do the authentication dance and exchange some frames.

0x6e6562:
>Does it make sense to only run a single test though with the authentication?

I thought that was the point of the $AUTH matrix entry. When the $AUTH flag is set, then the only test that should get executed is `TestAuthentication`. So this would be the only test with authentication.

Zariel:
True.

Getting somewhere at last :) 

`2015/04/07 15:33:56 connectionpool.go:214: connect: failed to connect to "127.0.0.1:9042": Username and/or password are incorrect`

Zariel:
I think that might be caused by the fact that the

Zariel:
Sorry didnt mean to do that

0x6e6562:
I wonder what the real password is - I'd assumed that by default it is cassandra/cassandra.

Zariel:
That should be the default

Zariel:
I think the user is created after Cassandra starts listening for clients, its the last thing in the log when I try to run it locally.

Zariel:
I dont know what else to do here except add a sleep?

Zariel:
might need a bit longer than 15ms, try 3 seconds. Also the cqlsh line is causing it to fail before it gets to the tests

0x6e6562:
Does `sleep 15` not mean 15 seconds?

0x6e6562:
Yes, I was trying to use `cqlsh` as a way of testing connectivity to Cassandra without any of our code, i.e. in a fail fast fashion.

The underlying issue is

```
Connection error: TSocket read 0 bytes
```

which indicates that the listener in Cassandra may not yet have started.

Zariel:
I get that when doing it locally when Cassandra has defiantly started up so im not sure.

Zariel:
The cqlsh issue is because of requiring SSL, cqlsh needs setting up to use it too, I couldn't get it to work locally.

Zariel:
So that looks like cqlsh from 2.0.x which doesnt have ssl flags, not sure it has ssl config either.

I think you instead want to use `ccm node1 cqlsh`

Zariel:
I get this error when I try locally.

`vagrant@vagrant-ubuntu-trusty-64:/vagrant/testdata/pki$ SSL_VALIDATE=false SSL_CERTFILE=$(pwd)/cassandra.crt ccm node1 cqlsh -- --ssl
Connection error: ('Unable to connect to any servers', {'127.0.0.1': error(8, "Tried connecting to [('127.0.0.1', 9042)]. Last error: _ssl.c:510: EOF occurred in violation of protocol")})`

0x6e6562:
Looking at `sslhandling.py`, it seems that we might need to set the user key and password in `~/.cqlshrc` as well - if so, I wonder how we script that?

0x6e6562:
Build 776 (with client auth turned off in the server) gives an interesting error:

```
+ccm node1 cqlsh --version
cqlsh 5.0.1
+sleep 3s
+SSL_VALIDATE=false
+SSL_CERTFILE=/home/travis/gopath/src/github.com/gocql/gocql/testdata/pki/cassandra.crt
+ccm node1 cqlsh --ssl -u cassandra -p cassandra
+echo 'list users;'
Connection error: ('Unable to connect to any servers', {'127.0.0.1': AuthenticationFailed(u'Failed to authenticate to 127.0.0.1: code=0100 [Bad credentials] message="Username and/or password are incorrect"',)})
```

Zariel:
Yeah it occurred to me that if cqlsh is using the python driver and we cant authenticate then I doubt it will be able to either :)

Interestingly when I run this locally, heres the logs. Check the time between listening for clients and creating the superuser

`INFO  [main] 2015-04-09 13:59:13,739 Server.java:180 - Starting listening for CQL clients on /127.0.0.1:9042...
INFO  [main] 2015-04-09 13:59:13,850 ThriftServer.java:119 - Binding thrift service to /127.0.0.1:9160
INFO  [main] 2015-04-09 13:59:13,868 CustomTThreadPoolServer.java:250 - enabling encrypted thrift connections between client and server
INFO  [Thread-2] 2015-04-09 13:59:14,038 ThriftServer.java:136 - Listening for thrift clients...
WARN  [NonPeriodicTasks:1] 2015-04-09 13:59:23,115 FBUtilities.java:349 - Trigger directory doesn't exist, please create it and try again.
INFO  [NonPeriodicTasks:1] 2015-04-09 13:59:23,122 PasswordAuthenticator.java:213 - PasswordAuthenticator created default user 'cassandra'
INFO  [NonPeriodicTasks:1] 2015-04-09 13:59:23,143 Auth.java:239 - Created default superuser 'cassandra'`

Zariel:
And I can get it to pass if I add a 20s sleep, lol, timing bugs. I wonder if this is a bug in cassandra in that it should create that user before listening? I might see if I can fix it.

Zariel:
!()[http://gifs.joelglovier.com/napoleon/kip_yes.gif]

Zariel:
Might be a good idea to squash this monster of a PR into a single commit :) its been a hell of a ride!

0x6e6562:
Yeah, good idea - is there some button on GitHub somewhere to do that?

Zariel:
unfortunately you have to do the good old rebase squash

0x6e6562:
This patch was superseded by #354.

PR #317, opened 2015-03-05
Use HSHA, tweak rpc_min/max_threads

Zariel:
Use HSHA instead of the SYNC connection pool for the RPC threads,
allow a max of 8 threads which is the default in the cluster config

### DIFF
--- a/cass1batch_test.go
+++ b/cass1batch_test.go
@@ -9,6 +9,8 @@ import (
 
 func TestProto1BatchInsert(t *testing.T) {
 	session := createSession(t)
+	defer session.Close()
+
 	if err := session.Query("CREATE TABLE large (id int primary key)").Exec(); err != nil {
 		t.Fatal("create table:", err)
 	}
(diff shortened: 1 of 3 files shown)


0x6e6562:
I'll be interested in seeing what the fix for this is, since I got stalled on running `go test` with ccm as part of #316.

Zariel:
`io.netty.handler.codec.DecoderException: org.apache.cassandra.transport.ProtocolException: Invalid or unsupported protocol version: 22`

Zariel:
Ive been able to reproduce the travis failures in vagrant locally, but not sure why they happen. It seems fine if I launch a cluster with ccm then run go test, odd.

0x6e6562:
It would be interesting to get to the bottom on the ccm errors in general as this seems to cause us issues every time we want to change things :-)

Zariel:
Yeah, ive also realised when we run in proto=1 we never set the protocol to cassandra as protocol 1, we still operate with protocol = 0x82/0x02

0x6e6562:
Yes, as with the last major effort to get the integration tests to work, it turned out to be bugs in our own code rather than anything wrong with ccm or anything else.

That proto 1 bug sounds like a pull request in its own right. Or we just get rid of v1 support all together.

Zariel:
I don't think is necessary anymore

PR #318, opened 2015-03-07
Support protocol v3 wire format

### LINKED ISSUE #176: Increase number of streams per connection (2.1)
Another 2.1 change is that clients can use more than the current 127 streams per connection, will be a nice little performance booster reusing connections more - https://issues.apache.org/jira/browse/CASSANDRA-7231

Zariel:
closes #176

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -330,15 +330,20 @@ func (c *Conn) execSimple(op operation) (interface{}, error) {
 		// this should be a noop err
 		return nil, err
 	}
-	f.setLength(len(f)-headerProtoSize[c.version], c.version)
+
+	bodyLen := len(f) - headerProtoSize[c.version]
+	f.setLength(bodyLen, c.version)
 
 	if _, err := c.Write([]byte(f)); err != nil {
 		c.Close()
 		return nil, err
 	}
+
+	// here recv wont timeout waiting for a header, should it?
 	if f, err = c.recv(); err != nil {
 		return nil, err
 	}
+
 	return c.decodeFrame(f, nil)
 }
 
@@ -347,6 +352,7 @@ func (c *Conn) exec(op operation, trace Tracer) (interface{}, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	if trace != nil {
 		req[1] |= flagTrace
 	}
@@ -360,7 +366,8 @@ func (c *Conn) exec(op operation, trace Tracer) (interface{}, error) {
 		req = append(req[:headerSize], frame(body)...)
 		req[1] |= flagCompress
 	}
-	req.setLength(len(req)-headerSize, c.version)
+	bodyLen := len(req) - headerSize
+	req.setLength(bodyLen, c.version)
 
 	id := <-c.uniq
 	req.setStream(id, c.version)
--- a/conn.go
+++ b/conn.go
@@ -21,9 +21,11 @@ import (
 	"time"
 )
 
-const defaultFrameSize = 4096
-const flagResponse = 0x80
-const maskVersion = 0x7F
+const (
+	defaultFrameSize = 4096
+	flagResponse     = 0x80
+	maskVersion      = 0x7F
+)
 
 //JoinHostPort is a utility to return a address string that can be used
 //gocql.Conn to form a connection with a host.
--- a/conn.go
+++ b/conn.go
@@ -389,6 +389,7 @@ func (c *Conn) exec(op operation, trace Tracer) (interface{}, error) {
 	if reply.err != nil {
 		return nil, reply.err
 	}
+
 	return c.decodeFrame(reply.buf, trace)
 }
 
@@ -683,13 +684,14 @@ func (c *Conn) decodeFrame(f frame, trace Tracer) (rval interface{}, err error) 
 		}
 	}()
 
-	if len(f) < headerProtoSize[c.version] {
-		return nil, NewErrProtocol("Decoding frame: less data received than required for header: %d < %d", len(f), headerProtoSize[c.version])
+	headerSize := headerProtoSize[c.version]
+	if len(f) < headerSize {
+		return nil, NewErrProtocol("Decoding frame: less data received than required for header: %d < %d", len(f), headerSize)
 	} else if f[0] != c.version|flagResponse {
 		return nil, NewErrProtocol("Decoding frame: response protocol version does not match connection protocol version (%d != %d)", f[0], c.version|flagResponse)
 	}
 
-	flags, op, f := f[1], f.Op(c.version), f[headerProtoSize[c.version]:]
+	flags, op, f := f[1], f.Op(c.version), f[headerSize:]
 	if flags&flagCompress != 0 && len(f) > 0 && c.compressor != nil {
 		if buf, err := c.compressor.Decode([]byte(f)); err != nil {
 			return nil, err
--- a/conn.go
+++ b/conn.go
@@ -10,7 +10,9 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
+	"log"
 	"net"
 	"strconv"
 	"strings"
@@ -88,7 +90,7 @@ type Conn struct {
 	r       *bufio.Reader
 	timeout time.Duration
 
-	uniq  chan uint8
+	uniq  chan int
 	calls []callReq
 	nwait int32
 
@@ -123,14 +125,17 @@ func Connect(addr string, cfg ConnConfig, pool ConnectionPool) (*Conn, error) {
 				return nil, errors.New("Failed parsing or appending certs")
 			}
 		}
+
 		mycert, err := tls.LoadX509KeyPair(cfg.SslOpts.CertPath, cfg.SslOpts.KeyPath)
 		if err != nil {
 			return nil, err
 		}
+
 		config := tls.Config{
 			Certificates: []tls.Certificate{mycert},
 			RootCAs:      certPool,
 		}
+
 		config.InsecureSkipVerify = !cfg.SslOpts.EnableHostVerification
 		if conn, err = tls.Dial("tcp", addr, &config); err != nil {
 			return nil, err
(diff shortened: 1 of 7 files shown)


Zariel:
This became bigger as it touches quite a lot of code, I think the frame and conn files could do with a tidy up and improvement as well

conn.go
```diff
@@ -139,16 +146,25 @@ func Connect(addr string, cfg ConnConfig, pool ConnectionPool) (*Conn, error) {
...
-	if cfg.NumStreams <= 0 || cfg.NumStreams > 128 {
-		cfg.NumStreams = 128
-	}
-	if cfg.ProtoVersion != 1 && cfg.ProtoVersion != 2 {
+	// going to default to proto 2
+	if cfg.ProtoVersion < protoVersion1 || cfg.ProtoVersion > protoVersion3 {
+		log.Printf("unsupported protocol version: %d using 2\n", cfg.ProtoVersion)
 		cfg.ProtoVersion = 2
 	}
+
```
0x6e6562: This block of code is duplicated from `cluster.go` - can we refactor this a bit?

conn.go
```diff
@@ -88,7 +92,7 @@ type Conn struct {
 	r       *bufio.Reader
 	timeout time.Duration
 
-	uniq  chan uint8
+	uniq  chan int
```
0x6e6562: Is this really a signed integer in the wire protocol?

0x6e6562:
I'm wondering whether we want to maintain these test servers going forwards - I would rather put the effort into testing against real severs instead of spending time on mock servers.

conn.go
Zariel: on the wire its either an int8 or an int16, but incase they change it in the future I thought an int would fit best. Could potentially use a uint though as we should only ever send >= 0 stream ids

Zariel:
True, most of the tests in conn_test are actually testing a session with a mock db. I think have the mocks does have benefits, but I would like to rework the test server.

conn.go
Zariel: I think we still need to have it in both places, maybe a common function is better. The reason we need it in both places is that Conn is exported so Session is not the only creator of connections, but if we dont care about this use case then it can be omitted from Conn along with the protocol validation.

0x6e6562:
What are the benefits of a mock server over the real thing?

Zariel:
We can validate that the client reacts as we expect it to in certain situations which are hard to replicate with a real server. We can validate that the protocol is how we expect it.

0x6e6562:
I take your point about being able to induce an erroneous server behavior in order to verify the client's handling of this situation. If there were multiple implantations of the CQL protocol (e.g. something other than Apache Cassandra), this kind of checking might have some value. But in our case, we are talking about one single implementation. The fact that it is called a protocol is misleading IMHO - it is merely the documented wire format of a particular database. Hence why I feel it is more pragmatic to stick to the real server side peer implementation, rather than an abstract variant.

0x6e6562:
s/implantations/implementations/

Zariel:
We can move the discussion about the TestServer to the mailing list and sort that out, I would also like to in the future move the current tests which are really session tests into session_test.go

Outside of these issues are you ok with the changes?

0x6e6562:
I'm fine with pushing ahead with supporting v3, but I wanted to clarify the test server scope before doing a fine grained check of the code for any bugs we might be introducing.

The main issue I have is there is no coverage for this change in any of the integration tests. When we only have mock server coverage, this doesn't prove that much from a practical perspective.

Basically I'd like to reduce the likelihood that the first real test of the new code will be somebody trying it out in the wild and finding that it doesn't work with a real Cassandra instance.

Is there a relatively cheap way this can get incorporated into the integration tests?

Zariel:
We could change the test suite to run the integration tests for every protocol version the Cassandra version supports

0x6e6562:
OK, then I vote for adding v2 and v3 support to the test suite. v1 is not so interesting since we ultimately want to drop support for it (as soon as somebody completes a patch for it).

Zariel:
So only using the test server gives a false sense of security, though now the integration tests look more reliable (has anything changed?). The code now works and should be compliant with v3 of the spec, but without any of the extra features (except stream count), which I added tickets for.

Before adding these features I think the frame code needs a refactor.

0x6e6562:
> Before adding these features I think the frame code needs a refactor.

Are you saying that otherwise, you think this patch is good to go?

BTW - what would happen if we merged this patch before #309 lands?

0x6e6562:
The other significant question I'd like to clarify is the role of the mock server. Since we we've just had a case whereby the real server highlighted a short coming in the mock server, I'd just like to be clear about their value, given that it makes up a fair amount of code that we're going to end up having to maintain. There is also protocol v1 specific code in this patch, which to be seems questionable, given that we think we're going to be sunsetting support for this version in the not too distant future.

Zariel:
id be happy to merge this after and handle the rebase

Zariel:
#309 applies relatively cleanly over this, conflicts in the config and the AUTHORS file.

conn.go
```diff
@@ -312,6 +350,8 @@ func (c *Conn) exec(op operation, trace Tracer) (interface{}, error) {
 	if trace != nil {
 		req[1] |= flagTrace
 	}
+
+	headerSize := headerProtoSize[c.version]
```
0x6e6562: This is the 2nd usage of

```
headerSize := headerProtoSize[c.version]
```

in this function.

conn.go
```diff
@@ -254,47 +270,69 @@ func (c *Conn) serve() {
...
 
 func (c *Conn) execSimple(op operation) (interface{}, error) {
 	f, err := op.encodeFrame(c.version, nil)
-	f.setLength(len(f) - headerSize)
-	if _, err := c.conn.Write([]byte(f)); err != nil {
+	if err != nil {
+		// this should be a noop err
+		return nil, err
+	}
+	f.setLength(len(f)-headerProtoSize[c.version], c.version)
```
0x6e6562: It would be more readable (and hence idiomatic Go) if this was split up into 2 lines.

conn.go
Zariel: I cant see another?

Zariel:
im not sure whats causing the <v3 tests to fail, will investigate. Though it seems they pass if you run them individually.

Zariel:
@justinretailnext any ideas why the metadata tests dont pass on proto < 3?

Zariel:
locally they seem to fail on every proto as well, without any particularly obvious error messages, looks like somewhere we are not handling the protocol correctly along the query path as it doesnt do any thing directly with frames

justinretailnext:
I took a look at running the integration.sh script with just "1.2.19" and just looking at TestGetTableMetadata's result in particular. What the test failure for TestGetTableMetadata is indicating that one of the string values returned by the query:
```
SELECT
    columnfamily_name,
    key_validator,
    comparator,
    default_validator,
    key_aliases,
    column_aliases,
    value_alias
FROM system.schema_columnfamilies
WHERE keyspace_name = ?
```
Contains a 0x00 (i.e. string termination character) within the string's bytes. It may be that since TestGetTableMetadata queries for a large result set (more than a fewr rows), that it is triggering something in the new code in this patch that causes a 0x00 byte to find it's way into one of the strings in the results.

I've so far reduced the tests being run in the integration test which still reproduce the issue with the following parameter to the "go test" command:
```
-test.run="^(TestProto1BatchInsert|TestShouldPrepareFunction|TestCreateSessionTimeout|TestMapScanWithRefMap|TestSliceMap|TestScanWithNilArguments|TestRebindQueryInfo|TestStaticQueryInfo|TestBoundQueryInfo|TestMissingSchemaPrepare|TestReprepareStatement|TestQueryInfo|TestPreparedCacheEviction|TestPreparedCacheKey|TestMarshalFloat64Ptr|TestNilInQuery|TestEmptyTimestamp|TestGetTableMetadata)\$"
```

justinretailnext:
Okay, threw together a test that recreates the issue when run alone based on my hunch above:
```
func TestZeroInStrings(t *testing.T) {
	session := createSession(t)
	defer session.Close()

	if err := createTable(session, "CREATE TABLE test_zero_in_string (id int PRIMARY KEY, str text)"); err != nil {
		t.Fatalf("failed to create table with error '%v'", err)
	}

	rowsToCreate := 100
	strLenToCreate := 100

	// create a string of the form: 012345678911234567892123456789...
	str := ""
	for charCount := 0; charCount < strLenToCreate; charCount++ {
		num := charCount % 10
		if num == 0 {
			// indicate which ten's character
			num = (charCount / 10) % 10
		}
		str = str + strconv.Itoa(num)
	}

	for rowCount := 1; rowCount <= rowsToCreate; rowCount++ {
		if err := session.Query("INSERT INTO test_zero_in_string (id, str) VALUES (?,?)", rowCount, str).Exec(); err != nil {
			t.Fatalf("failed to insert with err: %v", err)
		}
	}

	iter := session.Query("SELECT * FROM test_zero_in_string").Iter()
	var id int
	var strBytes []byte
	var resultIndex = 0
	for iter.Scan(&id, &strBytes) {
		if id < 1 || id > rowsToCreate {
			t.Errorf("result[%d] id was expected to be in the range [1,%d] but was %d", resultIndex, rowsToCreate, id)
		}
		if len(strBytes) != strLenToCreate {
			t.Errorf("result[%d] strBytes was expected to have len %d but was %d", resultIndex, strLenToCreate, len(strBytes))
		}
		for chIndex, ch := range strBytes {
			if ch == 0 {
				t.Errorf("result[%d] strBytes[%d] was 0x00", resultIndex, chIndex)
			}
		}
		resultIndex++
	}

		t.Fatalf("failed to iterate with err: %v", err)
	}
}
```
Seems that all data decoded after a specific point is 0's since both last non-zero len string has 0's and the id column returned is zero

Zariel:
yeah i seem to be getting this,

`2015/03/12 17:27:12 conn.go:727: table=schema_columnfamilies col=comparator val="org.apache.cassandra.db.marshal.CompositeType(o\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"`

justinretailnext:
When I put rowsToCreate := 1 and strLenToCreate := 4096 into my test above, every byte over 4013 is a zero. So somewhere my guess is that there's a 4096 byte limit somewhere in the new protocol decoding.

justinretailnext:
Okay, got it. bufio.Reader has a default buffer size of 4096. Every time you call Read on bufio.Reader, it will at most return what it has left in it's buffer. So if your underlying reader might supply say 4100 bytes, and you call Read with a 4100 byte slice it the result will return n=4096 as the bytes read. The code then has to compare n to the number of bytes expected to read and try again until all bytes have been read. So in Conn.go, func recv, you need to change the for loop at line 307 to read until the sum of the returned var nn == bodySize.

justinretailnext:
```diff
diff --git a/conn.go b/conn.go
index e2b8ffa..ecd07f4 100644
--- a/conn.go
+++ b/conn.go
@@ -300,21 +300,26 @@ func (c *Conn) recv() (frame, error) {
        }
        resp.grow(bodySize)

-   const maxAttempts = 5
+ const maxRetryCount = 5

-   n := size
-   for i := 0; i < maxAttempts; i++ {
-           var nn int
+ respIndex := size
+ bytesToRead := bodySize
+ retryCount := 0
+ for retryCount < maxRetryCount && bytesToRead > 0 {
+         var bytesRead int
                c.conn.SetReadDeadline(time.Now().Add(c.timeout))
-           nn, err = c.Read(resp[n : size+bodySize])
-           if err == nil {
-                   break
-           }
-           n += nn
+         bytesRead, err = c.Read(resp[respIndex : respIndex+bytesToRead])
+         if err != nil {
+                 if verr, ok := err.(net.Error); !ok || !verr.Temporary() {
+                         break
+                 }

-           if verr, ok := err.(net.Error); !ok || !verr.Temporary() {
-                   break
+                 // try again
+                 retryCount++
+                 continue
                }
+         respIndex += bytesRead
+         bytesToRead -= bytesRead
        }

        if err != nil {
```

Zariel:
awesome, thanks!

Zariel:
@0x6e6562 ive tried to rekick off the builds, looks ok now but have created #324 along with a label to track the travis issues we encounter.

0x6e6562:
Nice work beating this into shape :-)

PR #319, opened 2015-03-07
lock mu when checking if the pool is closed

Zariel:
This patches a bug whereby a closed pool could be refilled during a race condition. It also makes access the the `quit` flag more consistent with other access paths.


0x6e6562:
Why do you need to lock on mu in this instance?

Zariel:
Because c.quit is accessed concurrently, for instance, 

https://github.com/Zariel/gocql/blob/edf935ca8ec50a2b1a35c454d1bffd60c994a1f0/connectionpool.go#L330

PR #320, opened 2015-03-07
extend the package docstring a bit

Zariel:


Zariel:
Do you think its a better idea to just use the relevant parts from the readme? That way it will show up on godoc.org as well.

Zariel:
Ill come back to this another time

PR #325, opened 2015-03-13
Default a Session's consistency to that of its ClusterConfig

obeattie:
I'm not 100% sure whether this is technically a bug or not (please close if not), but it seems to me that when constructing a `Session` with `NewSession`, its consistency level should be defaulted to that of its `ClusterConfig`.


0x6e6562:
This is a bug - good spot :-)

I've just re-triggered the CI build after a time out, but I don't think this has anything to do with the change.

0x6e6562:
Thank you for your contribution :-)

PR #327, opened 2015-03-13
conn: only setup the TLS config once

Zariel:
we can reuse the TLS Config between connections, so avoid the
expensive read and cert parsing when dialing connections, instead
do this in the connection pool once during startup.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -113,7 +113,10 @@ func (cfg *ClusterConfig) CreateSession() (*Session, error) {
 		cfg.NumStreams = maxStreams
 	}
 
-	pool := cfg.ConnPoolType(cfg)
+	pool, err := cfg.ConnPoolType(cfg)
+	if err != nil {
+		return nil, err
+	}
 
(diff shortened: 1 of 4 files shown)


Zariel:
Ive never seen this error in travis before, race condition?

`	cassandra_test.go:463: '`SELECT * FROM too_few_query_args WHERE id = ? and cluster = ?`, 1' should return an ErrQueryArgLength, but returned: unconfigured columnfamily not_enough_query_args
`

Zariel:
@0x6e6562 think its worth continuing with this change?

Zariel:
My hesitation for this change is that ConnConfig is exported so changing its fields are breaking API changes, but im not hugely concerned about the impact because I dont think we should have exported the connection related things (Connect, Conn, ConnConfig).

0x6e6562:
I take your point about not exporting connection related stuff - so given that this patch introduces a breaking change, why not down case the new struct member to completely unexport the field?

Other than that suggestion, this change LGTM.

Zariel:
@0x6e6562 good idea! done!

PR #331, opened 2015-03-13
Surface ErrNotFound Errors on Iterator

johannac:
When scanning for rows on an `*Iter`, if you get an `ErrNotFound` error, it is not returned or otherwise available in the interface. This patch fixes that problem, so when you call `Iter.Scan()` and get `ErrNotFound` it will return `false` to stop any loop, and store the error in `iter.err` for retrieval later with `Iter.Close()`. This includes helper functions like `Iter.MapScan()` and `Iter.SliceMap()` which make use of `Iter.Scan()`

*Note:* This kind of changes the previous functionality of the iterator. It doesn't technically break the API, as the types are all the same, but it will return an `ErrNotFound` where previously only critical/connection errors will have been returned, so some users might not be handling this kind of error.


Zariel:
This change makes sense and brings it more inline with database/sql which does the same thing. Though this change will break existing clients who expect to get `ErrNotFound` from `Scan()`. What do you think @0x6e6562 @phillipCouto

johannac:
`Query.Scan()` should still return `ErrNotFound` and any other errors like usual. I think the biggest impact will be to clients who call `Iter.Close()` and then check `if err != nil {//return critical failure}`, since this patch makes it also return `ErrNotFound` errors as well, which is not always critical. But based on the rest of the interface, it seemed to me like this should be expected behaviour, just seemed to be missing?

0x6e6562:
Notionally I'm fine with changing the `Scan()` semantics in this way. But I think the change should be accompanied by a test case that verifies the intended behavior of `ErrNotFound`. It would be even better if that test also explicitly tests for a critical error and differentiates itself from `ErrNotFound`, such that the semantics are clearly defined.

We should also put a note into the README about the change in the behavior.

Why is the Travis build for this patch failing BTW?

Zariel:
`--- FAIL: TestPaging (3.18 seconds)
	cassandra_test.go:240: close: not found`

The tests need updating to change the expected behavior.

Slightly off topic, i would have thought it makes more sense for the query API to be more like the database/sql api which has a Next()/Scan()/Close() methods, though it also has an Error() which I dont think we need.

0x6e6562:
Originally @tux21b prototyped the library based on database/sql but found that particular abstraction to be not such a great fit for CQL.

Zariel:
yeah, though I always found it odd that you had to allocate something to scan it even if there were no rows, which is avoided by using Next() then Scan().

Zariel:
Ive had a think about this and on reflection im not convinced this is a particularly useful change which will have a large breaking API impact, imagine if clients currently do this

``go
iter := session.Query("SELECT id FROM table").Iter()

for {
    if err := iter.Scan(&id); err != nil {
        if err == gocql.ErrNoRows {
             return nil, nil
        }
        return nil, err
    }
}

    return nil, err
}
```

Then this will end up breaking lots of existing code. Im not against making breaking API changes, but this seems like it works well enough currently, or am I missing a use case or pattern that makes it more required?

Like I said earlier I would be more in favour of Next/Scan/Close but thats rather out of scope for this.

Zariel:
@johannac sorry ignore my previous comment, got mildly confused :)

This change looks good, could you rebase this branch against master and update the failing test. This looks like a bug in that we wont also return a ErrNotFound from Exec() when it looks like we should.

@0x6e6562 do you think we should update the README for this? It makes the API more consistent, maybe just docs change?

johannac:
Thanks all for feedback! I am travelling at the moment, but will pick this back up in a week or two when I'm back, and will incorporate the changes and tests etc.

0x6e6562:
Yes, I think the the README should be updated to reflect the current API, especially if we are going to introduce a change in the behavior.

0x6e6562:
@johannac Is there likely to be any movement on this patch?

johannac:
Yes, sorry for the delay, I'll try to get this resolved this week.

0x6e6562:
No worries, there's no hurry - I just wanted to make sure it was still in the pipeline.

@Zariel As and when this is in a mergeable state, I'd like to road test the patch against [cqlc](https://github.com/relops/cqlc) since it is a heavy user of this API and this will break that dependency, so I'd like to be in position to supply a timely update.

Zariel:
Im going to close this as it hasnt been updated in over 6 months

azamikramullah:
Circling back on this, was this ever resolved? We are seeing this issue on our side

PR #332, opened 2015-03-16
Wait until we're done connecting before returning from fillPool

mdelah:
The changeset adds a mutex to the fillPool() function, designed to prevent multiple goroutines from repopulating the connection pool at once. At my organisation we've noticed gocql client programs would regularly open up tens or hundreds of times the configured number of connections to Cassandra. It seems that occassional network outages were causing many connections to be dropped from the pool at once, and each of the N goroutines responsible for the lost connections would race to re-open the full N, resulting in N * N new connections. We've been running this changeset in production for two weeks now, and the issue has disappeared. Note the existing mutex (c.hostMu) was ineffective at preventing this condition, because the connections are reopened in goroutines that continue to run after fillPool() returns.


0x6e6562:
Very interesting - thanks for contributing this patch. I take your point about the existing mutex and I can see how you've patched it to get a more deterministic read of the current pool size. It seems that you've introduced a semaphore pattern into the pool. I was wondering whether there is a more idiomatic way to do this in Go?

In any case I would to get @Zariel to comment on the concurrency implications of this change.

Zariel:
I dont think this will have an impact on the concurrency of the pool. Though this could have been done more simply with a `WaitGroup`?

mdelah:
Thanks @0x6e6562, @Zariel. Great suggestion to use WaitGroup; I've pushed the change to our fork.

connectionpool.go
```diff
@@ -251,7 +252,9 @@ func (c *SimplePool) fillPool() {
 
 		//This is reached if the host is responsive and needs more connections
 		//Create connections for host synchronously to mitigate flooding the host.
+		wg.Add(1)
 		go func(a string, conns int) {
+			defer func() { wg.Done() }()
```
Zariel: nit: this can be written as ```defer wg.Done()``` without needing the closure

Zariel:
LGTM @0x6e6562

0x6e6562:
LGTM as well. The only thing that needs to be done is to add yourself to the AUTHORS file. If you want to take the suggestion about the defer function syntax on board as well, that's fine as well. Otherwise, we should be good to merge this.

mdelah:
Updated. Cheers, guys.

0x6e6562:
This doesn't appear to merge cleanly - it is worth re-merging master into your branch?

0x6e6562:
Sometimes this due to line endings in the AUTHORS file.

mdelah:
Done @0x6e6562

0x6e6562:
Thanks very much for your contribution :-)

PR #333, opened 2015-03-18
Eagerly infer the last component of a column family

0x6e6562:
This patch adds some more metadata to the schema at compile time. The motivation behind this is to be able to infer the relative position of a column that forms either part of the partition key or clustering key. This can already be achieved using the current API, but that information needs to be computed on a per-column basis multiple times if this processing takes place in a client application. By moving this inference into the compilation phase, this computation is only done once. It also makes the metadata API richer.

### DIFF
--- a/metadata.go
+++ b/metadata.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -36,6 +37,7 @@ type TableMetadata struct {
 	PartitionKey      []*ColumnMetadata
 	ClusteringColumns []*ColumnMetadata
 	Columns           map[string]*ColumnMetadata
+	LastComponent     *ColumnMetadata
 }
 
 // schema metadata for a column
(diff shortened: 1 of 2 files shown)


0x6e6562:
Note to @Zariel : I've decided that I prefer the attribute variant, rather leaving this metadata at the level of the table (as it was originally in 063103e) - so 995f338 should be good to merge from my perspective.

justinretailnext:
I'm not sure I understand the use of the last component bool. Both PartitionKey and ClusteringColumns are meant to be [ordered by their ComponnentIndex](https://github.com/gocql/gocql/blob/master/metadata.go#L302) so the last component of either is simply the last index of either array:
```
lastPartitionKeyColumn := tableMetadata.PartitionKey[len(tableMetadata.PartitionKey)-1]
lastClusteringColumn := tableMetadata.ClusteringColumn[len(tableMetadata.ClusteringColumn)-1]
```
Possibly this wasn't clear from my PR. How would this boolean flag be used?

0x6e6562:
Interesting point - is this order guaranteed when the PartitionKey and ClusteringColumns are populated from the underlying schema metadata (i.e. does Cassandra guarantee that the columns are returned in component index order)?

I'm not sure that this was always the case in previous versions of Cassandra, but if it is currently an order we can rely on, then this change is unnecessary (since we can just your suggestion).

0x6e6562:
I forgot to answer the 2nd bit of your question: The usage of this flag is to determine when a column can used for WHERE IN clauses.

justinretailnext:
Cassandra proto V2+ does provide the component index in the schema metadata, which is how the compile code reconstructs the size and ordering of partition key and clustering columns.

The column metadata for protocol V1 is completely inferred from column_aliases and key_aliases values which are expected to be an ordered list of column names.

Zariel:
@0x6e6562 do we still need this change then? I dont know much about how Cassandra is storing its schema metadata.

0x6e6562:
It's still on my todo list - I haven't yet had the time to make a decision yet.

0x6e6562:
On reflection, I've decided that this change is not necessary, because the metadata is now available in a correctly ordered form (i.e. order by component index), as @justinretailnext has helpfully pointed out.

So I'm closing this pull request.

PR #338, opened 2015-03-27
Added routing key computation to Query

justinretailnext:
This PR adds some incremental functionality which would be needed for token awareness in a connection pool. It depends on the metadata functionality from my previous PR. Since gocql doesn't differentiate between prepared and unprepared statements, this functionality is a departure in some respects from the Datastax driver methodology, which it is loosely based on.

The top-level functionality added in this PR is the addition of the `GetRoutingKey() []byte` function to `Query`. This function returns the computed routing key for a given statement and its bound values. The function `RoutingKey([]byte)` is supplied in `Query` as a way to override the routing key, or as a simple short circuit to avoid the computation of the routing key (e.g. cqlc could probably generate code that creates appropriate routing keys without relying on the runtime computation of the partition key metadata). Computation of the routing key during runtime requires determining which bound parameters of a statement are part of the partition key. This is done in the `routingKeyInfo() *routingKeyInfo` function of `Session` which utilizes both the cached `QueryInfo` for a statement combined with the `TableMetadata`. The function `routingKeyInfo` is cached so that the routing key indexes and type information is only collected once for each unique statement.

Some things to note in this PR are the slight modification to queries in metadata.go which set the routing key to an empty byte slice:
```
query.RoutingKey([]byte{})
```
This disables computation of the routing key since the routing key computation calls the metadata queries thus avoiding an infinite loop. In the Datastax drivers the metadata queries are performed on a uniquely configured "control" connection which does not use the routing key, thus avoiding this workaround.

Another item to notice is that the routing key info cache is owned by the session since this information isn't bound to a specific connection unlike the prepared statement cache.

[Review on Reviewable](https://reviewable.io/reviews/gocql/gocql/338)

### DIFF
--- a/metadata.go
+++ b/metadata.go
@@ -331,6 +331,9 @@ func getKeyspaceMetadata(
 		`,
 		keyspaceName,
 	)
+	// Set a routing key to avoid GetRoutingKey from computing the routing key
+	// TODO use a separate connection (pool) for system keyspace queries.
+	query.RoutingKey([]byte{})
 
 	keyspace := &KeyspaceMetadata{Name: keyspaceName}
 	var strategyOptionsJSON []byte
@@ -375,6 +378,9 @@ func getTableMetadata(
 		`,
 		keyspaceName,
 	)
+	// Set a routing key to avoid GetRoutingKey from computing the routing key
+	// TODO use a separate connection (pool) for system keyspace queries.
+	query.RoutingKey([]byte{})
 	iter := query.Iter()
 
 	tables := []TableMetadata{}
@@ -511,6 +517,9 @@ func getColumnMetadata(
 	var indexOptionsJSON []byte
 
 	query := session.Query(stmt, keyspaceName)
+	// Set a routing key to avoid GetRoutingKey from computing the routing key
+	// TODO use a separate connection (pool) for system keyspace queries.
+	query.RoutingKey([]byte{})
 	iter := query.Iter()
 
 	for scan(iter, &column, &indexOptionsJSON) {
--- a/session.go
+++ b/session.go
@@ -5,6 +5,8 @@
 package gocql
 
 import (
+	"bytes"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
@@ -12,6 +14,8 @@ import (
 	"sync"
 	"time"
 	"unicode"
+
+	"github.com/golang/groupcache/lru"
 )
 
 // Session is the interface used by users to interact with the database.
@@ -24,13 +28,14 @@ import (
 // and automatically sets a default consinstency level on all operations
 // that do not have a consistency level set.
 type Session struct {
-	Pool            ConnectionPool
-	cons            Consistency
-	pageSize        int
-	prefetch        float64
-	schemaDescriber *schemaDescriber
-	trace           Tracer
-	mu              sync.RWMutex
+	Pool                ConnectionPool
+	cons                Consistency
+	pageSize            int
+	prefetch            float64
+	routingKeyInfoCache routingKeyInfoLRU
+	schemaDescriber     *schemaDescriber
+	trace               Tracer
+	mu                  sync.RWMutex
 
 	cfg ClusterConfig
 
@@ -40,7 +45,12 @@ type Session struct {
 
 // NewSession wraps an existing Node.
 func NewSession(p ConnectionPool, c ClusterConfig) *Session {
-	return &Session{Pool: p, cons: c.Consistency, prefetch: 0.25, cfg: c}
+	session := &Session{Pool: p, cons: c.Consistency, prefetch: 0.25, cfg: c}
+
+	// create the query info cache
+	session.routingKeyInfoCache.lru = lru.New(c.MaxRoutingKeyInfo)
+
+	return session
 }
 
 // SetConsistency sets the default consistency level for this session. This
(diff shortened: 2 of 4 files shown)


Zariel:
As a note once my refactor of the framing code lands I plan on working on adding a control connection, which should help with this and various other features for the connection pool.

Thanks for the PR, I will take a look :)

session.go
```diff
@@ -185,6 +195,104 @@ func (s *Session) KeyspaceMetadata(keyspace string) (*KeyspaceMetadata, error) {
...
+			table := queryInfo.Args[0].Table
+			var keyspaceMetadata *KeyspaceMetadata
+			keyspaceMetadata, inflight.err = s.KeyspaceMetadata(s.cfg.Keyspace)
+			if inflight.err == nil {
+				tableMetadata, found := keyspaceMetadata.Tables[table]
+				if !found {
+					inflight.err = ErrNoMetadata
+				}
+
+				partitionKey = tableMetadata.PartitionKey
```
Zariel: If found is false this will nil pointer

session.go
```diff
@@ -185,6 +195,104 @@ func (s *Session) KeyspaceMetadata(keyspace string) (*KeyspaceMetadata, error) {
...
+	s.routingKeyInfoCache.lru.Add(cacheKey, inflight)
+	s.routingKeyInfoCache.mu.Unlock()
+
+	var queryInfo *QueryInfo
+	var partitionKey []*ColumnMetadata
+
+	// get the query info for the statement
+	conn := s.Pool.Pick(nil)
+	if conn != nil {
+		queryInfo, inflight.err = conn.prepareStatement(stmt, s.trace)
```
Zariel: here should we not pass a nil tracer as it has no user impact?

Zariel:
lgmt baring comments, my only concern is that `routingKeyInfo` goes to the right quite a lot, would it be possible to split it out into separate functions?

justinretailnext:
Concerning the line length, looking at the code I may be able to even rearrange some of the branches so that the main non-error code branch is at the first indent. Otherwise I'll format or split it.

Zariel:
Yeah that was my concern, particularly just the braching around line 232

0x6e6562:
Overall this looks good to me.

My main question would be whether it is good for the new API functions to not return errors, i.e.:

```
func (q *Query) RoutingKey(routingKey []byte) *Query
func (q *Query) GetRoutingKey() []byte
```

instead of 

```
func (q *Query) RoutingKey(routingKey []byte) (*Query, error)
func (q *Query) GetRoutingKey() ([]byte, error)
```

As it stands right now, it is up to the caller to test for nil. Also, the underlying error will not be available to the caller. So this is a slightly different to a lot of the APIs we currently have, but there might be a good reason for this?

justinretailnext:
There's no error right now primarily because the code for token awareness didn't need to act on any error  differently than if there was no routing key. I do agree that this is an omission in general use and I've added the error to the signature. The code still can return a nil routing key for a non-error condition where the routing key can't be determined, and I've documented that.

justinretailnext:
Also, `RoutingKey([]byte) *Query` does not return an error since it cannot produce an error; this also makes sense since it follows with the other fluent style value setters such as `PageSize(int) *Query`.

metadata.go
```diff
@@ -331,6 +331,9 @@ func getKeyspaceMetadata(
 		`,
 		keyspaceName,
 	)
+	// Set a routing key to avoid GetRoutingKey from computing the routing key
+	// TODO use a separate connection (pool) for system keyspace queries.
+	query.RoutingKey([]byte{})
```
Zariel: should this be set for the RingDescriber as well?

justinretailnext: It's only necessary here since these methods are in the call path of GetRoutingKey.

More specifically, each of these get***Metadata functions execute a query using the connection pool, which may be implemented with token-awareness, which will in turn call GetRoutingKey, and therefore would infinitely loop recursively; except that by overriding the routing key here in this line, this short circuits the behavior of GetRoutingKey to return the override value immediately and avoid the recursion.

The ideal situation, as indicated by the TODO is that these functions execute their queries on an isolated connection (e.g. control connection) that wouldn't ever call GetRoutingKey.

Zariel: I see. I have ideas to replace the pluggable connection pool with pluggable host selectors for a query, so that should be replaceable on a per query basis.

justinretailnext: My token aware connection pool implementation (future PR) actually introduces a HostSelectionPolicy and ConnSelectionPolicy interfaces with a new policy driven connection pool. When we get to that I figure we can hash out any alterations to that.

session.go
```diff
@@ -610,6 +787,40 @@ type ColumnInfo struct {
...
+		r.lru.RemoveOldest()
+	}
+	r.lru.MaxEntries = max
+	r.mu.Unlock()
+}
+
+type inflightCachedEntry struct {
+	wg    sync.WaitGroup
+	err   error
+	value interface{}
```
Zariel: Is there a reason to store the value as an `interface{}` not a concrete type?

justinretailnext: This is mostly a copy of code from the prepared statement cache code. I suppose in the long term it would make sense to consolidate both cache code bases into a single abstract inflight cache.

Zariel:
Thanks for taking the time to contribute this :)

PR #339, opened 2015-03-30
refactor all framing protocol code into frame

Zariel:
Simplify the protocol implementation so that all the logic for handling different protocol versions is now handled inside frame.go. This should make it easier to improve and change the protocols going forward without having to make lots of changes elsewhere.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -177,6 +177,15 @@ func Connect(addr string, cfg ConnConfig, pool ConnectionPool) (*Conn, error) {
 	return c, nil
 }
 
+func (c *Conn) Write(p []byte) (int, error) {
+	c.conn.SetWriteDeadline(time.Now().Add(c.timeout))
+	return c.conn.Write(p)
+}
+
+func (c *Conn) Read(p []byte) (int, error) {
+	return c.r.Read(p)
+}
+
 func (c *Conn) startup(cfg *ConnConfig) error {
 	m := map[string]string{
 		"CQL_VERSION": cfg.CQLVersion,
@@ -242,6 +251,96 @@ func (c *Conn) authenticateHandshake(authFrame *authenticateFrame) error {
 	}
 }
 
+// Serve starts the stream multiplexer for this connection, which is required
+// to execute any queries. This method runs as long as the connection is
+// open and is therefore usually called in a separate goroutine.
+func (c *Conn) serve() {
+	var (
+		err    error
+		header frameHeader
+	)
+
+	for {
+		header, err = c.recv()
+		if err != nil {
+			break
+		}
+		c.dispatch(header)
+	}
+
+	c.Close()
+	for id := 0; id < len(c.calls); id++ {
+		req := &c.calls[id]
+		if atomic.CompareAndSwapInt32(&req.active, 1, 0) {
+			req.resp <- struct{}{}
+			close(req.resp)
+		}
+	}
+
+	if c.started {
+		c.pool.HandleError(c, err, true)
+	}
+}
+
+func (c *Conn) recv() (frameHeader, error) {
+	// not safe for concurrent reads
+
+	// read a full header, ignore timeouts, as this is being ran in a loop
+	// TODO: TCP level deadlines? or just query level deadlines?
+
+	// were just reading headers over and over and copy bodies
+	head, err := readHeader(c.r, c.headerBuf)
+	if err != nil {
+		return frameHeader{}, err
+	}
+
+	call := &c.calls[head.stream]
+
+	call.mu.Lock()
+	log.Printf("readframe stream=%v\n", head.stream)
+	err = call.framer.readFrame(&head)
+	call.mu.Unlock()
+	if err != nil {
+		return frameHeader{}, err
+	}
+
+	if head.version.version() != c.version {
+		return frameHeader{}, NewErrProtocol("unexpected protocol version in response: got %d expected %d", head.version.version(), c.version)
+	}
+
+	return head, nil
+}
+
+func (c *Conn) dispatch(header frameHeader) {
+	id := header.stream
+	if id >= len(c.calls) {
+		// should this panic?
+		return
+	}
+
+	// TODO: replace this with a sparse map
+	call := &c.calls[id]
+
+	call.resp <- struct{}{}
+	atomic.AddInt32(&c.nwait, -1)
+
+	c.uniq <- id
+}
+
+type callReq struct {
+	active int32
+	// could use a waitgroup but this allows us to do timeouts on the read/send
+	resp   chan struct{}
+	mu     sync.Mutex
+	framer *framer
+	err    error
+}
+
+type callResp struct {
+	framer *framer
+	err    error
+}
+
 func (c *Conn) exec(req frameWriter, tracer Tracer) (frame, error) {
 	// TODO: move tracer onto conn
 	stream := <-c.uniq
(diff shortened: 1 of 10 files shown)


Zariel:
@0x6e6562 this is a fairly large change, and took me a while :)

This is a major refactor of how the framing code (frame.go) and connection (conn.go) are interacting, it removes all knowledge of the protocol from the connection and changes how it reads / write frames to the underlying connection.

Im not sure how relevant they are as I didnt spend anytime on optimising anything, but the bench results are interesting.

`benchmark                           old ns/op     new ns/op     delta
BenchmarkWikiCreateSchema-4         184891883     157283266     -14.93%
BenchmarkWikiCreatePages-4          638084        771324        +20.88%
BenchmarkWikiSelectAllPages-4       1624179       2544581       +56.67%
BenchmarkWikiSelectSinglePage-4     315850        408556        +29.35%
BenchmarkWikiSelectPageCount-4      527972        531931        +0.75%

benchmark                           old allocs     new allocs     delta
BenchmarkWikiCreateSchema-4         32             29             -9.38%
BenchmarkWikiCreatePages-4          37             32             -13.51%
BenchmarkWikiSelectAllPages-4       446            1142           +156.05%
BenchmarkWikiSelectSinglePage-4     55             60             +9.09%
BenchmarkWikiSelectPageCount-4      27             25             -7.41%

benchmark                           old bytes     new bytes     delta
BenchmarkWikiCreateSchema-4         11393         1982          -82.60%
BenchmarkWikiCreatePages-4          6389          1699          -73.41%
BenchmarkWikiSelectAllPages-4       50782         50292         -0.96%
BenchmarkWikiSelectSinglePage-4     7509          3243          -56.81%
BenchmarkWikiSelectPageCount-4      5748          1021          -82.24%`

0x6e6562:
Is this change ready for a review or are you still working on it?

conn.go
```diff
@@ -179,66 +185,113 @@ func Connect(addr string, cfg ConnConfig, pool ConnectionPool) (*Conn, error) {
...
 
+	go c.serve()
+
 	if err := c.startup(&cfg); err != nil {
 		conn.Close()
 		return nil, err
 	}
-
-	go c.serve()
+	c.started = true
```
0x6e6562: Just as a point of interest, do we know that the connection has actually been started at this point in time, since `c.serve()` is executed asynchronously? Or is this irrelevant?

conn.go
```diff
@@ -248,170 +301,121 @@ func (c *Conn) startup(cfg *ConnConfig) error {
...
+	head, err := readHeader(c.r, c.headerBuf)
+	if err != nil {
+		return err
 	}
 
-	bodySize := resp.Length(c.version)
-	if bodySize == 0 {
-		return resp, nil
+	call := &c.calls[head.stream]
+	call.mu.Lock()
```
0x6e6562: What's the scope of this lock, given that we state above that this is "not safe for concurrent reads"?

conn.go
```diff
@@ -248,170 +301,121 @@ func (c *Conn) startup(cfg *ConnConfig) error {
...
+		framer.trace()
 	}
 
-	// here recv wont timeout waiting for a header, should it?
-	if f, err = c.recv(); err != nil {
+	// there is a race that we can read and write to the same buffer, I dont think
+	// the data will actually corrupt but to be safe and appease the race detector gods,
+	// guard it.
+	// We could fix this by using seperate read and write buffers, which may end up
+	// being faster and easier to reason about.
```
0x6e6562: The idea about separate buffers sounds good on face value - would there be much involved in splitting this out?

frame.go
```diff
@@ -16,602 +19,1302 @@ const (
...
+	opExecute               = 0x0A
+	opRegister              = 0x0B
+	opEvent                 = 0x0C
+	opBatch                 = 0x0D
+	opAuthChallenge         = 0x0E
+	opAuthResponse          = 0x0F
+	opAuthSuccess           = 0x10
+)
+
+func (f frameOp) String() string {
```
0x6e6562: If we were only supporting Go 1.4, I guess we could auto-generate these constants:

//go:generate stringer -type=frameOp

0x6e6562:
I've had a good look through this change and I've noted the things that may require some follow up - other than that the patch looks pretty good. I can't help thinking that because of the size of this change, there is likely to be subtle bugs we've both missed. I guess end users are likely to tease these out. Because of this, I'm wondering if we temporarily change the top of the README to reflect the fact that we are trying to mature out new framing code and that if people require more tried and tested code, that we provide a tag to the last commit of the previous generation framing code?

Zariel:
Yeah, I think we should wait for #316 to merge so that we can verify the authentication handshake works correctly.

With regards to a tag, that makes sense.

frame.go
Zariel: yeah that would be nice, many of the types could have auto-gen strings

conn.go
Zariel: Not really, just the change on read and write paths, currently they just share the same buffer and assume the owner of framer is not reading/writing concurrently, as it should be used for a single stream on a connection.

conn.go
Zariel: This lock protects the buffer in framer so that we guarantee that we dont read/write concurrently, this is race condition mentioned in the framer

Zariel:
Given that #316 is a bit blocked, I ran the tests manually setting the authenticator, found a bug in the type switch for the authentication. Added some more stringing functions to the frames as well.

I think its in a state to merge now @0x6e6562

Zariel:
@0x6e6562 I made some changes, namely revert back to copying byte slices when parsing frames, which makes the code much easier to reason about. I also changed it to use separate read/write buffers which also makes it easier to reason about.

conn.go
Zariel: This is to prevent errors going back to the connection pool from startup, which is the same behaviour as the previous implementation.

0x6e6562:
I'm fine to leap frog #316 to get this going and code wise this LGTM, but I think we should put a highly visible addendum at the top of the README (which we will delete after this code has been tested out enough in the wild), so that people will be able to see that they're dealing with new code. We should also provide a tag to the previous "stable" revision.

Zariel:
What shall we call the tag?

0x6e6562:
>What shall we call the tag?

How about `1st_gen_framing`?

Zariel:
created

Zariel:
For the readme, how about this under the website/docs link

Stability
---------
The underlying framing code was rewritten as part of #339 and as such may have
unforseen bugs, if you would like to use the previously known stable framing code the tag `1st_gen_framing` (180456fef0a3c6d02c51dc7211f49b55e9315867) has
been created.

0x6e6562:
How about:

"The underlying framing code was rewritten as part of #339 and as such may have
unforseen bugs. If you run into a bug related to wire framing, please raise a ticket and we will try to resolve this as soon as we can. If you require a stable version to pin your production app against, we have tagged the previous stable version in source code, so you can build against this. The tag is called 1st_gen_framing (180456f). This note will be removed as the new generation framing code base matures."

0x6e6562:
I might also change the section title to "Production Stability"?

Zariel:
Committed

Zariel:
@0x6e6562 merge?

0x6e6562:
Very nice work - this has laid a flexible foundation for the library going forwards :-)

Zariel:
Thanks :+1:

PR #340, opened 2015-04-01
Token partitioners and a token ring

justinretailnext:
This PR adds implementations of token partitioners for the 3 cassandra partitioning strategies as well as a token ring implementation. These implementations are mostly based on the DataStax cassandra drivers, with some references to the cassandra codebase itself. My original Murmur3Token implementation depended on github.com/spaolacci/murmur3 package, though since the DataStax driver includes the murmur3 algorithm in its implementation, I though it might be simpler to include the murmur3 implementation here as well and avoid the dependency ; the murmur3 code is based on [MurmurHash3_x64_128 code in the C++ reference implementation](https://code.google.com/p/smhasher/source/browse/trunk/MurmurHash3.cpp#255) along with the [unsafe getblock optimization from the spaolacci murmur3 code](https://github.com/spaolacci/murmur3/blob/master/murmur128.go#L61).

For the murmur3 tests I [copied the murmur3 code from the Java DataStax implementation into a JUnit test](https://gist.github.com/justinretailnext/1315cca9361fcdeea979) to generate the expected values for the data values I tested.

[Review on Reviewable](https://reviewable.io/reviews/gocql/gocql/340)

### DIFF
--- a/token.go
+++ b/token.go
@@ -1,0 +1,315 @@
+// Copyright (c) 2015 The gocql Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package gocql
+
+import (
+	"bytes"
+	"crypto/md5"
+	"fmt"
+	"math/big"
+	"sort"
+	"strconv"
+	"strings"
+	"unsafe"
+)
+
+// a token partitioner
+type Partitioner interface {
+	Hash([]byte) Token
+	ParseString(string) Token
+}
+
+// a token
+type Token interface {
+	fmt.Stringer
+	Less(Token) bool
+}
+
+// murmur3 partitioner and token
+type Murmur3Partitioner struct{}
+type Murmur3Token int64
+
+func (p Murmur3Partitioner) Hash(partitionKey []byte) Token {
+	h1 := murmur3H1(partitionKey)
+	return Murmur3Token(int64(h1))
+}
+
+// murmur3 little-endian, 128-bit hash, but returns only h1
+func murmur3H1(data []byte) uint64 {
+	length := len(data)
+
+	var h1, h2, k1, k2 uint64
+
+	const c1 = 0x87c37b91114253d5
+	const c2 = 0x4cf5ad432745937f
+
+	// body
+	nBlocks := length / 16
+	for i := 0; i < nBlocks; i++ {
+		block := (*[2]uint64)(unsafe.Pointer(&data[i*16]))
+
+		k1 = block[0]
+		k2 = block[1]
+
+		k1 *= c1
+		k1 = (k1 << 31) | (k1 >> 33) // ROTL64(k1, 31)
+		k1 *= c2
+		h1 ^= k1
+
+		h1 = (h1 << 27) | (h1 >> 37) // ROTL64(h1, 27)
+		h1 += h2
+		h1 = h1*5 + 0x52dce729
+
+		k2 *= c1
+		k2 = (k2 << 33) | (k2 >> 31) // ROTL64(k2, 33)
+		k2 *= c2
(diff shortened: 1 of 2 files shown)


justinretailnext:
I'm going to add documentation to the exported types and funcs, and see if I can liven up the tests a bit more.

token.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+func (o OrderPreservingToken) Less(token Token) bool {
+	return -1 == bytes.Compare(o, token.(OrderPreservingToken))
+}
+
+// random partitioner and token
+type RandomPartitioner struct{}
+type RandomToken struct {
+	*big.Int
```
Zariel: can we avoid embedding this type?

token_test.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+}
+
+func BenchmarkMurmur3H1(b *testing.B) {
+	var h1 uint64
+	var data [1024]byte
+	for i := 0; i < 1024; i++ {
+		data[i] = byte(i)
+	}
+	for i := 0; i < b.N; i++ {
```
Zariel: should have a reset timer call before this loop

token.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+// a data structure for organizing the relationship between tokens and hosts
+type TokenRing struct {
+	partitioner Partitioner
+	tokens      []Token
+	hosts       []*HostInfo
+}
+
+func NewTokenRing(partitioner string, hosts []*HostInfo) (*TokenRing, error) {
```
Zariel: Should the TokenRing have methods to change the hosts? What happens if it becomes out of sync with the connection pool?

token.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+// murmur3 little-endian, 128-bit hash, but returns only h1
+func murmur3H1(data []byte) uint64 {
+	length := len(data)
+
+	var h1, h2, k1, k2 uint64
+
+	const c1 = 0x87c37b91114253d5
+	const c2 = 0x4cf5ad432745937f
```
Zariel: should use a const block here

token.go
justinretailnext: The TokenRing struct is meant to be replaced instead of updated. This is something that should be documented.

Zariel: I see, so the TokenRing will belong to a session, and when the hosts change the token ring will be replaced?

justinretailnext: It'll be owned by the TokenAwareHostSelectionPolicy and when the policy get's updates from the connection pool when the host info changes, then it will replace its token ring.

0x6e6562:
Does the spoalacci implementation have a test suite at all that we can feasibly use to further verify this patch (a.k.a. pirate testing)?

token.go
Zariel: In that case we we make the methods to create, and the types for the TokenRing not exported. My reasoning is that as this is not used internally yet, I would rather not expose the API until it is used internally, and also it allows us to make more sweeping changes to the API if it is not exposed.

0x6e6562:
Not exporting the API is a good idea if there is no initial need to use it externally - it could always be exported at a later point.

token_test.go
```diff
@@ -0,0 +1,189 @@
...
+package gocql
+
+import (
+	"math/big"
+	"strconv"
+	"testing"
+)
+
+func TestMurmur3H1(t *testing.T) {
+	// assertMurmur3H1(t, []byte{}, 0x000000000000000)
```
Zariel: Why were these tests left commented out? Is that intentional?

justinretailnext: oops

Zariel:
Thanks for contributing this!

PR #348, opened 2015-04-07
Policy based ConnectionPool with token-aware, and round-robin policy implementation

justinretailnext:
Based on the work of PRs #340, #338, and #309, this PR adds a new ConnectionPool implementation which uses replaceable policy implementations of new interfaces HostSelectionPolicy and ConnSelectionPolicy. 
```
//HostSelectionPolicy is an interface for selecting
//the most appropriate host to execute a given query.
type HostSelectionPolicy interface {
	//SetHosts notifies this policy of the current hosts in the cluster
	SetHosts(hosts []HostInfo)
	//SetPartitioner notifies this policy of the current token partitioner
	SetPartitioner(partitioner string)
	//Pick returns an iteration function over selected hosts
	Pick(*Query) NextHost
}

//NextHost is an iteration function over picked hosts
type NextHost func() *HostInfo

//ConnSelectionPolicy is an interface for selecting an
//appropriate connection for executing a query
type ConnSelectionPolicy interface {
	SetConns(conns []*Conn)
	Pick(*Query) *Conn
}
```

The new connection pool (see type policyConnPool) is implemented as a pool of hosts which each have a pool of connections and makes a clearer distinction between selecting a host for a query and selecting a connection for the selected hosts. This allows for the implementation of the selection policies for host selection and connection selection to be implemented specifically in terms of hosts or connections while allowing the polices to still be composible; for example, the provided NewTokenAwareConnPool "ConnPoolType" is a composition of a token aware host selection policy, which falls back to a round-robin host selection policy when a routing key cannot be determined, combined with the trivial round-robin connection selection policy:

```
func NewTokenAwareConnPool(cfg *ClusterConfig) ConnectionPool {
	return newPolicyConnPool(
		cfg,
		NewTokenAwareHostPolicy(NewRoundRobinHostPolicy()),
		NewRoundRobinConnPolicy,
	)
}
```
Host selection also (in comparison to connection selection) returns an iterator over hosts allowing for say the token aware policy to return first the token replicas, followed by a shuffled round-robin.

The SimplePool ConnectionPool has been left untouched since it is mature code, however the new pool implementation can replace SimplePool (see NewRoundRobinConnPool in connectionpool.go)

I've also added the connectionpool_systems_test.sh and connectionpool_systems_test.go to verify that the new connection pool implementation behaves correctly when the cluster topology changes. Prior to PR #332, the SimplePool actually failed this test.

One more thing to note in this change is that there is no "distance" property to the HostSelectionPolicy so, as-is the interface can't support an a datacenter aware policy. Adding such support would be a good follow-on to this PR.

[Review on Reviewable](https://reviewable.io/reviews/gocql/gocql/348)


connectionpool.go
```diff
@@ -435,3 +439,404 @@ func (c *SimplePool) removeHostLocked(addr string) {
...
+	connCfg  ConnConfig
+	keyspace string
+
+	mu            sync.RWMutex
+	hostPolicy    HostSelectionPolicy
+	connPolicy    func() ConnSelectionPolicy
+	hostConnPools map[string]*hostConnPool
+}
+
+func NewPolicyConnPool(
```
Zariel: Could you add a short document this function

connectionpool.go
```diff
@@ -435,3 +439,404 @@ func (c *SimplePool) removeHostLocked(addr string) {
...
+	// ok fill the pool
+	pool.filling = true
+
+	// allow others to access the pool while filling
+	pool.mu.Unlock()
+
+	// fill only the first connection synchronously
+	if startCount == 0 {
+		err := pool.connect()
+		if opErr, ok := err.(*net.OpError); ok && opErr.Op == "read" {
```
Zariel: isnt this error usually just an `io.EOF`?

connectionpool.go
```diff
@@ -435,3 +439,404 @@ func (c *SimplePool) removeHostLocked(addr string) {
...
+	defer pool.mu.Unlock()
+
+	// find the connection index
+	for i, candidate := range pool.conns {
+		if candidate == conn {
+			// remove the connection, not preserving order
+			pool.conns[i], pool.conns = pool.conns[len(pool.conns)-1], pool.conns[:len(pool.conns)-1]
+
+			// update the policy
+			pool.policy.SetConns(pool.conns)
```
Zariel: I get the feeling that `Session` should implement `HandleError` as well, then have all errors go back to the session which in turn passes them to the pool and the policies, that way we wont have dependencies between the connection and pools and then implicitly to the policy.

connectionpool.go
```diff
@@ -435,3 +439,404 @@ func (c *SimplePool) removeHostLocked(addr string) {
...
+type hostConnPool struct {
+	host     string
+	port     int
+	addr     string
+	size     int
+	connCfg  ConnConfig
+	keyspace string
+	policy   ConnSelectionPolicy
+	// protection for conns, closed, filling
+	mu      sync.Mutex
```
Zariel: This should be a `RWMutex` so that we dont restrict reads in the case that nothing is being mutated

connectionpool.go
```diff
@@ -435,3 +439,404 @@ func (c *SimplePool) removeHostLocked(addr string) {
...
+
+			// lost a connection, so fill the pool
+			go pool.fill()
+			break
+		}
+	}
+}
+
+// removes and closes all connections from the pool
+func (pool *hostConnPool) drain() {
```
Zariel: should drain set a closed flag?

connectionpool_systems_test.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+set -e
+
+function run_tests() {
```
Zariel: instead of having another test runner, would it not just be easier to add the `conn_pool` tag to the `integration.sh` runner?

policies.go
```diff
@@ -42,3 +48,184 @@ type SimpleRetryPolicy struct {
 func (s *SimpleRetryPolicy) Attempt(q RetryableQuery) bool {
 	return q.Attempts() <= s.NumRetries
 }
+
+//HostSelectionPolicy is an interface for selecting
+//the most appropriate host to execute a given query.
+type HostSelectionPolicy interface {
+	//SetHosts notifies this policy of the current hosts in the cluster
+	SetHosts(hosts []HostInfo)
```
Zariel: Here you can just embed the previously defined interfaces which would make it more portable

connectionpool_systems_test.go
```diff
@@ -0,0 +1,266 @@
...
+	"strings"
+	"sync"
+	"testing"
+	"time"
+)
+
+// connection pool behavior test when nodes are removed from the cluster
+// to run this test, see connectionpool_systems_test.sh
+
+var (
```
0x6e6562: This feels like a lot of duplication from `cassandra_test.go` - is there something obvious I'm missing that makes it necessary to maintain separate test plumbing for these new tests?

policies.go
```diff
@@ -42,3 +48,184 @@ type SimpleRetryPolicy struct {
...
+}
+
+func (t *tokenAwareHostPolicy) SetPartitioner(partitioner string) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	t.fallback.SetPartitioner(partitioner)
+	t.partitioner = partitioner
+
+	t.resetTokenRing()
```
Zariel: should this only happen if the partitioner has changed?

policies.go
```diff
@@ -42,3 +48,184 @@ type SimpleRetryPolicy struct {
...
+	}
+	if routingKey == nil {
+		return t.fallback.Pick(qry)
+	}
+
+	var host *HostInfo
+
+	t.mu.RLock()
+	if t.tokenRing != nil {
+		host = t.tokenRing.GetHostForPartitionKey(routingKey)
```
Zariel: maybe change the tokenRing type so the method for this looks like

```go
func (t *tokenRing) GetHostForPartitionKey([]byte) *HostInfo {
    if t == nil {
        return nil
    }
}
```

policies.go
```diff
@@ -42,3 +48,184 @@ type SimpleRetryPolicy struct {
...
+	var once sync.Once
+	var fallbackIter NextHost
+	return func() *HostInfo {
+		if !hostReturned {
+			hostReturned = true
+			return host
+		}
+
+		// fallback
+		once.Do(func() { fallbackIter = t.fallback.Pick(qry) })
```
Zariel: is there a reason why this must be done inside the function? If so couldnt it be done simply by checking if `fallbackIter` is nil?

policies.go
```diff
@@ -42,3 +48,184 @@ type SimpleRetryPolicy struct {
...
+	if t.tokenRing != nil {
+		host = t.tokenRing.GetHostForPartitionKey(routingKey)
+	}
+	t.mu.RUnlock()
+
+	if host == nil {
+		return t.fallback.Pick(qry)
+	}
+
+	var hostReturned bool = false
```
Zariel: should use a var block here, and no need to define the bool type and assign it to be false

policies.go
```diff
@@ -42,3 +48,184 @@ type SimpleRetryPolicy struct {
...
+}
+
+type roundRobinConnPolicy struct {
+	conns []*Conn
+	pos   uint32
+	mu    sync.RWMutex
+}
+
+func NewRoundRobinConnPolicy() ConnSelectionPolicy {
+	return &roundRobinConnPolicy{conns: []*Conn{}}
```
Zariel: nit: here I dont think its necessary to provide an empty list

policies.go
```diff
@@ -42,3 +48,184 @@ type SimpleRetryPolicy struct {
...
+}
+
+func (r *roundRobinHostPolicy) SetPartitioner(partitioner string) {
+	// noop
+}
+
+func (r *roundRobinHostPolicy) Pick(qry *Query) NextHost {
+	pos := atomic.AddUint32(&r.pos, 1)
+	var i uint32 = 0
+	return func() *HostInfo {
```
Zariel: To get better query distribution here I think it will be better to have the function increment `pos` each call

Zariel:
Looks good, this change came about pretty nicely. 

I think in the future it will be worth to move to have the session own the hosts and partitioner, then it can delegate those to the connection pool and policy.

policies_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	hosts := []HostInfo{
+		HostInfo{HostId: "0", Peer: "0", Tokens: []string{"00"}},
+		HostInfo{HostId: "1", Peer: "1", Tokens: []string{"25"}},
+		HostInfo{HostId: "2", Peer: "2", Tokens: []string{"50"}},
+		HostInfo{HostId: "3", Peer: "3", Tokens: []string{"75"}},
+	}
+
+	policy.SetHosts(hosts)
+	policy.SetPartitioner("OrderedPartitioner")
```
0x6e6562: Would it be a good idea to have test coverage for `Murmur3Partitioner` and `RandomPartitioner` (in addition to `OrderedPartitioner` as implemented here), or is that covered somewhere else?

token.go
```diff
@@ -239,7 +239,7 @@ type tokenRing struct {
 	hosts       []*HostInfo
 }
 
-func newTokenRing(partitioner string, hosts []*HostInfo) (*tokenRing, error) {
+func newTokenRing(partitioner string, hosts []HostInfo) (*tokenRing, error) {
```
0x6e6562: Is it preferable to use `hosts []HostInfo` rather than `hosts []*HostInfo`?

0x6e6562:
This change is looking pretty good, I especially liked the way that the `ringDescriber` didn't have to change too much. My inline review comments are relatively minor in nature.

connectionpool_systems_test.go
Zariel: yeah, I think this file should just have the build tags `all, integration` that way they are ran during travis as well, or is that not a good idea?

Zariel:
Looks like this will require a rebase given #339 landed

token.go
justinretailnext: Originally I used ```[]*HostInfo``` but I had to change the ringDescriber to provide ```[]*HostInfo``` since it provides ```[]HostInfo```. However, I decided to avoid changing the ringDescriber or the SetHosts function in ```ConnectionPool``` and do it this way. In its favor, ```[]HostInfo``` does have the benefit of being more compact in memory.

policies_test.go
justinretailnext: Most test coverage for the variations in token partitioners should be in token_test.go; I'll take a look into the coverage profile and see if I can improve the coverage for all of the partitioners in use with the tokenRing struct.

connectionpool.go
justinretailnext: This check at this line is meant to avoid logging errors typically seen when the pool is trying to connect to a host which is down or unreachable since the pool attempts to connect so often (though less often than SimplePool). For net.Dial and net.DialTimeout mostly it seems like net.OpError is returned for most issues, such as "connection refused".

On reviewing this code while running the connectionpool_systems_test.sh I noticed that I was not suppressing errors with Op == "dial" so I'll be fixing that.

connectionpool.go
justinretailnext: Only Close sets closed to true, then it starts a goroutine call to drain; essentially drain is the async portion of closing a pool. I suppose when I wrote this code I liked the simple idea of there being a drain func as the reverse of the fill func. I suppose this means that you could drain the pool without closing it, but this is actually safe since Pick always checks if the pool is empty and fills it if it is.

connectionpool_systems_test.go
justinretailnext: When I created this test the idea was that I wanted to just test the behavior of NewSimplePool, NewRoundRobinConnPool and NewTokenAwareConnPool when the ccm cluster was changed by adding or removing a number of nodes. The go test itself is pretty sloppy since I was exploring situations where policyConnPool performed sub par compared to SimplePool and I wasn't sure what exactly par was until I found situations where SimplePool actually returns nil connections and policyConnPool doesn't.

So the problems with this test that make it less than ideal to merge into integration.sh:
- SimplePool has a race condition where occasionally nil will be returned by Pick.
- The test actually makes calls to ccm unlike the tests in cassandra_test.go and I'm sure in some cases the cluster may be left in an changed state (e.g. nodes stopped).

I'm totally open to moving these tests into the integration though. These tests though would need to be written a bit cleaner and with more time efficiency. Also the race condition in SimplePool which leads to nil being returned by Pick would need to be resolved (or SimplePool should just be removed and replaced with NewRoundRobinConnPool)

justinretailnext: I'll add that the race condition in SimplePool seems to be between concurrent calls to ```Pick``` after the call to ```c.mu.Unlock()``` and another goroutine which is handling a node disconnect via the call to ```HandleError```. The intention is that when all connections for a host are removed, the host pool is removed, but since the later part of ```Pick``` isn't keeping the c.mu lock it seems to be able to get off a call to RoundRobin.Pick while another routine is holding the c.mu lock within removeConnLock which is modifying that same RoundRobin object, this can create a situation where Pick can return nil even if there are other valid connections to the cluster.

I didn't put any effort into fixing this race condition since the new policyConnPool obsoletes SimplePool, but if SimplePool is kept along side the new policyConnPool, then the race condition should be fixed.

policies.go
justinretailnext: Which interfaces?

Zariel: Isnt there a SetHosts one? If not I guess that was in an old branch of mine

connectionpool_systems_test.go
Zariel: ah I see what its doing now. Its expecting that its being ran in as part of a build against `ccm`. Our test harness is becoming rather complex now :)

It would be nice if the flags were only defined once though.

justinretailnext:
This PR has been rebased and includes some of the fixes for your code review comments. Right now I am trying to improve the test coverage in both the token_test.go tests and the tests elsewhere relating to this and previous PRs.

connectionpool.go
```diff
@@ -435,3 +445,428 @@ func (c *SimplePool) removeHostLocked(addr string) {
...
+			fillCount--
+		}
+
+		// mark the end of filling
+		pool.fillingStopped()
+	}()
+}
+
+func (pool *hostConnPool) logConnectErr(err error) {
+	if opErr, ok := err.(*net.OpError); ok && (opErr.Op == "dial" || opErr.Op == "red") {
```
Zariel: should this be "red" or "read"?

justinretailnext:
Alright, hopefully this PR is now good enough to merge! I've done my best to improve the unit test coverage for this PR and the supporting changes to metadata, and tokens. I copied a couple tests in conn_test.go for testing policyConnPool and the round robin host policy, however I stopped at that when I realized a number of the tests in that file are marked as skipped for being flaky. I've also gone back and added documentation strings to all of the tests I've added as part of this series of PRs.

Zariel:
Thanks, ill take a look!

connectionpool.go
```diff
@@ -435,3 +449,433 @@ func (c *SimplePool) removeHostLocked(addr string) {
...
+
+	// avoid filling a full (or overfull) pool
+	if fillCount <= 0 {
+		pool.mu.RUnlock()
+		return
+	}
+
+	// switch from read to write lock
+	pool.mu.RUnlock()
+	pool.mu.Lock()
```
Zariel: Im not sure this is a problem, but you can't get upgrade the RWLock safely, here many readers can see the same state then all wait for the write lock, then go on filling even though one has already done it.

Here I think it would be better to do something along the lines of

```go
if !atomic.CompareAndSwapInt32(&pool.filling, 0, 1) {
    // already filling
    return
}
defer atomic.StoreInt32(&pool.filling, 0)

// check if should fill,

pool.mu.Lock()

// do filling
```

Zariel:
@justinretailnext thanks for this, will wait for @0x6e6562 to have a look if he has time but im happy to merge this in current state.

@justinretailnext could you submit the connection pool test you had written in the previous commits in a separate PR without the extra test runner script please, I would like to add it to the integration suite

0x6e6562:
This LGTM - @Zariel are you also happy with the current state of the RWLock upgrade that you commented on?

Zariel:
It's not really a bug, it just means that the check to see if we actually should be filling the pool happens twice, once with a read lock and once with a write lock, which means we acquire the write lock in cases where we will not actually do any filling. Guess to wait and see if @justinretailnext agrees and can make the change.

0x6e6562:
OK, I see - let's wait for @justinretailnext to comment on this before merging.

BTW, I just manually re-triggered build 812 in Travis to re-test the merge commit from #369.

justinretailnext:
Yeah, I made an initial attempt to add my connection pool systems tests to cassandra_test.go but their's some issues with my tests to work out. I'll PR that when I figure that out.

I'm fine with leaving the locking with the double acquire / double check; I'll take another shot at rewriting that with another PR.

BTW, I already started writing code for handling replication strategies which will be a needed step for #356.

0x6e6562:
OK, cool let's merge this then - many thanks for all of your hard work to land this feature :-)

0x6e6562:
I've created #370 as a follow up for the lock upgrades.

Zariel:
@justinretailnext im happy to add another test file which just tests the connection by manipulating nodes via ccm which can be ran from integration.sh separately from the integration tests which currently run. Ill create an issue to discuss this as I think it would very useful to ensure that the pool works correctly with node changes.

PR #349, opened 2015-04-07
Enhancements to integration.sh for ease-of-use in development

justinretailnext:
I've found that while testing changes locally that these changes simplify running integration.sh.

- Replaced the sed commands with usage of the --jvm_arg parameter of 'ccm create' for GC tuning
- Added 'ccm remove test' command as an optional step to remove the previous test cluster if it
already exists
- Changed the 'ccm clear' command to 'ccm remove'

YMMV, and I understand this is for travis builds too testing so not sure if they're the best generic changes.

[Review on Reviewable](https://reviewable.io/reviews/gocql/gocql/349)

### DIFF
--- a/integration.sh
+++ b/integration.sh
@@ -24,11 +24,10 @@ function run_tests() {
 	    "read_request_timeout_in_ms: 5000"
 	)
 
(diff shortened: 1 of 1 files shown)


Zariel:
Thanks!

PR #352, opened 2015-04-09
V3 tuples

### LINKED ISSUE #177: Tuple type (2.1)
Final new 2.1 feature that we're looking at (https://issues.apache.org/jira/browse/CASSANDRA-6875) requires the new CQL tuple type: https://issues.apache.org/jira/browse/CASSANDRA-7248

Zariel:
closes #177

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -25,22 +25,22 @@ var (
 // Marshaler is the interface implemented by objects that can marshal
 // themselves into values understood by Cassandra.
 type Marshaler interface {
-	MarshalCQL(info *TypeInfo) ([]byte, error)
+	MarshalCQL(info TypeInfo) ([]byte, error)
 }
 
 // Unmarshaler is the interface implemented by objects that can unmarshal
 // a Cassandra specific description of themselves.
 type Unmarshaler interface {
-	UnmarshalCQL(info *TypeInfo, data []byte) error
+	UnmarshalCQL(info TypeInfo, data []byte) error
 }
 
 // Marshal returns the CQL encoding of the value for the Cassandra
 // internal type described by the info parameter.
-func Marshal(info *TypeInfo, value interface{}) ([]byte, error) {
+func Marshal(info TypeInfo, value interface{}) ([]byte, error) {
 	if value == nil {
 		return nil, nil
 	}
-	if info.Proto < protoVersion1 {
+	if info.Version() < protoVersion1 {
 		panic("protocol version not set")
 	}
 
@@ -56,7 +56,7 @@ func Marshal(info *TypeInfo, value interface{}) ([]byte, error) {
 		}
 	}
 
-	switch info.Type {
+	switch info.Type() {
 	case TypeVarchar, TypeAscii, TypeBlob:
 		return marshalVarchar(info, value)
 	case TypeBoolean:
@@ -91,7 +91,7 @@ func Marshal(info *TypeInfo, value interface{}) ([]byte, error) {
 // Unmarshal parses the CQL encoded data based on the info parameter that
 // describes the Cassandra internal data type and stores the result in the
 // value pointed by value.
-func Unmarshal(info *TypeInfo, data []byte, value interface{}) error {
+func Unmarshal(info TypeInfo, data []byte, value interface{}) error {
 	if v, ok := value.(Unmarshaler); ok {
 		return v.UnmarshalCQL(info, data)
 	}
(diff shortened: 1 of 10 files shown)


Zariel:
This isnt really as large as it looks like, the problem is that `TypeInfo` is scattered around so many places, which is unfortunate. As noted in the commit this will break implementations of Un/MarshallCQL

marshal.go
```diff
@@ -25,22 +25,22 @@ var (
 // Marshaler is the interface implemented by objects that can marshal
 // themselves into values understood by Cassandra.
 type Marshaler interface {
-	MarshalCQL(info *TypeInfo) ([]byte, error)
+	MarshalCQL(info TypeInfo) ([]byte, error)
```
0x6e6562: Why is `TypeInfo` no longer a pointer?

marshal.go
```diff
@@ -1148,29 +1169,107 @@ func unmarshalInet(info *TypeInfo, data []byte, value interface{}) error {
...
-	Elem   *TypeInfo // only used for TypeMap, TypeList and TypeSet
-	Custom string    // only used for TypeCustom
+type TypeInfo interface {
+	Type() Type
+	Version() byte
+	Custom() string
+
+	// New creates a pointer to an empty version of whatever type
+	// is referenced by the TypeInfo receiver
+	New() interface{}
```
0x6e6562: Is `New()` as function prototype idiomatic Go?

session.go
```diff
@@ -632,20 +632,46 @@ func (iter *Iter) Scan(dest ...interface{}) bool {
...
+			i++
 			continue
 		}
-		err := Unmarshal(iter.columns[i].TypeInfo, iter.rows[iter.pos][i], dest[i])
-		if err != nil {
-			iter.err = err
+
+		// how can we allow users to pass in a single struct to unmarshal into
+		if col.TypeInfo.Type() == TypeTuple {
+			// this will panic, actually a bug, please report
```
0x6e6562: What does the comment about the panic mean?

session.go
```diff
@@ -632,20 +632,46 @@ func (iter *Iter) Scan(dest ...interface{}) bool {
...
+			iter.err = Unmarshal(col.TypeInfo, iter.rows[iter.pos][c], dest[i:i+count])
+			if iter.err != nil {
+				return false
+			}
+			i += count
+			continue
+		}
+
+		iter.err = Unmarshal(col.TypeInfo, iter.rows[iter.pos][c], dest[i])
+		if iter.err != nil {
```
0x6e6562: I imagine this return value would be different in #331 had landed already?

0x6e6562:
On the whole LGTM - the patch is noisy because you've plumbed through a change to the central type hierarchy of the library, but those changes are mainly mechanical.

I've just remarked on a few items of a minor nature.

marshal.go
Zariel: It was changed to an interface

marshal.go
Zariel: No, but its the best way I could think of the keep the current behaviour of the helper functions, which I didnt want to change in this patch.

session.go
Zariel: If we're unmarshalling a tuple but the `TypeInfo` isn't a tuple type then the type assertion will fail.

marshal.go
0x6e6562: Ah yes, you're right :-)

session.go
0x6e6562: I don't get that - which invocation are you talking about:

```go
if col.TypeInfo.Type() == TypeTuple {
```

or

```go
tuple := col.TypeInfo.(TupleTypeInfo)
```

?

marshal.go
0x6e6562: OK, fair enough.

session.go
Zariel: `tuple := col.TypeInfo.(TupleTypeInfo)`

0x6e6562: If you think that is possible, why not be defensive, check the type cast and return some error in the failure scenario:

```go
tuple, ok := col.TypeInfo.(TupleTypeInfo)
if !ok {
    return an error ....
}
```

Zariel: Then it breaks the `iter.Scan()` interface which would not be ideal, this case should not happen in practice, I left that comment there to make it explicit that it wasn't doing the safe type assertion.

session.go
Zariel: I don't think so, as it should already be set before the call the `Scan()`

0x6e6562:
Nice work :-)

Zariel:
v3 support is coming along nicely :+1:

PR #353, opened 2015-04-09
Add with serial consistency support

### LINKED ISSUE #343: Add support for with serial consistency
It sees that when implementing v2 of the protocol we missed adding support for with serial consistency flag in queries.

Zariel:
Add an option to Query to set the serial consistency for the
conditional update phase of a query.

closes #343

### DIFF
--- a/session.go
+++ b/session.go
@@ -524,9 +524,6 @@ func (q *Query) Bind(v ...interface{}) *Query {
 // SERIAL. This option will be ignored for anything else that a
 // conditional update/insert.
 func (q *Query) SerialConsistency(cons Consistency) *Query {
-	if !(cons == Serial || cons == LocalSerial) {
-		panic("only acceptable consistency for serial_consistency is SERIAL or LOCAL_SERIAL")
-	}
 	q.serialCons = cons
 	return q
 }
--- a/session.go
+++ b/session.go
@@ -96,7 +96,7 @@ func (s *Session) Query(stmt string, values ...interface{}) *Query {
 	s.mu.RLock()
 	qry := &Query{stmt: stmt, values: values, cons: s.cons,
 		session: s, pageSize: s.pageSize, trace: s.trace,
-		prefetch: s.prefetch, rt: s.cfg.RetryPolicy}
+		prefetch: s.prefetch, rt: s.cfg.RetryPolicy, serialCons: s.cfg.SerialConsistency}
 	s.mu.RUnlock()
 	return qry
 }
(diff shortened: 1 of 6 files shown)


Zariel:
I cant really think of a better test without it becoming very complicated

Zariel:
also, shall I get rid of the panic? As cassandra will return an error anyway, that way you get a consistent error from setting on the query to the cluster

session.go
```diff
@@ -517,6 +518,19 @@ func (q *Query) Bind(v ...interface{}) *Query {
 	return q
 }
 
+// SerialConsistency sets the consistencyc level for the
+// serial phase of conditional updates. That consitency can only be
+// either SERIAL or LOCAL_SERIAL and if not present, it defaults to
+// SERIAL. This option will be ignored for anything else that a
+// conditional update/insert.
+func (q *Query) SerialConsistency(cons Consistency) *Query {
```
0x6e6562: Is it such a great idea to panic here? I was under the impression that we refrained from panicking in this library.

0x6e6562:
I'm not 100% sure how this new field plumbs down to the framing layer - can you point me to the bit in the patch where this gets framed?

Zariel:
When I implemented the framer I pretty much just completed the spec so its already there, https://github.com/gocql/gocql/blob/master/frame.go#L911

0x6e6562:
Thanks for the clarification about the plumbing.

I can't really see how this can be feasibly tested with our integration test - that would be a significant undertaking.

My only other remark regards the panic policy - did you have any ideas on whether that is the bets way to handle this?

session.go
Zariel: yeah just removed the panic, Cassandra will return an error anyway

Zariel:
We should only panic in exceptional circumstances which would lead us into a buggy state, so here its not necessary as Cassandra will return an error with an invalid query, the docs are clear as well.

Zariel:
Hold off merging this, I forgot about batches supporting serial consistency in v3, will add that too

Zariel:
Can you use `SERIAL` or `LOCAL_SERIAL` in a regular query?

Zariel:
Looks like a good plan, from the cql docs

`You cannot configure this level as a normal consistency level, configured at the driver level using the consistency level field. You configure this level using the serial consistency field as part of the native protocol operation. See failure scenarios.`

Zariel:
@0x6e6562 anything blocking this merging?

0x6e6562:
LGTM

Zariel:
Thanks!

PR #354, opened 2015-04-09
Introduce an integration test for authentication

0x6e6562:
This patch squashes #316.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,10 @@ env:
   global:
     - GOMAXPROCS=2
   matrix:
-    - CASS=1.2.19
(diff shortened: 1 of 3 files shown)


Zariel:
I would remove the cqlsh checks, if gocql doesnt work then the python driver shouldnt

0x6e6562:
Yes, you're right - the cqlsh check was only introduced to debug the test run and shed more light on the situation.

0x6e6562:
@Zariel It looks like this variant might work.

Zariel:
The struggle has a happy ending :) good job!

PR #358, opened 2015-04-10
Fix reading the streamID sign

Zariel:
In v3 the stream is an int16 not a uint16, int v2 ensure we handle
the sign of the int8 correctly.


Zariel:
spotted this whilst messing around with events, shouldn't affect any users.

Issue #278, opened 2014-10-30
Re-adjust test coverage quota


0x6e6562:
As part of the integration test stabilization in #275, the test coverage got lowered to 55%. The motivation was stop the build from failing due to having a too low coverage threshold. The reason why #275 lowered the quota was because it skipped some tests that were not running consistently. So we need to review the overall coverage approach and re-adjust the minimum level of coverage as appropriate. One suggestion from @phillipCouto was to move the test coverage checks into a separate leg of the build matrix, which would avoid complex parameterization based whether we were testing against 1.2 or 2+.

PR #362, opened 2015-04-13
enable the race detector

Zariel:


0x6e6562:
This needs re-merging.

Zariel:
dont know what caused this to fail

Zariel:
With having the cover profile running it makes it pretty much impossible to figure out what is going wrong in that test to cause it to fail.

0x6e6562:
Is it worth splitting this into two test runs (for coverage, one for races) for debugging purposes (or just disabling coverage for now)?

With regards to coverage, I think on average we are higher than 55% right now, but that's a separate issue: #278

Zariel:
created #368 to discuss this

Issue #323, opened 2015-03-10
Add support for named param binding


Zariel:
In protocol v3 we can bind params in queries using names instead of ?

Zariel:
@0x6e6562 how would this interact with the current query binding? Could we replace the current logic in v3 with this?

Zariel:
This ties into #175 as they both require params to have names

Zariel:
Im not entirely sure this is worth adding to be honest

0x6e6562:
Do you think you state why not (for the sake of posterity when somebody is asking for this in the future)?

Zariel:
As we always prepare statements we get back the arguments in the correct order from the PREPARE call which means we wont need to send the value names.

It will complicate the API as we go from having a query interface by positional arguments to one which is identified by named values, which i'm not too keen on.

The spec has this to say about it as well,

`Note that this is only useful for
              QUERY requests where named bind markers are used; for EXECUTE statements,
              since the names for the expected values was returned during preparation,
              a client can always provide values in the right order without any names
              and using this flag, while supported, is almost surely inefficient.`

So unless someone comes up with a use case and a good API I don't think this is worth adding.

0x6e6562:
Fair enough, let's close this issue then.

PR #369, opened 2015-04-17
Add support for UDT's

### LINKED ISSUE #175: Support for UDTs (2.1)
So one of the upcoming features of 2.1 is UDTs (User Defined Types https://issues.apache.org/jira/browse/CASSANDRA-5590).

Given the CQL:

```
CREATE TYPE name (
    first_name text,
    last_name text
);
  
CREATE TABLE user (
    id uuid PRIMARY KEY,
    user_name name
    email text
);
```

And a Go struct:

```
type Name struct {
    First string,
    Last string
}

type User struct {
    Id gocql.UUID
    Name Name
    Email string
}
```

We need to be able to do this:

```
var user User

err := session.Query(`SELECT id, name, email FROM users WHERE id = ? LIMIT 1`, id).Scan(&user.Id, &user.Name, &user.Email)
```

Otherwise the Driver won't be CQL compliant. There is the option to do dot notation for nested fields, like:

```
var user User

err := session.Query(`SELECT id, name.first, name.last, email FROM users WHERE id = ? LIMIT 1`, id).Scan(&user.Id, &user.Name.First, &user.Name.Last, &user.Email)
```

However drivers have to implement the first option too.

Previously the general consensus has been against struct marshalling, any idea how you'd like this handled? I'm happy to build this once I know what will get merged.

Zariel:
Add support to read and write UDT's at the framing and scanning level.

To scan or marshal a UDT type the type must implement the Marshal/
UnmarshalUDT type.

Fixes #175

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -612,6 +612,22 @@ func (f *framer) readTypeInfo() TypeInfo {
 
 		return tuple
 
+	case TypeUDT:
+		udt := UDTTypeInfo{
+			NativeType: simple,
+		}
+		udt.KeySpace = f.readString()
+		udt.Name = f.readString()
+
+		n := f.readShort()
+		udt.Elements = make([]UDTField, n)
+		for i := 0; i < int(n); i++ {
+			field := &udt.Elements[i]
+			field.Name = f.readString()
+			field.Type = f.readTypeInfo()
+		}
+
+		return udt
 	case TypeMap, TypeList, TypeSet:
 		collection := CollectionType{
 			NativeType: simple,
--- a/marshal.go
+++ b/marshal.go
@@ -1203,13 +1203,22 @@ func unmarshalTuple(info TypeInfo, data []byte, value interface{}) error {
 }
 
 // UDTMarshaler is an interface which should be implemented by users wishing to
-// handle encoding UDT types to sent to Cassandra.
+// handle encoding UDT types to sent to Cassandra. Note: due to current implentations
+// methods defined for this interface must be value receivers not pointer receivers.
 type UDTMarshaler interface {
-	EncodeUDTField(name string, info TypeInfo) ([]byte, error)
+	// MarshalUDT will be called for each field in the the UDT returned by Cassandra,
+	// the implementor should marshal the type to return by for example calling
+	// Marshal.
+	MarshalUDT(name string, info TypeInfo) ([]byte, error)
 }
 
+// UDTUnmarshaler should be implemented by users wanting to implement custom
+// UDT unmarshaling.
 type UDTUnmarshaler interface {
-	DecodeUDTField(name string, info TypeInfo, data []byte) error
+	// UnmarshalUDT will be called for each field in the UDT return by Cassandra,
+	// the implementor should unmarshal the data into the value of their chosing,
+	// for example by calling Unmarshal.
+	UnmarshalUDT(name string, info TypeInfo, data []byte) error
 }
 
 func marshalUDT(info TypeInfo, value interface{}) ([]byte, error) {
@@ -1219,7 +1228,7 @@ func marshalUDT(info TypeInfo, value interface{}) ([]byte, error) {
 	case UDTMarshaler:
 		var buf []byte
 		for _, e := range udt.Elements {
-			data, err := v.EncodeUDTField(e.Name, e.Type)
+			data, err := v.MarshalUDT(e.Name, e.Type)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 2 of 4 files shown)


Zariel:
@0x6e6562 thoughts on the API? This style of API can used to implement #323 as well. It lends itself to being implemented by a third party library, but is pretty flexible. Also naming of the interfaces, methods.

Looks like I ran into a bug that you cant use none pointer receivers for Marshal, but im not sure it is a bug.

al3xdm:
On the implementation itself, I think it looks good. My only comment would be the interface names could be clearer and more consistent with the std library etc.

```
type UDTMarshaler interface {
    EncodeUDTField(name string, info TypeInfo) ([]byte, error)
}

type UDTUnmarshaler interface {
    DecodeUDTField(name string, info TypeInfo, data []byte) error
}
```

Could change to:

```
type UDTMarshaler interface {
    MarshalUDT(name string, info TypeInfo) ([]byte, error)
}

type UDTUnmarshaler interface {
    UnmarshalUDT(name string, info TypeInfo, data []byte) error
}
```

However, I'd favour the automatic Marshaling / Unmarshaling based on struct tags as went over in #175 rather than adding more interfaces to implement. If we did that, then of the use cases I can think of, the combination of the existing MarshalCQL / UnmarshalCQL and struct tags would achieve the same goal as adding this interface.

Zariel:
Im not against using reflection with struct tags to unmarshal UDT types, but I am weary of using them to replace the current `Scan(...interface{})`, to that end if they were implemented they would only be usable to marshal/unmarshal UDT types.

nemothekid:
Struct Tags (like JSON) should be supported here - 

```go
struct MyUDTName type {
  First `cql:"first"`
  Last `cql:"last"`
}
```

I can't quite parse your last comment, but I'd imagine the API would work like this:
```go
query := mySession.Query("SELECT textType, addressType FROM table LIMIT 1").Scan(&myString, &myUDTStructType)
```

AFAIK, it would not replace the current `Scan(...interface{})`, but simply allow you to use UDTs. Something like 
```go
var myUDTStructType struct MyUDTName type {
  First `cql:"first"`
  Last `cql:"last"`
}

query := mySession.Query("SELECT first, last FROM table LIMIT 1").Scan(&myUDTStructType)
```

**shouldn't** be allowed.

Zariel:
This PR now adds support for marshalling and unmarshalling structs for UDT types, without using struct tags though.

Like so,

```go
	type horse struct {
		Name  string
		Owner string
	}

	insertedHorse := &horse{
		Name:  "pony",
		Owner: "jim",
	}

	err = session.Query("INSERT INTO horse_race(position, horse) VALUES(?, ?)", 1, insertedHorse).Exec()
	if err != nil {
		t.Fatal(err)
	}

	retrievedHorse := &horse{}
	err = session.Query("SELECT horse FROM horse_race WHERE position = ?", 1).Scan(retrievedHorse)
	if err != nil {
		t.Fatal(err)
	}
```

Zariel:
Adding struct tags complicates it someone by requiring the scanner to scan every field in the struct even if tags are not used, and will probably make it more complicated too.

Zariel:
Added struct tag support

0x6e6562:
It feels like the sentiment towards support struct tagging with the core library has evolved over the course of this pull request. @Zariel has implementing a basic version of struct tag processing given you a different perspective?

Zariel:
I'm happy with struct tagging for UDT's as they lend themselves to it and fits the API very nicely. I don't want to adjust Scan to support struct scanning for arbitrary types. I agree with @nemothekid 

Adding struct tagging wasn't much extra work on top of the current implementation, as the TypeInfo has all the fields for a single type it makes more sense to unmarshal into a single type.

I'm happy with the state this PR is in now, as long as others are happy with the API then I think we should go ahead with this.

0x6e6562:
Yes, when I implemented struct tagging in the [cqlr](https://github.com/relops/cqlr) proof of concept, it turned out to be around about the same amount of code as what is required in this patch, which isn't a great deal.

The main difference between this implementation and that implementation was the re-use of struct metadata in subsequent API calls (back in the day this type of thing was called caching). But that's an optimization that doesn't have to be made now, and quite possibly may not be required in the future either.

So I'm happy to go ahead with this patch for now.

I agree that we should not implement any more complex support for arbitrary struct tagging options at this point in time. People could cook up all sorts of different tag conventions which might make sense in a particular use case, but is different to support generically.

Hence I suggest that we:

* Merge this patch.
* Invite any body in the future who wants/needs a more complicated binding mechanism to suggest a clean and maintainable way of doing this in the core that can't be done at a higher layer.

0x6e6562:
Very nice work BTW - you've implemented something that a whole bunch of people have been wanting for a while :-)

Zariel:
Thanks :) Most of the hard work was done in #352 and #339

PR #374, opened 2015-04-24
ensure that the value asertion does not panic

Zariel:
If there is no routing key available then the function will
return `nil, nil` and release the waitgroup. The waiters will
then type assert `inflight.value` which is a nil interface
value which then panics. In this case we should just return
nil as the originating function did.

### DIFF
--- a/session.go
+++ b/session.go
@@ -223,7 +223,9 @@ func (s *Session) routingKeyInfo(stmt string) (*routingKeyInfo, error) {
 			return nil, inflight.err
 		}
 
-		return inflight.value.(*routingKeyInfo), nil
+		key, _ := inflight.value.(*routingKeyInfo)
+
+		return key, nil
 	}
 
 	// create a new inflight entry while the data is created


session.go
```diff
@@ -223,7 +223,9 @@ func (s *Session) routingKeyInfo(stmt string) (*routingKeyInfo, error) {
 			return nil, inflight.err
 		}
 
-		return inflight.value.(*routingKeyInfo), nil
+		key, _ := inflight.value.(*routingKeyInfo)
+
+		return key, nil
```
0x6e6562: Why don't you return an error if the type assertion fails?

Zariel: the rest of the function just does a `return nil, nil` so we should do the same

0x6e6562:
Sure, that would be consistent. In that case I guess I  am questioning the precident that has been set - why does the function signature return an error when there is no code path in the function to actually return an error?

Zariel:
There are paths which return errors, but this case the return is `nil, nil` so it misses the previous `if inflight.err != nil {` which guards the usual case when `inflight.value` is `nil` but the case when you have a `nil` error and `nil` value was not accounted for.

PR #375, opened 2015-04-24
Consolidate new session

Zariel:
CreateSession on the cluster config is a helper method to setup
defaults for the session, but it also does a lot of the setup, such
as creating the connection pool and creating the host discovery
goroutine.

Move all this logic into NewSession so that CreateSession now just
returns NewSession.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -101,52 +101,7 @@ func NewCluster(hosts ...string) *ClusterConfig {
 // CreateSession initializes the cluster based on this config and returns a
 // session object that can be used to interact with the database.
 func (cfg *ClusterConfig) CreateSession() (*Session, error) {
-
-	//Check that hosts in the ClusterConfig is not empty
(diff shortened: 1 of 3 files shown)


Zariel:
I never really understood why `cluster.CreateSession` did more work than `NewSession`

0x6e6562:
I think #376 landing requires this to be rebased.

Zariel:
I think it should be fine as the commit will be ignored as it already exists in master

Zariel:
rebased

0x6e6562:
Did this become unmergeable again (presumably due to other stuff landing on master)?

0x6e6562:
>Did this become unmergeable again (presumably due to other stuff landing on master)?
Actually, rollback this last comment - this must have been a caching bug in iOctocat.

Anyway, this LGTM.

Zariel:
thanks

PR #378, opened 2015-04-26
Use timeouts when dialing tls sockets

Zariel:
When using tls we should set timeouts when dialing to prevent
blocking indefintely on downed hosts.


Zariel:
Now that we dont need to support go1.2 we can use `tls.DialWithDialer`

Issue #370, opened 2015-04-21
Lock upgrades whilst refilling a connection pool 


0x6e6562:
[As discussed](https://github.com/gocql/gocql/pull/348#discussion-diff-28655667) during the review of #348, we should revisit lock upgrades during a connection pool refill.

nemothekid:
Is there any reason why we should with @Zariel proposed solution? Its small enough that I can write it this weekend, and if the implementation is the same, the current tests should suffice?

0x6e6562:
>Is there any reason why we should with @Zariel proposed solution?

Did you mean to say "should not go with"?

If so, then I think it is fine to try that out - after all it would be a small patch, so feel free to contribute it.

PR #379, opened 2015-04-26
Hostpool fill block

Zariel:


Zariel:
This includes the changes to fix for #370

0x6e6562:
The first pass review of this LGTM but because this is a tricky change and has the potential to introduce subtle bugs, I'd like to re-review this after having mediated over a few things.

I was wondering where it's worth elaborating in the description of this pull request as to why we're making this change. I think this would serve two purposes:

1. Re-iterating the motivation will help to provide context for the review - i.e. it can help when we're trying to think of edge cases.
2. In the future it will provide us with the concrete memory of why we made this change, rather than having to back and think our way through multiple tickets.

Zariel:
Ill split this PR in two to make it more logically separated and write a better description.

PR #381, opened 2015-04-26
When the hostpool is empty dont block in fill

Zariel:
If the host is down and its connect blocks until timeout the
caller should not be blocked because it may be coming from a call
to execute a query.

### DIFF
--- a/connectionpool.go
+++ b/connectionpool.go
@@ -679,7 +679,8 @@ func (pool *hostConnPool) Pick(qry *Query) *Conn {
 
 	if empty {
 		// try to fill the empty pool
-		pool.fill()
+		go pool.fill()
+		return nil
 	}
 
 	return pool.policy.Pick(qry)


0x6e6562:
Interesting - what is the intended behavior when a nil connection is returned?

Zariel:
No available connections for this host, try another

0x6e6562:
Is it worth putting in a verification for this that we ensure during the test suite? I know this goes back on what I've previously said about mock servers, but this feels like the type of behavior that we would want to survive a potential future refactor. What do you think?

0x6e6562:
After having discussed in the gocql room, I turns out this could be difficult to test automatically, so I think this is just going to have to get tested in the wild.

justinretailnext:
The blocking behavior that was changed here was a copy over from SimplePool, however that empty check was at the entire pool level, not at the host connection pool level. So, we could move the empty and the closed check up to the policyConnPool.Pick method and just have the one read lock.

That said, I can't think of a reason that this change isn't going to work well enough.

Issue #376, opened 2015-04-25
When a host is down it blocks the PolicyAwareConnectionPool readers until timeout


Zariel:
As discovered by @nemothekid when a host is down (down in gossip) it will still exist in the `system.peers` table, this means that we will try to add it to the pools continually if discovery is enabled.

The cause of this is that when `SetHosts` is called it acquires the write lock, then tries to creating a host pool which dials the down node. Which in turn blocks until timeout before returning.

Zariel:
@nemothekid I have a commit to fix this but I want to write some tests to verify that it works correctly and fixes this issue.

The test suite fails on my machine due to 108f44b198d3cb9d069876575a5a3fe7454d674a which I think I will need to submit as a separate PR.

Zariel:
I havnt been able to replicate that this, my theory was that if a host is down and just hangs indefinitely then the pool will block to dial it. But this doesnt seem to be the case.

Though I have noticed that we dont set the timeout when dialing a TLS conn.

Zariel:
Im not entirely sure this is actually an issue after spending the day trying to replicate this.

nemothekid:
I think the proper solution here would to be implement proper server side events and only query system.peers once - much like the datastax drivers do. 

OTOH, I think your understanding of the issue is wrong - `SetHosts` is only called every 30 seconds, and I see the problem every 10 seconds - the issue potentially happens every query.

Specifically the issue occurs on [connectionpool.go#819](https://github.com/gocql/gocql/blob/master/connectionpool.go#L819) 

1.) `Query.Exec` is called.
2.)  [session.go#159](https://github.com/gocql/gocql/blob/master/session.go#L159) attempts to pick a connection from the pool.
3.) [connectionpool.go#594](https://github.com/gocql/gocql/blob/master/connectionpool.go#L594) chooses a host
4.) [policies.go#161](https://github.com/gocql/gocql/blob/master/policies.go#L161) returns a host based on the token. Now because the lists of hosts to choose from include all the hosts (incl. those which are down) this function may return a down host.
6.) [connectionpool.go#604](https://github.com/gocql/gocql/blob/master/connectionpool.go#L604) calls Pick, which call `pool.fill()` @ [connectionpool.go#682](https://github.com/gocql/gocql/blob/master/connectionpool.go#L682) which eventually calls connect @  [connectionpool.go#807](https://github.com/gocql/gocql/blob/master/connectionpool.go#L807).
[...]

Issue #368, opened 2015-04-15
update integration tests in travis


Zariel:
Currently we run the unit test then the integration tests separately in travis for each build, calculating the test coverage % based on the integration tests.

Having the integration tests run with the code coverage means that when tests fail they are impossible to debug because the line numbers are nonsensical. It also means that we miss some of the coverage % when checking what level we are at overall because some of the unit tests contribute to the overall %.

Im creating this issue to discuss the idea of running the integration tests with the tags '-all' to run everything, and not have a separate run of the unit tests beforehand. I also think it will be a good idea to remove the coverage profile so that we can debug issues if they arise.

PR #382, opened 2015-04-26
Reduce lock use when updating host pools

Zariel:
In the PolicyAwareConnection pool we shouldn't hold the write lock
whilst dialing hosts as it will block all readers. In the event that
a host is down we will try to connect to it and wait until timeout
before stopping. This means that whilst the connection is dialing
all readers will be blocked acquiring the read lock to access the
pools.

Instead change the policy aware pools to use a copy on write map
which allows us to do the mutations to the map without having the
write lock.

Use CAS to prevent concurrent calls to fill both attempting to fill
under a write lock.

Improve concurrency when mutating the pool when multiple concurrent
calls are occurring.

### DIFF
--- a/connectionpool.go
+++ b/connectionpool.go
@@ -14,6 +14,7 @@ import (
 	"math/rand"
 	"net"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -476,6 +477,7 @@ type policyConnPool struct {
 	numConns int
 	connCfg  ConnConfig
 	keyspace string
+	filling  int32
 
 	mu            sync.RWMutex
 	hostPolicy    HostSelectionPolicy
@@ -534,7 +536,13 @@ func NewPolicyConnPool(
 }
 
 func (p *policyConnPool) SetHosts(hosts []HostInfo) {
-	p.mu.Lock()
+	if !atomic.CompareAndSwapInt32(&p.filling, 0, 1) {
+		return
+	}
+	defer atomic.StoreInt32(&p.filling, 0)
+
+	// SetHosts calcuates the new set of hosts which should be inside the pool
+	p.mu.RLock()
 
 	toRemove := make(map[string]struct{})
 	for addr := range p.hostConnPools {
@@ -544,35 +552,50 @@ func (p *policyConnPool) SetHosts(hosts []HostInfo) {
 	// TODO connect to hosts in parallel, but wait for pools to be
 	// created before returning
 
+	// the old pool, retain a reference because we will replace this with the
+	// new pool later. This is essential now a copy on write map, when we drop
+	// support for 1.3 we can use atomic.Value to implement this without locks.
+	existingPool := p.hostConnPools
+
+	// new host pool which will be set once it has connected to the hosts
+	hostPool := make(map[string]*hostConnPool)
+
 	for i := range hosts {
-		pool, exists := p.hostConnPools[hosts[i].Peer]
+		addr := hosts[i].Peer
+		pool, exists := p.hostConnPools[addr]
 		if !exists {
+			// TODO: do this async
 			// create a connection pool for the host
 			pool = newHostConnPool(
-				hosts[i].Peer,
+				addr,
 				p.port,
 				p.numConns,
 				p.connCfg,
 				p.keyspace,
 				p.connPolicy(),
 			)
-			p.hostConnPools[hosts[i].Peer] = pool
+			hostPool[addr] = pool
 		} else {
+			hostPool[addr] = pool
 			// still have this host, so don't remove it
-			delete(toRemove, hosts[i].Peer)
+			delete(toRemove, addr)
 		}
 	}
+	p.mu.RUnlock()
 
+	p.mu.Lock()
+	p.hostConnPools = hostPool
+	p.mu.Unlock()
+
+	// we dont need to hold a lock to this pool as we are the only one with a refernce
+	// to it.
 	for addr := range toRemove {
-		pool := p.hostConnPools[addr]
-		delete(p.hostConnPools, addr)
+		pool := existingPool[addr]
 		pool.Close()
 	}
 
 	// update the policy
 	p.hostPolicy.SetHosts(hosts)
-
-	p.mu.Unlock()
 }
 
 func (p *policyConnPool) SetPartitioner(partitioner string) {
@@ -594,15 +617,20 @@ func (p *policyConnPool) Pick(qry *Query) *Conn {
 	nextHost := p.hostPolicy.Pick(qry)
 
 	p.mu.RLock()
-	var host *HostInfo
-	var conn *Conn
+	var (
+		host *HostInfo
+		conn *Conn
+	)
+
 	for conn == nil {
 		host = nextHost()
 		if host == nil {
 			break
 		}
+
 		conn = p.hostConnPools[host.Peer].Pick(qry)
 	}
+
 	p.mu.RUnlock()
 	return conn
 }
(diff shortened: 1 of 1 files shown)


connectionpool.go
```diff
@@ -719,76 +754,26 @@ func (pool *hostConnPool) fill() {
...
-	pool.mu.Unlock()
 	// only this goroutine should make calls to fill/empty the pool at this
 	// point until after this routine or its subordinates calls
 	// fillingStopped
 
-	// fill only the first connection synchronously
-	if startCount == 0 {
+	for fillCount > 0 {
 		err := pool.connect()
+		// TODO(zariel): if err != nil mark host DOWN so that we dont try to connect again
```
0x6e6562: Should this TODO get implemented as part of this patch?

0x6e6562:
> Use atomics to protect concurrent calls to fill

I'm assuming the intention is to use a CAS mechanism as a light weight barrier to check whether the fill operation is already taking place? If so, could we update the summary to reflect this?

justinretailnext:
I think generally, this is a bigger change than it needs to be, and introduces two significant differences from how `SimplePool` works: (1) copy on write hosts change, (2) changes the filling behavior of the per-host connection pool.

The real underlying problem in the `SetHosts` method of `policyConnPool` is that the fill method of `hostConnPool` is called within `newHostConnPool`; so my thought is what did `SimplePool` do to avoid that:
```go
func (c *SimplePool) SetHosts(hosts []HostInfo) {
	c.hostMu.Lock()
	toRemove := make(map[string]struct{})
	for k := range c.hosts {
		toRemove[k] = struct{}{}
	}

	for _, host := range hosts {
		host := host
		delete(toRemove, host.Peer)
		// we already have it
		if _, ok := c.hosts[host.Peer]; ok {
			// TODO: Check rack, dc, token range is consistent, trigger topology change
			// update stored host
			continue
		}

		c.hosts[host.Peer] = &host
	}

	// can we hold c.mu whilst iterating this loop?
	for addr := range toRemove {
		c.removeHostLocked(addr)
	}
	c.hostMu.Unlock()

	c.fillPool()
}
```
To summarize the above, `SetHosts` in `SimplePool` adds the new hosts as empty pools, then fills the host pools after releasing the lock. So instead I would suggest these changes to avoid calling `fill` while holding the `policyConnPool` write lock:
```diff
 func newHostConnPool(
@@ -660,9 +681,6 @@ func newHostConnPool(
 		closed:   false,
 	}
 
-	// fill the pool with the initial connections before returning
-	pool.fill()
-
 	return pool
 }
 
@@ -541,9 +543,7 @@ func (p *policyConnPool) SetHosts(hosts []HostInfo) {
 		toRemove[addr] = struct{}{}
 	}
 
-	// TODO connect to hosts in parallel, but wait for pools to be
-	// created before returning
-
+	added := make([]*hostConnPool, 0, len(hosts))
 	for i := range hosts {
 		pool, exists := p.hostConnPools[hosts[i].Peer]
 		if !exists {
@@ -556,6 +556,7 @@ func (p *policyConnPool) SetHosts(hosts []HostInfo) {
 				p.keyspace,
 				p.connPolicy(),
 			)
+			added = append(added, pool)
 			p.hostConnPools[hosts[i].Peer] = pool
 		} else {
 			// still have this host, so don't remove it
@@ -569,10 +570,23 @@ func (p *policyConnPool) SetHosts(hosts []HostInfo) {
 		pool.Close()
 	}
 
+	p.mu.Unlock()
+
 	// update the policy
 	p.hostPolicy.SetHosts(hosts)
 
-	p.mu.Unlock()
+	// force the new host connection pools to start filling
+	// this ensures that SetHosts has tried to connect at least once
+	// to new hosts before returning (an expectation of NewPolicyConnPool)
+	wg := sync.WaitGroup{}
+	for _, pool := range added {
+		wg.Add(1)
+		go func(pool *hostConnPool) {
+			pool.fill()
+			wg.Done()
+		}(pool)
+	}
+	wg.Wait()
 }
```

Following on that, I realized that the `hostPolicy` field of `policyConnPool` need not be protected by the lock as long as we protect against looking for a host conn pool that's not in the map:
```go
 func (p *policyConnPool) SetPartitioner(partitioner string) {
@@ -601,19 +615,24 @@ func (p *policyConnPool) Pick(qry *Query) *Conn {
 		if host == nil {
 			break
 		}
-		conn = p.hostConnPools[host.Peer].Pick(qry)
+
+		hostPool, found := p.hostConnPools[host.Peer]
+		if !found {
+			break
+		}
+
+		conn = hostPool.Pick(qry)
 	}
 	p.mu.RUnlock()
 	return conn
 }
 
 func (p *policyConnPool) Close() {
-	p.mu.Lock()
-
 	// remove the hosts from the policy
 	p.hostPolicy.SetHosts([]HostInfo{})
 
 	// close the pools
+	p.mu.Lock()
 	for addr, pool := range p.hostConnPools {
 		delete(p.hostConnPools, addr)
 		pool.Close()

```

The change to the `filling` field in `hostConnPool` is great, it basically creates a semaphore, but the rest of the change I think breaks the expectation of `fill` which is that it only synchronously makes the first connection to a host and asynchronously makes the rest of the connections. By breaking out that first connect call by itself it avoids the repeatedly failing to connect to a host if there's a serious issue.

I put together just these limited changes based on this PR into a [new branch as a comparison](https://github.com/retailnext/gocql/tree/connectionpool_locking_alt).

connectionpool.go
Zariel: No, its so that once we have the concept of hosts being UP or DOWN (from server events) we can detect that a host is not reachable and mark it such that it wont be used for queries.

Zariel:
I don't have any intention of keeping SimplePool around for much longer, and dont consider it the reference implementation of how the pools should behave, so I'm not much concerned about changes to how it works.

Would be happy to accept the changes you suggest @justinretailnext , though I do want to replace the current pools with a copy on write structure using atomic.Value so that reads don't incur a lock, but that requires dropping support for go1.3.

@0x6e6562 changed the description.

0x6e6562:
This is an extremely valuable discussion and I'm glad that you've chimed in with your suggestion @justinretailnext - I'm just wondering how we can arrive at a tractable outcome from this discussion. It seems that you like one aspect of the change (i.e. the `filling` field) but are not generally happy with the overall synchronicity of the approach.

So I'm wondering how we can cherry pick the bits that we do agree on and defer changes to more contentious issues to a subsequent patch (especially given the backdrop that `SimplePool` is likely to get EOL'ed sometime soon).

Maybe splitting out this patch might:

1. Make you more happy with the approach
2. Give us a basis to take a more peu à peu approach to whole subject matter (as opposed to having to land all concerns in one single pull request)

What do you think?

justinretailnext:
To be clear, the reason I refer to SimplePool is that it's been in gocql for a while longer, it's relatively more mature and I modeled policyConnPool and hostConnPool on SimplePool.

I'd suggest reducing the scope of this change to (1) a fix for SetHosts so that it does not hold the write lock while waiting for network connection to be established and (2) the change to the filling field of hostConnPool. That way we can resolve the likely cause of issue #376 and address issue #370.

It seems like it would be useful to have some tests to gain test coverage and profiling data. Particularly with as many locks involved as there are in the Pick function call hierarchy.

Zariel:
Thanks for the commends, I will split this PR into 2. Regarding testing I agree that having more tests for this specific case would be beneficial but it is very tricky to get to a point where this is being tested correctly, I run the test suites locally with the race detector and havn't seen any races for a while which is nice. But I think we should be using the race detector in travis, see #362 and #368.

PR #384, opened 2015-05-02
fuzz fixes

Zariel:
This contains provions to run go-fuzz and includes some fixes to issues it found.

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -328,6 +328,10 @@ func (f *framer) trace() {
 
 // reads a frame form the wire into the framers buffer
 func (f *framer) readFrame(head *frameHeader) error {
+	if head.length < 0 {
+		return fmt.Errorf("frame body length can not be less than 0: %d", head.length)
+	}
+
 	if cap(f.readBuffer) >= head.length {
 		f.rbuf = f.readBuffer[:head.length]
 	} else {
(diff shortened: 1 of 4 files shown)


0x6e6562:
Very interesting - I assume that the changes to `frame.go` are bugs that the fuzzer found?

Zariel:
Yes, pretty expected, I never added checking to what was read from the
buffer.

frame.go
```diff
@@ -1151,18 +1160,28 @@ func (f *framer) readByte() byte {
 }
 
 func (f *framer) readInt() (n int) {
+	if len(f.rbuf) < 4 {
+		panic(fmt.Errorf("not enough bytes in buffer to read int require 4 got: %d", len(f.rbuf)))
```
0x6e6562: We seem to be introducing panics into the library - is this the best way to exit out of these situations?

Zariel: The panics are caught and if they are not a `runtime.Error` then they are assigned as the error return from `parseFrame` so that they propagate out to the caller without having to have lots of control flow in the parse chain. It's the same mechanism that the standard library uses in the same situations such as in json encoding/decoding.

0x6e6562:
Ah yes, this is happening in `parseFrame`.

In that case, this patch LGTM.

0x6e6562:
Nice work - introducing this fuzzing seems like a pretty cool way to flush out latent bugs.

PR #385, opened 2015-05-05
be more careful when reading the headers, check sizes of the buffer and ...

Zariel:
...for valid versions


Zariel:
more bugs from go-fuzz

0x6e6562:
Nice :-)

Zariel:
more coming :)

PR #389, opened 2015-05-11
Ensure that frames are not bigger than the max

Zariel:
When writing frames ensure that we dont write a frame which is
invalid as it is too big, ensure that we dont read a frame which
is too large.

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -5,8 +5,10 @@
 package gocql
 
 import (
+	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


0x6e6562:
Not sure whether merging #393 made this unmergeable?

0x6e6562:
Actually, rollback my last comment - I saw this issue on iOctocat and I think it must have a caching bug.

Zariel:
thanks!

Issue #390, opened 2015-05-11
Failed to unmarshall UserType


nguyener:
I created a table which has one of the columns is User Defined Type (actually a map of User Defined Type).  Let says my table name is 'templates' and the UDT is 'File' which has two attributes: file_name and file_content (both as text).  The UDT column is defined as map[string]File
I have no problem retrieving data from this table using cqlsh.  However, when wrote some code with gocql, I got the following error:
"can not unmashal custom (org.apache.cassandra.db.marshal.UserType(.....) in to *templates.File

Zariel:
Could you post a snippet of code which is doing the query?

nguyener:
type File struct {
    File_name string `json: "file_name"`
    File_content string `json: "file_content"`
}


 var files = make(map[string]File)
    
get_template_str := "SELECT files FROM templates limit 1"
if err:=Conn.Query(get_template_str).Scan(&files); err != nil{
        log.Fatal(err)
}

Here is what my data look likes

cqlsh:demo> select template_name, template_hash, commit_hash, files from templates;

 template_name | template_hash                            | commit_hash                              | files
---------------+------------------------------------------+------------------------------------------+--------------------------------------------------------
    template_1 | 61d09ee12cdfe9b85003ea37dbc8c090fdb5c6f5 | 61d09ee12cdfe9b85003ea37dbc8c090fdb5c6f5 | {'file_1': {file_name: 'file_1', file_content: 'abc'}}
    template_1 | 62e0b4c4f14fea76d120c5827becdedb2a45af03 | 62e0b4c4f14fea76d120c5827becdedb2a45af03 | {'file_2': {file_name: 'file_2', file_content: 'edf'}}
    template_1 | 34e7688cda9d475056e2f206fb446317cb91167d | 34e7688cda9d475056e2f206fb446317cb91167d | {'file_3': {file_name: 'file_3', file_content: 'xyz'}}

nguyener:
cqlsh:demo> select files from templates;

 files
--------------------------------------------------------
 {'file_1': {file_name: 'file_1', file_content: 'abc'}}
 {'file_2': {file_name: 'file_2', file_content: 'edf'}}
 {'file_3': {file_name: 'file_3', file_content: 'xyz'}}

Zariel:
Thanks. This is because currently the UDT unmarshalling wont try to fuzzy match the field names like the json decoder does (ill look at adding that). You can get around this by adding `cql` tags to map the column names.

[...]

PR #394, opened 2015-05-12
Return a helpful error when using UDT in proto 2

Zariel:
When using UDT's in a protocol which does not support UDT's
return a more helpful error indicating that protocol version 3
or above should be used.


Zariel:
This fixes #390 and now returns `ErrorUDTUnavailable`

Zariel:
So UDT's are defiantly only available on 2.1 and above

PR #395, opened 2015-05-13
Fix list unmarshaling

c0va23:
This is an updated and an cleared version of PR #302. This version can be merged without #287.


Zariel:
Is there an example test case which the current code fails for? I would say this looks good to me but could do with a bit of convincing as to what this is solving. Also is your email in the `AUTHORS.md` file?

c0va23:
Example described into #302.
I have already indicated in the list of authors.

Zariel:
Thanks for the contribution!

c0va23:
Thank you for your promptness.

PR #396, opened 2015-05-13
Fixed marshaling of pointer to interface.

c0va23:
Including marshaling typed nils (for example (*CustomString)(nil)).


Zariel:
Thanks!

PR #397, opened 2015-05-15
Fix bitshift order in udt length calculation

nemothekid:
Currently, when marshaling UDTs, the marshaling of the `length` of the UDT's data uses the wrong bit shift order - which prevents UDTs from being larger than 256 bytes. Cassandra returns back with `Invalid remaining data after end of UDT value`.

This PR fixes the shift order, and also provides a test that trigger the big.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -1252,9 +1252,9 @@ func marshalUDT(info TypeInfo, value interface{}) ([]byte, error) {
 			}
 
 			n := len(data)
-			buf = append(buf, byte(n<<24),
-				byte(n<<16),
-				byte(n<<8),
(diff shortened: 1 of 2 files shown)


0x6e6562:
Nice spot - is it worth adding an extra verification to the test that reads the data back out again and hence verifies the complete round trip?

nemothekid:
Apologies for the delay in responding - we caught this bug trying to move a service into Cassandra w/ UDTs - now that this has been merged I can create a new PR for the test.

nemothekid:
Actually it looks like @Zariel extended the test a bit at df143093796ea46746d47f36db9ff829e91de0b0

Issue #398, opened 2015-05-18
Failed to unmarshall "null" UDT object


nguyener:
My table has a UDT column call ReleaseInfo defined as:
CREATE TYPE release_info(release_role text, firmware text)
create table templates (template_id uuid primary key,template_hash text, template_name text, release_info frozen<release_info>);

In my code, I have similar structs:
```go
type ReleaseInfo struct {
    Release_role string `cql:"release_role"`
    Firmware     string `cql:"firmware"`
}

type Template struct {
    Template_id       string            `cql:"template_id"`
    Template_name     string            `cql:"template_name"`
    Template_hash     string            `cql:"template_hash"`
    Release_info      ReleaseInfo       `cql:"release_info"`
}
```

I then executed a simple query such as "SELECT * from templates"
my table is currently has only one row.  When the record has NOT NULL release_info object, the query is success.  However, when the release_info object is NULL, I got error

```
goroutine 1 [running]:
github.com/gocql/gocql.unmarshalUDT(0x7f6916a9e040, 0xc20893a4e0, 0x0, 0x0, 0x0, 0x73dda0, 0xc20805c1d0, 0x0, 0x0)
	/home/nguyener/golang/src/github.com/gocql/gocql/marshal.go:1371 +0xf6c
github.com/gocql/gocql.Unmarshal(0x7f6916a9e040, 0xc20893a4e0, 0x0, 0x0, 0x0, 0x73dda0, 0xc20805c1d0, 0x0, 0x0)
	/home/nguyener/golang/src/github.com/gocql/gocql/marshal.go:136 +0x7e7
github.com/gocql/gocql.(*Iter).Scan(0xc2084ea300, 0xc20802e440, 0x4, 0x4, 0x0)
	/home/nguyener/golang/src/github.com/gocql/gocql/session.go:738 +0x6c8
factorytools.microsoft.com/tf/templates.(*Template).Get_all_releases(0xc2080b9d98, 0x0, 0x0, 0x0)
	/home/nguyener/golang/src/factorytools.microsoft.com/tf/templates/templates.go:251 +0x24b
main.main()
	/home/nguyener/golang/src/factorytools.microsoft.com/tf/tf-server.go:41 +0x10d
```

Zariel:
Thanks for the report, I've reproduced this and added a test case and a fix in #399

Zariel:
Im going to close this, if it still an issue please reopen.

PR #399, opened 2015-05-19
Correctly unmarshal null UDT responses

Zariel:
When a UDT column is null it's data will be empty, when this is the
case set the unmarshal target to its zero value if possible and
return.


Zariel:
This fixes #398

Issue #402, opened 2015-05-27
import error from code.google.com


gingerhot:
I run into a error when running `go get github.com/gocql/gocql` in a Ubuntu 14.04 node, while in my Mac all is fine.

```
package github.com/gocql/gocql
        imports code.google.com/p/snappy-go/snappy: unable to detect version control system for code.google.com/ path
```

How can I fix this? 
Thanks!

0x6e6562:
I'm just going to re-trigger the Travis build to see if it experiences the same issue with `go get`.

0x6e6562:
Yes, as suspected, Travis can also no longer reference the snappy dependency, so we have an issue - from memory there was something about google code being EOL'ed in some fashion, but I might be wrong.

gingerhot:
Thanks a lot!

PR #403, opened 2015-05-27
Redefine snappy import

0x6e6562:
We seem to be getting the following error importing snappy:

```
package code.google.com/p/snappy-go/snappy: unable to detect version control system for code.google.com/ path
```

So this is an attempt to fix this (see #402 for context).

### DIFF
--- a/compressor.go
+++ b/compressor.go
@@ -1,7 +1,7 @@
 package gocql
 
 import (
-	"code.google.com/p/snappy-go/snappy"
+	"github.com/golang/snappy/snappy"
 )
 
 type Compressor interface {
--- a/compressor_test.go
+++ b/compressor_test.go
@@ -4,7 +4,7 @@ package gocql
 
 import (
 	"bytes"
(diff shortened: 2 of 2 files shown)


gingerhot:
Thanks!
And I'll have a try.

0x6e6562:
I've just raised an issue with the apparent current upstream to see if this is the official way forwards: https://github.com/golang/snappy/issues/12

gingerhot:
Ok, I see. Thanks anyway.

0x6e6562:
Ideally I don't want to merge this type of change directly to master without getting the opinion of @Zariel - but in this case the master build is currently broken and this appears to be the fastest way to fix it. So @Zariel - could you please post-review this patch to see whether I've overlooked something.

gingerhot:
I tried again and succeed now.
Thanks again!

0x6e6562:
@gingerhot Nice to see it working for you - do you think we can close #402 now?

PR #405, opened 2015-05-27
Fixing Snappy Imports

dahernan:


dahernan:
Sorry, wrong project, blame github UI

PR #407, opened 2015-06-03
Do not require a client certificate in SslOptions

uhoh-itsmaciek:
The Certificates field in a tls.Config can be empty for a TLS client; this allows that configuration to work for SslOptions.

This simplifies setup of a TLS connection to Cassandra.

If this sounds reasonable, I can do the other changes in CONTRIBUTING.md:

- [x] - add self to AUTHORS
- [x] - add test for TLS connection without CertPath and KeyPath options
- [x] - document the new SslOptions behavior 

(I think everything else is covered; please let me know if I missed anything else)

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -59,7 +59,7 @@ func TestSSLSimple(t *testing.T) {
 	srv := NewSSLTestServer(t, defaultProto)
 	defer srv.Stop()
 
-	db, err := createTestSslCluster(srv.Address, defaultProto).CreateSession()
+	db, err := createTestSslCluster(srv.Address, defaultProto, true).CreateSession()
 	if err != nil {
 		t.Fatalf("0x%x: NewCluster: %v", defaultProto, err)
 	}
@@ -69,13 +69,27 @@ func TestSSLSimple(t *testing.T) {
 	}
 }
 
-func createTestSslCluster(hosts string, proto uint8) *ClusterConfig {
+func TestSSLSimpleNoClientCert(t *testing.T) {
+	srv := NewSSLTestServer(t, defaultProto)
+	defer srv.Stop()
+
+	db, err := createTestSslCluster(srv.Address, defaultProto, false).CreateSession()
+	if err != nil {
+		t.Fatalf("0x%x: NewCluster: %v", defaultProto, err)
+	}
+
+	if err := db.Query("void").Exec(); err != nil {
+		t.Fatalf("0x%x: %v", defaultProto, err)
+	}
+}
+
+func createTestSslCluster(hosts string, proto uint8, useClientCert bool) *ClusterConfig {
 	cluster := NewCluster(hosts)
 	sslOpts := &SslOptions{
 		CaPath:                 "testdata/pki/ca.crt",
 		EnableHostVerification: false,
 	}
-	if os.Getenv("REQ_CLIENT_AUTH") != "false" {
+	if useClientCert {
 		sslOpts.CertPath = "testdata/pki/gocql.crt"
 		sslOpts.KeyPath = "testdata/pki/gocql.key"
 	}
@@ -451,7 +465,7 @@ func TestPolicyConnPoolSSL(t *testing.T) {
 	srv := NewSSLTestServer(t, defaultProto)
 	defer srv.Stop()
 
-	cluster := createTestSslCluster(srv.Address, defaultProto)
+	cluster := createTestSslCluster(srv.Address, defaultProto, true)
 	cluster.ConnPoolType = NewRoundRobinConnPool
 
 	db, err := cluster.CreateSession()
--- a/conn_test.go
+++ b/conn_test.go
@@ -12,7 +12,6 @@ import (
 	"io"
 	"io/ioutil"
 	"net"
-	"os"
 	"strings"
 	"sync"
 	"sync/atomic"
--- a/conn_test.go
+++ b/conn_test.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net"
+	"os"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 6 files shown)


0x6e6562:
Just to clarify your intention for this change: is it that you want to be able to set up a TLS connection without certificates?

uhoh-itsmaciek:
@0x6e6562 right, yeah. If you look at the way the Postgres driver `pq` does it, the client cert config [is optional](https://github.com/lib/pq/blob/master/conn.go#L972) (disclaimer: I co-maintain pq, although I didn't write or review that code).

0x6e6562:
OK, cool, that's fine - I just wasn't 100% sure what the intention was, but this makes sense. The critical bit is this pull request may well be the integration testing - hopefully `ccm` can be configured to not mandate certificates from the client.

Zariel:
Yep in ccm you should be able to just remove `"client_encryption_options.require_client_auth: true"` from `integration.sh`. Though now im starting to get slightly concerned that our test matrix is becoming huge, but this change shouldn't be too much of a problem to test.

I would suggest removing the force client certs option then add a test which checks that it is possible to dial without a cert, I don't think it requires a rigorous test as this is more of a semantics change than a functional one.

(btw, love the formatting of the PR)

uhoh-itsmaciek:
> OK, cool, that's fine - I just wasn't 100% sure what the intention was, but this makes sense.

Great.

> Yep in ccm you should be able to just remove "client_encryption_options.require_client_auth: true" from integration.sh.

@Zariel You mean, as another env var for the test matrix? So add this to all the versions?

> (btw, love the formatting of the PR)

You mean the checklist? Thanks! I learned it from @tef

uhoh-itsmaciek:
Also, if we do just add this as a dimension to the test matrix, should we check to see if client certs are accepted if they're not required? That is, do I just add a dimension to the matrix and change `createTestSslCluster` to take it into account? Or do I actually add a separate test in `conn_test.go`, add an option to omit certs to `createTestSslCluster`, run the existing test as is and have the new test pass an option to omit certs if `require_client_auth` is `false` and do nothing otherwise.

Zariel:
I think the second option should suffice.

uhoh-itsmaciek:
I've implemented the above. It looks like Travis failures are due to `go get` problems with `groupcache`, and not with my patch. However, I initially accidentally reversed the test [here](https://github.com/gocql/gocql/pull/407/files#diff-f9df1a974544ddcc7ccada9d10f705a4R78) (I was testing for `== "false"`), and everything passed, so there's something strange going on. Any ideas? There's no way both of those configurations can pass, right?

Also, I just realized that my usage of the env var is totally different from how the AUTH test matrix is passed; I can fix that.

conn_test.go
```diff
@@ -70,12 +71,15 @@ func TestSSLSimple(t *testing.T) {
...
 func createTestSslCluster(hosts string, proto uint8) *ClusterConfig {
 	cluster := NewCluster(hosts)
-	cluster.SslOpts = &SslOptions{
-		CertPath:               "testdata/pki/gocql.crt",
-		KeyPath:                "testdata/pki/gocql.key",
+	sslOpts := &SslOptions{
 		CaPath:                 "testdata/pki/ca.crt",
 		EnableHostVerification: false,
 	}
+	if os.Getenv("REQ_CLIENT_AUTH") != "false" {
```
Zariel: I guess this should be

`if os.Getenv("REQ_CLIENT_AUTH") == "true"`

Zariel:
I would rather not add more entries to the travis build as it already takes long enough with 8. I think it would make sense to have the test suite either run in a mode which requires SSL certs or one which doesn't and then have an explicit test for the other. I think having the test suite run without SSL client auth forced makes sense because then any failure in the SSL auth code will be apparent in its test. 

I can make these changes in a follow up PR though you like.

The failure in travis looks like github returned a 503 so no issue there.

conn_test.go
uhoh-itsmaciek: That depends on whether we want to default to running with or without, right? Happy to change it, though.

uhoh-itsmaciek:
Ok. I thought this is what you were suggesting, but yeah, it does blow out the matrix.

> I think it would make sense to have the test suite either run in a mode which requires SSL certs or one which doesn't and then have an explicit test for the other.

Isn't the problem that the backend actually needs to be configured differently for this (i.e., changing `client_encryption_options.require_client_auth` in `integration.sh`)? I'm still very new to Cassandra so maybe I don't know what I'm talking about. And yeah, if you know what this should look like and can adapt my patch, that'd be great--I'm kind of stumbling around now =).

Zariel:
>Ok. I thought this is what you were suggesting, but yeah, it does blow out the matrix.
Sorry for the confusion!

>> I think it would make sense to have the test suite either run in a mode which requires SSL certs or one which doesn't and then have an explicit test for the other.

>Isn't the problem that the backend actually needs to be configured differently for this (i.e., changing `client_encryption_options.require_client_auth` in `integration.sh`)? I'm still very new to Cassandra so maybe I don't know what I'm talking about. And yeah, if you know what this should look like and can adapt my patch, that'd be great--I'm kind of stumbling around now =).

From my memory and previous testing having the certs enabled will allow the client to connect via TLS without enforcing the client to supply a cert.

Looks like you need to rebase to fix the authors file after #408 was merged.

uhoh-itsmaciek:
Ok, in that case I think I know what to do; I can update the patch later today or tomorrow.

uhoh-itsmaciek:
Ok, I've rebased, made the changes (or at least what I understood you to be suggesting) and force-pushed. I can squash everything if this looks good.

uhoh-itsmaciek:
Hmm, Travis doesn't like it on one of the matrix flavors, but I'm not sure what's up:

```
=== RUN TestMarshalInet-2

--- FAIL: TestMarshalInet-2 (11.36 seconds)

	cassandra_test.go:1316: insert string inet: unconfigured columnfamily inet_test

=== RUN TestVarint-2

--- FAIL: TestVarint-2 (11.49 seconds)

	cassandra_test.go:1367: insert varint: unconfigured columnfamily varint_test

=== RUN TestQueryStats-2

--- PASS: TestQueryStats-2 (0.27 seconds)

=== RUN TestBatchStats-2

--- FAIL: TestBatchStats-2 (11.51 seconds)

	cassandra_test.go:1478: query failed. unconfigured columnfamily batchstats

=== RUN TestNilInQuery-2

--- FAIL: TestNilInQuery-2 (12.49 seconds)

	cassandra_test.go:1499: failed to insert with err: unconfigured columnfamily testnilinsert

=== RUN TestEmptyTimestamp-2

--- PASS: TestEmptyTimestamp-2 (11.54 seconds)

=== RUN TestGetKeyspaceMetadata-2

--- PASS: TestGetKeyspaceMetadata-2 (0.26 seconds)

=== RUN TestGetTableMetadata-2

--- FAIL: TestGetTableMetadata-2 (1.51 seconds)

	cassandra_test.go:1627: Expected table metadata for name 'test_table_metadata'
```

0x6e6562:
Sometimes we get the odd test failure due to resource issues in the VMs, so I've kicked off the build again to see if we can reproduce this.

0x6e6562:
OK so after manually re-triggering the build, it looks like we now have a mergeable patch.

0x6e6562:
Many thanks for your contribution, much appreciated, especially for the way you used the to-do list in the patch summary :-)

Zariel:
Thanks!

uhoh-itsmaciek:
No problem! Thanks for reviewing and merging.

PR #408, opened 2015-06-08
Added closeChan to clean up ring describer when session is closed

zachbadgett:
Currently, when setting 'DiscoverHosts' to 'true', you open up a goroutine that periodically checks for host updates; this goroutine is never cleaned up (goroutine leak) when the session is closed and it will display 'RingDescriber: unable to get ring topology: use of closed network connection' continuously until the program exits.

My pull request adds a channel to 'ringDescriber'; adds a select statement to the 'run' method to check when the channel is closed; 'hostSource' variable is added to 'Session' (I'm not sure if keeping 'Session' agnostic to 'ringDescriber' is important or not); on session close we check to see if 'session.hostSource' is nil, if not, close the 'closeChan' that will clean up the goroutine.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -45,3 +45,4 @@ Dan Kinder <dkinder.is.me@gmail.com>
 Oliver Beattie <oliver@obeattie.com>
 Justin Corpron <justin@retailnext.com>
 Miles Delahunty <miles.delahunty@gmail.com>
+Zach Badgett <zach.badgett@gmail.com>
--- a/host_source.go
+++ b/host_source.go
@@ -97,23 +97,21 @@ func (h *ringDescriber) run(sleep time.Duration) {
 	}
 
 	for {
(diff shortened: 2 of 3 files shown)


nemothekid:
Might be cleaner with something like - 

```go
check := time.NewTimer(sleep) 
for {
  select {
    case <-check.C:
      //  h.GetHosts() 
    case <-h.closeChan:
      return
  } 
}
```

Also has the advantage that the function returns ASAP instead of holding onto that goroutine for the sleep period.

nemothekid:
Actually your current code with the `select` will actually block for the lifetime of the function and without a `default` clause it will never run `getHosts()`

zachbadgett:
Yeah, fixing.

Zariel:
Thanks for the contribution!

0x6e6562:
Nice spot @nemothekid

PR #412, opened 2015-06-12
Dont allocate all the columninfo upfront

Zariel:
If the column count for a row result is huge it will trigger either
a makeslice: len out of range or the program will do lots of gc
which make the application unresponsive, even if there are not really
that many columns.


Zariel:
This was found by fuzzing, the frame says that it has 808464432 columns, which on wind64 causes a makeslice: len out of range panic, on darwin64 it causes an 80s garbage collection.

PR #414, opened 2015-06-15
Remove limitations of SslOptions by making it inherit from tls.Config.

jefferai:
The helper functions simply toggle internal state, now, and all the
flexibility of tls.Config is available.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -47,3 +47,4 @@ Justin Corpron <justin@retailnext.com>
 Miles Delahunty <miles.delahunty@gmail.com>
 Zach Badgett <zach.badgett@gmail.com>
 Maciek Sakrejda <maciek@heroku.com>
+Jeff Mitchell <jeffrey.mitchell@gmail.com>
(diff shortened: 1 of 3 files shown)


0x6e6562:
Many thanks for your contribution - this looks like a worthwhile simplification to me. Could you add yourself to the AUTHORS file as well please (as per the contribution guidelines)?

Zariel:
I wonder if it is worth keeping the SSLOptions struct around any longer?

0x6e6562:
> I wonder if it is worth keeping the SSLOptions struct around any longer?

Can the fields they currently provide get provided by `tis.Config`?

Zariel:
It seems to now only provide the convenience of loading certs from paths, that can be done pretty easily by the calling application.

jefferai:
I've added myself to AUTHORS and rewrote the commit since it was so minor.

Regarding keeping SSLOptions around -- it's really up to you. The TLS package is complex and things are split between that and the x509 package so for basic uses/users it probably provides some beneficial simplification. But it does mean maintaining more code, and some users could miss the fact that it has tls.Config as an anonymous member, and think that it can't do what they need.

Zariel:
I think we should leave it in for now as I agree, it does make it easier to use.

Thanks for the contribution!

jefferai:
Thanks for merging  :-)

PR #417, opened 2015-06-16
apply a timeout for every query at the connection

### LINKED ISSUE #416: connection does not timeout queries
Currently the conn will only set a read deadline once it has received a frame header, which means that quries can hang indefinably if nothing is received from the server.

Zariel:
Start a timeout when waiting for the response from cassandra
in the call to exec() so that if no header is read the query
will not hang for ever.

If there are more than a limit of timeouts then the connection will
be closed. This may need tuning in the future if there are queries
which take a long time.

fixes #416

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -15,6 +15,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -82,6 +83,11 @@ type ConnErrorHandler interface {
 	HandleError(conn *Conn, err error, closed bool)
 }
 
+// How many timeouts we will allow to occur before the connection is closed
+// and restarted. This is to prevent a single query timeout from killing a connection
+// which may be serving more queries just fine.
+const timeoutLimit = 10
+
 // Conn is a single connection to a Cassandra node. It can be used to execute
 // queries, but users are usually advised to use a more reliable, higher
 // level API.
@@ -105,6 +111,8 @@ type Conn struct {
 
 	closedMu sync.RWMutex
 	isClosed bool
+
+	timeouts int64
 }
 
 // Connect establishes a connection to a Cassandra node.
@@ -296,7 +304,16 @@ func (c *Conn) serve() {
 		}
 	}
 
+	c.closeWithError(err)
+}
+
+func (c *Conn) closeWithError(err error) {
+	if c.Closed() {
+		return
+	}
+
 	c.Close()
+
 	for id := 0; id < len(c.calls); id++ {
 		req := &c.calls[id]
 		// we need to send the error to all waiting queries, put the state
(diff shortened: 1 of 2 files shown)


Zariel:
@0x6e6562 any opposition to removing TestSlowQuery as I dont think it actually tests anything.

conn.go
```diff
@@ -337,7 +354,11 @@ func (c *Conn) recv() error {
 
 	// once we get to here we know that the caller must be waiting and that there
 	// is no error.
-	call.resp <- nil
+	select {
+	case call.resp <- nil:
+	default:
+		// incase the caller timedout
```
0x6e6562: Should this read "in case the caller times out"?

conn.go
```diff
@@ -82,6 +83,11 @@ type ConnErrorHandler interface {
 	HandleError(conn *Conn, err error, closed bool)
 }
 
+// How many timeouts we will allow to occur before the connection is closed
+// and restarted. This is to prevent a single query timeout from killing a connection
+// which may be serving more queries just fine.
+const timeoutLimit = 10
```
0x6e6562: How is this value set?

conn_test.go
```diff
@@ -486,6 +471,40 @@ func TestPolicyConnPoolSSL(t *testing.T) {
...
+
+	ch := make(chan error, 1)
+
+	go func() {
+		err := db.Query("timeout").Exec()
+		if err != nil {
+			ch <- err
+			return
+		}
+		t.Error("error was nil for timeout")
```
0x6e6562: I think this should say that we expected a timeout, but did not get one, hence we are failing this test.

conn_test.go
```diff
@@ -182,21 +182,6 @@ func TestQueryRetry(t *testing.T) {
 	}
 }
 
-func TestSlowQuery(t *testing.T) {
```
0x6e6562: Removing this is fine by me.

conn_test.go
```diff
@@ -486,6 +471,40 @@ func TestPolicyConnPoolSSL(t *testing.T) {
 	}
 }
 
+func TestQueryTimeout(t *testing.T) {
+	srv := NewTestServer(t, protoVersion2)
+	defer srv.Stop()
+
+	cluster := NewCluster(srv.Address)
+	cluster.Timeout = 250 * time.Millisecond
```
0x6e6562: What is the significance of 250 ms?

conn.go
Zariel: its pretty arbitrary, im not so keen on it being configurable at the session level, maybe globally?

conn_test.go
Zariel: just an arbitrary timeout which is low

0x6e6562: Is 250ms low enough - why not set it to 1ms, to make quite clear that you want the query to time out? Also, why not put a comment in the code so that we remember why we set the timeout to that particular value?

conn.go
0x6e6562: Globally is fine, as long as it is settable - somebody is bound to come along and want to tweak this.

PR #419, opened 2015-06-17
Marshal fixed for pointers implementing Marshaler

bfontaine:
This is a fix for the following situation:

```go
type T struct { /* ... */ }

// MarshalCQL implements the gocql.Marshaler interface
func (t *T) MarshalCQL(info TypeInfo) ([]byte, error) {
    // ...
}

func laterInTheCode() {

    t := &T{ /* ... */ }

    data, err := gocql.Marshal(info, t) // <-- won't work

    // ...
}
```

`Marshal` resolves a pointer value before checking if it implements the `Marshaler` interface. The code above won’t work because `T` doesn’t implement the `Marshaler` interface, only `*T` does.

I added a test that shows the bug. The test fails without the fix and passes with it.

---

Running `go fmt` on `marshal_test.go` fixes other parts of the code but it polluted the diff so I didn’t include the changes here. Do you want me to include them?

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -48,3 +48,4 @@ Miles Delahunty <miles.delahunty@gmail.com>
 Zach Badgett <zach.badgett@gmail.com>
 Maciek Sakrejda <maciek@heroku.com>
 Jeff Mitchell <jeffrey.mitchell@gmail.com>
+Baptiste Fontaine <b@ptistefontaine.fr>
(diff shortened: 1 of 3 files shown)


Zariel:

Zariel:
Thanks for pointing out the fmt issues, I've fixed them in master.

balaweblog:
Hi, How to pass the info  to gocql.Marshal(), I am trying to typecast data, what object should I pass for the "info". My call is outside the gocql package.

data, err := gocql.Marshal(info, t)

Zariel:
Hi @balaweblog this isn't the best place to ask this question, in the future these types of questions are best for the mailinglist or gitter. That being said TypeInfo is a representation of Cassandra's type information, which we use to work out how to convert a Go type into a valid Cassandra type binary representation.

PR #420, opened 2015-06-19
SimpleRetryPolicy doc format fixed

bfontaine:
This fixes the indentation of the example code, which was wrongly indented in the online Godoc.

With a local `godoc` server, here’s how it’s now:

before

And how it’s with this patch:

after


Zariel:
Thank you!

Issue #426, opened 2015-06-21
nil pointer in exec after timeouts


Zariel:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x5eac21]
goroutine 1393738 [running]:
github.com/gocql/gocql.(Conn).exec(0xc20813cb40, 0x7f4eeaf03110, 0xc2083803c0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
/go/src/github.com/gocql/gocql/conn.go:393 +0x391
github.com/gocql/gocql.(Conn).executeQuery(0xc20813cb40, 0xc2083fc9c0, 0x9c76e0)
/go/src/github.com/gocql/gocql/conn.go:529 +0x5f4
github.com/gocql/gocql.(Session).executeQuery(0xc208038820, 0xc2083fc9c0, 0x7f4650)
/go/src/github.com/gocql/gocql/session.go:222 +0x1da
github.com/gocql/gocql.(Query).Iter(0xc2083fc9c0, 0xc2083fc9c0)
/go/src/github.com/gocql/gocql/session.go:613 +0x107
github.com/gocql/gocql.(*Query).Exec(0xc2083fc9c0, 0x0, 0x0)
/go/src/github.com/gocql/gocql/session.go:603 +0x3a
```

Zariel:
@xoraes what conditions trigger this? Is it after timeouts? I think there could be a data race when there is a timeout accessing the call.framer, but not sure. Could you run with the race detector?

xoraes:
Sorry I had to make some major changes to our prod env (moved single dc to multi-dc cluster) to prevent timeouts and consequently I am no longer seeing the issue as of today. I no longer have timeouts and I do not see panic anymore. Looks like your have disabled timeout for now?

Zariel:
@xoraes yeah I decided to remove timeouts until I could reproduce this error and clearly understand the races going on.

I think I have been able to create a test that reproduces this and have a fix in place to enable timeouts again.

The failure shown is as follows,
```
=== RUN TestQueryTimeoutReuseStream-4
==================
WARNING: DATA RACE
Read by goroutine 9:
  github.com/gocql/gocql.(*Conn).recv()
      /Users/chris/dev/go/src/github.com/gocql/gocql/conn.go:328 +0x295
  github.com/gocql/gocql.(*Conn).serve()
[...]

PR #427, opened 2015-06-21
Stream timeouts

Zariel:


Zariel:
This might be related to #426

PR #429, opened 2015-06-22
Reinstate query timeouts

### LINKED ISSUE #426: nil pointer in exec after timeouts

Zariel:
Correctly synchronise the response timeout handling between waiting for the response and sending the response back. Improve handling of streams so they are returned in the correct place, ensure that streams are returned always. Improve handling when a connection is closed so queries don't get blocked in exec.

Add a test which timesout on a stream then execute another query before the first query returns which triggers the previous error when reusing streams.

Fixes #426 #417

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -348,6 +348,7 @@ func (c *Conn) recv() error {
 	select {
 	case call.resp <- err:
 	case <-call.timeout:
+		c.releaseStream(head.stream)
 	case <-c.quit:
 	}
 
--- a/conn.go
+++ b/conn.go
@@ -611,12 +611,11 @@ func (c *Conn) closeWithError(err error) {
 		req := &c.calls[id]
 		// we need to send the error to all waiting queries, put the state
 		// of this conn into not active so that it can not execute any queries.
+		atomic.StoreInt32(&req.waiting, 0)
 		select {
 		case req.resp <- err:
 		default:
 		}
-
-		close(req.resp)
 	}
 
 	c.conn.Close()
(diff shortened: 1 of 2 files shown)


Zariel:
@xoraes I think this should fix the issue you have been seeing, could you try this?

Zariel:
Looking at previous travis runs we were htiting the timeout but queries were still running, will look into what needs changing. I think I might just remove the test streams vX tests as they dont really add much value and end up taking a long time.

Zariel:
Fixed the issues which were causing tests to fail, improved synchronisation when waiting for queries to finish and timeouts, and improved handling of closing connections.

Issue #430, opened 2015-06-24
Receiving error: Connection closed waiting for response -- version issue?


masonoise:
I'm wondering if this could be a version mismatch error, though that seems unlikely -- perhaps I'm missing something. I'm using Cassandra 2.0.10, go version `go1.4.2 darwin/amd64`.

Connecting via cqlsh works:
````
$ CQLSH_PORT=9042 ./cqlsh
Connected to Test Cluster at localhost:9042.
[cqlsh 4.1.1 | Cassandra 2.0.10 | CQL spec 3.1.1 | Thrift protocol 19.39.0]
Use HELP for help.
cqlsh>
````

Connecting with gocql gives an error as follows:
````
2015/06/23 21:29:15 connect: failed to connect to "localhost:9042": gocql: connection closed waiting for response
runtime error: invalid memory address or nil pointer dereference
````

With the code:
````
cluster := gocql.NewCluster("localhost")
cluster.Keyspace = "testing"
session, _ := cluster.CreateSession()
defer session.Close()
err := session.Query("SELECT summary,time from testing where foo = 'bar'").Scan(&summary, &time)
````

I must be missing something, but everything matches up so it seems like it should be working. Any thoughts would be great, I'm happy to test things out if it'll help. Thanks very much.

Zariel:
Thanks for the report. Could you specify what version of gocql you are using, and the stack trace from the nil pointer. Also it looks like you are missing an error being returned from createSession.

masonoise:
Sure, thanks for checking. Here's the full stack trace. I'm very new to go so I'm not sure how to check the version of gocql, but I just pulled it down yesterday so it's presumably the latest. The commit sha in the gocql dir is `46cf8d711076fd4e4bdd89f680b127a9b46ed90a`. I will investigate the createSession call and see if I can find any error that's not showing up there.

````
2015/06/23 21:29:15 connect: failed to connect to "localhost:9042": gocql: connection closed waiting for response
runtime error: invalid memory address or nil pointer dereference
/Users/mason/workspace/gowork/src/github.com/ant0ine/go-json-rest/rest/recover.go:41 (0x46558)
        func.019: trace := debug.Stack()
/usr/local/Cellar/go/1.4.2/libexec/src/runtime/asm_amd64.s:401 (0x3cbf5)
        call16: CALLFN(·call16, 16)
/usr/local/Cellar/go/1.4.2/libexec/src/runtime/panic.go:387 (0x14b28)
[...]

PR #431, opened 2015-06-25
dont swallow the error from closeWithError

Zariel:
Try to send the error to the waiting requests so that their
callers will see the correct error before closing the quit
channel.

Ensure that this error is not nil so that exec() will not see
a nil frame.


Zariel:
This might help with #430

PR #438, opened 2015-07-04
Set Cluster Default PageSize to 5000

phstudy:
If we do not set default `PageSize` to a reasonable number, the query may cause whole cassandra cluster OOM. The [default page size of cassandra java driver is 5000](http://docs.datastax.com/en/drivers/java/2.0/com/datastax/driver/core/QueryOptions.html#DEFAULT_FETCH_SIZE). 

More information can be found in this issue: https://issues.apache.org/jira/browse/CASSANDRA-9607

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -72,7 +72,7 @@ type ClusterConfig struct {
 	DiscoverHosts     bool              // If set, gocql will attempt to automatically discover other members of the Cassandra cluster (default: false)
 	MaxPreparedStmts  int               // Sets the maximum cache size for prepared statements globally for gocql (default: 1000)
 	MaxRoutingKeyInfo int               // Sets the maximum cache size for query info about statements for each session (default: 1000)
-	PageSize          int               // Default page size to use for created sessions (default: 0)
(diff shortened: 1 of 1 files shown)


Zariel:
Thanks

15astro:
Is it recommended to increase PageSize than the default value(5000) to reduce no to retrieve large results?

Example: To fetch 50000 records from the table, having page size 5000 will result in 10 iterations to return the results. However, page size 20000 may result in 3 iterations to fetch watching records.

phstudy:
@15astro It depends on yours row type and the number of columns.

If you query single partition with few columns, it should fine to increase the PageSize. However, If you query multiple partitions with many columns, increasing PageSize may cause read timeout and Cassandra OOM.

15astro:
This specific use case makes multiple queries to Cassandra. Every query targets only a single distinct partition. Every partition has 40,000 records
Case 1: Fetch data for a single partition using default PageSize as 5k
Case 2: Fetch data for a single partition using default PageSize as 20k

Is it recommended to change the PageSize for this use case?

PR #441, opened 2015-07-08
Fix timestamp type marshaling

mattheath:
The `Timestamp` CQL type is represented by a 64bit integer number of ms since the epoch, stored as 8 bytes. Currently marshaling of the `time.Time` type into this uses the `UnixNano()` func, which can only represent times between approximately the years 1677 and 2262 before overflowing.

This means that the zero time (`0001-01-01 00:00:00 +0000 UTC`) can _not_ be represented within this range, as mentioned in the [time pkg documentation](http://golang.org/pkg/time/#Time.UnixNano): _"the result of calling UnixNano on the zero Time is undefined"_. And as a side effect, marshaling a zero time into the cassandra results in a time of approx `1754-08-30 22:43:41.129 +0000 UTC` rather than zero. This is then correctly unmarshalled back out of cassandra, so the data stored does not equal the expected read data. An example can be seen here: http://play.golang.org/p/mY0KhtD1DF

This PR changes this behaviour to take the seconds from the epoch, and separately combine the `ms` component from `Nanoseconds`, preventing overflowing. There are test cases which demonstrate the overflow with the addition or subtraction of a single nanosecond.

In addition I think it makes sense to store the actual zero time, as a blank `[]byte` rather than its numerical value of `-62135596800000` (or bytes "\xff\xff\xc7\x7c\xed\xd3\x28\x00") , so this has been added as a special case in the marshaling step. Note that the `unmarshalTimestamp` method already supports unmarshaling data with a len of 0.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -49,3 +49,4 @@ Zach Badgett <zach.badgett@gmail.com>
 Maciek Sakrejda <maciek@heroku.com>
 Jeff Mitchell <jeffrey.mitchell@gmail.com>
 Baptiste Fontaine <b@ptistefontaine.fr>
+Matt Heath <matt@mattheath.com>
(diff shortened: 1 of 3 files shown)


Zariel:
Thanks!

PR #442, opened 2015-07-15
Improve support for complicated cassandra schemas.

justinretailnext:
In some schemas there may be "partition_key" type columns returned in the V2 protocol metadata which have the same component index; in our case we have a couple column families with aliased columns "KEY" and "key" for historical reasons. When the token-aware connection pool is used with these column families, a nil-pointer panic occurs since the partition key length (column count) was being incorrectly calculated; This change is based on the DataStax driver's behavior for columns with the same component index, which prefers the last column in order defined for that component index; the consequence of this behavior is that other column names defined for the same component index will not be identified as part of a routing key.

Also, "compact_value" has been added as a column kind/type.

[Review on Reviewable](https://reviewable.io/reviews/gocql/gocql/442)

### DIFF
--- a/metadata.go
+++ b/metadata.go
@@ -36,6 +36,7 @@ type TableMetadata struct {
 	PartitionKey      []*ColumnMetadata
 	ClusteringColumns []*ColumnMetadata
 	Columns           map[string]*ColumnMetadata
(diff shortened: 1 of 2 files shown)


Zariel:

Issue #439, opened 2015-07-06
panic: nil reference in framer.readFrame on f.readBuffer


dankinder:
We recently had gocql panic with the trace below (go 1.4, GOMAXPROCS=8). Our application does many millions of reads and writes daily and gocql has been stable; I have only seen this exception once.

It seems to be a nil reference in frame.go:365, which in my version of the codebase (we generally just use master) is this:
```go
    if cap(f.readBuffer) >= head.length {
```
It seems unlikely that head would be nil since it was referenced previously in this function so `f.readBuffer` seems the likely culprit. Two lines down I do see:
```go
        f.readBuffer = make([]byte, head.length)
```
If readFrame could be called concurrently then I believe this operation is not safe. I haven't personally run the [race detector](https://golang.org/doc/articles/race_detector.html) on this code but that may help.

```
2015/07/03 07:01:39 connect: failed to connect to "10.5.24.113:9042": gocql: no response recieved from cassandra within timeout period                                                                             
panic: runtime error: invalid memory address or nil pointer dereference                                                                                                                                            
[signal 0xb code=0x1 addr=0x70 pc=0x727b99]                                                                                                                                                                        
                                                                                                                                                                                                                   
goroutine 20188720 [running]:                                                                                                                                                                                      
github.com/gocql/gocql.(*framer).readFrame(0x0, 0xc2657174a0, 0x0, 0x0)                                                                                                                                            
    /var/lib/jenkins/jobs/walker-master-build/gopath/src/github.com/gocql/gocql/frame.go:365 +0x309                                                                                                                
github.com/gocql/gocql.(*Conn).recv(0xc27dc3d6c0, 0x0, 0x0)                                                                                                                                                        
    /var/lib/jenkins/jobs/walker-master-build/gopath/src/github.com/gocql/gocql/conn.go:330 +0x1c7                                                                                                                 
github.com/gocql/gocql.(*Conn).serve(0xc27dc3d6c0)                                                                                                                                                                 
    /var/lib/jenkins/jobs/walker-master-build/gopath/src/github.com/gocql/gocql/conn.go:305 +0x3a                                                                                                                  
created by github.com/gocql/gocql.Connect                                                                                                                                                                          
    /var/lib/jenkins/jobs/walker-master-build/gopath/src/github.com/gocql/gocql/conn.go:187 +0x727
```

Zariel:
[...]

PR #448, opened 2015-07-20
conn: ensure that call.framer is not nil

Zariel:
If a connection is closed then call to closeWithError will send
the error to the callers resp channel, which the caller in exec
will unblock, release the stream which will set the framer to nil.
Then if the response for that stream comes back from Cassandra into
recv() then the call.framer will be nil and will panic.

Instead of always releasing the stream when getting a reponse from
the channel only do it if there is no error.

Updates #439


clutchski:
I gave it another try, and alas, it still had the error. I'll try race one more time as well.

PR #450, opened 2015-07-26
conn: dont use stream 0 for outgoing requests

Zariel:
If Cassandra has an error and does not honor the protocol and returns
a response on a different stream than the request then the stream will
be 0, this can happen if there is a corrputed frame or an invalid
protocol version.

Detect the responses on stream 0 and cause it to be sent to all
outstanding requests so that they don't block indefinatly, as they may
have caused the original error.


Zariel:
This might be also related to #439

PR #451, opened 2015-08-07
Implement stringer interface for the Query type

johnweldon:
To make logging a little nicer in downstream usage.

### DIFF
--- a/session.go
+++ b/session.go
@@ -435,7 +435,7 @@ type Query struct {
 }
 
 // String implements the stringer interface.
-func (q *Query) String() string {
+func (q Query) String() string {
 	return fmt.Sprintf("[query statement=%q values=%+v consistency=%s]", q.stmt, q.values, q.cons)
 }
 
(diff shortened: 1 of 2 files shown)


Zariel:
Hi, thanks for this. It might be worth making this consistent with the other string methods in frame, something like

`[query statment=%q values=%+#v consistency=%s]`

johnweldon:
makes sense - will do
Thanks

Zariel:
Also if you are not already could you add yourself to the AUTHORS file, thanks.

session.go
```diff
@@ -434,6 +434,11 @@ type Query struct {
 	defaultTimestamp bool
 }
 
+// String implements the stringer interface.
+func (q *Query) String() string {
```
bfontaine: Why did you implement it on a pointer receiver?

johnweldon: Good catch @bfontaine - it was actually a mistake I've made for a few months now.  I mis-interpreted [Effective Go](https://golang.org/doc/effective_go.html#pointers_vs_values) and have been "doing it wrong" until you asked the question :)

I'll fix this.

johnweldon:
Okay @bfontaine and @Zariel - I think this is finally ready to merge :)

Zariel:

PR #454, opened 2015-08-25
Routing key improvements

abustany:
This speeds up things quite a bit when doing lots of small insert queries against Cassandra, by reducing the load on the garbage collector. I'm not 100% sure about the correctness of "no need to include keyspace in routingKeyInfoCache key", but since the session keyspace seems to be immutable, including it in a map key when the map is per-session doesn't seem needed.

benchcmp (when running with benchtime 30s to let some time for the GC to kick in):
```
benchmark                   old ns/op     new ns/op     delta
BenchmarkConnRoutingKey     654           421           -35.63%

benchmark                   old allocs     new allocs     delta
BenchmarkConnRoutingKey     5              4              -20.00%

benchmark                   old bytes     new bytes     delta
BenchmarkConnRoutingKey     112           48            -57.14%
```

### DIFF
--- a/session.go
+++ b/session.go
@@ -422,6 +422,7 @@ type Query struct {
 	cons             Consistency
 	pageSize         int
 	routingKey       []byte
+	routingKeyBuffer []byte
 	pageState        []byte
 	prefetch         float64
 	trace            Tracer
@@ -532,8 +533,14 @@ func (q *Query) GetRoutingKey() ([]byte, error) {
 		return routingKey, nil
 	}
 
+	// We allocate that buffer only once, so that further re-bind/exec of the
+	// same query don't allocate more memory.
+	if q.routingKeyBuffer == nil {
+		q.routingKeyBuffer = make([]byte, 0, 256)
+	}
+
 	// composite routing key
-	buf := bytes.NewBuffer(make([]byte, 0, 256))
+	buf := bytes.NewBuffer(q.routingKeyBuffer)
 	for i := range routingKeyInfo.indexes {
 		encoded, err := Marshal(
 			routingKeyInfo.types[i],
--- a/session.go
+++ b/session.go
@@ -542,7 +542,9 @@ func (q *Query) GetRoutingKey() ([]byte, error) {
 		if err != nil {
 			return nil, err
 		}
-		binary.Write(buf, binary.BigEndian, int16(len(encoded)))
+		lenBuf := []byte{0x00, 0x00}
+		binary.BigEndian.PutUint16(lenBuf, uint16(len(encoded)))
+		buf.Write(lenBuf)
 		buf.Write(encoded)
 		buf.WriteByte(0x00)
 	}
--- a/session.go
+++ b/session.go
@@ -533,7 +533,7 @@ func (q *Query) GetRoutingKey() ([]byte, error) {
 	}
 
 	// composite routing key
-	buf := &bytes.Buffer{}
+	buf := bytes.NewBuffer(make([]byte, 0, 256))
 	for i := range routingKeyInfo.indexes {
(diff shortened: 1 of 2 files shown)


abustany:
Looks like the CI failed due to some unrelated error... Also, this code won't work with Go 1.3, it uses sync.Pool which appeared in 1.4. As 1.5 has just been released, do we want to keep compatibility with Go 1.3?

abustany:
Another side note: the next "offenders" when it comes to CPU cost when doing lots of small queries are:

- `Session.routingKeyInfo` that uses a lru cache, that lru cache uses a map[interface] as a cache, which is much slower than using a map[string] (we use the statement as a key). Forking that LRU and adapting it to use strings as keys could be an option.
- CQL marshaling. I haven't looked yet deeply into that issue, but since it's based on reflection we probably can't make it much faster.

There also seems to be quite some load on the `select` statements, but I don't see a trivial way to reduce it either.

abustany:
Adding one more commit, also up for discussion, that uses an LRU cache that does not wrap keys in `interface{}`. Not sure how feasible it is to include an Apache licensed file in an MIT project?

benchcmp:
```
benchmark                   old ns/op     new ns/op     delta
BenchmarkConnRoutingKey     421           335           -20.43%

benchmark                   old allocs     new allocs     delta
BenchmarkConnRoutingKey     4              3              -25.00%

benchmark                   old bytes     new bytes     delta
BenchmarkConnRoutingKey     48            32            -33.33%
```

Zariel:
Thanks for submitting this, I will take a look. IANAL but I think its OK to include the source with the license, though I would prefer to have it go gettable, is the source available anywhere elese for the lru library? Could you split this PR into 3 parts? One for the changes to the routing key, one for the use statement improvement and finally one for the LRU change.

The test failures look to be caused by Github timing out due to DDoS.

abustany:
The LRU source is a copy of the original one (from github.com/golang/groupcache/lru), with `interface{}` replaced by `string` for keys. Since I just copied the folder from another repository, it should still be go gettable? ie. `go get github.com/gocql/gocql/lru` once the commit is merged.

I'll split that PR as requested, that's no problem :-)

abustany:
OK, looks like Travis cached the tests results, so we're still with the spurious errors from last run (I just removed the last two commits that ended up in two separate PRs). Not sure if you can retrigger it? Else, I can rebase it if some other commits get merged...

session.go
```diff
@@ -260,9 +260,8 @@ func (s *Session) KeyspaceMetadata(keyspace string) (*KeyspaceMetadata, error) {
 // returns routing key indexes and type info
 func (s *Session) routingKeyInfo(stmt string) (*routingKeyInfo, error) {
 	s.routingKeyInfoCache.mu.Lock()
-	cacheKey := s.cfg.Keyspace + stmt
```
Zariel: im not sure this is a great idea, even though the cache uses the config keyspace which is static, it should really use the keyspace from the connection pool which is a bug. Can you remove this change and check the performance again?

abustany:
Here is a benchcmp before and after the commit (before = include keyspace, after = without the keyspace):
```
benchmark                     old ns/op     new ns/op     delta
BenchmarkConnRoutingKey-4     537           326           -39.29%

benchmark                     old allocs     new allocs     delta
BenchmarkConnRoutingKey-4     4              3              -25.00%

benchmark                     old bytes     new bytes     delta
BenchmarkConnRoutingKey-4     112           48            -57.14%
```

So clearly doing a string concatenation every time has a cost :-) If we want to have a global cache, we should maybe use a "map[keyspace]LRU"?

Zariel:
After some though this is fine, thanks very much!

PR #455, opened 2015-08-26
query: Improve efficiency of 'USE' statement detection

abustany:
Benchmark results for a 80 character statement that does not start with
'use':

Before: 1927 ns/op   192 B/op   2 allocs/op
After:  142 ns/op      6 B/op   2 allocs/op


Zariel:
Just checking, if you are currently in the AUTHORS file or not, if you are not can you addyourself, if you are just say so and ill merge. This is a really nice change :)

abustany:
OK, added myself to the AUTHORS file

Zariel:
Thanks!

PR #463, opened 2015-08-30
Proto v4

Zariel:


Zariel:
@0x6e6562 fancy taking a look? This doesnt implement the smallint and other new types, the overall changes were pretty simple with the new framing layer which is good.

PR #471, opened 2015-09-12
Expose page state

### LINKED ISSUE #400: Allow clients to access the page state and pagingate manually
Currently the only way to do pagination is by using the auto paginator, this means that a result set must be fully read by the client before returning, or they must keep the `Iter` around. An alternative will be to allow users to get the `page_state` for the query from the `Iter`, then allow setting the `page_state` on the query in subsequent queries.

Zariel:
Exposes PageState on the Iter which can then be used to create a new Query which will page from a specific point, disabling automatic page fetching.

Fixes #400


Zariel:
Will probably need to come back and improve the API later

PR #473, opened 2015-09-21
Add support for "LexicalUUIDType".

muirrn:
Support it by treating it the same as UUIDType.

AFAIK originally there was LexicalUUIDType and TimeUUIDType. Lexical did
simple byte-wise comparison, where Time sorted using the UUID's time
value. UUIDType was added as a convenience type (CASSANDRA-2233) which
compares using the time value for time UUIDs, else byte-wise.
LexicalUUIDType was deprecated but not removed. We still have some
schema around explicitly using the LexicalUUIDType comparator.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -2126,3 +2126,34 @@ func TestManualQueryPaging(t *testing.T) {
 		t.Fatalf("expected to fetch %d rows got %d", fetched, rowsToInsert)
 	}
 }
+
+func TestLexicalUUIDType(t *testing.T) {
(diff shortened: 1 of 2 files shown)


muirrn:
Looks like one of the travis runs flaked (I see similar failure in previous job #988.5).

Zariel:
Thanks, can you add yourself to the AUTHORS file and ill merge this

muirrn:
I'm already in the AUTHORS file :)

Zariel:
Thanks :)

PR #476, opened 2015-09-23
if we read a frame header and dont have a handler, consume the frame …

Zariel:
…and ignore it

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -350,6 +350,14 @@ func (c *Conn) serve() {
 	c.closeWithError(err)
 }
 
+func (c *Conn) discardFrame(head frameHeader) error {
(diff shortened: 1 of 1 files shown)


Zariel:
This should fix #439 but still unsure how this can happen, if it is caused by the driver or other network issues.

conn.go
```diff
@@ -350,6 +350,14 @@ func (c *Conn) serve() {
 	c.closeWithError(err)
 }
 
+func (c *Conn) discardFrame(head frameHeader) error {
+	_, err := io.CopyN(ioutil.Discard, c, int64(head.length))
+	if err != nil {
```
clutchski: return err?

Zariel: What do you mean?

Issue #359, opened 2015-04-10
Features requiring server events


Zariel:
This is a meta issue to track which features require supporting server sent events. The driver is getting to a stage of stability and maturity that this is the last outstanding feature to bring gocql on par with the official drivers.
- [x] Non polled ring topology
- [ ] Ability to send 'no_metadata' flag for requests #205
- [ ] Non polled schema information, #356

PR #480, opened 2015-10-03
Remove SimpleConnPool, remove custom pools

Zariel:
Remove custom pools so that the driver can be simplified and
make more assumptions about the underlying conneciton pool whilst
still allowing configuration of host selection and conn selection policies.


Zariel:
The motivation behind this is to simplify the internal pools and to make #359 possible

PR #483, opened 2015-10-06
Ability to call NewCluster() with string hosts separated with commas

sm9sh:
For usability
Example, NewCluster("1.1.1.1, 2.2.2.2, 3.3.3.3")

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/gocql/gocql/lru"
+	"strings"
 )
 
 const defaultMaxPreparedStmts = 1000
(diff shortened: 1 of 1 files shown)


Zariel:
I think its better for users to be able to do this them selves if they want without the driver making assumptions and trying to handle all the use cases for passing in the hosts. Something like this is easy

```go
cluster := gocql.NewCluster()
cluster.Hosts = strings.Split("127.0.0.1,127.0.0.2")

// or
cluster := gocql.NewCluster(strings.Split("127.0.0.1,127.0.0.2")...)
```

Thanks for the patch though.

sm9sh:
Oh! Thanks for advice.

PR #489, opened 2015-10-08
Add support for tuple marshalling

smira:
Also add unit-tests for tuple marshalling and unmarshalling.

Marshalling tuples is required when tuple is part of collection
(e.g. `set<frozen<tuple<text,text>>>`) and Cassandra doesn't support
placeholders inside collections when parsing queries.


Zariel:
Thanks

PR #490, opened 2015-10-09
add ParseConsistency() to frame.go

adamweiner:
Added a function to frame.go that takes a string representation of a Cassandra consistency level and returns the proper gocql constant. If an unrecognized string is provided, it defaults to ALL - when in doubt provide the highest consistency level.

One use case would be when reading the desired Cassandra consistency level from a configuration file.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -55,3 +55,4 @@ Adrian Casajus <adriancasajus@gmail.com>
 John Weldon <johnweldon4@gmail.com>
 Adrien Bustany <adrien@bustany.org>
 Andrey Smirnov <smirnov.andrey@gmail.com>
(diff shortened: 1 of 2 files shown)


Zariel:
I think it is best in this situation to return an error indicating an invalid consistency, like the functions in strconv do. Otherwise patch looks good

adamweiner:
Updated my implementation to panic if given an invalid consistency @Zariel

Zariel:
@adamweiner can you add yourself to the AUTHORS file and ill merge

adamweiner:
Awesome - just added myself, @Zariel

Zariel:
thanks

PR #491, opened 2015-10-16
Add ExecuteBatchCAS and MapExecuteBatchCAS functions for using batches with LWT. Fixes #453

### LINKED ISSUE #453: Batch does not correctly handle LWT response
Currently we expect BATCH frames to only respond with RESULT VOID frames, but when a query has a LWT then it will respond with RESULT ROWS like a normal CAS will. Reported on the mailing list.

```go
CREATE KEYSPACE foo WITH replication = {
  'class': 'SimpleStrategy',
  'replication_factor': '1'
};

USE foo;

CREATE TABLE bar (
  pcol int,
  ccol int,
  col1 int,
  scol int static,
  PRIMARY KEY ((pcol), ccol)
);

the following code,

func main() {
	cluster := gocql.NewCluster("localhost")
	cluster.Keyspace = "foo"

	sess, err := cluster.CreateSession()
	if err != nil {
		log.Fatalln("Error connecting:", err)
	}

	batch := gocql.NewBatch(gocql.LoggedBatch)
	batch.Query("insert into bar (pcol, ccol, col1) values (1, 1, 1)")
	batch.Query("update bar set scol = 1 where pcol = 1 if scol = 0")

	if err := sess.ExecuteBatch(batch); err != nil {
		log.Fatalln("Error executing batch:", err)
	}
	fmt.Println("Success")
}
```

```
2015/08/18 08:16:27 Error executing batch: Unknown type in response to batch statement: [result_rows meta=[metadata flags=0x1 paging_state= columns=[[column keyspace=foo table=bar name=[applied] type=boolean] [column keyspace=foo table=bar name=pcol type=int] [column keyspace=foo table=bar name=ccol type=int] [column keyspace=foo table=bar name=scol type=int]]]]
```

nemothekid:
I quickly put this together thinking I would need something like ExecuteBatchCAS, and while in the middle of writing the tests, I realized it wouldn't work for the use case I had imagined. 

In any case, this PR adds two functions, `ExecuteBatchCAS` and `MapExecuteBatchCAS`. They function like `ScanCAS`, except they also return an `*Iter` allowing you to read more rows if needed. The `[applied]` column is taken care of in the first function call, but further calls to `.Scan` need to take care of the `applied` column. 

Potentially what we could also do is add a `.ScanCAS` helper function on `*Iter`. What do you guys think?

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -340,6 +340,60 @@ func TestCAS(t *testing.T) {
 	} else if !applied {
 		t.Fatal("delete should have been applied")
 	}
+
+	if err := session.Query(`TRUNCATE cas_table`).Exec(); err != nil {
+		t.Fatal("truncate:", err)
+	}
+
+	successBatch := session.NewBatch(LoggedBatch)
+	successBatch.Query("INSERT INTO cas_table (title, revid, last_modified) VALUES (?, ?, ?) IF NOT EXISTS", title, revid, modified)
+	if applied, _, err := session.ExecuteBatchCAS(successBatch, &titleCAS, &revidCAS, &modifiedCAS); err != nil {
+		t.Fatal("insert:", err)
+	} else if !applied {
+		t.Fatal("insert should have been applied")
+	}
+
+	successBatch = session.NewBatch(LoggedBatch)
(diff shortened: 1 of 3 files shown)


nemothekid:
Linking #453

Zariel:
Thanks

Issue #492, opened 2015-10-17
Reduce GC pressure in framer.parseResultRows() and framer.readBytes()


mhamrah:
I've been doing some performance profiling with the gocql driver and for applications which do frequent reads (specifically on time-series data) I've noticed gc pressure caused by the three allocations which take place inside the parseResultRows function and the readBytes() function, which is called from parseResultRows (readBytes is also called from other places, but the calls from parseResultRows are the most excessive for reads).

In looking at the code I've noticed this comment from @Zariel :

	// we cant make assumptions about the length of the life of the supplied byte
	// slice so we defensivly copy it out of the underlying buffer. This has the
	// downside of increasing allocs per read but will provide much greater memory
	// safety. The allocs can hopefully be improved in the future.
	// TODO: dont copy into a new slice
	l := make([]byte, size)
	copy(l, f.rbuf[:size])

My idea is to create a memory pool that's somehow tied to the underlying session--or possibly-connection, with hooks to get, put, and reset on calls to Iter.Scan. Before venturing down this path I'm wondering if maintainers have ideas on this problem or other possible solutions?

Zariel:
Yep this is my plan, when I rewrote the framing layer I scarified some performance in byte slice copying for a simpler and easier to maintain and extend.

I have an idea to fix this, remove the copies and have the iter release the framer on close. Let me try this and see if it helps.

Zariel:
@mhamrah So `readBytes` no longer copies the buffer which should reduce allocations a lot, and memory pressure in general. According to my benchmarks the next thing to remove would be all the strings from `readTypeInfo` but the best way to do that would be to implement #205.

mhamrah:
Thanks, I'll re-profile my app with this change and see how it looks.
Appreciate the quick turnaround!

Zariel:
@mhamrah any update on this?

mhamrah:
@Zariel Thanks for checking in. This does improve performance, and I'm happy to close the issue. There's still some performance pain when allocating the 2d byte slice for columns in parseResultRows, however. I think this can be remedied by allocating a column buffer in a similar way as read buffer- I've been experimenting locally.

As an FYI we're reading time-series data, so taking slices of very wide columns.

PR #493, opened 2015-10-17
Remove []byte copy in framer

Zariel:
Dont defensively copy framers buf when reading bytes during parseFrame.

Updates #492


Zariel:
Before and after

```
benchmark                     old ns/op     new ns/op     delta
BenchmarkParseRowsFrame-4     76868         34321         -55.35%

benchmark                     old allocs     new allocs     delta
BenchmarkParseRowsFrame-4     609            110            -81.94%

benchmark                     old bytes     new bytes     delta
BenchmarkParseRowsFrame-4     31904         20304         -36.36%
```

PR #496, opened 2015-10-17
Add method to await for schema to become consistent

### LINKED ISSUE #422: unconfigured column family failure in travis
I've seen this one itermitantly in the past, not sure what is causing it.

```
--- FAIL: TestTooManyQueryArgs-2 (1.45s)
	cassandra_test.go:520: '`SELECT * FROM too_many_query_args WHERE id = ?`, 1, 2' should return an ErrQueryArgLength, but returned: unconfigured columnfamily too_many_query_args
```

Full log is available at https://gist.github.com/Zariel/ab6ed48081e7e3ff9c80

### LINKED ISSUE #495: Should wait for schema to be consistent when making changes to a cluster
When changing schema we should wait for the cluster to have a consistent view, so that we dont end up using old schemas or duplicating create table calls, or keyspaces.

The log from a travis run shows this is a cause of some of our test failures https://gist.github.com/Zariel/fd58c6bb984c3068b976

See also https://issues.apache.org/jira/browse/CASSANDRA-8387

Zariel:
Wait for schema to become consistent across the cluster when
creating tables and keyspaces. Dont export this for now.

Fixes #495
Fixes #422

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ install:
 script:
   - set -e
   - go test -v -tags unit
-  - PATH=$PATH:$HOME/.local/bin bash -x integration.sh $CASS $AUTH
+  - PATH=$PATH:$HOME/.local/bin bash integration.sh $CASS $AUTH
   - go vet .
 
 notifications:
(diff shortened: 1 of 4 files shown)


Zariel:
So this appears to bring all sorts of bugs out of the woodwork

PR #505, opened 2015-10-27
Update host policies to support host pools, fixes #499

### LINKED ISSUE #499: Detect host failure in connection pool
We were using a custom connection pool which used [bitly/hostpool](https://github.com/bitly/go-hostpool), this allowed us to detect issues with any of our hosts and direct queries to other hosts in the cluster. 

Unfortunately a recent change (139a10f05ec6cf5cf66bda1aace1356b71fa3711) removed the ability to use custom connection pool which broke our ability to use this hostpool. We looked into using the new host/connection policies however with the current interface it is not possible as there is no way to return the result of the query to the hostpool.

Having this feature built-in to gocql could be useful as it should make the client much more resilient to failure. If this is not a possibility can you think of any other ways of implementing something similar given the current interface?

dancannon:
This PR updates the `HostPolicy` interface to support host pools, more info in #499.

There are some small API changes that were required to allow host status to be reported back to the pool.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -56,3 +56,4 @@ John Weldon <johnweldon4@gmail.com>
 Adrien Bustany <adrien@bustany.org>
 Andrey Smirnov <smirnov.andrey@gmail.com>
 Adam Weiner <adamsweiner@gmail.com>
+Daniel Cannon <daniel@danielcannon.co.uk>
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -1137,7 +1137,7 @@ func injectInvalidPreparedStatement(t *testing.T, session *Session, table string
 		t.Fatal("create:", err)
 	}
 	stmt := "INSERT INTO " + table + " (foo, bar) VALUES (?, 7)"
-	conn := session.pool.Pick(nil)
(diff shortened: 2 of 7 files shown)


host_source.go
```diff
@@ -33,7 +33,9 @@ func (r *ringDescriber) GetHosts() (hosts []HostInfo, partitioner string, err er
 		return r.prevHosts, r.prevPartitioner, nil
 	}
 
-	conn := r.session.pool.Pick(nil)
+	// TODO check if we are assuming that pool will return a different connection
+	// each time.
+	_, conn := r.session.pool.Pick(nil)
```
dancannon: As mentioned in the comment I am not too sure about this change.

Zariel:
@dancannon looks like the tests dont compile

dancannon:
Sorry about that, I will work at fixing that now.

dancannon:
I didn't realise there were integration tests, the tests should all be passing now.

Zariel:
Can you add yourself to the AUTHORS file and ill merge, thanks.

dancannon:
Done.

Zariel:
Thanks

PR #508, opened 2015-10-28
Set rpc_address for discovered hosts

balboah:
According to [documentation](http://docs.datastax.com/en/cassandra/2.1/cassandra/configuration/configCassandra_yaml_r.html?scroll=reference_ds_qfg_n1r_1k__start_native_transport), the rpc_address is what should be used
for native transport. We should not use the peer address as the code currently suggest (listen address).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -57,3 +57,4 @@ Adrien Bustany <adrien@bustany.org>
 Andrey Smirnov <smirnov.andrey@gmail.com>
 Adam Weiner <adamsweiner@gmail.com>
 Daniel Cannon <daniel@danielcannon.co.uk>
+Johnny Bergström <johnny@joonix.se>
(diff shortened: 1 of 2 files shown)


Zariel:
@balboah if your not already could you add yourself to the AUTHORS file?

balboah:
Sure, I'll add that commit

balboah:
@Zariel uppdated

Zariel:
Thanks

Zariel:
This fixes #509

m-m-f:
Please note that `rpc_address` is not an address you should use to contact a server: it is the address the server listens/binds to. As such it can contain invalid. i.e., non routable addresses like `0.0.0.0`. Please see here: https://github.com/gocql/gocql/issues/509

balboah:
Is 0.0.0.0 possible for `system.peers`?

PR #510, opened 2015-10-30
Connection pool deadlock / invalid discovery rpc address

### LINKED ISSUE #509: use of system.local rpc_address breaks gocql for rpc_address 0.0.0.0 configuration
Commit a70d1ac9aaf9d1b373cbf3133893b9ff44f8cb33 changes strategy for retrieving IP information of nodes in a cassandra cluster. It uses the `rpc_address` information from the `system.local` table.

Unfortunately, a `rpc_address` is not ideal way of retrieving this information (information below is from http://docs.datastax.com/en/cassandra/2.2/cassandra/configuration/configCassandra_yaml.html)

1) It can be set to 0.0.0.0 to configure Cassandra to listen to all interfaces,

2) broadcast_rpc_address can be used as a mechanism to overwrite the advertised address for the node (even in cases where rpc_address is *not* set to 0.0.0.0).

According to the cassandra documentation, it seems like clients really should be looking at broadcast_rpc_address information:

```
(Default: unset)note RPC address to broadcast to drivers and other Cassandra nodes.
This cannot be set to 0.0.0.0. If blank, it is set to the value of the rpc_address or rpc_interface.
If rpc_address or rpc_interfaceis set to 0.0.0.0, this property must be set.
```

Now, I'm not quite sure if this information corresponds to the `broadcast_rpc` field in the `system.local` table or the `rpc_address` of the `system.peers`. But I can say that the current behavior defined broke our tests since in our deployment rpc_address is set to 0.0.0.0, causing the gocql client to break communications with the cluster after an initial fetch of the `system.local` table information.

balboah:
- Using broadcast address instead of rpc adress (which might not be unicast IP for client connections)
- Small refactoring of mutex locking

From [Cassandra docs](http://docs.datastax.com/en/cassandra/2.1/cassandra/configuration/configCassandra_yaml_r.html?scroll=reference_ds_qfg_n1r_1k__broadcast_rpc_address):
```
RPC (remote procedure call) settings

Settings for configuring and tuning client connections.

broadcast_rpc_address 
    (Default: unset)note RPC address to broadcast to drivers and other Cassandra nodes. This cannot be set to 0.0.0.0. If blank, it is set to the value of the rpc_address or rpc_interface. If rpc_address or rpc_interfaceis set to 0.0.0.0, this property must be set. 
```

Fixes #509

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -35,7 +35,7 @@ type ringDescriber struct {
 }
 
 func checkSystemLocal(control *controlConn) (bool, error) {
-	iter := control.query("SELECT rpc_address FROM system.local")
+	iter := control.query("SELECT broadcast_address FROM system.local")
 	if err := iter.err; err != nil {
 		if errf, ok := err.(*errorFrame); ok {
 			if errf.code == errSyntax {
@@ -58,7 +58,7 @@ func (r *ringDescriber) GetHosts() (hosts []HostInfo, partitioner string, err er
 	const (
 		legacyLocalQuery = "SELECT data_center, rack, host_id, tokens, partitioner FROM system.local"
 		// only supported in 2.2.0, 2.1.6, 2.0.16
-		localQuery = "SELECT rpc_address, data_center, rack, host_id, tokens, partitioner FROM system.local"
+		localQuery = "SELECT broadcast_address, data_center, rack, host_id, tokens, partitioner FROM system.local"
 	)
 
 	var localHost HostInfo
--- a/connectionpool.go
+++ b/connectionpool.go
@@ -121,6 +121,7 @@ func newPolicyConnPool(session *Session, hostPolicy HostSelectionPolicy,
 
 func (p *policyConnPool) SetHosts(hosts []HostInfo) {
 	p.mu.Lock()
+	defer p.mu.Unlock()
 
 	toRemove := make(map[string]struct{})
 	for addr := range p.hostConnPools {
@@ -159,7 +160,6 @@ func (p *policyConnPool) SetHosts(hosts []HostInfo) {
 	// update the policy
 	p.hostPolicy.SetHosts(hosts)
 
-	p.mu.Unlock()
 }
 
 func (p *policyConnPool) SetPartitioner(partitioner string) {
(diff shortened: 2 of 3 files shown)


czertbytes:
+1

host_source.go
```diff
@@ -96,6 +96,11 @@ func (r *ringDescriber) GetHosts() (hosts []HostInfo, partitioner string, err er
 		localHost.Peer = addr
 	}
 
+	// Make sure we didn't screw up discovering our local peer.
+	if localHost.Peer == "" || localHost.Peer == "0.0.0.0" {
+		return nil, "", ErrHostQueryFailed
```
Zariel: If this is the case then we should use conn.Addr() which is less bad than just never discovering

Zariel:
Instead of patching over the issue why not just add a check which is causing a NPE? I guess the calling code is catching the panic and continuing the application?

host_source.go
balboah: this was added as a safety check, but it could be removed. It was 0.0.0.0 only when selecting on rpc_address in previous version.

balboah:
I'm not deep enough into the code base yet, but the nil pointer exception happened when we selected 0.0.0.0 as localhost in the discovery. The defer Unlocks was just so that I could actually see the panic from the beginning, it just froze before that.

I think the nil came through roundRobinHostPolicy.Pick(). Where host was nil.
But is no longer happening after the select on broadcast_address

host_source.go
```diff
@@ -58,7 +58,7 @@ func (r *ringDescriber) GetHosts() (hosts []HostInfo, partitioner string, err er
 	const (
 		legacyLocalQuery = "SELECT data_center, rack, host_id, tokens, partitioner FROM system.local"
 		// only supported in 2.2.0, 2.1.6, 2.0.16
-		localQuery = "SELECT rpc_address, data_center, rack, host_id, tokens, partitioner FROM system.local"
+		localQuery = "SELECT broadcast_address, data_center, rack, host_id, tokens, partitioner FROM system.local"
```
balboah: @Zariel this is the real fix, you can scrap other parts if you don't like them

Zariel: This change + the check that we can use this query and the check if the returned host is "0.0.0.0" are fine. In the case where the host is "0.0.0.0" that implies that the user has a misconfigured Cluster, we could panic in this case.

Zariel:
@balboah this also needs a rebase

balboah:
@Zariel rebased and removed the 0.0.0.0 check which shouldn't happen, as I was not sure panic was a good choice either

Zariel:
Thanks

PR #513, opened 2015-10-30
Refresh ring only if DiscoverHosts is true in control connection

nemothekid:
This prevents a panic on calling a `nil` `ringDescriber` when you call `refreshRing`.


Zariel:
Thanks

PR #530, opened 2015-11-18
added a list of approved authenticators 

### LINKED ISSUE #529: Hardcoded class name check for authentication
What is the purpose of the following line: https://github.com/gocql/gocql/blob/master/conn.go#L44
Connection to instaclustr nodes fail because they use com.instaclustr.cassandra.auth.SharedSecretAuthenticator, yet simply removing that line works perfectly.

Aorioli:
Fixes #529

It's not the ideal solution, but without knowing anything about the internals of instaclustr there is no point in doing anything else.


Zariel:
ccm is having issues downloading from the archives. Thanks for the contribution!

PR #531, opened 2015-11-18
conn: lazily allocate stream call

### LINKED ISSUE #446: session: Reduce default number of streams w/ Protocol version 3
Protocol version 3 bumps up the number of default streams from 128 and 32768, nearly eradicating the need to have have multiple connections.

Currently gocql defaults to opening the maximum number of streams (if you don't set `cluster.NumStreams`) which can really blow up the amount of memory used, and we may want to look at this tradeoff.

Currently with default settings, a bare application (creates a session and nothing else) uses around ~220MiB of resident RAM with 32768, but only 4.5MiB with 256 streams (2x the default with protocol 1, 2 connections, everything else default). Maybe others can test this, but this means to me you can't really throw a gocql app on a $5/DO box without causing some swapping.

Zariel:
Use sync.Pool to recycle inflight call requests.

Fixes #446

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -112,15 +112,17 @@ var TimeoutLimit int64 = 10
 // queries, but users are usually advised to use a more reliable, higher
 // level API.
 type Conn struct {
-	conn    net.Conn
-	r       *bufio.Reader
-	timeout time.Duration
-	cfg     *ConnConfig
+	conn       net.Conn
+	r          *bufio.Reader
+	timeout    time.Duration
+	cfg        *ConnConfig
+	numStreams int
 
 	headerBuf []byte
 
 	uniq  chan int
-	calls []callReq
+	mu    sync.RWMutex
+	calls map[int]*callReq
 
 	errorHandler    ConnErrorHandler
 	compressor      Compressor
@@ -188,7 +190,7 @@ func Connect(addr string, cfg *ConnConfig, errorHandler ConnErrorHandler, sessio
 		r:            bufio.NewReader(conn),
 		cfg:          cfg,
 		uniq:         make(chan int, streams),
-		calls:        make([]callReq, streams),
+		calls:        make(map[int]*callReq),
 		timeout:      cfg.Timeout,
 		version:      uint8(cfg.ProtoVersion),
 		addr:         conn.RemoteAddr().String(),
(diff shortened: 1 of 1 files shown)


conn.go
```diff
@@ -172,7 +172,7 @@ func Connect(addr string, cfg *ConnConfig, errorHandler ConnErrorHandler, sessio
 		r:            bufio.NewReader(conn),
 		cfg:          cfg,
 		uniq:         make(chan int, streams),
-		calls:        make([]callReq, streams),
+		calls:        make([]*callReq, streams), // TODO(zariel): use a sparse map instead?
```
dojodominator: using a map[int]*callReq would make sense now.  Delete the entry from the map when you put it in the sync.Pool to free up memory.  We don't need to hold references to where every possible stream may be help anymore.

dojodominator: We would have to protect the map with a mutex though...  so we can do it like this if we don't want to have to lock.

conn.go
```diff
@@ -324,8 +323,7 @@ func (c *Conn) closeWithError(err error) {
 	if err != nil {
 		// we should attempt to deliver the error back to the caller if it
 		// exists
-		for id := 0; id < len(c.calls); id++ {
-			req := &c.calls[id]
+		for _, req := range c.calls {
```
dojodominator: I might have clicked cancel by accident making this comment earlier.  Because the slice is holding pointers now, and most of them will be nil, we will need to check if the req is nil before attempting to send on the channel.

conn.go
nemothekid: IMO, i'd be great if you could get rid of allocating `streams` upfront entirely. The map suggestion is apt, but the lock might be undesirable. Motivation being is this is allocated for every connection + combined with something like DiscoverHosts, and 2 hosts per connection, large clusters tend to require a lot of idle memory. Tracking the current inflight streams seems to be the deal breaker - however, if you could use the `calls` data structure to enforce `cfg.NumStreams` (possibly blocking once we reach `cfg.NumStreams`, you could turn `uniq` into an atomic counter.

Zariel:
@nemothekid I would be interested to know if you see any performance changes after applying this

dojodominator:
This is great.  My app went from 700+MB to 150MB after building with this.

PR #532, opened 2015-11-20
idea: Don't allocate all streams upfront / use a condition

nemothekid:
I tinkered around with the stream code to scratch an itch I've had that had been rekindled by the [lazy-streams](https://github.com/gocql/gocql/commit/83d38bd442ed90486aa76743a182b81d6f709bbe) PR and my lack of ability to sleep atm.

Motivation for this PR is as you grow your cluster gocql creates a new connection (or 2, the default) for each node, and each connection allocates a full buffered channel `numStreams` large (and unril recently one for `callReq`'s as well). A back of an envelope calculation puts the size of `uniq chan` at 16 hosts * 2conns/host * 32768 streams/conn * 8bytes/stream = 8.388608MiB. Not a lot (I imagine the new lazy-streams code saves memory too in the worst case, as `callReq` was ~24x bigger) - but I still think its interesting if that could go away.

This PR generates streams sequentially and then uses a combination of atomic operations and a `sync.Cond` to manage how many active streams there are. Performance wise, it should be as fast as using a channel. In the majority of cases (which is if you aren't starved for streams), you are simply doing an atomic increment/decrement instead of a send/recv. In the starved case a Lock/Cond is used which the general knowledge states they are faster than `chan`s while being less idiomatic.

Some other caveats include 

1. It's possible (don't know how possible) that we reuse a stream. If we are set to use maxStreams *and* we wrap around all the stream ids, and a stream we issued long ago is still in use, we may reuse that stream. I'm doubtful this can occur on Protocol 3+, but on Protocol 2 this might be a possibility?
    1. Instead of panicing on an already used stream, we can keep walking the stream list until we fine an unused one. Given that this will probably be an extreme case, the performance implications of doing so may not be important.
1. It's not idiomatic Go. One of the classic "resource-bound" examples in Go, like this one, involves creating a chan of *n* length. While its not idiomatic, its probably faster and uses less memory.

And as mentioned the tradeoffs with the above are performance & memory.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -120,9 +120,11 @@ type Conn struct {
 
 	headerBuf []byte
 
-	uniq  chan int
-	mu    sync.RWMutex
-	calls map[int]*callReq
+	availableStreams int32
+	streamCounter    int32
+	streamBlocker    *sync.Cond
+	mu               sync.RWMutex
+	calls            map[int]*callReq
 
 	errorHandler    ConnErrorHandler
 	compressor      Compressor
@@ -186,31 +188,27 @@ func Connect(addr string, cfg *ConnConfig, errorHandler ConnErrorHandler, sessio
 	}
 
 	c := &Conn{
-		conn:         conn,
-		r:            bufio.NewReader(conn),
-		cfg:          cfg,
-		uniq:         make(chan int, streams),
-		calls:        make(map[int]*callReq),
-		timeout:      cfg.Timeout,
-		version:      uint8(cfg.ProtoVersion),
-		addr:         conn.RemoteAddr().String(),
-		errorHandler: errorHandler,
-		compressor:   cfg.Compressor,
-		auth:         cfg.Authenticator,
-		headerBuf:    make([]byte, headerSize),
-		quit:         make(chan struct{}),
-		session:      session,
-		numStreams:   streams,
+		conn:             conn,
+		r:                bufio.NewReader(conn),
+		cfg:              cfg,
+		availableStreams: int32(maxStreams - 1), // reserve stream 0
+		streamCounter:    0,
+		streamBlocker:    sync.NewCond(&sync.Mutex{}),
+		calls:            make(map[int]*callReq),
+		timeout:          cfg.Timeout,
+		version:          uint8(cfg.ProtoVersion),
+		addr:             conn.RemoteAddr().String(),
+		errorHandler:     errorHandler,
+		compressor:       cfg.Compressor,
+		auth:             cfg.Authenticator,
+		headerBuf:        make([]byte, headerSize),
+		quit:             make(chan struct{}),
+		session:          session,
+		numStreams:       streams,
 	}
 
 	if cfg.Keepalive > 0 {
 		c.setKeepalive(cfg.Keepalive)
-	}
-
-	// reserve stream 0 incase cassandra returns an error on it without us sending
-	// a request.
-	for i := 1; i < streams; i++ {
-		c.uniq <- i
 	}
 
 	go c.serve()
@@ -410,9 +408,7 @@ func (c *Conn) recv() error {
 		return err
 	}
 
-	if head.stream > c.numStreams {
-		return fmt.Errorf("gocql: frame header stream is beyond call exepected bounds: %d", head.stream)
-	} else if head.stream == -1 {
+	if head.stream == -1 {
 		// TODO: handle cassandra event frames, we shouldnt get any currently
 		return c.discardFrame(head)
 	} else if head.stream <= 0 {
@@ -485,9 +481,8 @@ func (c *Conn) releaseStream(stream int) {
 
 	streamPool.Put(call)
 
-	select {
-	case c.uniq <- stream:
-	case <-c.quit:
+	if atomic.AddInt32(&c.availableStreams, 1) <= 0 {
+		c.streamBlocker.Signal()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


nemothekid:
For Point 1i, I'd probably do something like the following in `conn.go`

```go
var stream int
var call *callReq
maxStreams := 128 - 1
if c.cfg.ProtoVersion > protoVersion2 {
	maxStreams = 32768 - 1
}
for stream == 0 {
	if atomic.AddInt32(&c.availableStreams, -1) >= 0 {
		stream = 1 + (((int(atomic.AddInt32(&c.streamCounter, 1)) % maxStreams) + maxStreams) % maxStreams)
		c.mu.Lock()
		call = c.calls[stream]
		if call != nil {
			stream = 0
			atomic.AddInt32(&c.availableStreams, 1)
			continue
		}
		call = streamPool.Get().(*callReq)
		c.calls[stream] = call
		c.mu.Unlock()
	} else {
		c.streamBlocker.L.Lock()
		c.streamBlocker.Wait()
		c.streamBlocker.L.Unlock()
		select {
		case <-c.quit:
			return nil, ErrConnectionClosed
		default:
		}
	}
}
```

Zariel:
Interesting, ill take a fuller look later. As a note here is how the java driver does it, https://github.com/datastax/java-driver/blob/fb7b0e7b6a48adbfc3880a41fae8aa7f57d3c5d1/driver-core/src/main/java/com/datastax/driver/core/StreamIdGenerator.java

conn.go
```diff
@@ -120,9 +120,11 @@ type Conn struct {
 
 	headerBuf []byte
 
-	uniq  chan int
-	mu    sync.RWMutex
-	calls map[int]*callReq
+	availableStreams int32
```
Zariel: I think it would be best to move the stream ID related things into its own struct as conn is starting to get quite large.

conn.go
```diff
@@ -410,9 +408,7 @@ func (c *Conn) recv() error {
 		return err
 	}
 
-	if head.stream > c.numStreams {
```
Zariel: I dont see a reason to remove this check?

conn.go
```diff
@@ -510,10 +505,24 @@ var (
 func (c *Conn) exec(req frameWriter, tracer Tracer) (*framer, error) {
 	// TODO: move tracer onto conn
 	var stream int
-	select {
-	case stream = <-c.uniq:
-	case <-c.quit:
-		return nil, ErrConnectionClosed
+	maxStreams := 128 - 1
+	if c.cfg.ProtoVersion > protoVersion2 {
```
Zariel: should move this out of the exec path as it wont change

conn.go
```diff
@@ -510,10 +505,24 @@ var (
...
+	for stream == 0 {
+		if atomic.AddInt32(&c.availableStreams, -1) >= 0 {
+			stream = 1 + (((int(atomic.AddInt32(&c.streamCounter, 1)) % maxStreams) + maxStreams) % maxStreams)
+		} else {
+			atomic.AddInt32(&c.availableStreams, 1)
+			c.streamBlocker.L.Lock()
+			c.streamBlocker.Wait()
+			c.streamBlocker.L.Unlock()
+			select {
+			case <-c.quit:
```
Zariel: here we should check the closed variable not the quit chan

conn.go
```diff
@@ -799,7 +808,11 @@ func (c *Conn) Address() string {
 }
 
 func (c *Conn) AvailableStreams() int {
-	return len(c.uniq)
+	n := int(c.availableStreams)
```
Zariel: this should be an atomic load

conn.go
```diff
@@ -510,10 +505,24 @@ var (
...
-	case stream = <-c.uniq:
-	case <-c.quit:
-		return nil, ErrConnectionClosed
+	maxStreams := 128 - 1
+	if c.cfg.ProtoVersion > protoVersion2 {
+		maxStreams = 32768 - 1
+	}
+	for stream == 0 {
+		if atomic.AddInt32(&c.availableStreams, -1) >= 0 {
+			stream = 1 + (((int(atomic.AddInt32(&c.streamCounter, 1)) % maxStreams) + maxStreams) % maxStreams)
```
Zariel: I dont really follow what this is doing

Zariel:
I really dont want to have the possibility that the driver reuses streams whilst waiting for a response, that will be very bad.

conn.go
nemothekid: The new code only guarantees that there will be `c.numStreams` active streams, but doesn't guarantee what range they will be. They will likely be contiguous, but could be 500-600 if numStreams is 100.

conn.go
nemothekid: [Positive Modulus / Remainder function](http://stackoverflow.com/a/14997413/807701).

Since `streamCounter`/`stream` there is a possibility after a long enough time, `streamCounter` will be negative -> x & -n = -stream = panic index out of bounds, and we don't want that.

conn.go
```diff
@@ -186,33 +189,30 @@ func Connect(addr string, cfg *ConnConfig, errorHandler ConnErrorHandler, sessio
...
+		errorHandler:     errorHandler,
+		compressor:       cfg.Compressor,
+		auth:             cfg.Authenticator,
+		headerBuf:        make([]byte, headerSize),
+		quit:             make(chan struct{}),
+		session:          session,
+		numStreams:       streams,
+		maxStreams:       maxStreams,
+	}
+	c.streamWaiter = sync.NewCond(&c.mu)
```
Zariel: I dont think its a good idea to use the same mutex as the inflight map, as that would prevent incoming responses being dispatched while requests are searching for a stream

nemothekid:
Thinking this over, I realized that there was a possible race condition in the code.

If all streams were in use you could have the process:
```
A (acquirer): Decrement Available Streams & Available Streams == -1
B (releaser): Increment Available Streams Counter & Available Streams == 0
B: Signal()
A: Wait() (Now Deadlocked)
```

Basically if A were to be context switched out right after decrementing the available streams, but before `Wait`ing, it may never receive a Signal. The new code locks the whole section with the mutex for the calls (since they are related and cuts down on the number of critical sections).

Since we also know in the critical section that we must have at least 1 available stream we can fall back on the rare case that next sequential stream is in use, to just scanning for an available stream. The only way we can end up reusing a stream is when we wrap around on the protocol's `maxStreams`, and if that first stream id is still in use (this would mean however we made so many queries, so quickly that we exhausted the entire pool of streams before the first query we made actually completed - which is why I believe this case is rare).

If we do panic here on the inability to find an empty stream, it would indicate more serious problems in the synchronization primitives.

Thoughts?

conn.go
nemothekid: Ignoring the wrap case (see my most recent comment, in short that case should be rare), the only extra statements to that critical section would be
```
c.availableStreams--
stream = 1 + (((c.streamCounter % (c.maxStreams - 1)) + (c.maxStreams - 1)) % (c.maxStreams - 1)) // positive modulo
c.streamCounter++
```

It's not very expensive.

nemothekid: Also note that `streamWaiter.Wait()` internally calls `c.mu.Unlock()` before the goroutine is parked.

Zariel:
This is pretty neat, but I think its superseded by #533

PR #533, opened 2015-11-21
conn: replace streams chan with an ID generator

Zariel:
Replace the bounched chan uniq with a stream ID generator based
very closely on the StreamIDGenerator in the java-driver.

Updates #446

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -18,6 +18,8 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+
+	"github.com/gocql/gocql/internal/streams"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


Zariel:
@nemothekid this should further reduce the size overhead of head connection.

Zariel:
Though this removes the ability to configure number of streams to use, but we wont preallocate them now.

nemothekid:
AFAICT, On one hand this does bring us inline with the "official" implementation, on the other hand, we lose the ability to specify the number of streams (might not be a bad thing, I don't think anyone changed the value, and it might be a useless option) and we lose the ability for connections to wait for an available stream.

What happens when we get `ErrNoStreams`, do we just Retry according to the RetryPolicy or tell the application to deal with it?

Zariel:
I think the round robin conn policy doesn't check for available streams on a conn, will fix that in another PR.

PR #535, opened 2015-11-21
RoundrobbinConnPool check streams

Zariel:
RoundRobbin connection pool should check the available streams on a connection to avoid using busy connections


Zariel:
This needs to land after #533

PR #543, opened 2015-11-25
Added a method which returns true if iterator reached end of the page

Unix4ever:
and new page is available.

It enables counting how many pages we gone through calling `Next`.
It is useful in case if you want to stop iteration if you've gone through too many pages, so this method tells when calling `Next` will cause new page fetching.


Zariel:
Thank you!

Issue #205, opened 2014-07-23
Skip meta data flag on EXECUTE


0x6e6562:
Right now we're missing a trick with statement preparation. Currently we eagerly fetch the result column metadata to perform unmarshaling, but if we take advantage of the the v2 protocol (in conjunction with #204), we could cache this metadata on PREPARE and ask the server not to send it on EXECUTE.

This is what the v2 protocol says about statement preparation:

```
4.2.5.4. Prepared

  The result to a PREPARE message. The rest of the body of a Prepared result is:
    <id><metadata><result_metadata>
  where:
    - <id> is [short bytes] representing the prepared query ID.
    - <metadata> is defined exactly as for a Rows RESULT (See section 4.2.5.2; you
      can however assume that the Has_more_pages flag is always off) and
      is the specification for the variable bound in this prepare statement.
    - <result_metadata> is defined exactly as <metadata> but correspond to the
      metadata for the resultSet that execute this query will yield. Note that
      <result_metadata> may be empty (have the No_metadata flag and 0 columns, See
      section 4.2.5.2) and will be for any query that is not a Select. There is
      in fact never a guarantee that this will non-empty so client should protect
      themselves accordingly. The presence of this information is an
      optimization that allows to later execute the statement that has been
      prepared without requesting the metadata (Skip_metadata flag in EXECUTE).
      Clients can safely discard this metadata if they do not want to take
      advantage of that optimization.

  Note that prepared query ID return is global to the node on which the query
  has been prepared. It can be used on any connection to that node and this
  until the node is restarted (after which the query must be reprepared).
```

Zariel:
This will be easier once we have a control connection, the logic in Conn to read the rows will need to be updated to not rely on it being in the metadata, which shouldnt be too hard.

PR #551, opened 2015-12-25
Add support for node events for up and down hosts

Zariel:
Replace the host discovery polling mechanism with server sent events. Remove the option for disabling host discovery.

Will add a new method to filter hosts for queries.


Zariel:
I'll get back to this, but first I need to rework how hosts are filtered and selected for queries as this removes the ability to filter DC/Racks.

Zariel:
updates #205
updates #359 
updates #554 
fixes #552

PR #573, opened 2016-01-12
DSE reports >3 elements in the version string

robmccoll:
When using this driver with the DataStax edition of Cassandra the connect bombs out after the most recent version validation change (was working as of 2 days ago).  This just allows for extra sections in the version string (which should work for DSE and OSS Cassandra).

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -32,7 +32,7 @@ type cassVersion struct {
 func (c *cassVersion) UnmarshalCQL(info TypeInfo, data []byte) error {
 	version := strings.TrimSuffix(string(data), "-SNAPSHOT")
 	v := strings.Split(version, ".")
-	if len(v) != 3 {
+	if len(v) < 3 {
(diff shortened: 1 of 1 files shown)


robmccoll:
Also for me at least:
```
$go test
PASS
ok  	github.com/gocql/gocql	1.527s
```

Zariel:
fyi for running the tests you need to do `go test -tags=all` thanks for the patch, never occurred to me that DSE would do different things, are they still sensible version numbers that relate to the Cassandra version>

robmccoll:
Ah, thanks - I'll be sure to run them all moving forward.

Since I don't have access to the source I can only guess, but looking at the version string it looks like its a schema version from the 2.x line with perhaps an internal build number appended as an extra field (which seems sensible enough to me if you're shipping binary).

Zariel:
Thanks!

PR #576, opened 2016-01-13
Fix Unlock of unlocked RWMutex when removing a host from a host policy

mattheath:
I'm guessing this is a typo - we're calling `Unlock`, then defering another `Unlock`, which results in the following panic:

```panic: sync: Unlock of unlocked RWMutex```

### DIFF
--- a/policies.go
+++ b/policies.go
@@ -450,7 +450,7 @@ func (r *hostPoolHostPolicy) AddHost(host *HostInfo) {
 }
 
 func (r *hostPoolHostPolicy) RemoveHost(addr string) {
-	r.mu.Unlock()
+	r.mu.Lock()
 	defer r.mu.Unlock()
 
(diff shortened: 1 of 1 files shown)


mattheath:
This apparently fails on Go 1.4 with Cassandra 2.2.3 but the problem was introduced when `hostPoolHostPolicy.RemoveHost` was added in #551 in which the tests passed. Given this would have caused a runtime panic, I'm not sure how this could have previously passed on travis - the code can't have been executed?

Zariel:
I think the event tests might be a bit flaky in travis, ill need to keep an eye on them and tweak them. Thanks for the patch.

Zariel:
@mattheath as to why the code never executed, the tests in travis only do host up/down checks, I will look at adding new node and removed node

PR #577, opened 2016-01-13
Concurrent connect to hosts

tecbot:
Currently, if a new cluster should be created all pools and connections are created in sequence which can take a few seconds in a big cluster. This PR implements concurrent pool creation which speedup cluster initialization.


Zariel:
Thanks very much!

PR #578, opened 2016-01-14
prepared INSERT JSON only works once

Nomon:
When inserting JSON via prepared statement only first insert is successful. The second insert fails silently without error. I modified the TestJSON function to make 2 inserts and verify both to test against this case. Tested to fail against cassandra 2.2.1. 

Seems to work in travis but tested locally in OSX and inside vagrant, both give:

go test -tags integration -v -proto 4 -run JSON
=== RUN   TestJSONSupport
2016/01/14 12:59:43 common_test.go:113: closing keyspace session
--- FAIL: TestJSONSupport (0.85s)
	cassandra_test.go:2148: not found
FAIL
exit status 1
FAIL	github.com/gocql/gocql	0.858s

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -2118,6 +2118,11 @@ func TestJSONSupport(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	err = session.Query("INSERT INTO test_json JSON ?", `{"id": "user1234", "age": 43, "state": "CA"}`).Exec()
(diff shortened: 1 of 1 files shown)


Nomon:
Works with cassandra 2.2.4. Closing.

Zariel:
Thanks, will be handy to know about that issue

PR #583, opened 2016-01-15
mark the selected host with the real query response

tecbot:
This PR fixes the incorrect `SelectedHost.Mark()` behaviour if a failed query occurs. Currently, all errors are mistagged as `nil`.


Zariel:
Thanks

PR #589, opened 2016-01-15
mark the selected host with the real batch response

tecbot:
There is the same issue with batch operation. Didn't see it in the previous PR. Would be good to create a follow up ticket to merge some logic into one function because executeQuery/executeBatch looks almost the same.


Zariel:
Yeah, also we should move the bad host detection into the connection pool directly as well.

Zariel:
@tecbot let me fix the batch api as it really should just return `*Iter` like queries do

Zariel:
Thanks

Issue #582, opened 2016-01-14
Poor read/write performance with latest commit 496fdda


ttqin:
Hi there,

We seem to be getting poor C* read and write performance after building against latest gocql/gocql, commit hash 496fdda0436cec5c0f0f361b34a686e501662d80. Reverting back to fe1bd1bc33e6a89c5cdbf0bf8847aea79fde19a2 seems to fix this issue, so the 4 PR's that have been merged in may have caused the issue.

read rate 100,000/s dropped to 4,200/s
write rate 10,000/s dropped to 600/s

I'm not sure what potential impact PR's #551 #577 have, nothing glaringly obvious.

cc @dancannon

Zariel:
Thanks ill take a look at this, do you have any pprofs?

Zariel:
@ttqin Ive been doing some benchmarking using pre prepared frames and so far I don't see such a large performance regression. My 4 core i5 macbook gives these numbers,

```
benchmark                        old ns/op     new ns/op     delta
BenchmarkSession_ReadPerf        597996        593589        -0.74%
BenchmarkSession_ReadPerf-2      435783        436608        +0.19%
BenchmarkSession_ReadPerf-4      501313        469419        -6.36%
BenchmarkSession_ReadPerf-8      512366        500448        -2.33%
BenchmarkSession_ReadPerf-16     576868        562769        -2.44%

benchmark                        old allocs     new allocs     delta
BenchmarkSession_ReadPerf        1974           1974           +0.00%
BenchmarkSession_ReadPerf-2      1974           1974           +0.00%
BenchmarkSession_ReadPerf-4      1975           1975           +0.00%
BenchmarkSession_ReadPerf-8      1976           1976           +0.00%
BenchmarkSession_ReadPerf-16     1976           1976           +0.00%

benchmark                        old bytes     new bytes     delta
BenchmarkSession_ReadPerf        209444        209449        +0.00%
BenchmarkSession_ReadPerf-2      210101        210073        -0.01%
BenchmarkSession_ReadPerf-4      211333        211262        -0.03%
BenchmarkSession_ReadPerf-8      212266        212099        -0.08%
BenchmarkSession_ReadPerf-16     212221        212312        +0.04%
```

This load is using `select * from system.schema_columns`. Can you give any more details on what sort of load you see such a drop in performance with.

ttqin:
@Zariel apologies for the delay in replying. Unfortunately, I don't have any pprofs readily available.

[...]

PR #623, opened 2016-02-02
dont fill the host pool whilst holding the connection pools lock

Zariel:


Zariel:
This might impact #582

Issue #627, opened 2016-02-02
nil pointer reference on connectionpool failure


robzienert:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x1c880a]

goroutine 21 [running]:
github.com/PhysicalGraph/lever/vendor/github.com/gocql/gocql.(*policyConnPool).Close(0x0)
	/Users/xxxx/go/src/github.com/PhysicalGraph/lever/vendor/github.com/gocql/gocql/connectionpool.go:229 +0x1ca
github.com/PhysicalGraph/lever/vendor/github.com/gocql/gocql.(*Session).Close(0xc8201d0240)
	/Users/xxxx/go/src/github.com/PhysicalGraph/lever/vendor/github.com/gocql/gocql/session.go:251 +0xae
github.com/PhysicalGraph/lever/vendor/github.com/gocql/gocql.NewSession(0xc8201d22c0, 0x3, 0x4, 0x7305b8, 0x5, 0x2, 0x12a05f200, 0x2352, 0xc8201ca1c0, 0x5, ...)
	/Users/xxxx/go/src/github.com/PhysicalGraph/lever/vendor/github.com/gocql/gocql/session.go:100 +0x2e4
github.com/PhysicalGraph/lever/vendor/github.com/gocql/gocql.(*ClusterConfig).CreateSession(0xc82019cdc0, 0x3, 0x0, 0x0)
	/Users/xxx/go/src/github.com/PhysicalGraph/lever/vendor/github.com/gocql/gocql/cluster.go:150 +0x4b
github.com/PhysicalGraph/lever/cassandra.trySession(0xc8201d4960, 0xc82019cdc0, 0xc8201d22c0)
```

I'm using a config that looks like this:

```
	cluster := gocql.NewCluster(spec.Seeds...)
	cluster.Keyspace = spec.Keyspace
	cluster.CQLVersion = "3.2.0"
	cluster.Consistency = gocql.Quorum
	cluster.Timeout = time.Second * 5
	if spec.SSLCertPath != "" && spec.SSLKeyPath != "" {
		cluster.SslOpts = &gocql.SslOptions{
			CertPath:               spec.SSLCertPath,
			KeyPath:                spec.SSLKeyPath,
			EnableHostVerification: false,
		}
	}
	if spec.Username != "" && spec.Password != "" {
		cluster.Authenticator = gocql.PasswordAuthenticator{
			Username: spec.Username,
[...]

PR #628, opened 2016-02-02
session: connection pool may be nil on close

Zariel:
When the connConfig errors the connection pool will not be created,
check that it is not nil before closing it.


Zariel:
fixes #627

PR #644, opened 2016-02-13
conn: closeWithError should always try to deliver

### LINKED ISSUE #581: TestStream0 is racy
Seen this a few times in travis

```
--- FAIL: TestStream0-2 (0.81s)
	cassandra_test.go:1948: expected to get error prefix "gocql: error on stream 0:" got "gocql: connection closed waiting for response"
```

Zariel:
Remove the redundant err != nil check inside the loop and
use call.timeout to check if the caller has gone away.

This fixes the root cause issue of Stream0 test flakes

fixes #581

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -328,18 +328,16 @@ func (c *Conn) closeWithError(err error) {
 		return
 	}
 
+	// we should attempt to deliver the error back to the caller if it
+	// exists
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Zariel:
Ran with stress for 10k runs with no failures, before the rate was about 5 failures per 1000 runs.

PR #654, opened 2016-02-17
Fix token aware metadata fetching deadlock.

muirrn:
The token aware host policy needs to fetch schema metadata so it can
hash row keys properly. Before the internal "control" conn was added,
metadata.go would set Query().RoutingKey([]byte{}) before it performed
the metadata queries so the metadata queries themselves didn't re-enter
the token aware host selection code (and deadlock). The control conn
still works most of the time, though, because the control conn performs
queries directly on a conn object, bypassing the token aware selection
stuff. That is, unless there is more than one page of metadata, since
iter.next.fetch() executes the query via the session, which re-enters
the token aware host selection code.

Fix by setting RoutingKey to []byte{} on the control conn's
queries. This is the same tactic the old code was using to avoid this
problem.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -222,7 +222,7 @@ func TestCAS(t *testing.T) {
 
 	if _, err := session.Query(`DELETE FROM cas_table WHERE title = ? and revid = ? IF last_modified = ?`,
 		title, revid, tenSecondsLater).ScanCAS(); !strings.HasPrefix(err.Error(), "gocql: not enough columns to scan into") {
-		t.Fatal("delete: was expecting count mismatch error but got: %q", err.Error())
+		t.Fatalf("delete: was expecting count mismatch error but got: %q", err.Error())
 	}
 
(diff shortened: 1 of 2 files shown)


muirrn:
I'm not sure my fix is the best approach, so feedback is appreciated. Could the control conn switching to another conn during paging cause other problems/undesired behavior?

Zariel:
Its probably not great that the control connection will swap to another connection during paging, another fix would be to have a hidden `*Conn` on Query which the control connection sets. But this will solve the issue for now. If not already can you add yourself to the `AUTHORS` file and I will merge. Thanks.

muirrn:
Closing for now (don't merge yet). I've encountered another token aware related issue during testing. Will reopen this pull request once I've figured it out.

I'm already in AUTHORS.

muirrn:
Reopening as I have figured out the other issues I was having (will open separate issues/pull requests). None are directly related to token aware stuff, so I think this change stands on its own.

Zariel:
Thanks for the contribution

PR #659, opened 2016-02-19
Don't crash on bad peers.

muirrn:
It's possible (or at least was possible), to be left with invalid
system.peers entries in your db after changing your
ring (https://issues.apache.org/jira/browse/CASSANDRA-6053 is at least
one way it used to be able to happen). If this happens (or someone who
is already in this state upgrades gocql and is trying to doing peer
discovery), ideally gocql would continue working on the good peers
instead of crashing trying to parse the bad peers. I made two changes
with this goal in mind:

1. Fix version string parsing to not crash if version is too short.
2. Skip peers if we fail parsing the version string (but still return
   the peers we were able to parse).
3. When waiting for schema agreement, skip peers that have an empty
   schema_version (otherwise it times out and fails).

Our ring had peer entries with null "rack", "release_version",
"schema_version", and "tokens".

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -47,7 +47,7 @@ func (c *cassVersion) unmarshal(data []byte) error {
 	version = strings.TrimPrefix(version, "v")
 	v := strings.Split(version, ".")
 
-	if len(v) < 3 {
+	if len(v) < 2 {
 		return fmt.Errorf("invalid version string: %s", data)
 	}
 
--- a/host_source.go
+++ b/host_source.go
@@ -2,6 +2,7 @@ package gocql
 
 import (
 	"fmt"
+	"log"
 	"net"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


muirrn:
My log.Prints add some noise to the test output. Maybe gocql needs a Logger config field where all the random warning/error messages go? That way, people can redirect the out-of-band error messages to their log files instead of wherever stderr goes.

host_source.go
```diff
@@ -38,10 +39,18 @@ func (c *cassVersion) Set(v string) error {
...
 func (c *cassVersion) UnmarshalCQL(info TypeInfo, data []byte) error {
+	return c.unmarshal(data)
+}
+
+func (c *cassVersion) unmarshal(data []byte) error {
 	version := strings.TrimSuffix(string(data), "-SNAPSHOT")
 	version = strings.TrimPrefix(version, "v")
 	v := strings.Split(version, ".")
 
+	if len(v) < 3 {
```
Zariel: a valid version can have < 3 fields, such as 3.2.

Zariel:
Thank you for the contribution

PR #660, opened 2016-02-20
Reuse a timer to prevent excessive allocations

jbardin:
- Every call to Conn.exec creates a new timer with the call to
  time.After. These are alive in memory for the duration of the timer,
  and can add a lot of overhead on a busy connection with a long
  timeout.
- Add a timer to the Conn struct that can be reused on every call to
  exec.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -9,7 +9,6 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
-	"github.com/gocql/gocql/internal/lru"
 	"io"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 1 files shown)


jbardin:
I saw this issue here: http://stackoverflow.com/questions/35512541/timer-overhead-causing-memory-overflow

Zariel:
Thanks for looking at this, unfortunately this will apply a timeout globally on a connection wheras the previous implementation will set a timeout per query. Queries are multiplexed onto connections so there can be many connections on the same connection. A fix could be to move the timer onto `callReq` which are recycled in a `sync.Pool`

jbardin:
Sorry, should have mentioned I'm not familiar with this code base at all, and might have missed something. Feel free to close this and push a better change in another PR if you want. I might be a few days before I have some time to look at it again.

Zariel:
Thanks for the patch and pointing to the StackOverflow issue, I've filed #661

Issue #658, opened 2016-02-18
No available streams error after upgrading gocql


muirrn:
We are running c\* 2.0.X so we only have 128 streams per connection. When all streams are in use, it looks like the old behavior was to block waiting for a new stream. The new behavior (via IDGenerator) is to return ErrNoStreams. One of our services makes a lot of queries very quickly (easily in the thousands). After upgrading gocql, the service gets an ErrNoStreams once it's kicked off 128 queries.

I think gocql should wait up to config.Timeout for an available stream rather than returning an error immediately. I don't see how my application can efficiently handle the exhausted streams situation since the only option as I see it is to retry queries in a busy CPU loop as long as you keep getting ErrNoStreams.

I looked at fixing this myself, but it seemed like introducing a channel into the IDGenerator would be prone to leaking goroutines or leaking streams, depending on how you implemented it. With some guidance, I would be happy to implement the solution.

As a side-note, we want to upgrade to 2.1.X (and get the 32K streams, which would probably alleviate this issue), but that might not be able to happen for some time.

Zariel:
This is more likely the query execution is not trying hard enough to get a connection with streams available. There should be `NumConns * NumHosts * (127 - 1)` streams available to the application, which should be lots even on proto2. Is this easily reproducible?

muirrn:
Yes, it happens every time I run my test program against our production ring. I will see if I can make a test case to reproduce it.

Trying different connections/different hosts would certainly improve things, but I still think it should block waiting for a stream if it can't get one. If you have a 3 node ring with the default 2 connections per-host, you have ~768 streams total which is still easily exhaustible, even if the query code is smarter and looks across connections for available streams.

muirrn:
Here is a test case:

```
func TestProtocolV2StreamExhaustion(t *testing.T) {
	cluster := createCluster()
	cluster.ProtoVersion = 2

	session := createSessionFromCluster(cluster, t)
	defer session.Close()

	if err := createTable(session, "CREATE TABLE gocql_test.test_stream_exhaustion (id int, data text, PRIMARY KEY (id))"); err != nil {
		t.Fatalf("failed to create test_stream_exhaustion with err: %v", err)
	}

	query := session.Query("INSERT INTO test_stream_exhaustion (id, data) VALUES (?,?)", 351, "Calliphora-piously")
	if err := query.Exec(); err != nil {
		t.Fatalf("failed to insert with err: %v", err)
	}

	var wg sync.WaitGroup

[...]

PR #666, opened 2016-03-01
Fix protocol v2 stream exhaustion.

muirrn:
Protocol 2 only gets 128 streams per-connection, so it is easy to
exhaust all your streams if you have a lot of parallel
queries. Previously, gocql would wait for an available stream, but after
a refactoring of how stream allocation works, an error is returned
immediately if no stream is available. I made two change to address
this:

- change round robin connection policy to fall back to plain round robin
  logic if none of the connections have available streams (rather than
  returning nil and causing the client to get a mysterious "no
  connections available" error)
- change stream allocation to wait up to config.Timeout for an available
  stream. I tried to make it do timeout accounting
  properly (i.e. subtract stream wait time from available timeout for
  remainder of query)

Issue #658

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -1977,6 +1977,40 @@ func TestNegativeStream(t *testing.T) {
 	}
 }
 
+func TestProtocolV2StreamExhaustion(t *testing.T) {
+	cluster := createCluster()
+	cluster.ProtoVersion = 2
+
+	session := createSessionFromCluster(cluster, t)
+	defer session.Close()
+
+	if err := createTable(session, "CREATE TABLE gocql_test.test_stream_exhaustion (id int, data text, PRIMARY KEY (id))"); err != nil {
+		t.Fatalf("failed to create test_stream_exhaustion with err: %v", err)
+	}
+
+	query := session.Query("INSERT INTO test_stream_exhaustion (id, data) VALUES (?,?)", 351, "Calliphora-piously")
+	if err := query.Exec(); err != nil {
+		t.Fatalf("failed to insert with err: %v", err)
+	}
+
+	var wg sync.WaitGroup
+
+	for i := 0; i < 1000; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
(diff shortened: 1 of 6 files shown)


Zariel:
I dont think using a timeout chan here is a good idea, as every query will potentially allocate a new timer here. Another issue is that why block on the first connection for a stream, what if the next one has available streams? Im not keen on making this the default, maybe it should be configurable on the session to be able to use the old implementation of the ID generator, or possibly use that just for protocol < 3?

But I get the feeling that if the need arises to limit the total concurrency then the app could use its own semaphore to limit usage.

muirrn:
First let me point out my change won't affect behavior at all if there are available streams (which must be the overwhelmingly normal case if I am the only one having this problem).

> I dont think using a timeout chan here is a good idea, as every query will potentially allocate a new timer here

It only makes a timer if there are no available streams. Doesn't exec() already create a timer for every query?

> why block on the first connection for a stream

It blocks on the first conn it picks because it was simple and easy to implement. If the conn policy waits for a connection to have an available stream it would need to allocate the stream itself (otherwise there's no guarantee the stream would be available when you start the query), and that seemed like a bigger/harder change.

> maybe it should be configurable on the session to be able to use the old implementation of the ID generator, or possibly use that just for protocol < 3?

There should be no difference for users that weren't running out of streams before, so I'm not sure extra config/stream logic is beneficial.

dax0:
@muirrn I'm having this issue as well getting "no connections available" and forced to reboot the service

cassandra_test.go
```diff
@@ -1977,6 +1977,40 @@ func TestNegativeStream(t *testing.T) {
 	}
 }
 
+func TestProtocolV2StreamExhaustion(t *testing.T) {
```
Zariel: A simpler test would be to use the unit test functions and just exhaust a connections streams.

muirrn: I'm not too familiar with the testing functions - which unit test functions should I use?

I wanted a roundtrip test for my use case because I wasn't sure that stream exhaustion was the only problem.

PR #667, opened 2016-03-02
Add support for Cassandra 3 system_schema metadata

nemothekid:
This PR is a *very* quick fix to get cluster metadata and TokenAwareRouting working for Cassandra 3.0 clusters. It needs tests, and the code isn't very clean (a lot of names were removed/moved, and I opted to just co-opt the old names from Cassandra 2).

The cluster detection just tests for a single Cassandra 3 node, as the new `system_schema` keyspace depends on Cassandra version, as opposed to the protocol version.


Zariel:
Thanks

PR #669, opened 2016-03-03
Fix 100ms sleep when creating session.

muirrn:
hostConnPool.fill() was synchronously filling the pool and calling
fillingStopped(), which sleeps for ~100ms.

- don't synchronously fill the entire conn pool (connect once
  synchronously, then fill the rest in a separate goroutine)
- only sleep in hostConnPool.fillingStopped if fill() encountered an
  error
- fix cleanup of bad system.peer in TestInvalidPeerEntry

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -88,9 +88,6 @@ func TestInvalidPeerEntry(t *testing.T) {
 		"169.254.235.45",
 	)
 
-	// clean up naughty peer
-	defer session.Query("DELETE from system.peers where peer == ?", "169.254.235.45").Exec()
-
 	if err := query.Exec(); err != nil {
 		t.Fatal(err)
 	}
@@ -100,7 +97,10 @@ func TestInvalidPeerEntry(t *testing.T) {
 	cluster := createCluster()
 	cluster.PoolConfig.HostSelectionPolicy = TokenAwareHostPolicy(RoundRobinHostPolicy())
 	session = createSessionFromCluster(cluster, t)
-	defer session.Close()
+	defer func() {
+		session.Query("DELETE from system.peers where peer = ?", "169.254.235.45").Exec()
+		session.Close()
+	}()
 
 	// check we can perform a query
 	iter := session.Query("select peer from system.peers").Iter()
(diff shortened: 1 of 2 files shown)


Zariel:
Thanks, I will merge this but first the tests need adjusting, `TestTokenAwareConnPool ` expects that the pool creation is synchronous for the initial connect.

cassandra_test.go
```diff
@@ -1920,8 +1920,18 @@ func TestTokenAwareConnPool(t *testing.T) {
 	session := createSessionFromCluster(cluster, t)
 	defer session.Close()
 
-	if expected := cluster.NumConns * len(session.ring.allHosts()); session.pool.Size() != expected {
-		t.Errorf("Expected pool size %d but was %d", expected, session.pool.Size())
+	expectedPoolSize := cluster.NumConns * len(session.ring.allHosts())
+
+	// wait for pool to fill
+	for i := 0; i < 10; i++ {
```
Zariel: this could lead to test flakes if implemented this way, which I would rather not risk.

connectionpool.go
```diff
@@ -439,19 +439,14 @@ func (pool *hostConnPool) fill() {
 
 		// filled one
 		fillCount--
-		// connect all remaining connections to this host
```
Zariel: Is there a reason to make this not finish synchronously? The returned session should be fully usable once created but in this case there may have been connection errors which are not caught here. Could you leave this code to return synchronously when the pool is empty to retain previous behaviour.

muirrn:
I see in 23e622bb3385d3a6c820471548997844494cf992 the pool filling was changed to be synchronous. The only reason mentioned was to deflake a test. Is there a user-facing reason for this change?

As a gocql user, I certainly do not want gocql to synchronously open all connections to my entire ring when I start my app. If one connection or host has a problem, my app is forced to wait the entire connection timeout, right?

connectionpool.go
muirrn: The reason is to reduce gocql startup time. There were conflicting comments in that code, so I restored the original behavior because I thought opening a single connection synchronously was a good balance (it gives synchronous feedback to user, but doesn't make app wait for all n connections to be established).

The previous code did not really handle errors differently (connectMany() simply logs errors).

PR #698, opened 2016-03-30
Adds parsing / handling for CQLv4 Custom Payload, bytes map, access via Iter

robmccoll:
This is pretty straightforward - needed it to be able to work with a custom QueryHandler C*-side.

### DIFF
--- a/session.go
+++ b/session.go
@@ -996,6 +996,24 @@ func (iter *Iter) Scan(dest ...interface{}) bool {
 	return true
 }
 
+// GetCustomPayload returns any parsed custom payload results if given in the
+// response from Cassandra. Note that the result is not a copy.
+//
+// This additional feature of CQL Protocol v4
+// allows additional results and query information to be returned by
+// custom QueryHandlers running in your C* cluster.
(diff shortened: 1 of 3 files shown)


session.go
```diff
@@ -996,6 +996,24 @@ func (iter *Iter) Scan(dest ...interface{}) bool {
...
+//
+// This additional feature of CQL Protocol v4
+// allows additional results and query information to be returned by
+// custom QueryHandlers running in your C* cluster.
+// See https://datastax.github.io/java-driver/manual/custom_payloads/
+func (iter *Iter) GetCustomPayload() map[string][]byte {
+	if iter == nil ||
+		iter.framer == nil ||
+		iter.framer.header == nil ||
+		iter.framer.header.customPayload == nil {
```
robmccoll: Great - dropped the nil checks.  Yeah - panics off of things that should never be nil have bitten me, but either way propagating the nil will probably still panic elsewhere, so finding the real cause is probably easier without the checks anyway :-)

Zariel:
this looks great, can you add yourself to the authors file and if your already there ill merge.

Zariel:
Thanks very much for the contribution

PR #703, opened 2016-04-07
Fixed IgnorePeerAddr config option

zachbadgett:
Currently the IgnorePeerAddr config option does not work, which causes starting up a session to take a long time since the hosts need to be changed and host pools need to be filled again. I'm sure there's a better solution to handling the host lookup and peer addresses, but for now this pull request fixes the issue.


zachbadgett:
Fairly certain that test fail is a false positive.

Zariel:
Yep, thanks for the fix. If you are not already could you add yourself to the AUTHORS file and I will merge.

zachbadgett:
Already in there :)

Zariel:
thanks very much :)

PR #706, opened 2016-04-11
Implemented WithTimestamp function in gocql.Query

Unix4ever:
This function is an addition to `DefaultTimestamp` function.
Using this function enables defaultTimestamp flag, but instead of
generating new timestamp, it is possible to set it's value explicitly.

It's equal to `USING TIMESTAMP x` in the statement, but does not break
prepared query optimization.


smira:
:+1:

Zariel:
Thanks very much for the contribution, if you are not already could you add yourself to the authors file.

Unix4ever:
I'm already in that list :)

Zariel:
I need to automate this :) thank you very much!

PR #707, opened 2016-04-13
Fix two bugs in stream id allocation.

muirrn:
- Previously if GetStream() found an available stream but the CAS
  failed, it was skipping that stream and trying the next. However, a
  CAS failure doesn't mean the current bit is taken, just that some bit
  in the bucket changed. Fix to keep retrying the CAS while the bit it's
  checking continues to be available in the bucket.
- When the CAS failed in GetStream(), it was re-loading the bucket using
  "offset" instead of "pos". This wouldn't end up causing any incorrect
  behavior, but it greatly reduces the chance of finding an available
  bit in the current bucket since the CAS will only succeed if
  c.streams[pos] happens to be the same as c.streams[offset] (assuming
  pos != offset).

### DIFF
--- a/internal/streams/streams.go
+++ b/internal/streams/streams.go
@@ -62,12 +62,12 @@ func (s *IDGenerator) GetStream() (int, bool) {
 
 		for j := 0; j < bucketBits; j++ {
 			mask := uint64(1 << streamOffset(j))
-			if bucket&mask == 0 {
+			for bucket&mask == 0 {
 				if atomic.CompareAndSwapUint64(&s.streams[pos], bucket, bucket|mask) {
 					atomic.AddInt32(&s.inuseStreams, 1)
 					return streamFromBucket(int(pos), j), true
 				}
-				bucket = atomic.LoadUint64(&s.streams[offset])
+				bucket = atomic.LoadUint64(&s.streams[pos])
 			}
 		}
 	}


muirrn:
I couldn't come up with a good testing strategy. The only symptom of the problem was you might get a stream even though there were some available under heavy stream contention. However, even after my changes that is still the case (just less common).

nemothekid:
>"you might get a stream even though there were some available under heavy stream contention"

Not sure what you mean here? Do you mean, you might not get a stream even though there some available?

I just did a quick look of the code - seems like your problem might be solved by randomizing the offset (are consecutive queries on the same connection trying to find a stream on the same offset)?

muirrn:
Yes, I meant you might *not* get a stream. I'm not sure randomizing the offset will help. The (starting) bucket offset already increments every time you call GetStream (i.e. per-query), plus GetStream() checks all the buckets every time. My changes are more helpful if you are still using protocol v2 and only have 128 streams per-connection (so, only 2 stream buckets).

Zariel:
The reason it skips the bucket is to reduce contention on that bucket

muirrn:
Can you see how it is problematic for protocol v2 where there are only 2 buckets? There is always going to be contention on both buckets, so skipping available bits under contention can easily cause you to not get any stream at all. Do you have any suggestions?

Zariel:
Thanks, ive got a test that this fixes, but im not sure how reliable the test is :)

Issue #554, opened 2015-12-26
Use server sent events to maintain a client side view of the keyspace schemas


Zariel:

martin-sucha:
Currently we clear schema on the schema change events:
https://github.com/gocql/gocql/blob/93eedddb64669e218f31cff48059066b03fabf83/events.go#L106-L123

 Do I understand correctly that this issue is for updating the individual schema changes based on the contents of messages without clearing the cache?

OleksiienkoMykyta:
The issue has a pretty poor description, the last update was on Jan 12 2016 with the PR, but the issue reporter didn't respond.
Candidate to be closed.

PR #715, opened 2016-04-27
Use server sent events to maintain a client side view of the keyspace schemas #554

tonkolviktor:
Hey,

Thanks for the lot of work put into this project.
We needed the shcema update feature, so i implemented based on our requirements.

Hope it helps and can be merged (or at least starts a conversation :))

### DIFF
--- a/events.go
+++ b/events.go
@@ -107,6 +107,9 @@ func (s *Session) handleEvent(framer *framer) {
 }
 
 func (s *Session) handleSchemaEvent(frames []frame) {
+	if s.schemaDescriber == nil {
+		return
+	}
 	for _, frame := range frames {
 		switch f := frame.(type) {
 		case *schemaChangeKeyspace:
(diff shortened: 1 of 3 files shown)


events.go
```diff
@@ -107,6 +107,14 @@ func (s *Session) handleEvent(framer *framer) {
 }
 
 func (s *Session) handleSchemaEvent(frames []frame) {
+	for _, frame := range frames {
```
Zariel: It looks like there are cases where the sessions `schemaDescriber` is nil, we should check and not clear metadata.

events.go
```diff
@@ -107,6 +107,9 @@ func (s *Session) handleEvent(framer *framer) {
 }
 
 func (s *Session) handleSchemaEvent(frames []frame) {
+	if s.schemaDescriber == nil {
```
tonkolviktor: thanks good point, fixed

Zariel:
Thanks, if not already can you add yourself to the AUTHORS file and I will merge.

tonkolviktor:
done, thanks

Zariel:
Thank you very much for the contribution!

PR #723, opened 2016-05-02
Attempt to reconnect known DOWN nodes

matope:
Hi, guys.

I've found that gocql doesn't try to reconnect to nodes which connection has been closed with TooManyTimeouts error.

For now, once a node goes DOWN, gocql can't get it back to UP status without UP server-push events.

But in some cases, gocql make nodes DOWN by itself. (e.g. after timeouts : https://github.com/gocql/gocql/blob/51009831ae7a79fd8f1b160d7359916a7a044afa/connectionpool.go#L399 )
In those case, UP events will not occur naturally so the node can't get back to UP forever.

So, I think polling for the DOWN nodes to make them UP would be a good way to keep connections.

Thanks.

### DIFF
--- a/session.go
+++ b/session.go
@@ -9,8 +9,8 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"golang.org/x/net/context"
 	"io"
+	"log"
 	"net"
 	"strconv"
 	"strings"
@@ -18,6 +18,8 @@ import (
 	"sync/atomic"
 	"time"
 	"unicode"
+
+	"golang.org/x/net/context"
 
 	"github.com/gocql/gocql/internal/lru"
 )
@@ -173,6 +175,31 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 
 			s.handleNodeUp(net.ParseIP(host.Peer()), host.Port(), false)
 		}
+	}
+
+	if cfg.ReconnectInterval > 0 {
+		go func() {
+			for !s.Closed() {
+				time.Sleep(cfg.ReconnectInterval)
+				hosts := s.ring.allHosts()
+
+				// Print session.ring for debug.
+				if gocqlDebug {
+					buf := bytes.NewBufferString("Session.ring:")
+					for _, h := range hosts {
+						buf.WriteString("[" + h.Peer() + ":" + h.State().String() + "]")
+					}
+					log.Println(buf.String())
+				}
+
+				for _, h := range hosts {
+					if h.IsUp() {
+						continue
+					}
+					s.handleNodeUp(net.ParseIP(h.peer), h.port, true)
+				}
+			}
+		}()
 	}
 
 	// TODO(zariel): we probably dont need this any more as we verify that we
--- a/session.go
+++ b/session.go
@@ -178,28 +178,7 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 	}
 
 	if cfg.ReconnectInterval > 0 {
-		go func() {
-			for !s.Closed() {
-				time.Sleep(cfg.ReconnectInterval)
-				hosts := s.ring.allHosts()
-
-				// Print session.ring for debug.
-				if gocqlDebug {
-					buf := bytes.NewBufferString("Session.ring:")
-					for _, h := range hosts {
-						buf.WriteString("[" + h.Peer() + ":" + h.State().String() + "]")
-					}
-					log.Println(buf.String())
-				}
-
-				for _, h := range hosts {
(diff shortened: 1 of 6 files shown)


matope:
Also, I added some debug logs to make this patch.
I'd like to remove these if they're too verbose.

matope:
Following is the result of my experiment with current master branch (with some debug print).

- At 18:15:27, I shut down 127.0.0.3 (`sudo iptables -A INPUT -p tcp --dest 127.0.0.3 --dport 9042 -j DROP`).
- At 18:15:47, I recover 127.0.0.3 (`sudo iptables -D INPUT 1 `)
- But after that, `127.0.0.3` doesn't come back to the roundRobinHostPolicy's pool.

```
2016/05/02 18:15:07 RoundRobinHostPolicy:[127.0.0.1, UP],[127.0.0.3, UP],[127.0.0.2, UP],
2016/05/02 18:15:17 RoundRobinHostPolicy:[127.0.0.1, UP],[127.0.0.3, UP],[127.0.0.2, UP],
2016/05/02 18:15:27 RoundRobinHostPolicy:[127.0.0.1, UP],[127.0.0.3, UP],[127.0.0.2, UP],    // a.
2016/05/02 18:15:27 TimeoutLimit Exceeded! 0x6be640
2016/05/02 18:15:27 unable to dial "127.0.0.3": dial tcp 127.0.0.3:9042: i/o timeout 
2016/05/02 18:15:37 RoundRobinHostPolicy:[127.0.0.1, UP],[127.0.0.2, UP],
2016/05/02 18:15:47 RoundRobinHostPolicy:[127.0.0.1, UP],[127.0.0.2, UP], // b.
2016/05/02 18:15:57 RoundRobinHostPolicy:[127.0.0.1, UP],[127.0.0.2, UP],
2016/05/02 18:16:07 RoundRobinHostPolicy:[127.0.0.1, UP],[127.0.0.2, UP],
2016/05/02 18:16:17 RoundRobinHostPolicy:[127.0.0.1, UP],[127.0.0.2, UP],
2016/05/02 18:16:27 RoundRobinHostPolicy:[127.0.0.1, UP],[127.0.0.2, UP],
2016/05/02 18:16:37 RoundRobinHostPolicy:[127.0.0.1, UP],[127.0.0.2, UP],
2016/05/02 18:16:47 RoundRobinHostPolicy:[127.0.0.1, UP],[127.0.0.2, UP],
```

This patch fixes this issue.

session.go
```diff
@@ -175,6 +177,31 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 		}
 	}
 
+	if cfg.ReconnectInterval > 0 {
+		go func() {
```
Zariel: pull this goroutine into its own method on session

session.go
```diff
@@ -175,6 +177,31 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
...
+						buf.WriteString("[" + h.Peer() + ":" + h.State().String() + "]")
+					}
+					log.Println(buf.String())
+				}
+
+				for _, h := range hosts {
+					if h.IsUp() {
+						continue
+					}
+					s.handleNodeUp(net.ParseIP(h.peer), h.port, true)
```
Zariel: need to use `h.Peer()` and `h.Port()`, though they are unlikely to change.

Zariel:
Thanks very much for the patch, looks good. I had planned to add something similar, but have it attempt to dial each downed node using a back off strategy, this is much simpler and straightforward.

session.go
```diff
@@ -215,6 +194,32 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 	return s, nil
 }
 
+func (s *Session) startReconnection(itvl time.Duration) {
+	go func() {
```
Zariel: I meant to define a method like, 

```go
func (s *Session) recconectDownedHosts() {
    for {
...
}

go s.recconectDownedHosts()
```

Zariel:
The test failures look unrelated

session.go
matope: I see. I'll rewrite this like that.

Zariel: sorry for not making it clearer the first time :(

matope: No problem.
I've overwritten the last commit. 2ea86d3ae9baa67e82490af4c8caaf622f76794c

Zariel:
Thanks very much!

matope:
Thanks for the marge!
This time I took a polling strategy for simplicity and smaller impact but I think the back off strategy would be a better way for more quick reconnecting. So I think making it back off would be a good next step :)

PR #726, opened 2016-05-08
Notify the HostSelectionPolicy of the partioner

### LINKED ISSUE #724: SetPartitioner never called on the HostSelectionPolicy
Looks like the call to SetPartitioner on the HostSelectionPolicy was lost in commit a04d2a5e3f4076edc106347f1624809c0733a1a1. The tokenAwareHostPolicy needs SetPartitioner to be called before it will be effective.

jncorpron:
Adds a call to SetPartitioner on the session's HostSelectionPolicy. 
This is mostly needed so that the token aware policy initializes 
and updates its token ring representation.

Fixes #724 

- I've updated my email address in the AUTHORS file.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -43,7 +43,7 @@ James Maloney <jamessagan@gmail.com>
 Ashwin Purohit <purohit@gmail.com>
 Dan Kinder <dkinder.is.me@gmail.com>
 Oliver Beattie <oliver@obeattie.com>
-Justin Corpron <justin@retailnext.com>
+Justin Corpron <jncorpron@gmail.com>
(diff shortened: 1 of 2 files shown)


jncorpron:
I see that clusterMetadata has the partitioner class value, but
that doesn't have access to call the session's HostSelectionPolicy.
Maybe I can provide a further PR with a change that integrates
the policy with the clusterMetadata, but this fix seemed the simplest
to fix the TokenAwareSelectionPolicy functionality.

Zariel:
Thanks, my intention was to have the policies use clusterMetadata, but looks like I never got round to it :)

PR #728, opened 2016-05-10
Avoid ErrTooManyTimeouts if TimeoutLimit is 0

matope:
Hi, again!

In my environment, many ErrTooManyTimeouts occurs and many in-flight requests are closed with the error abruptly.

I'm not sure if counting timeouts would be a good way to detect failure for the coordinator nodes.
Timeouts can be occurred by not only coordinator nodes but also specific heavy query( like scanning ) or heavy-loading of data-nodes.

Also, I'm thinking of HostSelectionPolicy might be a good place to handle these kind things.

So, In this branch, I'd like to add an option to avoid this way to ErrTooManyTimeouts by setting TimeoutLimit zero.

I'd appreciate it if you could review this PR.
Thanks.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -113,7 +113,7 @@ func (fn connErrorHandlerFn) HandleError(conn *Conn, err error, closed bool) {
 	fn(conn, err, closed)
 }
 
-// How many timeouts we will allow to occur before the connection is closed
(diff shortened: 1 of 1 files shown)


Zariel:
Thanks very much, I agree it is not ideal and I guess now we could just rely on the cluster to tell us when nodes are down.

PR #732, opened 2016-05-13
New: added the possibility to define the type needed to convert an uuid

Dogild:
Previously, when getting a uuid from cassandra, the uuid was always converted in gocql.UUID.
This PR adds the possibility to define a custom type for that. Right now you have the choice between string and gocql.UUID (default). When using the string choice, every uuid will be converted in a string.
You can define this behavior with the variable gocql.DefaultUUIDType, CustomUUIDTypeUUID and CustomUUIDTypeString are available.

The goal of this method is the possibility to create model which will talk with different storages. For instance, objects of a model could be store in a reddis and cassandra database, the reddis library would not want to handle the type gocql.UUID...


Zariel:
Is this not already possible by implementing the `MarshalCQL` and `UnmarshalCQL` methods? The UUID also takes a `string` or a `[]byte` which are compatible between many libraries

Dogild:
Could you give more informations about this feature please ? I'm kind of new with this lib :D
To add more details about the UUID, this feature is to control the type got in the SliceMap of the iterator.

PR #736, opened 2016-05-18
Pool query objects

i:
Queries account for a lot of unnecessary allocations. This PR exposes a `Release` method on the `Query` type so that callers of `Session.Query()` can reduce the number of allocations.

### DIFF
--- a/session.go
+++ b/session.go
@@ -68,8 +68,12 @@ type Session struct {
 
 	closeMu  sync.RWMutex
 	isClosed bool
+}
 
-	queryPool *sync.Pool
+var queryPool = &sync.Pool{
+	New: func() interface{} {
+		return new(Query)
+	},
 }
 
 func addrsToHosts(addrs []string, defaultPort int) ([]*HostInfo, error) {
@@ -100,12 +104,11 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 	}
 
 	s := &Session{
-		cons:      cfg.Consistency,
-		prefetch:  0.25,
-		cfg:       cfg,
-		pageSize:  cfg.PageSize,
-		stmtsLRU:  &preparedLRU{lru: lru.New(cfg.MaxPreparedStmts)},
-		queryPool: &sync.Pool{New: func() interface{} { return new(Query) }},
+		cons:     cfg.Consistency,
+		prefetch: 0.25,
+		cfg:      cfg,
+		pageSize: cfg.PageSize,
+		stmtsLRU: &preparedLRU{lru: lru.New(cfg.MaxPreparedStmts)},
 	}
 
 	connCfg, err := connConfig(s)
@@ -262,7 +265,7 @@ func (s *Session) SetTrace(trace Tracer) {
 // if it has not previously been executed.
 func (s *Session) Query(stmt string, values ...interface{}) *Query {
 	s.mu.RLock()
-	qry := s.queryPool.Get().(*Query)
+	qry := queryPool.Get().(*Query)
 	qry.stmt = stmt
 	qry.values = values
 	qry.cons = s.cons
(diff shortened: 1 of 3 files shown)


nemothekid:
On a very quick look through, doesn't it make more sense to eschew the `Release` method and just invalidate/`Put` the `Query` reference after `Close()` is called?

session.go
```diff
@@ -98,11 +100,12 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
...
-		prefetch: 0.25,
-		cfg:      cfg,
-		pageSize: cfg.PageSize,
-		stmtsLRU: &preparedLRU{lru: lru.New(cfg.MaxPreparedStmts)},
+		cons:      cfg.Consistency,
+		prefetch:  0.25,
+		cfg:       cfg,
+		pageSize:  cfg.PageSize,
+		stmtsLRU:  &preparedLRU{lru: lru.New(cfg.MaxPreparedStmts)},
+		queryPool: &sync.Pool{New: func() interface{} { return new(Query) }},
```
Zariel: this pool can be package local

session.go
```diff
@@ -947,6 +956,44 @@ func (q *Query) MapScanCAS(dest map[string]interface{}) (applied bool, err error
...
+		return
+	}
+	pool := q.session.queryPool
+	q.reset()
+	pool.Put(q)
+}
+
+// reset zeroes out all fields of a query so that it can be safely pooled.
+func (q *Query) reset() {
+	q.stmt = ""
```
Zariel: i dont think all these fields need to be reset, only ones which dont get set by `Query()`

Zariel:
I agree with @nemothekid it would be better to do this in Close and update the documentation to state that the Query must not be reused after close is called.

i:
@nemothekid and @Zariel,

Are you guys referring to to `Iter.Close()`? What happens in the case where a query is used more than once? Although it isn't explicitly supported, I don't think it's unreasonable to do something like this in order to reuse a query:
```go
qry := session.Query("select * from health_check")
for ;; time.Sleep(time.Second) {
  if err := qry.Exec(); err != nil {
    ...
  }
}

```

Am I wrong in my assumption?

session.go
```diff
@@ -947,6 +956,44 @@ func (q *Query) MapScanCAS(dest map[string]interface{}) (applied bool, err error
...
 
+// Release releases a query back into a pool of queries. Released Queries
+// cannot be reused.
+//
+// Example:
+// 		qry := session.Query("SELECT * FROM my_table")
+// 		qry.Exec()
+// 		qry.Release()
+func (q *Query) Release() {
+	if q == nil || q.session == nil || q.session.queryPool == nil {
```
Zariel: this can be simplified if the pool is a package local variable then it will never be nil, how worthwhile is the query nil check?

Zariel:
Ah sorry I miss understood/forgot how the query api was working in the case without iterators:)

Patch looks good once the suggested changes to move the pool out of session. If you are not already could you please add yourself to the AUTHORS file and ill merge once the changes have been made.

session.go
i: Assuming this is invalid since we're pooling at the package-level now.

session.go
```diff
@@ -947,6 +959,40 @@ func (q *Query) MapScanCAS(dest map[string]interface{}) (applied bool, err error
...
+// Release releases a query back into a pool of queries. Released Queries
+// cannot be reused.
+//
+// Example:
+// 		qry := session.Query("SELECT * FROM my_table")
+// 		qry.Exec()
+// 		qry.Release()
+func (q *Query) Release() {
+	q.reset()
+	pool.Put(q)
```
Zariel: this should be `queryPool.Put(q)`

i: good catch, thanks

Zariel:
Thanks very much for the contribution :)

i:
Thanks for the opportunity 😃

PR #737, opened 2016-05-19
Add Session.DisableSkipMetadata to avoid panic by CASSANDRA-10786

matope:
Hi.

After we `ALTER TABLE ADD COLUMEN...` on a live cluster, almost all C* clients using gocql failed with panic.
As a result of some investigation, we found that the panic has been caused by [CASSANDRA-10786](https://issues.apache.org/jira/browse/CASSANDRA-10786).

The metadata for EXECUTE ROWS results have been changed after a `ALTER TABLE` but only one client can handle the change of the metadata. Other clients failed to unmarshal (especially for parse of collection type) and panic.

So, In this ticket, I'd like to add a way to disable SkipMetadata within the session to avoid this issue.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -110,6 +110,14 @@ type ClusterConfig struct {
 		DisableSchemaEvents bool
 	}
 
+	// DisableSkipMetadata will override the internal result metadata cache so that the driver does not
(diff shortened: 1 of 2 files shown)


Zariel:
Thanks!

PR #740, opened 2016-05-25
Implemented WithTimestamp function in gocql.Batch

Unix4ever:
This function is an addition to DefaultTimestamp function.
Using this function enables defaultTimestamp flag, but instead of
generating new timestamp, it is possible to set it's value explicitly.

It's equal to USING TIMESTAMP x in the statement, but does not break
prepared query optimization.

Similar to #706 PR but for batches.

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -1414,8 +1414,9 @@ type writeBatchFrame struct {
 	consistency Consistency
 
 	// v3+
-	serialConsistency SerialConsistency
-	defaultTimestamp  bool
+	serialConsistency     SerialConsistency
+	defaultTimestamp      bool
+	defaultTimestampValue int64
 }
 
 func (w *writeBatchFrame) writeFrame(framer *framer, streamID int) error {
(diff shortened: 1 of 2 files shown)


smira:
👍

frame.go
```diff
@@ -1469,9 +1470,15 @@ func (f *framer) writeBatchFrame(streamID int, w *writeBatchFrame) error {
 		if w.serialConsistency > 0 {
 			f.writeConsistency(Consistency(w.serialConsistency))
 		}
+
 		if w.defaultTimestamp {
-			now := time.Now().UnixNano() / 1000
-			f.writeLong(now)
+			var ts int64
+			if w.defaultTimestampValue != 0 {
```
Zariel: sometimes it is useful to do an insert with the timestamp set to 0, I guess users could use the `USING TIMESTAMP 0` method, but I think this could be improved, while this problem exists in the query method as well

smira: @Zariel it is used to insert with "as old as possible" timestamp, so that any recent write would "overwrite" it?

Zariel: Yes, for example you are backfilling from another database, you can insert all data from the old database while writes continue, any write that is newer will be the one returned from reads.

Issue #696, opened 2016-03-28
2 seconds for a call to CreateSession


mattmassicotte:
I've been using (and loving) gocql for a while now. Normally, I use Cassandra 2.x, but I've been trying out 3.3 recently. After changing the ProtoVersion to 4, things are working. However, every query now takes almost exactly 2 seconds. The same queries from cqlsh don't have the same problem. Any ideas?

MalteJ:
how do you change the protoversion to 4?

mattmassicotte:
I followed the procedure from this issue:

https://github.com/gocql/gocql/issues/538

Without this, gocql is unable to connect at all.

mattmassicotte:
Welp, I'm mistaken. Obviously something else has changed on my end, because this is now reproducing with 2.x as well. Apologies.

mattmassicotte:
Actually, I narrowed this down a little. Me changing Cassandra versions was a red herring. I also updated gocql, and that was what actually did it. The perf change is in the call to ClusterConfig.CreateSession. It appears that as of recent versions, this call now blocks for almost two seconds.

Is that now expected?

MalteJ:
Same here. I have checked out a commit at Dec 31st 2015 and it's very fast.
But the current version is very slow.
@Zariel do you know what's wrong here?

Zariel:
is it related to #669?

MalteJ:
I am not sure but I don't think so.

mattmassicotte:
More clues. I run cassandra inside a docker container on my mac. There is a single node listed as a seed. This seed IP is visible to the cassandra process, but not from my mac. This makes it look to gocql like the one seed node is down.

MalteJ:
I was using Docker on Mac with the _/cassandra image, too!
So I have created my own cassandra docker image and now I do not have the problems anymore.
You can try my image: `docker run -it -p 9042:9042 parce/cassandra:3.4`
You can find my Dockerfile at https://github.com/parce-iot/docker-cassandra

Zariel:
This related to #703?

MalteJ:
With the official docker cassandra image and current gocql release I still have problems with the delay.
Maybe there are some strange settings in the docker image.

MalteJ:
Anyhow I will not further investigate as I have solved this issue by creating an own image.

Highstead:
Underlying cause is from the initial host lookup.   The quick and dirty solution is:
[...]

PR #751, opened 2016-07-13
Stop retrying to connect on permanent failure

Highstead:
Minor change to the connection pool code that uses the *net.OpError type
to pull of retrying to connect to a host if the error is a permanent
erorr.

Reduces startup time on creating a new sesession when we try to
establish a connection using gocql that does not have visibility to the
host.

Performance improvement easily reproduced when using default cassandra
docker container with gocql.

Relates to issue #696


Zariel:

PR #754, opened 2016-07-20
Fix incorrect random partioner hash function

### LINKED ISSUE #753: Random partitioner hash function does not match the reference python implementation
The tokens produced by this function: https://github.com/gocql/gocql/blob/master/token.go#L92 do not match the reference python implementation: https://github.com/datastax/python-driver/blob/3fa4768304165fb9429abf8b341f2221d8149f62/cassandra/metadata.py#L1516-L1525 .

Because of this we were seeing requests hitting cassandra nodes unevenly. In my opinion the best course of action would be to match the python implementation.

Steps to reproduce:

    † pip install cassandra-driver
    † python

    >>> from cassandra.metadata import MD5Token
    >>> MD5Token.hash_fn("test")
    12707736894140473154801792860916528374L

The equivalent gocql result:

    † go run ~/Desktop/main.go
    664505362348203461494978569832354303401755427454

I will create a PR fixing the issue shortly.

calebdoxsey:
Fixes #753

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -71,3 +71,5 @@ Viktor Tönköl <viktor.toenkoel@motionlogic.de>
 Ian Lozinski <ian.lozinski@gmail.com>
 Michael Highstead <highstead@gmail.com>
 Sarah Brown <esbie.is@gmail.com>
+Caleb Doxsey <caleb@datadoghq.com>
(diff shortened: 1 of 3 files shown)


Zariel:
Great thanks!

PR #758, opened 2016-07-27
marshal: Fix unmarshalUDT for less values than UDT fields

### LINKED ISSUE #757: UDT not work with ScyllaDB
Env:
ScyllaDB [https://github.com/scylladb/scylla] version `1.2.2`
gocql version `94f1315e46b0b2eae3a48a77a49e3e1cc9377923`

user following User Define Type
```
CREATE TYPE A.F (
	m   ascii,
	ft   ascii,	 
	fn  text,	 
	sz  bigint,
	td   ascii
);
```
and following Table
```

CREATE TABLE A.s (
  si  bigint PRIMARY KEY,
  ru                  frozen<A.F>,  
```

with the guide in `https://github.com/gocql/gocql/blob/master/udt_test.go`

I use the following method

```
import "fmt"
import "github.com/gocql/gocql"

type F struct {
	M      string `cql:"m"`
	Ft           string `cql:"ft"`
	Fn              string `cql:"fn"`
	Sz            uint64 `cql:"sz"`
	Td           string `cql:"td"`
}

func (p *F) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
	switch name {
	case "m":
		return gocql.Marshal(info, p.M)
	case "ft":
		return gocql.Marshal(info, p.Ft)
	case "fn":
		return gocql.Marshal(info, p.Fn)
	case "sz":
		return gocql.Marshal(info, p.Sz)
	case "td":
		return gocql.Marshal(info, p.Td)
	default:
		return nil, fmt.Errorf("unknown column for position: %q", name)
	}
}

func (p *F) UnmarshalUDT(name string, info gocql.TypeInfo, data []byte) error {
	switch name {
	case "m":
		return gocql.Unmarshal(info, data, &p.M)
	case "ft":
		return gocql.Unmarshal(info, data, &p.Ft)
	case "fn":
		return gocql.Unmarshal(info, data, &p.Fn)
	case "sz":
		return gocql.Unmarshal(info, data, &p.Sz)
	case "td":
		return gocql.Unmarshal(info, data, &p.Td)
	default:
		return fmt.Errorf("unknown column for position: %q", name)
	}
}


s:=&F{}
err = session.Query(
		`SELECT 
				ru
				FROM A.s limit 1` ).Scan(s); 
```

But It came up err
```
panic: gocql: not enough columns to scan into: have 1 want 5

goroutine 1 [running]:
main.main()
	<PATH>.go:114 +0x272


```

a deep view digs out it came from `session.go:1053`. 

when remove these lines. It cameout the result's `col.TypeInfo.Type` is `TypeTuple`. Not expected `UDTTypeInfo`.  So I can't convert result to struct F.

I stucked here. any help is welcomed.

penberg:
The CQL binary protocol specification states that a serialized UDT is
allowed to have _less_ values than it has fields:

> A UDT value will generally have one value for each field of the type it represents, but it is allowed to have less values than the type has fields.

This is also evident in the Cassandra Java driver's UDTCodec class:

```java
  public UDTValue deserialize(ByteBuffer bytes) {
      ByteBuffer input = bytes.duplicate();
      UDTValue value = definition.newValue();

      int i = 0;
      while (input.hasRemaining() && i < value.values.length) {
          int n = input.getInt();
          value.values[i++] = n < 0 ? null : readBytes(input, n);
      }
      return value;
  }
```

Fix the unmarshalUDT() function to check whether there's more data to
read to avoid accessing a slice out of bounds:

```
  panic: runtime error: slice bounds out of range

  goroutine 1 [running]:
  github.com/gocql/gocql.unmarshalUDT(0x7ffff7fbc380, 0xc208030c60, 0x0, 0x0, 0x0, 0x694f60, 0xc208094da0, 0x0, 0x0)
      /disk/GOPATH/src/github.com/gocql/gocql/marshal.go:1654 +0x574
  github.com/gocql/gocql.Unmarshal(0x7ffff7fbc380, 0xc208030c60, 0x0, 0x0, 0x0, 0x694f60, 0xc208094da0, 0x0, 0x0)
      /disk/GOPATH/src/github.com/gocql/gocql/marshal.go:157 +0x9b7
  github.com/gocql/gocql.(*Iter).Scan(0xc20806c630, 0xc20800adc0, 0x1, 0x1, 0x7ffff7fa9000)
      /disk/GOPATH/src/github.com/gocql/gocql/session.go:1084 +0x6ff
  github.com/gocql/gocql.(*Query).Scan(0xc208082700, 0xc20800adc0, 0x1, 0x1, 0x0, 0x0)
      /disk/GOPATH/src/github.com/gocql/gocql/session.go:918 +0xab
  main.main()
      /home/a.go:86 +0x4cd
```

Fixes #757.

### DIFF
--- a/udt_test.go
+++ b/udt_test.go
@@ -503,3 +503,42 @@ func TestUDT_UpdateField(t *testing.T) {
 		t.Errorf("expected %+v: got %+v", *writeCol, *readCol)
 	}
 }
+
+func TestUDT_ScanNullUDT(t *testing.T) {
+	if *flagProto < protoVersion3 {
+		t.Skip("UDT are only available on protocol >= 3")
+	}
+
+	session := createSession(t)
+	defer session.Close()
+
+	err := createTable(session, `CREATE TYPE gocql_test.scan_null_udt_position(
+		lat int,
+		lon int,
+		padding text);`)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = createTable(session, `CREATE TABLE gocql_test.scan_null_udt_houses(
+		id int,
+		name text,
+		loc frozen<position>,
+		primary key(id)
+	);`)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = session.Query("INSERT INTO scan_null_udt_houses(id, name) VALUES(?, ?)", 1, "test" ).Exec()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	pos := &position{}
+
+	err = session.Query("SELECT loc FROM scan_null_udt_houses WHERE id = ?", 1).Scan(pos)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
--- a/AUTHORS
+++ b/AUTHORS
@@ -73,3 +73,4 @@ Michael Highstead <highstead@gmail.com>
 Sarah Brown <esbie.is@gmail.com>
 Caleb Doxsey <caleb@datadoghq.com>
 Frederic Hemery <frederic.hemery@datadoghq.com>
+Pekka Enberg <penberg@scylladb.com>
--- a/marshal.go
+++ b/marshal.go
@@ -1651,6 +1651,9 @@ func unmarshalUDT(info TypeInfo, data []byte, value interface{}) error {
 		udt := info.(UDTTypeInfo)
 
 		for _, e := range udt.Elements {
+			if len(data) == 0 {
+				return nil
+			}
 			size := readInt(data[:4])
 			data = data[4:]
 
@@ -1689,6 +1692,9 @@ func unmarshalUDT(info TypeInfo, data []byte, value interface{}) error {
 		m := *v
 
 		for _, e := range udt.Elements {
+			if len(data) == 0 {
+				return nil
+			}
 			size := readInt(data[:4])
 			data = data[4:]
 


penberg:
I have tested this with both Cassandra 2.1.8 and ScyllaDB 1.2.2 using the reproducer found in the issue: https://gist.github.com/slayercat/d2820d46bfbd5f34424dacad4175443a

Zariel:
Thanks, the map unmarshalling code for udt seems to have the same bug, but the reflection case does not. Could you add the change to the map logic and also add the repro as a test case?

penberg:
@Zariel done!

Zariel:
Great, if not already can you add yourself to the AUTHORS file and ill merge

Zariel:
Actually looks like there is an issue with the test, I think its reusing udt names?

`--- FAIL: TestUDT_ScanNullUDT (0.21s)
	udt_test.go:520: A user type of name gocql_test.position already exists
`

udt_test.go
```diff
@@ -503,3 +503,42 @@ func TestUDT_UpdateField(t *testing.T) {
...
+
+func TestUDT_ScanNullUDT(t *testing.T) {
+	if *flagProto < protoVersion3 {
+		t.Skip("UDT are only available on protocol >= 3")
+	}
+
+	session := createSession(t)
+	defer session.Close()
+
+	err := createTable(session, `CREATE TYPE gocql_test.position(
```
Zariel: is there already a udt called position defined? Probably best to prefix this with the test name

penberg:
@Zariel Fixed!

Zariel:
Great thanks for the contribution!

PR #761, opened 2016-08-01
Use correct CQL version even if `DisableInitialHostLookup` is true

mark-i-m:
When `DisableInitialHostLookup` is true, `addrToHost` is called instead of doing a lookup in `system.local` and `system.peers`. This means that the host's version is the zero value (version 0) for all hosts.

Thus, calls to `KeyspaceMetadata` fail for cassandra versions >= 3.x because gocql attempts to make queries on `system.schema_keyspaces`, which no longer exists.

This PR simply stipulates that if `DisableInitialHostLookup` is true, we should use CQL version specified by the ClusterConfig that was used to create the session. Otherwise, we use the version of the first host, as before.

### DIFF
--- a/session.go
+++ b/session.go
@@ -195,7 +195,14 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 		return nil, ErrNoConnectionsStarted
 	}
 
-	s.useSystemSchema = hosts[0].Version().Major >= 3
+	// If initial host lookup is disabled, then we should use the ClusterConfig CQL version
+	if cfg.DisableInitialHostLookup {
+		var v cassVersion
+		v.Set(cfg.CQLVersion)
+		s.useSystemSchema = v.Major >= 3
+	} else {
+		s.useSystemSchema = hosts[0].Version().Major >= 3
+	}
 
 	return s, nil
 }
--- a/session.go
+++ b/session.go
@@ -195,11 +195,13 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 		return nil, ErrNoConnectionsStarted
 	}
 
-	// If initial host lookup is disabled, then we should use the ClusterConfig CQL version
-	if cfg.DisableInitialHostLookup {
-		var v cassVersion
-		v.Set(cfg.CQLVersion)
-		s.useSystemSchema = v.Major >= 3
+	// If we disable the initial host lookup, we need to still check if the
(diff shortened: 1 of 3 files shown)


mark-i-m:
Any news?

session.go
```diff
@@ -195,7 +195,14 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
...
 	}
 
-	s.useSystemSchema = hosts[0].Version().Major >= 3
+	// If initial host lookup is disabled, then we should use the ClusterConfig CQL version
+	if cfg.DisableInitialHostLookup {
+		var v cassVersion
+		v.Set(cfg.CQLVersion)
+		s.useSystemSchema = v.Major >= 3
+	} else {
+		s.useSystemSchema = hosts[0].Version().Major >= 3
```
Zariel: I dont think we should set this here as it should already be set

mark-i-m: Hmmm... when I try this, I get cassandra's `unconfigured table ...` error. Whereas with this patch, it seems to work

mark-i-m: To the best that I can tell, the problem is on line 159... `addrsToHosts` creates `HostInfo` with version set to the zero-value of `cassVersion`

Zariel: My mistake sorry. Im not sure tying the schema to the cql version as it will always be 3.x.y for a while, maybe we should have an option to pass in cassandra version for this?

mark-i-m: I'm not sure I understand what you mean... The CQL version is what determines the system schema layout, to the best of my knowledge, so schema is tied to CQL version. Am I misunderstanding?

Zariel: The CQL version is the version of the query language. Helpfully Cassandra likes to change the system table schemas without changing the major version bump, the only reliable way I can think to detect is to either query the schema or to use the version of Cassandra that is running

mark-i-m: Ah, that makes sense
Thanks :)

mark-i-m:
Hello, is there any news?

Zariel:
The CQLVersion does not indicate what the internal table structure is like, this is tied only to the Cassandra version

mark-i-m:
Ah, sorry... I had forgotten about the previous thread of conversation. I have made the change you suggested. Please let me know what you think.

Zariel:
Thanks for the contribution!

PR #763, opened 2016-08-03
Issue 743 Fix

marcusking01:
Added a nil check before trying to use a host to execute a query.  This was causing nil pointer dereference exceptions as described at https://github.com/gocql/gocql/issues/743. Also made some optimizations to reduce lock scoping in the TokenAwareHostPolicy

### DIFF
--- a/policies.go
+++ b/policies.go
@@ -13,7 +13,7 @@ import (
 	"github.com/hailocab/go-hostpool"
 )
 
-// cowHostList implements a copy on write host list, its equivilent type is []*HostInfo
+// cowHostList implements a copy on write host list, its equivalent type is []*HostInfo
 type cowHostList struct {
 	list atomic.Value
 	mu   sync.Mutex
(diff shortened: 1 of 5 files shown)


policies.go
```diff
@@ -263,9 +263,6 @@ type tokenAwareHostPolicy struct {
 }
 
 func (t *tokenAwareHostPolicy) SetPartitioner(partitioner string) {
-	t.mu.Lock()
```
Zariel: im not sure removing this lock is correct, as its protecting concurrent access to `t.partitioner`

marcusking01: The only thing that SetPartitioner does is call to the fallback policies Partitioner (which in the case of HostPool and RoundRobin are no-ops)  then calls to the resetTokenRing which is where I moved the lock to.

Zariel:
Ill merge this once I confirm there are no data races around the partitioner on the token aware policy

Zariel:
Thanks for the contribution, ive merged this manually. Sorry for the delay!

PR #769, opened 2016-08-16
Non-panic ParseConsistency added (#749)

### LINKED ISSUE #749: Non-panic ParseConsistency
ParseConsistency() will panic (instead of returning an error) if you give it a non-standard value

Can you add this function to gocql?

```go
// ParseConsistencyWrapper wraps gocql.ParseConsistency to provide an err return instead of a panic
func ParseConsistencyWrapper(s string) (consistency gocql.Consistency, err error) {
	defer func() {
		if r := recover(); r != nil {
			var ok bool
			err, ok = r.(error)
			if !ok {
				err = fmt.Errorf("ParseConsistencyWrapper: %v", r)
			}
		}
	}()
	consistency = gocql.ParseConsistency(s)
	return consistency, nil
}
```

bx2:
This PR adds the suggested ParseConsistencyWrapper including a simple test for it.

[Fixes #749]

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -205,6 +205,22 @@ func ParseConsistency(s string) Consistency {
 	}
 }
 
+// ParseConsistencyWrapper wraps gocql.ParseConsistency to provide an err
+// return instead of a panic
+func ParseConsistencyWrapper(s string) (consistency Consistency, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			var ok bool
+			err, ok = r.(error)
+			if !ok {
+				err = fmt.Errorf("ParseConsistencyWrapper: %v", r)
+			}
+		}
+	}()
+	consistency = ParseConsistency(s)
+	return consistency, nil
+}
+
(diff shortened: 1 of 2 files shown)


frame.go
```diff
@@ -205,6 +205,22 @@ func ParseConsistency(s string) Consistency {
 	}
 }
 
+// ParseConsistencyWrapper wraps gocql.ParseConsistency to provide an err
+// return instead of a panic
+func ParseConsistencyWrapper(s string) (consistency Consistency, err error) {
+	defer func() {
```
Zariel: To be honest, it would be better to do this the other way around and have ParseConsistency panic which calls this.

bx2: Yea, was thinking the same, but orders are orders ;) Based on the comment, I thought that this is what you guys want.

Zariel:
Thanks for the contribution!

nlacey:
Thanks guys

PR #775, opened 2016-08-23
Fix leaking goroutine in events_test.go

malandrew:
This just eliminates the leaking goroutine in events_test

### DIFF
--- a/events_test.go
+++ b/events_test.go
@@ -16,6 +16,7 @@ func TestEventDebounce(t *testing.T) {
 		defer wg.Done()
 		eventsSeen += len(events)
 	})
+	defer debouncer.stop()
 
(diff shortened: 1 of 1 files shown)


events_test.go
```diff
@@ -29,4 +29,5 @@ func TestEventDebounce(t *testing.T) {
 	if eventCount != eventsSeen {
 		t.Fatalf("expected to see %d events but got %d", eventCount, eventsSeen)
 	}
+	debouncer.stop()
```
Zariel: Move this to defer so it stops in the fatal case too

malandrew:
@zariel done and done!

Zariel:
Thanks very much!

PR #780, opened 2016-09-10
Allow nil lists.

nicerobot:
Similar to #750 

- [ ] Update AUTHORS.
- [ ] Add tests.


Zariel:
Thanks, can you add a test to the marshal suite, and if not already could you add yourself to the AUTHORS file?

Zariel:
Thanks for the contribution!

PR #797, opened 2016-09-26
Add support for tinyint

### LINKED ISSUE #796: Cassandra TinyInt Does Not Map to Go int8
When I try to use .SliceMap() with a query result that passes back to tiny ints, I get the panic: reflect: New(nil) error.
I'm using Cassandra 3.7, Go 1.7, and the latest pull of gocql using Debian Jessie.
```sql
cqlsh> create keyspace github with replication = {'class': 'SimpleStrategy', 'replication_factor' : 1};
cqlsh> use github;
cqlsh:github> create table test_tiny_int( a tinyint, PRIMARY KEY (a));
cqlsh:github> insert into test_tiny_int (a) values (1);
cqlsh:github> select a from test_tiny_int;

 a
---
 1

(1 rows)
```
```go
package main

import (
    "fmt"
    "github.com/gocql/gocql"
    "log"
)

func main() {
    // connect to the cluster
    cluster := gocql.NewCluster("localhost")
    cluster.ProtoVersion = 4
    cluster.Keyspace = "github"
    cluster.Consistency = gocql.Quorum
    session, _ := cluster.CreateSession()
    defer session.Close()

    iter := session.Query(`SELECT a from test_tiny_int`).Iter()
    // the below code will parse a correctly
    //var a int8
    //for iter.Scan(&a){
    //  fmt.Println(a)
    //}

    //this will throw an error
    m, err := iter.SliceMap()

    if err != nil {
      log.Fatal(err)
    }
    fmt.Println(m[0])

        log.Fatal(err)
    }

}
```

nathantchan:
Fixes #796

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -39,6 +39,8 @@ func goType(t TypeInfo) reflect.Type {
 		return reflect.TypeOf(*new(int))
 	case TypeSmallInt:
 		return reflect.TypeOf(*new(int16))
+	case TypeTinyInt:
+		return reflect.TypeOf(*new(int8))
 	case TypeDecimal:
 		return reflect.TypeOf(*new(*inf.Dec))
 	case TypeUUID, TypeTimeUUID:


penberg:
Looks good to me :+1:

Zariel:
Thanks very much!

Issue #789, opened 2016-09-18
TestUnpreparedBatch is flaky


Zariel:
Seems to be on 2.0 only

PR #798, opened 2016-09-29
adding AddressTranslator interface and impl for use in ec2

gus:
This change introduces the AddressTranslator interface, which is
intended for use during discovery and node-event processing to translate
peer addresses of nodes. The primary use -- which is driving the
change -- is to translate public IPs to private IPs (when possible)
in ec2. The ec2 translator will be implemented elsewhere and is not
provided here.

This solution is common among other CQL driver implementations. The
specific implementation here also follows the convention set by
HostFilter.

We try our best to have test coverage for the bits that changed.

Signed-off-by: Justin "Gus" Knowlden gus@gusg.us

### DIFF
--- a/address_translators.go
+++ b/address_translators.go
@@ -1,0 +1,26 @@
+package gocql
+
+import "net"
+
+// AddressTranslator provides a way to translate node addresses (and ports) that are
+// discovered or received as a node event. This can be useful in an ec2 environment,
+// for instance, to translate public IPs to private IPs.
+type AddressTranslator interface {
+	// Translate will translate the provided address and/or port to another
+	// address and/or port. If no translation is possible, Translate will return the
+	// address and port provided to it.
+	Translate(addr net.IP, port int) (net.IP, int)
+}
+
+type AddressTranslatorFunc func(addr net.IP, port int) (net.IP, int)
+
+func (fn AddressTranslatorFunc) Translate(addr net.IP, port int) (net.IP, int) {
+	return fn(addr, port)
+}
+
+// IdentityTranslator will do nothing but return what it was provided. It is essentially a no-op.
+func IdentityTranslator() AddressTranslator {
+	return AddressTranslatorFunc(func(addr net.IP, port int) (net.IP, int) {
+		return addr, port
+	})
+}
--- a/cluster.go
+++ b/cluster.go
@@ -6,6 +6,8 @@ package gocql
 
 import (
 	"errors"
+	"log"
+	"net"
 	"time"
 )
 
@@ -74,6 +76,10 @@ type ClusterConfig struct {
 	// the filter will be ignored. If set will take precedence over any options set
 	// via Discovery
 	HostFilter HostFilter
+
+	// AddressTranslator will translate addresses found on peer discovery and/or
+	// node change events.
+	AddressTranslator AddressTranslator
 
 	// If IgnorePeerAddr is true and the address in system.peers does not match
 	// the supplied host by either initial hosts or discovered via events then the
@@ -144,6 +150,21 @@ func NewCluster(hosts ...string) *ClusterConfig {
 // session object that can be used to interact with the database.
 func (cfg *ClusterConfig) CreateSession() (*Session, error) {
 	return NewSession(*cfg)
+}
+
+// translateAddressPort is a helper method that will use the given AddressTranslator
+// if defined, to translate the given address and port into a possibly new address
+// and port, If no AddressTranslator or if an error occurs, the given address and
+// port will be returned.
+func (cfg *ClusterConfig) translateAddressPort(addr net.IP, port int) (net.IP, int) {
+	if cfg.AddressTranslator == nil || len(addr) == 0 {
+		return addr, port
+	}
+	newAddr, newPort := cfg.AddressTranslator.Translate(addr, port)
+	if gocqlDebug {
+		log.Printf("gocql: translating address '%v:%d' to '%v:%d'", addr, port, newAddr, newPort)
+	}
+	return newAddr, newPort
 }
 
 var (
--- a/control.go
+++ b/control.go
@@ -318,8 +318,8 @@ func (c *controlConn) reconnect(refreshring bool) {
 		}
 	}
 
-	// TODO: should have our own roundrobbin for hosts so that we can try each
-	// in succession and guantee that we get a different host each time.
+	// TODO: should have our own round-robin for hosts so that we can try each
+	// in succession and guarantee that we get a different host each time.
 	if newConn == nil {
 		host := c.session.ring.rrHost()
 		if host == nil {
--- a/events.go
+++ b/events.go
@@ -205,7 +205,6 @@ func (s *Session) handleNewNode(ip net.IP, port int, waitForBinary bool) {
 	s.pool.addHost(hostInfo)
 	s.policy.AddHost(hostInfo)
 	hostInfo.setState(NodeUp)
-
 	if s.control != nil && !s.cfg.IgnorePeerAddr {
 		s.hostSource.refreshRing()
 	}
--- a/session_connect_test.go
+++ b/session_connect_test.go
@@ -1,0 +1,131 @@
+package gocql
+
+import (
+	"golang.org/x/net/context"
+	"net"
+	"strconv"
+	"sync"
+	"testing"
+	"time"
+)
+
+type OneConnTestServer struct {
+	Err  error
+	Addr net.IP
+	Port int
+
+	listener   net.Listener
+	acceptChan chan struct{}
+	mu         sync.Mutex
+	closed     bool
+}
+
+func NewOneConnTestServer() (*OneConnTestServer, error) {
+	lstn, err := net.Listen("tcp4", "localhost:0")
+	if err != nil {
+		return nil, err
+	}
+	addr, port := parseAddressPort(lstn.Addr().String())
+	return &OneConnTestServer{
+		listener:   lstn,
+		acceptChan: make(chan struct{}),
+		Addr:       addr,
+		Port:       port,
+	}, nil
+}
+
+func (c *OneConnTestServer) Accepted() chan struct{} {
+	return c.acceptChan
+}
+
+func (c *OneConnTestServer) Close() {
+	c.lockedClose()
+}
+
+func (c *OneConnTestServer) Serve() {
+	conn, err := c.listener.Accept()
+	c.Err = err
+	if conn != nil {
+		conn.Close()
+	}
+	c.lockedClose()
+}
+
+func (c *OneConnTestServer) lockedClose() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if !c.closed {
+		close(c.acceptChan)
+		c.listener.Close()
+		c.closed = true
+	}
+}
+
+func parseAddressPort(hostPort string) (net.IP, int) {
+	host, portStr, err := net.SplitHostPort(hostPort)
+	if err != nil {
+		return net.ParseIP(""), 0
+	}
+	port, _ := strconv.Atoi(portStr)
+	return net.ParseIP(host), port
+}
+
+func testConnErrorHandler(t *testing.T) ConnErrorHandler {
+	return connErrorHandlerFn(func(conn *Conn, err error, closed bool) {
+		t.Errorf("in connection handler: %v", err)
+	})
+}
+
+func assertConnectionEventually(t *testing.T, wait time.Duration, srvr *OneConnTestServer) {
+	ctx, cancel := context.WithTimeout(context.Background(), wait)
+	defer cancel()
+
+	select {
+	case <-ctx.Done():
+		if ctx.Err() != nil {
+			t.Errorf("waiting for connection: %v", ctx.Err())
+		}
+	case <-srvr.Accepted():
+		if srvr.Err != nil {
+			t.Errorf("accepting connection: %v", srvr.Err)
+		}
+	}
+}
+
+func TestSession_connect_WithNoTranslator(t *testing.T) {
+	srvr, err := NewOneConnTestServer()
+	assertNil(t, "error when creating tcp server", err)
+	defer srvr.Close()
+
+	session := createTestSession()
+	defer session.Close()
+
+	go srvr.Serve()
+
+	Connect(&HostInfo{
+		peer: srvr.Addr,
+		port: srvr.Port,
+	}, session.connCfg, testConnErrorHandler(t), session)
+
+	assertConnectionEventually(t, 500*time.Millisecond, srvr)
+}
+
+func TestSession_connect_WithTranslator(t *testing.T) {
+	srvr, err := NewOneConnTestServer()
+	assertNil(t, "error when creating tcp server", err)
+	defer srvr.Close()
+
+	session := createTestSession()
+	defer session.Close()
+	session.cfg.AddressTranslator = staticAddressTranslator(srvr.Addr, srvr.Port)
+
+	go srvr.Serve()
+
+	// the provided address will be translated
+	Connect(&HostInfo{
+		peer: net.ParseIP("10.10.10.10"),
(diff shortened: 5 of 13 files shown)


gus:
It does say there is a test failure, that's the [same failure](https://travis-ci.org/gocql/gocql/jobs/162855812) as gocql/master was getting in previous builds.

Zariel:
Yeah thats unrelated and caused by #789 which I haven't had time to dig into. Ill try and take a look at this over the weekend

address_translators.go
```diff
@@ -0,0 +1,53 @@
...
+		return addr, port
+	})
+}
+
+// EC2MultiRegionAddressTranslator will do a reverse DNS lookup on the address provided,
+// then do a DNS lookup on the first hostname found in order to potentially get the
+// private IP of the given address. If no results are found in either lookup,
+// the given address is returned. Port is always returned unchanged. Build with the
+// gocql_debug tag to see the address translation at work.
+func EC2MultiRegionAddressTranslator() AddressTranslator {
```
Zariel: While this is useful, its pretty easy to implement and I don't see a huge value in including it in the library as it does not directly relate to Cassandra and in most cases will be implementation specific.

cluster.go
```diff
@@ -125,6 +129,15 @@ func (cfg *ClusterConfig) CreateSession() (*Session, error) {
 	return NewSession(*cfg)
 }
 
+// translateAddress is a helper method wthat ill use the given AddressTranslator
```
Zariel: that will

events.go
```diff
@@ -133,6 +133,9 @@ func (s *Session) handleNodeEvent(frames []frame) {
 		// TODO: can we be sure the order of events in the buffer is correct?
 		switch f := frame.(type) {
 		case *topologyChangeEventFrame:
+			host, port := s.cfg.translateAddress(f.host.String(), f.port)
```
Zariel: One issue I can see with this is that if the translator is not deterministic then we are going to end up with orphaned connections or nodes in the ring, which will be bad. A few places index on a host's address, it might be worthwhile adding another field to `HostInfo` which stores the translated address and still keeps the address that Cassandra thinks a node is?

events.go
```diff
@@ -141,6 +144,9 @@ func (s *Session) handleNodeEvent(frames []frame) {
 			event.change = f.change
 
 		case *statusChangeEventFrame:
+			host, port := s.cfg.translateAddress(f.host.String(), f.port)
+			f.host = net.ParseIP(host)
+			f.port = port
 			event, ok := events[f.host.String()]
```
Zariel: we can just use `events[host]` here now

events.go
```diff
@@ -133,6 +133,9 @@ func (s *Session) handleNodeEvent(frames []frame) {
 		// TODO: can we be sure the order of events in the buffer is correct?
 		switch f := frame.(type) {
 		case *topologyChangeEventFrame:
+			host, port := s.cfg.translateAddress(f.host.String(), f.port)
+			f.host = net.ParseIP(host)
```
Zariel: events are defined as having a resolved IP address, if a user returns something which is not an IP address we will run into issues later on.

Zariel: I think it's probably best to move the translation to the layer below event deduplication that way we dont need to do this in every event.

events_test.go
```diff
@@ -31,3 +32,55 @@ func TestEventDebounce(t *testing.T) {
...
+		},
+	})
+	assertEqual(t, "new ring host length", 1, len(session.ring.hosts))
+	assertEqual(t, "translated host and port", 5432, session.ring.hosts["127.10.10.220"].port)
+}
+
+func TestHandleNodeEvent_StatusChange_TranslatesAddress(t *testing.T) {
+	defer (func (d time.Duration) { nodeUpDelayDuration = d })(nodeUpDelayDuration)
+	nodeUpDelayDuration = 0
+	session := createTestSession()
```
Zariel: defer close the session

events_test.go
```diff
@@ -31,3 +32,55 @@ func TestEventDebounce(t *testing.T) {
 		t.Fatalf("expected to see %d events but got %d", eventCount, eventsSeen)
 	}
 }
+
+func TestHandleNodeEvent_TopicChange_TranslatesAddress(t *testing.T) {
+	defer (func (d time.Duration) { nodeUpDelayDuration = d })(nodeUpDelayDuration)
+	nodeUpDelayDuration = 0
+	session := createTestSession()
```
Zariel: defer close the session

address_translators.go
gus: Hey @Zariel, I'm not sure what you mean by doing this outside of Cassandra. It directly relates to configuring a ring for multi-region support in ec2. You'll find similar support in the [Java driver](https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/policies/EC2MultiRegionAddressTranslator.java) and the [NodeJS driver](https://github.com/datastax/nodejs-driver/blob/master/lib/policies/address-resolution.js#L76), for instance. If you need it, I can describe in more detail our ring setup spread across 6 regions, but there are plenty of other people who have described this issue.

events.go
gus: True enough

events.go
gus: That's probably fair. In general we were trying to avoid altering too many of the existing structures and behaviors. Especially for testing reasons. But, keeping both addresses (broadcast and translated) makes sense.

address_translators.go
```diff
@@ -0,0 +1,53 @@
...
+	"net"
+)
+
+// AddressTranslator provides a way translate node addresses (and ports) when one is
+// discovered or received as a node event. This is especially useful in ec2 (when
+// using the EC2MultiRegionAddressTranslator) to translate public IPs to private IPs
+// when possible.
+type AddressTranslator interface {
+	// Translate will translate the provided address and/or port when a host is discovered.
+	Translate(addr string, port int) (string, int)
```
Zariel: This should accept a string addr and return a `net.IP`

address_translators.go
Zariel: I fully understand, my comment comes from the aspect of maintaining parts of the driver which do not directly relate to Cassandra, while the Datastax drivers provide this in the main package I don't feel the need to replicate every part of functionality they provide and maintain.

address_translators.go
Zariel: The addr here should always be a `net.IP` actually, so I think the function should look something like

```Translate(ip net.IP, port int) (ip net.IP, port int, err error)`

Returning an error will be useful as it will prevent the driver having to check down the chain if an address is valid and such, also allowing the translate the block hosts (though arguably that should be done by the host filter).

Comment (gus) [inline address_translators.go]: I'm trying to ensure that any issue with translation ends up returning what was provided. This also seems in keeping with the other translator implementations.

File: events.go (hunk 3 above)
Comment (gus) [inline events.go]: So ... in looking into your suggestion, I'm thinking it might actually just be easiest to avoid processing and storing the translated address anywhere. Instead, it might be easiest to simply be best to translate in `Session.connect` before calling `Connect`. I'll add some debugging in that method, too.

**Comment by gus** [comment]
@Zariel Made changes that hopefully address your concerns. Apparently I need to allow some more time for an event to occur in the new unit tests, since travis is failing so quickly; integrations pass locally though.

File: cluster.go [hunk 9]
```diff
@@ -125,6 +132,33 @@ func (cfg *ClusterConfig) CreateSession() (*Session, error) {
...
+	return net.JoinHostPort(newAddr.String(), strconv.Itoa(newPort))
+}
+
+func parseIPPort(ipPort string) (net.IP, int, error) {
+	addr, portStr, err := net.SplitHostPort(ipPort)
+	if err != nil {
+		return nil, 0, err
+	}
+	port, _ := strconv.Atoi(portStr)
+	return net.ParseIP(addr), port, nil
```
BlueDragonX: In the event that `addr` is a hostname, [ParseIP](https://golang.org/pkg/net/#ParseIP) will return `nil`. This in turn causes the translator to fail as it recieves `nil` for the IP.

Currently `NewCluster` supports accepting a list of host names. This change ungracefully breaks that functionality.

It may be that `NewCluster` is only intended to accept a list of IP addresses. If this is the case then a change should be included which causes a proper error to be surfaced in this case. Otherwise this function should be updated to to a name lookup in order to resolve an IP address.

gus: Dang it ... yes. I got tricked by "everything should be an IP". Thanks.

BlueDragonX: I should point out that there is a similar problem with the port number. Use `net.LookupPort` instead of `Atoi`.

Lookup errors should either be surfaced or handled by returning the original addr and logging.

Zariel: I agree, do we really need to translate the initial hosts? As they will be added through handleNoUp, though they may have a DNS name not an IP address, hmm.

BlueDragonX: Yeah, feeding a DNS name in via `NewCluster` will cause the translator to break. My inclination is to do the name resolution here in `translateHostPort`. Since `translateHostPort` is called before dialing we do the DNS lookup at connect time rather than at application start. I feel this is safer and accounts for changes in the DNS record.

Though an argument could be made that the user expects the name to be resolved at the time `NewCluster` is called. I believe doing the name resolution at the time the connection is made to be safer. If the user desires a different functionality he may resolve the names himself prior to calling `NewCluster`.

The chosen approach should be documented for the caller in any case.

address_translators.go
```diff
@@ -0,0 +1,24 @@
+package gocql
+
+// AddressTranslator provides a way to translate node addresses (and ports) that are
+// discovered or received as a node event. This is especially useful in ec2 (when
+// using the EC2MultiRegionAddressTranslator) to translate public IPs to private IPs
+// when possible.
+type AddressTranslator interface {
```
Zariel: Im still considering if this function should return an error, though I guess we should leave that for the HostFilter?

control.go
```diff
@@ -130,6 +130,7 @@ func (c *controlConn) shuffleDial(endpoints []string) (conn *Conn, err error) {
 		}
 
 		hostInfo, _ := c.session.ring.addHostIfMissing(host)
+		// TODO (gus) use translated address here?
```
Zariel: I dont think we need to, as `session.connect` will translate the address for us below

cluster.go
```diff
@@ -125,6 +132,33 @@ func (cfg *ClusterConfig) CreateSession() (*Session, error) {
 	return NewSession(*cfg)
 }
 
+// translateHostPort is a helper method that will use the given AddressTranslator
+// if defined, to translate the given host:port (addr) into a new host:port string.
+// If no AddressTranslator or if an error occurs, the given host:port will be returned.
+func (cfg *ClusterConfig) translateHostPort(hostPort string) (string) {
```
Zariel: gofmt

cluster.go
```diff
@@ -125,6 +132,33 @@ func (cfg *ClusterConfig) CreateSession() (*Session, error) {
...
+	if cfg.AddressTranslator == nil {
+		return hostPort
+	}
+	host, port, err := parseHostPort(hostPort)
+	if err != nil {
+		return hostPort
+	}
+	newHost, newPort := cfg.AddressTranslator.Translate(host, port)
```
Zariel: `if newHost == "" || newPort == 0` return hostPort`?

session_connect_test.go
```diff
@@ -0,0 +1,82 @@
...
+	assertNil(t, "error when creating tcp server", err)
+	testHost, testPort, err := parseHostPort(srvr.Addr().String())
+	log.Printf("using host=%s and port=%d for test tcp server (err=%v)", testHost, testPort, err)
+	return srvr, testHost, testPort
+}
+
+func tcpConnWaitAndClose(t *testing.T, srvr net.Listener, tchan chan<- error) {
+	conn, err := srvr.Accept()
+	if conn != nil {
+		log.Printf("connection received from addr=%+v", conn.RemoteAddr())
```
Zariel: `t.Logf`

session_connect_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"net"
+	"testing"
+	"time"
+)
+
+func setupTestTcpServer(t *testing.T) (net.Listener, string, int) {
+	srvr, err := net.Listen("tcp4", "0.0.0.0:0")
+	assertNil(t, "error when creating tcp server", err)
+	testHost, testPort, err := parseHostPort(srvr.Addr().String())
+	log.Printf("using host=%s and port=%d for test tcp server (err=%v)", testHost, testPort, err)
```
Zariel: `t.Logf`

session_connect_test.go
```diff
@@ -0,0 +1,82 @@
...
+	tchan := make(chan error)
+	defer close(tchan)
+	go tcpConnWaitAndClose(t, srvr, tchan)
+
+	session.connect(srvr.Addr().String(), testConnErrorHandler(t), &HostInfo{
+		peer: testHost,
+		port: testPort,
+	})
+
+	assertConnectionEventually(t, 500*time.Millisecond, tchan)
```
Zariel: im hesitant to add tests which rely on things happening in a time period, as they tend to flake in travis (see the node event tests)

session_connect_test.go
```diff
@@ -0,0 +1,82 @@
...
+	defer srvr.Close()
+
+	session := createTestSession()
+	defer session.Close()
+
+	tchan := make(chan error)
+	defer close(tchan)
+	go tcpConnWaitAndClose(t, srvr, tchan)
+
+	session.connect(srvr.Addr().String(), testConnErrorHandler(t), &HostInfo{
```
Zariel: Why not use the conn test test server? It should error out if the session cant connect

session_connect_test.go
```diff
@@ -0,0 +1,82 @@
...
+func TestSession_connect_WithTranslator(t *testing.T) {
+	srvr, testHost, testPort := setupTestTcpServer(t)
+	defer srvr.Close()
+
+	session := createTestSession()
+	defer session.Close()
+	session.cfg.AddressTranslator = staticAddressTranslator(testHost, testPort)
+
+	tchan := make(chan error)
+	defer close(tchan)
```
Zariel: if we hit the timeout then return from this function, `tchan` will be closed and potentially trigger a send on closed channel.

control.go
gus: No, we don't. That was for me as I was looking around and forgot to clean the comment out. Thanks.

cluster.go
gus: I get the defensiveness, but it feels unnecessary to me. I've update the documentation for `Translate` to state that the provided host:port should be returned if no translation is possible.

session_connect_test.go
gus: I always forget about `t.Logf`

session_connect_test.go
gus: Me, too. They are the bane of my coding existence. I'm happy to increase the value here. I'd be happier if there was an interface or some DI available such that I could provide a mock dialer, but I also don't personally want to do that in this PR. Finally, I can remove the tests, but I feel bad about that.

gus:
Well, it's been a while since doing anything with this PR, and some code has changed (`net.IP`s abount). So, I had to change some things and address previous concerns.

Zariel:
Thanks for updating this, due to the bug identified in #807 and fixed in #809 the driver change all its internal host addresses to `net.IP` which reflect how Cassandra stores them and notifies the driver via events. Ill take a look at this again shortly when I have time.

Zariel:
Thank you for the contribution! I like how this became simpler with the changes to use `net.IP` over string addrs.

zot24:
Is there any documentation or an example of how to implement the `AddressTranslator` somewhere? Thanks

zot24:
Just in case someone else it's interested I just found this package that use it and it's a good example https://github.com/compose/composeaddresstranslator

gus:
Oh wow, didn't realize anyone else had done anything with it. We split out
the EC2 address translator code here: https://github.com/signal/gocql-ec2

Issue #771, opened 2016-08-16
CCM Event tests are flaky in Travis


Zariel:
We should add some hooks to the connection pool or session to detect events to test instead of using timeouts.

nathany:
Does that refer to these (flaky) tests, or are these something else?

```
--- FAIL: TestUnpreparedBatch (1.00s)
	cassandra_test.go:429: count: expected 100, got 0
```

```
--- FAIL: TestStream0 (0.00s)
	conn_test.go:684: read tcp 127.0.0.1:41287->127.0.0.1:33392: read: connection reset by peer
```

build: https://travis-ci.org/gocql/gocql/builds/159111287

Zariel:
No something else, those 2 are other flakes :(

nathany:
Ok. I didn't see #789 at the time.

In this event test I don't see any timeouts -- I'm not sure why it fails sometimes.

```
--- FAIL: TestEventDiscovery (2.41s)
	events_ccm_test.go:25: status=map[node1:{State:UP Addr:127.0.0.1 Name:node1} node3:{State:UP Addr:127.0.0.3 Name:node3} node2:{State:UP Addr:127.0.0.2 Name:node2}]
	events_ccm_test.go:29: poolhosts=map[127.0.0.1:[filling=false closed=false conns=2 size=2 host=[hostinfo peer="127.0.0.1" port=9042 data_centre="datacenter1" rack="" host_id="3d8b84be-79ca-4cd3-9370-5af3b6b37f60" version="v2.0.17" state=UP num_tokens=256]]]
	events_ccm_test.go:33: did not discover "127.0.0.3"
	events_ccm_test.go:33: did not discover "127.0.0.2"
```

from: https://travis-ci.org/gocql/gocql/jobs/168777130

nathany:
I don't understand why this is failing.

```
--- FAIL: TestEventNodeUp (57.38s)
	events_ccm_test.go:174: node not added after node added event
```

From what I see, cmd.Run is synchronously running `ccm start --wait-for-binary-proto` which says it will:

> Wait for the binary protocol to start

Yet there is a Sleep that apparently isn't enough? 

```go
// cassandra < 2.2 needs 10 seconds to start up the binary service
time.Sleep(15 * time.Second)
```

Perhaps a temporary fix would be a retry loop that polls every few seconds:

```go
start := time.Now()

for time.Since(start) < 60 * time.Second {
    // ...
    time.Sleep(5 * time.Second)
}
```

[...]

PR #803, opened 2016-10-18
[WIP] use a Logger interface for logging

### LINKED ISSUE #766: library logging output
I noticed that the gocql library logs in several different spots. For our production logging we prefer to log in json so random text output is at odds with this. Have there been any thoughts of removing the internal logging and bubbling up errors or allowing a logger to be injected?

nathany:
Adds Log to ClusterConfig to use for all logging, allowing a custom logger like [Logrus](https://github.com/Sirupsen/logrus) to be used. fixes #766

This is partially implemented, as there are several uses of the standard logger that may be more difficult to get the Logger into (marked with TODOs). 
- [ ] resetTokenRing() in polices.go 
- [ ] parse() in metadata.go
- [ ] parseFrame() in frame.go
- [ ] debounce(frame frame) in events.go

I'm interested in feedback on how each of those situations should be resolved:
- Inject the logger somehow
- Bubble errors up, so the logging can be moved elsewhere
- Remove the logging
#### Finishing the implementation

If someone wants to take on the rest of the implementation, [hub](https://github.com/github/hub) makes it easy to pull this code into a new local branch:

``` console
git checkout -b logger
hub am -3 https://github.com/gocql/gocql/pull/803
```
#### How should this be manually tested?

Something like this:

``` go
log := logrus.New()
log.Formatter = &logrus.JSONFormatter{}
...
cfg.Log = log
```

Output like this:

```
{"level":"info","msg":"gocql: unable to dial control conn localhost:9042: dial tcp [::1]:9042: getsockopt: connection refused\n","time":"2016-10-18T17:27:53-06:00"}
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -11,7 +11,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"log"
 	"net"
 	"strconv"
 	"strings"
@@ -491,7 +490,7 @@ func (c *Conn) recv() error {
 	call, ok := c.calls[head.stream]
 	c.mu.RUnlock()
 	if call == nil || call.framer == nil || !ok {
-		log.Printf("gocql: received response for stream which has no handler: header=%v\n", head)
+		c.session.cfg.Log.Printf("gocql: received response for stream which has no handler: header=%v\n", head)
 		return c.discardFrame(head)
 	}
 
@@ -871,7 +870,7 @@ func (c *Conn) executeQuery(qry *Query) *Iter {
 		iter := &Iter{framer: framer}
 		if err := c.awaitSchemaAgreement(); err != nil {
 			// TODO: should have this behind a flag
-			log.Println(err)
+			c.session.cfg.Log.Println(err)
 		}
 		// dont return an error from this, might be a good idea to give a warning
 		// though. The impact of this returning an error would be that the cluster
@@ -1082,7 +1081,7 @@ func (c *Conn) awaitSchemaAgreement() (err error) {
 		var schemaVersion string
 		for iter.Scan(&schemaVersion) {
 			if schemaVersion == "" {
-				log.Println("skipping peer entry with empty schema_version")
+				c.session.cfg.Log.Println("skipping peer entry with empty schema_version")
 				continue
 			}
 
--- a/cluster.go
+++ b/cluster.go
@@ -6,6 +6,8 @@ package gocql
 
 import (
 	"errors"
+	"log"
+	"os"
 	"time"
 )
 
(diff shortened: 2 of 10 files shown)


conn.go
```diff
@@ -481,7 +480,7 @@ func (c *Conn) recv() error {
 	call, ok := c.calls[head.stream]
 	c.mu.RUnlock()
 	if call == nil || call.framer == nil || !ok {
-		log.Printf("gocql: received response for stream which has no handler: header=%v\n", head)
+		c.session.cfg.Log.Printf("gocql: received response for stream which has no handler: header=%v\n", head)
```
nathany: It may be worth forwarding these methods so that `c.Log` or `c.session.Log` work just as well.

nathany:
I think the Travis failures are due to flaky tests (#771), not changes in this branch. 

```
❯ go test -race ./...   
ok  	github.com/gocql/gocql	25.511s
ok  	github.com/gocql/gocql/internal/lru	1.022s
ok  	github.com/gocql/gocql/internal/murmur	1.025s
ok  	github.com/gocql/gocql/internal/streams	1.022s
```

nathany:
Merge conflicts have been resolved. 

I'm still not sure what to do with the remaining logging statements.

Zariel:
is there a reason why we cant pass the logger into them from the session?

nathany:
That would make sense. It's probably just that I haven't figured out the code well enough to see if that's possible for all 4 situations mentioned in the original post. I can take a closer look though.

nathany:
Just looking at the first case, this seems problematic because something like this is a breaking change:

```
func TokenAwareHostPolicy(fallback HostSelectionPolicy, log Logger) HostSelectionPolicy
```

nathany:
`newFramer` could accept a Logger to log warnings, but this question in the comment makes me wonder if that's the appropriate solution `// what to do with warnings?`

`parseType` could also require a Logger for logging an error when a non-hexadecimal name is encountered (it continues without decoding it)

`newEventDebouncer` could require a Logger as well which logs "buffer full" for a debounce

so maybe `TokenAwareHostPolicy` is the only problematic spot because it is an exported API.

Zariel:
Hm, passing into the policies may not be necessary, are the logs necessary? The framer log for warnings is a placeholder, in reality the warnings should be available on the `Iter` returned from the query, I'll make a ticket.

nathany:
I think the log in policies.go `resetTokenRing()` could be removed, or moved up the stack. It reports "Unable to update the token ring due to error", but the error is also returned

`SetPartioner`, `AddHost`, and `RemoveHost` call `resetTokenRing()` but they don't return the error, so it's swallowed up there. In most cases it wouldn't be a breaking change to return the `error` from those functions.

conn.go
Zariel: session.logf/log would be ok

nathany:
It looks like this change is unnecessary in light of the workaround at https://github.com/gocql/gocql/issues/766#issuecomment-257715015.

Zariel:
Thanks for working on this though it was not merged, I think logging can still be improved.

PR #807, opened 2016-10-20
Fix bug preventing hostnames from being added to hostpool correctly

dancannon:
Currently if multiple hostnames are passed to `Connect` only the last hostname is actually used by the host selection policy. This is because the hostname is passed by `net.ParseIP` which given anything other than an IP address returns a nil `net.IP` which results in adding the host `{Peer: "<nil>"}`. Any other hosts added to the pool will have the same peer value meaning that any previous hosts are overridden.

To fix this issue all calls to `net.ParseIP` are replaced by a call to the new function `lookupHostAddress` which will attempt to lookup the IP from the hostname if `net.ParseIP` fails.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -22,7 +22,6 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/gocql/gocql/internal/lru"
-
 	"github.com/gocql/gocql/internal/streams"
 )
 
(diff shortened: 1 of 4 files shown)


dancannon:
The failing test appears to be passing locally, have you got any idea why they might be failing on travis?

Zariel:
I don't quite understand the issue being fixed, could you add a test which demonstrates it? The address handling in the driver has gotten a bit too complex for my liking, some places assume its always an IP address and other accept any address. Which usage of `net.ParseIP` is causing issues? It might make sense to make the driver only use an IP address and port combination except for the initial hosts to connect to.

The test failures are unrelated and caused by #771 which im looking into.

Zariel:
Looking into this I can see issues with how host address are being handled and gone through and changed the driver to use a nodes ip address not a dns name, which I think should cuase the root issue of what you report here? #809

Zariel:
Something else to note, if a user uses a DNS address for the initial connection, we may have different peer address used in the pool and ring and for events. I think its probably best to have a note in the readme and doc for cluster config that the supplied hosts should ideally be the values of the nodes broadcast_address or listen_address.

dancannon:
#809 looks like it should fix the issue, unfortunately using the IP addresses of the Cassandra instances is not really possible for us right now (we are currently using `IgnorePeerAddr` to avoid the issues you mentioned).

Zariel:
Are you using docker so the address returned from the peer lookup is not a routable address from the client? If so I'd like to have a chat about it offline sometime.

dancannon:
Not at the moment no (although this might be something we look into at some point), we are using hostnames to simplify the discovery of the initial hosts.

PR #810, opened 2016-10-25
When unmarshalling timestamps set null values to 0 time to be consist…

clawconduce:
…ent with other types

Warning: I consider this a bug, but some people may be using this as a feature.

Below is a simplified version of some code I'm working with:

    //Print times files deleted
    iter := session.Query(`SELECT deleted FROM files.metadata WHERE owner = ?`, user).Iter()
    var deleted time.Time
    for iter.Scan(&deleted) {
        if !deleted.IsZero() {
            fmt.Println("File was deleted at", deleted)
        }
    }

The `deleted` variable is not re-zero'ed for null rows and so after seeing 1 non-null value, the deleted value remains non-zero.  Other types get reset.  Here it is for string: https://github.com/gocql/gocql/blob/master/marshal.go#L222 which gets set to an empty string.  Ints go through https://github.com/gocql/gocql/blob/master/marshal.go#L489, etc.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -80,3 +80,4 @@ Marcus King <marcusking01@gmail.com>
 Andrew de Andrade <andrew@deandrade.com.br>
 Robert Nix <robert@nicerobot.org>
 Nathan Youngman <git@nathany.com>
+Charles Law <charles.law@gmail.com>; <claw@conduce.com>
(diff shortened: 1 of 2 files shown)


Zariel:
Thank you for the contribution! Fyi that query can be rewritten to avoid this problem,

```go
rows := session.Query(`SELECT deleted FROM files.metadata WHERE owner = ?`, user).Iter().Scanner()
for rows.Next() {
    var delted time.Time
    if err := rows.Scan(&deleted); err != nil {
        return err
    }
    if !deleted.IsZero() {
        fmt.Println("File was deleted at", deleted)
    }
}
```

Issue #820, opened 2016-10-31
Unknown column kind 'static'


nathj07:
Cassandra: v.2.2.7
Go: go1.7.1 linux/amd64
gocql: 
```
  {
                        "checksumSHA1": "nFDfqGsNbhovyV6wW/r/fpLoqYc=",
                        "path": "github.com/gocql/gocql",
                        "revision": "64566e11bbd97365437f0eb1324131cf2b98a52e",
                        "revisionTime": "2016-10-30T12:41:10Z"
                },
```
So I think I'm all up to date. The trouble is in my schema I have columns like `exclude_methods boolean STATIC,` (I can't post the full schema for business reasons)

I am now getting errors:
```
failed to get table info: Error querying column schema: unknown column kind: "static"
```
I have oly just updated the `gocql` package from commit `b7b8a0e04b0cb0ca0b379421c58ec6fab9939b85`. This was not a problem prior to the upgrade. (upgrade done to get Scan to write time.Time{} when the column is empty)

Digging through the package I think the issue is caused by the change from 
```
// schema metadata for a column
type ColumnMetadata struct {
	Keyspace        string
	Table           string
	Name            string
	ComponentIndex  int
	Kind            string
	Validator       string
	Type            TypeInfo
	ClusteringOrder string
	Order           ColumnOrder
	Index           ColumnIndexMetadata
}
```
Where `Kind` is a string to:
```
/ schema metadata for a column
type ColumnMetadata struct {
	Keyspace        string
	Table           string
	Name            string
	ComponentIndex  int
	Kind            ColumnKind
	Validator       string
	Type            TypeInfo
	ClusteringOrder string
	Order           ColumnOrder
	Index           ColumnIndexMetadata
}
```
Where Kind is now a custom type - and iota as it turns out. The code to support this does not seem to expect `STATIC`.

I'm raising this issue and will take a look at patching it. In the mean time any further comment or input on this is greatly appreciated.

PR #821, opened 2016-10-31
Add support for STATIC

nathj07:
In response to issue #820 this patch adds ColumnStatic to the ColumKind enum

I have not added any new test for the extra iota value. I'm happy to if you can point me in the right direction for those. I was having some trouble running tests locally - 
```
package ghe.iparadigms.com/ndavies/gocql
	imports github.com/gocql/gocql/internal/lru: use of internal package not allowed
package ghe.iparadigms.com/ndavies/gocql
	imports github.com/gocql/gocql/internal/murmur: use of internal package not allowed
package ghe.iparadigms.com/ndavies/gocql
	imports github.com/gocql/gocql/internal/streams: use of internal package not allowed
```

The patch itself is fairly straightforward and I think I've got it in the right spot. Thanks

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -81,3 +81,4 @@ Andrew de Andrade <andrew@deandrade.com.br>
 Robert Nix <robert@nicerobot.org>
 Nathan Youngman <git@nathany.com>
 Charles Law <charles.law@gmail.com>; <claw@conduce.com>
+Nathan Davies <nathanjamesdavies@gmail.com>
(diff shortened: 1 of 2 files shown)


Zariel:
Dont you also need to update https://github.com/gocql/gocql/blob/64566e11bbd97365437f0eb1324131cf2b98a52e/metadata.go#L109?

The reason for the build failure is due to building under your fork not `github.com/gocql/gocql`.

nathj07:
Thanks for the response. I'll take care of that tomorrow when I'm back online.

nathj07:
This is ready for follow up review, thanks

Zariel:

Issue #833, opened 2016-11-16
Use a custom logger instead of the default from the log package


vrischmann:
Hi,

currently gocql logs using the default logger via `log.Printf` or `log.Println`, it makes it impossible for me to separate gocql's logs into its own file (I generally a log file for my app, a log file for sarama (Kafka client), a log file for service discovery, etc. I'd like to be able to put the gocql logs into a gocql file).

Sarama (the Kafka client I use) defines [its own interface](https://godoc.org/github.com/Shopify/sarama#StdLogger) that Logger from the `log` package satisfies, this then allows me to do this:

    logger := log.New(fileOutput, "sarama", log.LstdFlags)
    sarama.Logger = logger

It would be nice if gocql had something like this. A quick grep of the source code tells me there's not a lot of logging calls to change if one were to implement this.

What do you think ? I'm going to implement this for my needs anyway so if you're interested I'll make a PR.

vrischmann:
I'm closing this since I implemented it in #834

PR #834, opened 2016-11-16
Custom logger

vrischmann:
This add a StdLogger interface and a Logger global variable to allow users to change the logger they want to use.

This is the follow up to #833

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -1,9 +1,6 @@
 package gocql
 
-import (
-	"io/ioutil"
-	"log"
-)
+import "log"
 
 type StdLogger interface {
 	Print(v ...interface{})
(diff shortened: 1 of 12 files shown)


logger.go
```diff
@@ -0,0 +1,14 @@
...
+	"log"
+)
+
+type StdLogger interface {
+	Print(v ...interface{})
+	Printf(format string, v ...interface{})
+	Println(v ...interface{})
+}
+
+var Logger StdLogger = log.New(ioutil.Discard, "", log.LstdFlags)
```
Zariel: I feel this should be the default go logger and users can silence it in case of need to

vrischmann: I'll have to add a simple struct that forwards calls to `log.Println` and so on because the default from `log` is not accessible unfortunately

Zariel: Or just using `os.Stderr` are the log output

vrischmann: Oh you meant the default output. I may be wrong but os.Stderr and os.Stdout are not safe for concurrent use right ?

I've never tried it but I would think if someone uses `log.Println` in a goroutine and in another goroutine a new, different logger writes to os.Stderr there's a chance the output will be mangled.

vrischmann:
I changed the logger to forward calls to the standard log package. Like I said I don't think we should make a new logger that wraps `os.Stderr`.

Zariel:
Great, if not already could you add yourself to the authors file and ill merge.

vrischmann:
Okay, I just rebased on master and added myself to the authors file

Issue #730, opened 2016-05-10
lazy connecting (feature request)


clawconduce:
Right now you can create a `Session` to a Cassandra server, and if server goes down, gocql will return an error when a request using that `Session` executes.  If the server comes back up, the next request will reconnect and it act like the server was never down.

It would be nice to be able to jump directly to the state where all your `Session` configuration is saved ignoring the server state.  So even if the server is down, a user could create a `Session` and lazily connect to the server when executing a request.  This would allow users to startup programs even if their Cassandra server is down.

PR #838, opened 2016-11-23
Some pre-work for lazy loading #730

clawconduce:
While implementing lazy loading (#730) I made these changes that were independent of the lazy connection changes.  I pulled this out into this PR since they were small and probably good changes regardless of #730.

### DIFF
--- a/session.go
+++ b/session.go
@@ -127,15 +127,15 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 	}
 
 	if err := s.init(); err != nil {
-		// TODO(zariel): dont wrap this error in fmt.Errorf, return a typed error
 		s.Close()
-		return nil, fmt.Errorf("gocql: unable to create session: %v", err)
-	}
-
-	if s.pool.Size() == 0 {
-		// TODO(zariel): move this to init
-		s.Close()
-		return nil, ErrNoConnectionsStarted
+		if err == ErrNoConnectionsStarted {
+			//This error used to be generated inside NewSession & returned directly
+			//Forward it on up to be backwards compatible
+			return nil, ErrNoConnectionsStarted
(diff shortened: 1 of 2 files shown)


session.go
```diff
@@ -126,16 +126,24 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 		policy: cfg.PoolConfig.HostSelectionPolicy,
 	}
 
-	if err := s.init(); err != nil {
-		// TODO(zariel): dont wrap this error in fmt.Errorf, return a typed error
-		s.Close()
+	//Check the TLS Config before trying to connect to anything external
+	connCfg, err := connConfig(&s.cfg)
+	if err != nil {
+		//TODO: Return a typed error
```
clawconduce: does this need a `s.Close()` call?

Zariel:
Sorry for the delay in getting to this, can you go over a bit of what you are attempting to achieve because its not very obvious to me the benefit of this.

clawconduce:
Sure.  So just to get it out of the way, there’s 1 change unrelated to the rest - `connConfig` had `session` as a parameter, when it only needed the config, so I cleaned that up.

The rest of the changes are to make sure all the code related to connecting to a server is inside `init()`, so that it can be called multiple times for situations where a microservice starts while the cassandra server is not up yet - this will be hopefully be a PR in the future.  So errors that used to be in `NewSession` related to connecting to the server were moved (also marked as TODOs) inside `init()`, and the config error checking inside `init()` was moved out.  These changes should be invisible from the outside as well, so the config error checking that was moved out still returns the same error as the `init()` since it used to be inside there, and one error now inside the init (the `pool.Size()` check) returns `ErrNoConnectionsStarted` even though it’s inside the `init()` now.

PR #843, opened 2016-12-01
Support adding cluster contact points with specified port

trinchan:
In our cluster set up (hosted on [Compose](https://www.compose.com)), the public facing contact points for the cluster do not have consistent ports, due to how they are hosted. We implemented support for specifying the port for each host during cluster configuration.

Let me know if this is something you guys would consider useful or if you guys need any changes before getting this upstream.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -4,6 +4,7 @@ package gocql
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"math"
 	"math/big"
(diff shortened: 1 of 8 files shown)


Zariel:
Is this necessary? Whats stopping you from passing in strings for addresses?

trinchan:
@Zariel You are right, we were having some issues with the AddressTranslator functionality, but the `host:port` functionality in `hostInfo` is working properly

PR #847, opened 2016-12-21
use native context instead of golang.orx/x/net/context

mrwonko:
This was introduced in Go 1.7 and is preferred.

It is however a breaking change and requires Go 1.7, I'm not sure what your policy is on that.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -4,6 +4,7 @@ package gocql
 
 import (
 	"bytes"
+	"context"
 	"io"
 	"math"
 	"math/big"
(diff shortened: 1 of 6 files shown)


mrwonko:
Come to think about it, looks like the interfaces are identical, so you should be able to use either one anyway... Well done, Go!

tendervittles:
Other projects support both contexts with [build constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints). You don't force anyone to upgrade to the latest go version.

See [gorilla mux](https://github.com/gorilla/mux/blob/master/context_native.go) for an example.

If you mix build constraints with godep, the code copied over to vendor will depend on the go version. Your developers should all run the same version.

tendervittles:
Personally, I don't recommend gocql use build constraints.

Since gocql seems to officially support the last two go versions, I'd wait until go1.8 is released around February, then apply your commit.

Issue #849, opened 2017-01-11
KeyspaceMetadata() returns a struct even if the keyspace doesn't exist


jesse-c:
Was this a design decision? If not, I'd be happy to try and tackle this.

For example, return nil, nil, or nil, error("Keyspace doesn't exist") from KeyspaceMetadata().

PR #853, opened 2017-01-19
Change KeypsaceMetadata() to return an error if the keyspace does not exist 

jesse-c:
Previously if a keyspace did not exist a partially empty `KeyspaceMetadata` struct was returned. From what I could see, the more appropriate behaviour would be to return an error indicating so.

Detecting if the keyspace actual exists is done by checking the number of rows returned from the query to the system table `system_schema.keyspaces`. If there were no rows returned then that indicates the keyspace doesn't exist.

No unit tests exist for the called function (`geKeyspaceMetadata`) though an integration test exists for `KeyspaceMetadata()`. A subsequent integration test was added which attempts to get the metadata for a keyspace that is known to not exist.

A slight update to the godoc for KeyspaceMetadata() was added detailing this new behaviour.

Solves issue #849.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -84,3 +84,4 @@ Charles Law <charles.law@gmail.com>; <claw@conduce.com>
 Nathan Davies <nathanjamesdavies@gmail.com>
 Bo Blanton <bo.blanton@gmail.com>
 Vincent Rischmann <me@vrischmann.me>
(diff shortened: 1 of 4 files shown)


Zariel:
`cassandra_test.go:1640: keyspaceMetadata declared and not used`

Zariel:

jesse-c:
No problem!

PR #856, opened 2017-02-01
Get rid of ReconnectInterval

amelius0712:
ReconnectInterval parameter and reconnectDownedHosts goroutine was deprecated because of usage control conn. Furthermore, reconnectDownedHosts does not try to connect, but called handleNodeUp callback. Thus really downed host added into policy pool and marked us "UP" that of cause wasn't true.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -85,3 +85,4 @@ Nathan Davies <nathanjamesdavies@gmail.com>
 Bo Blanton <bo.blanton@gmail.com>
 Vincent Rischmann <me@vrischmann.me>
 Jesse Claven <jesse.claven@gmail.com>
+Pavel Sapezhko <amelius0712@gmail.com>
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -570,26 +570,6 @@ func TestCreateSessionTimeout(t *testing.T) {
 	}
 }
 
-func TestReconnection(t *testing.T) {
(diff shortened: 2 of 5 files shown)


matope:
Please read #723 which describes the background of ReconnectInterval.

amelius0712:
Already read. And also I read a code. You just mark node as UP. And then add this node to the policy pool. Thus, gocql tries to do requests to dead nodes.

matope:
@amelius0712 Thank you for your check. 

Yes, that's what I did. I think we can't completely rely on server-sent event for now. Because gocql sometimes mark nodes DOWN WITHOUT server-sent event. So, we have to recover from the status automatically.

> Thus, gocql tries to do requests to dead nodes.

It's not necessarily accurate. Actually "Thus, gocql tries to do requests to nodes which gocql **believes** it've been DEAD."

matope:
I admit that my patch could be a little dirty. It might be good if we call `ringDescriber.refreshRing()` periodically (I'm not sure if this idea will work fine).

Zariel:
While I agree in parts, this shouldn't cause the driver to send queries to hosts which are not actually up as they will have 0 connections in the pool. The driver needs a way to reconnect to hosts as events are not always reliable, for example all hosts go down but then come back.

A better solution would be to only mark the host as UP iff a connect succeeds. Closing this for now.

Issue #839, opened 2016-11-24
Long connect time when 1 node down in cluster


Highstead:
So I have not isolated the problem yet but if anyone wants to pop in with a guiding hand or tell me functioning as intended i'd appreciate it.

One of the cassandra nodes was terminated in AWS so we lost access to it at a network level.  This increased connection times and caused smoke tests to fail on deployment and thus automatic rollbacks.  

This seems to only be reproducible when the node is terminated / turned off.  connection pool seems to be populating everything asynchronously so it isn't immediately obvious as to why the initial connection would take additional time when it was not the seed node that was down.

Anyhow, more than happy to fix it if that is desirable and would appreciate any direction if available.  I'll keep this updated as i poke around.

thrawn01:
I ran into this issue as well. Turns out `NewSession()` will not return until all discovered peers in a cluster have connected or timed out after 3 attempts (3 attempts is hardcoded). So if your timeout is 5 seconds 5 * 3 = 15 seconds until `NewSession()` returns. Imagine if you have more than one node down!

Currently the connect and connection methods both share the same timeout. I'm working on a PR to change this. Will reference this issue when complete.

Zariel:
If a node is down the error reported from dial should indicate that it is not temporary and should not be retried.

An interesting thought is at what point can we say the driver is `connected` to Cassandra, in the sense that nodes can go up and down, that a DC containing a users data is not contactable, that not enough nodes are connected to serve all queries.

I think it could make sense to say that if we can connect to atleast 1 host, then we are good to go, as it will be Cassandra telling the user that not enough hosts are up instead of us.

thrawn01:
Turns out the issue I encountered was indeed a temporary one, but not one gocql could detect.. Some of the nodes where super busy and allowed the connection, but then never responded in a timely manner. `dial tcp: i/o timeout` which resulted in `gocql: no response to connection startup within timeout` for some nodes.

PR #858, opened 2017-02-09
Separate ConnectTimeout from Timeout

thrawn01:
## Purpose
In large clusters with busy nodes it is not uncommon to have a high connection timeout to tolerate large or longer running queries. With such a configuration and with nodes under heavy load `CreateSession()` can block for the duration of the high timeout `x3` for each node not responding in a timely manner. This can cause issues (See #839) where either a temporary network issue causes connection issues or the connection was successful but Cassandra doesn't respond with the startup frame in a timely manner.

## Implementation
* Introduced a new timeout `ClusterConfig.ConnectTimeout` which is only applied during the initial dial and exchange of startup frames.
* Added a test called `TestStartupTimeout` to ensure timeout while waiting for startup frame works and is separate from `ClusterConfig.Timeout`

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -85,3 +85,4 @@ Nathan Davies <nathanjamesdavies@gmail.com>
 Bo Blanton <bo.blanton@gmail.com>
 Vincent Rischmann <me@vrischmann.me>
 Jesse Claven <jesse.claven@gmail.com>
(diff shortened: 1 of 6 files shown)


Zariel:
Thanks for the contribution!

PR #859, opened 2017-02-14
Improved MapScan() documentation and added new MapScan test

thrawn01:
## Purpose
I spent many ours chasing down a bug that turned out to be a feature! To assist future users I've improved the documentation for `MapScan()` and added a new test demonstrating it's proper use.


Zariel:
Thanks!

PR #860, opened 2017-02-14
GetHosts() is now more version tolerant

thrawn01:
## Purpose
In order to preform more complex host filtering for large clusters, more information on cluster peers is required. This implementation follows the Java Driver method of collecting local and peer host information, which makes the collection of cluster node information more version tolerant. `system.local` and `system.peer` fields that are available for newer clusters versions are safely omitted when running against older clusters.

## Implementation
* Introduced `HostInfo.ConnectAddress()` as a consistent way of retrieving the desired node address. The previous use of `Peer()` as the connect address was overloaded with the `peer` field in the `system.peers` table. In addition it would appear the code was connecting via the `Peer()` address, unless you where aware the code explicitly overwrote `Peer()` with the broadcast address during `InitialHostLookup`.
* Removed the need for `hostSource.localHasRpcAddr` which was miss leading because it was actually the `broadcast_address`
* Removed the need for `checkSystemLocal()`. Previously `checkSystemLocal()` was used to determine if `system.local` had `broadcast_address`. This check is no longer required, as `hostInfoFromMap()` will only set `HostInfo` fields that exist in the table.
* Added `IsValidPeer()` as implemented by the Java driver to ensure we don't connect to new hosts that are not ready
* Moved `HostFilter` application into `GetHosts()` to consolidate code
* Took advice from the TODO and moved `controlConn.fetchHostInfo()` into `host_source` and renamed it `GetHostInfo()`

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -67,7 +67,7 @@ func TestRingDiscovery(t *testing.T) {
 
 	if *clusterSize != size {
 		for p, pool := range session.pool.hostConnPools {
-			t.Logf("p=%q host=%v ips=%s", p, pool.host, pool.host.Peer().String())
+			t.Logf("p=%q host=%v ips=%s", p, pool.host, pool.host.ConnectAddress().String())
 
 		}
 		t.Errorf("Expected a cluster size of %d, but actual size was %d", *clusterSize, size)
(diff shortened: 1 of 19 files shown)


Zariel:
Sorry for taking so long, thanks for the contribution.

PR #862, opened 2017-02-28
Fix data race between event processing and session when concurrent go…

nullaus:
…routines are potentially accessing schema details or creating, altering keyspaces and tables.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -20,9 +20,9 @@ import (
 
 	"golang.org/x/net/context"
 
-	"github.com/gocql/gocql/internal/lru"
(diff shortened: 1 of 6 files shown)


nullaus:
Happy to make a separate pull request, just had to commit import path fixes for my own usage for the time being.

nemothekid:
In general you should make PR's off of branches, instead of your master branch. I'm not sure you can edit your PR to use a different branch.

nullaus:
Yeah, thanks, I know. That's why I made the follow-up comment.

PR #863, opened 2017-03-01
Fix data race between event processing and session when concurrent go…

nullaus:
…routines are potentially accessing schema details or creating, altering keyspaces and tables.


Zariel:
Thanks :)

Issue #828, opened 2016-11-06
CQL date type is not supported?


olegabr:
I have a field of type `date` in my CQL table. When I try to insert data in this table from my `gocql` code, I have this error:

```
can not marshal time.Time into custom(org.apache.cassandra.db.marshal.SimpleDateType)
```

I've tried to use string representation since CQL docs states that `date - Date string, such as 2015-05-03` but has similar error:

```
can not marshal string into custom(org.apache.cassandra.db.marshal.SimpleDateType)
```

How to deal with date type from `gocql` code?

Zariel:
That should marshal, ill take a look

Zariel:
Yep it appears they are not implemented for marshal/unmarshal. Time is trivial, date is not so obvious. Ill need to add some tests before I can add these.

evildecay:
The latest has been supported

Zariel:
thanks

olegabr:
the commit https://github.com/gocql/gocql/commit/bdbe7c9b943cfd07dfeccccbaa04071ae7ce22b6 is very good, but no tests were added for the new functionality added.

evildecay:
@olegabr Yes, test file will be added later.

PR #884, opened 2017-03-24
Fix and test for unmarshal date type

jqin:
Related to #828


Zariel:
Thanks for the contribution

PR #885, opened 2017-03-29
Add hosts to policy pool on ring refreshring

amelius0712:
While ring refreshing if we have found hosts not added to ring yet, we need add them not only to connection pool but into policy poll too.


Zariel:
Thanks for the contribution

PR #887, opened 2017-03-31
Replace golang.org/x/net/context with stdlib context

jacobhaven:
Remove an external dependency since it's been incorporated into the stdlib.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -4,6 +4,7 @@ package gocql
 
 import (
 	"bytes"
+	"context"
 	"io"
 	"math"
 	"math/big"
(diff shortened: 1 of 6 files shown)


leighmcculloch:
The Travis CI tests for go1.6 are failing because `context` wasn't introduced until go1.7.

According to the projects README the last two versions are actively developed for. Is that still the case? If so, this should be fine and Travis CI should be updated to test only go1.7 and go1.8. I've opened #897 proposing this.

If we still really want to support go1.6, we could move the functions that use context into new files, duplicate the files, with one set of the copies having the old import and the build tag at the top `// +build go1.6` and the other set having `// +build go1.7 go1.8`. It would be simpler not to do that and deprecate use of go1.6.

Zariel:
Thanks!

chitrakojha:
This fails programs that are on Go versions < 1.7

chitrakojha:
Is the plan to support older versions ?

Issue #869, opened 2017-03-09
The hostinfo pull request broke connecting to our 2.0.15 cluster


JelteF:
Suddenly with the newest version we cannot connect to our cluster anymore. 
I'm getting this error:
```
panic: host missing connect ip address

goroutine 1 [running]:
github.com/gocql/gocql.Connect(0xc4200869a0, 0xc42001c480, 0x797e20, 0xc420064b80, 0xc42009c000, 0x0, 0x0, 0x0)
        /home/jelte/go/src/github.com/gocql/gocql/conn.go:160 +0xf5a
github.com/gocql/gocql.(*Session).connect(0xc42009c000, 0xc4200869a0, 0x797e20, 0xc420064b80, 0xc420064b80, 0xc42008f748, 0x2008f748)
        /home/jelte/go/src/github.com/gocql/gocql/session.go:629 +0x52
github.com/gocql/gocql.(*hostConnPool).connect(0xc420064b80, 0x0, 0x0)
        /home/jelte/go/src/github.com/gocql/gocql/connectionpool.go:482 +0x83
github.com/gocql/gocql.(*hostConnPool).fill(0xc420064b80)
        /home/jelte/go/src/github.com/gocql/gocql/connectionpool.go:392 +0x140
github.com/gocql/gocql.(*policyConnPool).addHost(0xc42001c420, 0xc4200869a0)
        /home/jelte/go/src/github.com/gocql/gocql/connectionpool.go:215 +0xe5
github.com/gocql/gocql.(*Session).handleNewNode(0xc42009c000, 0xc420010fc0, 0x10, 0x10, 0x2352, 0x2300)
        /home/jelte/go/src/github.com/gocql/gocql/events.go:199 +0x2e6
github.com/gocql/gocql.(*Session).handleNodeUp(0xc42009c000, 0xc420010fc0, 0x10, 0x10, 0x2352, 0x0)
        /home/jelte/go/src/github.com/gocql/gocql/events.go:254 +0x243
github.com/gocql/gocql.(*controlConn).setupConn(0xc420015e80, 0xc4200ae0e0, 0x1, 0x1)
        /home/jelte/go/src/github.com/gocql/gocql/control.go:261 +0x193
github.com/gocql/gocql.(*controlConn).connect(0xc420015e80, 0xc42000e050, 0x1, 0x1, 0x2, 0x0)
        /home/jelte/go/src/github.com/gocql/gocql/control.go:231 +0x18d
[...]

PR #888, opened 2017-04-06
Fix more hostinfo

thrawn01:
## Purpose
This is the fixed version of PR #860 (See #869)

During  initial discovery gocql will only use the user supplied address when picking a control host.
Peers that are discovered will have an `rpc_address` or `peer` address which will be used to connect. 
```go
// session.go
186:	for _, host := range hosts {
187:		host = s.ring.addOrUpdate(host)
188:		s.handleNodeUp(host.ConnectAddress(), host.Port(), false)
189:	}
```
Also
```go
// host_source.go
151: func (h *HostInfo) ConnectAddress() net.IP {
152: 	h.mu.RLock()
153: 	defer h.mu.RUnlock()
154: 
155: 	if h.connectAddress == nil {
156: 		// Use 'rpc_address' if provided and it's not 0.0.0.0
157: 		if h.rpcAddress != nil && !h.rpcAddress.IsUnspecified() {
158: 			return h.rpcAddress
159: 		}
160: 		// Peer should always be set if this from 'system.peer'
161: 		if h.peer != nil {
162: 			return h.peer
163: 		}
164: 	}
165: 	return h.connectAddress
166: }
```
If we receive a `NEW_NODE` or `UP` event the address provided by that event will be used to query the `system.peers` table then filtered by `hostFilter` before connecting.

If the up event is for the control connection,  always respect the provided control node address and disregard the ip address the cassandra node provides. We do this as we are already connected and have a known valid ip address. This insulates gocql from client connection issues stemming from node misconfiguration. For instance when a node is run from a container, by
default the node will report its ip address as 127.0.0.1 which is typically invalid.
```go
// host_source.go
610: host.SetConnectAddress(ip)
```

### Original Purpose for this PR
In order to perform more complex host filtering for large clusters, more information on cluster peers is required. This implementation follows the Java Driver method of collecting local and peer host information, which makes the collection of cluster node information more version tolerant. `system.local` and `system.peer` fields that are available for newer clusters versions are safely omitted when running against older clusters.

## Implementation
* Introduced `HostInfo.ConnectAddress()` as a consistent way of retrieving the desired node address. The previous use of `Peer()` as the connect address was overloaded with the `peer` field in the `system.peers` table. In addition it would appear the code was connecting via the `Peer()` address, unless you were aware the code explicitly overwrote `Peer()` with the broadcast address during `InitialHostLookup`.
* Removed the need for `hostSource.localHasRpcAddr` which was misleading because it was actually the looking for the`broadcast_address`
* Removed the need for `checkSystemLocal()`. Previously `checkSystemLocal()` was used to determine if `system.local` had `broadcast_address`. This check is no longer required, as `hostInfoFromMap()` will only set `HostInfo` fields that exist in the table.
* Added `IsValidPeer()` as implemented by the Java driver to ensure we don't connect to new hosts that are not ready.
* Moved `HostFilter` application into `GetHosts()` to consolidate code
* Took advice from the TODO and moved `controlConn.fetchHostInfo()` into `host_source` and renamed it `GetHostInfo()`

### DIFF
--- a/events.go
+++ b/events.go
@@ -174,23 +174,15 @@ func (s *Session) handleNodeEvent(frames []frame) {
 }
 
 func (s *Session) handleNewNode(ip net.IP, port int, waitForBinary bool) {
-	var hostInfo *HostInfo
-	if s.control != nil && !s.cfg.IgnorePeerAddr {
-		var err error
-		hostInfo, err = s.control.fetchHostInfo(ip, port)
-		if err != nil {
-			Logger.Printf("gocql: events: unable to fetch host info for (%s:%d): %v\n", ip, port, err)
-			return
-		}
-	} else {
-		hostInfo = &HostInfo{peer: ip, port: port}
+	// Get host info and apply any filters to the host
+	hostInfo, err := s.hostSource.GetHostInfo(ip, port)
+	if err != nil {
+		Logger.Printf("gocql: events: unable to fetch host info for (%s:%d): %v\n", ip, port, err)
+		return
 	}
 
-	if s.cfg.IgnorePeerAddr && hostInfo.Peer().Equal(ip) {
-		hostInfo.setPeer(ip)
-	}
-
-	if s.cfg.HostFilter != nil && !s.cfg.HostFilter.Accept(hostInfo) {
+	// If hostInfo is nil, this host was filtered out by cfg.HostFilter
+	if hostInfo == nil {
 		return
 	}
 
@@ -216,7 +208,7 @@ func (s *Session) handleRemovedNode(ip net.IP, port int) {
 	// we remove all nodes but only add ones which pass the filter
 	host := s.ring.getHost(ip)
 	if host == nil {
-		host = &HostInfo{peer: ip, port: port}
+		host = &HostInfo{connectAddress: ip, port: port}
 	}
 
 	if s.cfg.HostFilter != nil && !s.cfg.HostFilter.Accept(host) {
@@ -240,9 +232,9 @@ func (s *Session) handleNodeUp(ip net.IP, port int, waitForBinary bool) {
 
 	host := s.ring.getHost(ip)
 	if host != nil {
-		if s.cfg.IgnorePeerAddr && host.Peer().Equal(ip) {
+		if s.cfg.IgnorePeerAddr && host.ConnectAddress().Equal(ip) {
 			// TODO: how can this ever be true?
-			host.setPeer(ip)
+			host.SetConnectAddress(ip)
 		}
 
 		if s.cfg.HostFilter != nil && !s.cfg.HostFilter.Accept(host) {
@@ -269,7 +261,7 @@ func (s *Session) handleNodeDown(ip net.IP, port int) {
 
 	host := s.ring.getHost(ip)
 	if host == nil {
-		host = &HostInfo{peer: ip, port: port}
+		host = &HostInfo{connectAddress: ip, port: port}
 	}
 
 	if s.cfg.HostFilter != nil && !s.cfg.HostFilter.Accept(host) {
--- a/events.go
+++ b/events.go
@@ -232,8 +232,9 @@ func (s *Session) handleNodeUp(ip net.IP, port int, waitForBinary bool) {
 
 	host := s.ring.getHost(ip)
 	if host != nil {
-		if s.cfg.IgnorePeerAddr && host.ConnectAddress().Equal(ip) {
-			// TODO: how can this ever be true?
+		// If we receive a node up event and user has asked us to ignore the peer address use
+		// the address provide by the event instead the address provide by peer the table.
+		if s.cfg.IgnorePeerAddr && !host.ConnectAddress().Equal(ip) {
 			host.SetConnectAddress(ip)
 		}
 
--- a/host_source.go
+++ b/host_source.go
@@ -145,19 +145,19 @@ func (h *HostInfo) setPeer(peer net.IP) *HostInfo {
 	return h
 }
 
-// Returns the address that should be used to connect to the host
-// This defaults to 'broadcast_address', then falls back to 'peer'
-// This is to maintain existing functionality. If you wish to
-// override this, use an AddressTranslator or use a HostFilter
-// to SetConnectAddress()
+// Returns the address that should be used to connect to the host.
+// If you wish to override this, use an AddressTranslator or
+// use a HostFilter to SetConnectAddress()
 func (h *HostInfo) ConnectAddress() net.IP {
 	h.mu.RLock()
 	defer h.mu.RUnlock()
 
 	if h.connectAddress == nil {
-		if h.broadcastAddress != nil {
-			return h.broadcastAddress
+		// Use 'rpc_address' if provided and it's not 0.0.0.0
+		if h.rpcAddress != nil && !h.rpcAddress.IsUnspecified() {
+			return h.rpcAddress
 		}
+		// Peer should always be set if this from 'system.peer'
 		if h.peer != nil {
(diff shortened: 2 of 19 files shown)


host_source.go
```diff
@@ -271,106 +365,279 @@ func checkSystemLocal(control *controlConn) (bool, error) {
...
-// Returns true if we are using system_schema.keyspaces instead of system.schema_keyspaces
-func checkSystemSchema(control *controlConn) (bool, error) {
-	iter := control.query("SELECT * FROM system_schema.keyspaces")
-	if err := iter.err; err != nil {
-		if errf, ok := err.(*errorFrame); ok {
-			if errf.code == errReadFailure {
-				return false, nil
+// Given a map that represents a row from either system.local or system.peers
+// return as much information as we can in *HostInfo
+func (r *ringDescriber) hostInfoFromMap(row map[string]interface{}) (error, *HostInfo) {
```
Zariel: change the return definition to be `(*HostInfo, error)` to be consistent with Go code style

Zariel: change the return definition to be (*HostInfo, error) to be consistent with Go code style

host_source.go
```diff
@@ -271,106 +365,279 @@ func checkSystemLocal(control *controlConn) (bool, error) {
...
+	host := HostInfo{}
```
Zariel: allocate this like
```go
host := &HostInfo{
   port: r.session.cfg.Port,
}
```
that way it defaults simply

host_source.go
```diff
@@ -271,106 +365,279 @@ func checkSystemLocal(control *controlConn) (bool, error) {
...
-		if iter == nil {
-			return r.prevHosts, r.prevPartitioner, nil
-		}
+	// extract all available info about the host
+	err, host := r.hostInfoFromMap(row)
+	if err != nil {
+		return nil, err
+	}
+
+	return host, err
```
Zariel: no need to return here, instead just return directly `r.hostInfoFromMap(row)`

host_source.go
```diff
@@ -271,106 +365,279 @@ func checkSystemLocal(control *controlConn) (bool, error) {
...
+
+	// expect only 1 row
+	it.MapScan(row)
+	if err := it.Close(); err != nil {
+		return nil, err
 	}
 
-	localHost.port = r.session.cfg.Port
+	// extract all available info about the host
+	err, host := r.hostInfoFromMap(row)
```
Zariel: again return here

host_source.go
```diff
@@ -271,106 +365,279 @@ func checkSystemLocal(control *controlConn) (bool, error) {
...
 
-		if err = iter.Close(); err != nil {
-			return nil, "", err
-		}
+	it := r.session.control.query("SELECT * FROM system.peers WHERE peer=?", ip)
+	if it == nil {
+		return nil, errors.New("Attempted to query 'system.peers' on a closed control connection")
+	}
+
+	// expect only 1 row
```
Zariel: these functions all share common code, maybe split the common code out to something like

```go
func (r *ringDescriber) extractHostInfo(it Iter) (*HostInfo, error) {
...
}
```

host_source.go
```diff
@@ -271,106 +365,279 @@ func checkSystemLocal(control *controlConn) (bool, error) {
...
+	if err != nil {
+		return r.prevHosts, r.prevPartitioner, err
 	}
 
-	if err = rows.Err(); err != nil {
-		return nil, "", err
+	hosts = append(hosts, localHost)
+
+	// Filter the hosts if filter is provided
+	var filteredHosts []*HostInfo
```
Zariel:
```go
filteredHosts := hosts
if r.session.cfg.HostFilter != nil {
    filterHosts = filterHosts[:0]
    ...
}
```

host_source.go
```diff
@@ -271,106 +365,279 @@ func checkSystemLocal(control *controlConn) (bool, error) {
...
-	return hosts, partitioner, nil
+	return filteredHosts, localHost.partitioner, nil
 }
 
-func (r *ringDescriber) matchFilter(host *HostInfo) bool {
-	if r.dcFilter != "" && r.dcFilter != host.DataCenter() {
-		return false
+// Given an ip/port return HostInfo for the specified ip/port
+func (r *ringDescriber) GetHostInfo(ip net.IP, port int) (*HostInfo, error) {
+	var host *HostInfo
```
Zariel: define these in a var block and closer to where they are assigned

host_source.go
```diff
@@ -271,106 +365,279 @@ func checkSystemLocal(control *controlConn) (bool, error) {
...
+		return nil, errors.New("invalid control connection")
+	}
+
+	// If we are asking about the same node our control connection has a connection too
+	if controlHost.ConnectAddress().Equal(ip) {
+		host, err = r.GetLocalHostInfo()
+	} else {
+		host, err = r.GetPeerHostInfo(ip, port)
+	}
+	// Always respect the user provided or discovered address
```
Zariel: can you add a why to this comment

host_source.go
```diff
@@ -271,106 +365,279 @@ func checkSystemLocal(control *controlConn) (bool, error) {
...
+		// If it's not a valid peer
+		if !r.IsValidPeer(host) {
+			Logger.Printf("Found invalid peer '%+v' "+
+				"Likely due to a gossip or snitch issue, this host will be ignored", host)
+			continue
 		}
+		hosts = append(hosts, host)
+	}
+	if it.err != nil {
+		return nil, errors.Wrap(it.err, "GetClusterPeerInfo()")
```
Zariel: Is it necessary to add a new dependency to achieve this?

host_source.go
```diff
@@ -7,8 +7,12 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/pkg/errors"
 )
 
+const assertErrorMsg = "Assertion failed for %s"
```
Zariel: define this inside the function it is used

events.go
```diff
@@ -240,9 +232,9 @@ func (s *Session) handleNodeUp(ip net.IP, port int, waitForBinary bool) {
 
 	host := s.ring.getHost(ip)
 	if host != nil {
-		if s.cfg.IgnorePeerAddr && host.Peer().Equal(ip) {
+		if s.cfg.IgnorePeerAddr && host.ConnectAddress().Equal(ip) {
```
Zariel: looking at it this code looks wrong, should it be `s.cfg.IgnorePeerAddr && !host.ConnectAddress().Equal(ip)`?

host_source.go
thrawn01: As Dave Cheney lays out here https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully annotated errors are useful for end users, but since gocql does not provide annotated errors; I agree, this doesn't make much sense here.

host_source.go
thrawn01: I evaluated this statement again and found that the java driver appears to use the address provided by the `NEW_NODE` event for peer table lookup only. I will change the code to mirror this and improve the comment to better explain why we prefer the user provided addresses in the case of the control connection.

events.go
thrawn01: I think you are correct, I will add a comment on this code to help clarify.

thrawn01:
I'm sorry I have not been able to update this PR, been dealing with health problems in recent months. Thanks for the feedback, these are great changes.

Zariel:
No problem hope you doing ok. Looks like the refactoring went a little ary and messed up some returns?

thrawn01:
Doh! I don't have my testing environment setup so figured I would let the CI do the work for me. I should have atleast built the code before making a commit.  =/

Zariel:

PR #889, opened 2017-04-06
CreateSession() will now attempt to resolve all host names

### LINKED ISSUE #873: DNS request timeout appears skip moving on to next node and returns nil session
We use an internal DNS server, but as you can see from the picture when it was misconfigured to go out to Google DNS at 8.8.8.8, the gocql connector couldn't resolve the domain name and it didn't go on to any other node, it just returned a nil session which crashed the program, I put in a piece of code to continue trying to re-establish the session to force it to use a different node, but I figured this was a fix that could be incorporated into the gocql package itself. Here is some code that maybe could generate the error again if you correctly create a domain that could be resolved and a domain that couldn't be resolved on your DNS server. 

`var cluster *gocql.ClusterConfig`
`cluster = gocql.NewCluster("domainthatdoesexist.mynetwork","domainthatdoesntexist.mynetwork")`
`session, err := cluster.CreateSession()`

screen shot 2017-03-10 at 1 48 52 pm

thrawn01:
## Purpose
`CreateSession()` would exit prematurely if it encountered a hostname it could not resolve. With this patch all host names will be resolved before an error is returned. fixes #873 

## Implementation
* `addrsToHosts()` will log and continue to the next host if `hostInfo()` returns a `DNSError`
* Added an error if none of the host names can be resolved

### DIFF
--- a/session.go
+++ b/session.go
@@ -92,7 +92,7 @@ func addrsToHosts(addrs []string, defaultPort int) ([]*HostInfo, error) {
 		hosts = append(hosts, host)
 	}
 	if len(hosts) == 0 {
-		return nil, errors.New("gocql: failed to resolve any of the provided hostnames")
+		return nil, errors.New("failed to resolve any of the provided hostnames")
 	}
 	return hosts, nil
 }
--- a/session.go
+++ b/session.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


session.go
```diff
@@ -76,16 +77,23 @@ var queryPool = &sync.Pool{
...
 func addrsToHosts(addrs []string, defaultPort int) ([]*HostInfo, error) {
-	hosts := make([]*HostInfo, len(addrs))
-	for i, hostport := range addrs {
+	var hosts []*HostInfo
+	for _, hostport := range addrs {
 		host, err := hostInfo(hostport, defaultPort)
 		if err != nil {
+			// Try other hosts if unable to resolve DNS name
+			if _, ok := err.(*net.DNSError); ok {
+				Logger.Printf("gocql: %v\n", err)
```
Zariel: improve this log line or remove it

thrawn01: Thanks for the feedback!

### Given
```go
	cluster := gocql.NewCluster("blahasdf.com", "blahasdf2.com")
	_, err := cluster.CreateSession()
	if err != nil {
		fmt.Printf("err: %s\n", err)
		os.Exit(1)
	}
	os.Exit(0)
```
### Output
```bash
$ go run main.go
2017/04/07 07:14:56 gocql: lookup blahasdf.com: no such host
2017/04/07 07:14:56 gocql: lookup blahasdf2.com: no such host
err: gocql: unable to create session: gocql: failed to resolve any of the provided hostnames
exit status 1
```
I'll add `dns error` to the message to make it more verbose.

session.go
```diff
@@ -76,16 +77,23 @@ var queryPool = &sync.Pool{
...
+			}
 			return nil, err
 		}
 
-		hosts[i] = host
+		hosts = append(hosts, host)
+	}
+	if len(hosts) == 0 {
+		return nil, errors.New("gocql: failed to resolve any of the provided hostnames")
 	}
```
leighmcculloch: I think we should have tests that confirm this new behavior functions as expected too.

thrawn01: Great suggestion! I discoverd the error message duplicated the "gocql: " when writing the test.

PR #890, opened 2017-04-09
marshal: Add an error when UUID has wrong length

### LINKED ISSUE #868: Marshal/Unmarshal of struct fields of type []byte do not work for uuid cql type
Field of type `[]byte` that maps to an `uuid` type within Cassandra will failed to Marshal/Unmarshal with error `can not marshal []uint8 into uuid`.  
Likely because gocql's functions _marshalUUID_ and _unmarshalUUID_ test if the field is of type `[]byte`. But the go's `.(type)` switch returns a type `[]uint8` instead !  
The switch within _marshalUUID_ and _unmarshalUUID_ functions (and related) should have : 
```
case []uint8:
```

leighmcculloch:
What
===
Add a descriptive error when a UUID is marshaled or unmarshaled and the length of the `[]byte` is not exactly 16 bytes long.

Example
===

```
can not marshal []byte 6 bytes long into timeuuid, must be exactly 16 bytes long
```

Why
===
In cases where UUID `[]byte` was not the correct length the generic `can not marshal %T into %s` error was being returned. The error was misleading because it gave the impression the root cause was the type of the data and not it's length. Additionally the `%T` formatter for a `[]byte` prints out `[]uint8` which was even more confusing.

Notes
===
This PR contains three commits. It's easier to digest the change if you review them separately as the first refactors the tests to support expectations on the error itself.

Ref
===
Fixes #868

### DIFF
--- a/marshal_test.go
+++ b/marshal_test.go
@@ -18,44 +18,60 @@ import (
 type AliasInt int
 
 var marshalTests = []struct {
-	Info  TypeInfo
-	Data  []byte
-	Value interface{}
+	Info           TypeInfo
+	Data           []byte
+	Value          interface{}
+	MarshalError   error
+	UnmarshalError error
 }{
 	{
 		NativeType{proto: 2, typ: TypeVarchar},
 		[]byte("hello world"),
 		[]byte("hello world"),
+		nil,
+		nil,
 	},
 	{
 		NativeType{proto: 2, typ: TypeVarchar},
 		[]byte("hello world"),
 		"hello world",
+		nil,
+		nil,
 	},
 	{
 		NativeType{proto: 2, typ: TypeVarchar},
 		[]byte(nil),
 		[]byte(nil),
+		nil,
+		nil,
 	},
 	{
 		NativeType{proto: 2, typ: TypeVarchar},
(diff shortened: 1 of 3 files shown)


marshal_test.go
```diff
@@ -309,86 +416,120 @@ var marshalTests = []struct {
...
 		bigintize("123456789123456789123456789"), // From the datastax/python-driver test suite
+		nil,
+		nil,
 	},
 	{
 		NativeType{proto: 2, typ: TypeVarint},
 		[]byte(nil),
 		nil,
+		nil,
+		UnmarshalError("can not unmarshal into non-pointer <nil>"),
```
leighmcculloch: @Zariel: 👋 Aside to the main issue being addressed in this PR it looks like the `Marshal` and `Unmarshal` functions are not entire symmetrical. The `[]byte` here can be marshaled but not unmarshaled.

Zariel:

marshal_test.go
leighmcculloch: Actually, I guess the asymmetry here doesn't matter, since you wouldn't Unmarshal into nil.

PR #892, opened 2017-04-17
fix consistent high timeout err rate caused by prepareStatement

zhixinwen:
Currently gocql will cache the timeout err caused by network instability when prepare a statement.

As a result, if gocql fails to prepare a statement due to network issue at the first time on a connection, the rest of the call to that statement on that connection would return "context deadline exceeded”. 

It can result in high and consistent err rate and can only fixed by restart a host.

We have been suffering from this in production for a while.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -711,11 +711,8 @@ func (c *Conn) prepareStatement(ctx context.Context, stmt string, tracer Tracer)
 
 	framer, err := c.exec(ctx, prep, tracer)
 	if err != nil {
-		flight.err = err
 		flight.wg.Done()
-		if err.Error() == "context deadline exceeded" {
-			c.session.stmtsLRU.remove(stmtCacheKey)
-		}
+		c.session.stmtsLRU.remove(stmtCacheKey)
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


zhixinwen:
A more detailed description of the problem can be found at this [post](https://groups.google.com/forum/#!topic/gocql/YWEe_Sy3kbA).

conn.go
```diff
@@ -713,6 +713,9 @@ func (c *Conn) prepareStatement(ctx context.Context, stmt string, tracer Tracer)
 	if err != nil {
 		flight.err = err
 		flight.wg.Done()
+		if err.Error() == "context deadline exceeded" {
```
Zariel: instead of doing this we should compare the value of `context.ErrTimeout`, but I think in this case we can just always remove the value on errors

zhixinwen:
@Zariel Fixed

Zariel:
Great thanks!

rengawm:
lol, this has been plaguing us for months and I just spent the last day digging into this same bug. I was literally writing up an issue report to describe my findings and see what people thought a good fix would be... and I went to grab some links from the latest commit to the code which causes the problem, and noticed that it just got fixed. Thanks! :D

PR #897, opened 2017-04-20
Add go1.8 to CI tests

leighmcculloch:
What
===
Add go1.8 and remove go1.6 from CI tests.

Why
===
It's the latest Go release and we should be ensuring tests pass on it and according to the gocql README the last two major versions will be supported. The last two major versions are go1.7 and go1.8.


Zariel:
Thanks!

Issue #861, opened 2017-02-17
Support UNSET values in Prepared Statements


nemosupremo:
A useful feature ([CASSANDRA-7304](https://issues.apache.org/jira/browse/CASSANDRA-7304)) in 2.2.0+ is to treat `null` values as unset in prepared statements (the difference here is that sending a `null` in a prepared statement is the equivalent to a DELETE - which causes tombstones, unset just treats it as you not having set anything at all). I'm looking to start work on PR, but I don't think I can get to it soon, so I just opened an issue instead.

Zariel:
From what I can tell this just requires sending -2 instead of -1 for null values. How would the user indicate that it should send "unset" instead of "nil", we could add a type? `gocql.UnsetCol` which got marshalled to `-2` which would be relatively clean

nemothekid:
If we follow the Python driver, there are two methods we would implement

1. A custom type (like `gocql.UnsetCol`).
1. A session level setting that treats all `nil`s as `gocql.UnsetCol`/`-2`.

Going with 1 would be good, but when I last looked at it, the when we write the bytes to the connection we assume if the byte array is nil, we will always write -1.

nemothekid:
Maybe the fix could be added here - https://github.com/gocql/gocql/blob/5e9bfab8b94dac35b012c8bd39953e928299a86d/frame.go#L1326

Something like -
```diff
if n := len(opts.values); n > 0 {
	f.writeShort(uint16(n))
	for i := 0; i < n; i++ {
		if names {
			f.writeString(opts.values[i].name)
		}
+              if (opts.values[i].unset) {
+		     f.writeBytesButNullIsNeg2(opts.values[i].value)
+              } else {
+                    f.writeBytes(opts.values[i].value)
                }
	}
}
```

anshulr:
Any Updates on this ?

PR #903, opened 2017-05-04
Support for UnSet columns

Highstead:
# Purpose 
Adding support for the UNSET_VALUE in gocql.  See #861 and [Cassandra Jira issue](https://issues.apache.org/jira/browse/CASSANDRA-7304)

## Tests to Add:
- [x] Insert Unset column does not replace existing column
- [x] Insert Batch with unset

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -16,6 +16,12 @@ import (
 	"time"
 )
 
+type UnsetColumn struct{}
+
+func UnsetValue() UnsetColumn {
+	return UnsetColumn{}
+}
+
 const (
 	protoDirectionMask = 0x80
 	protoVersionMask   = 0x7F
@@ -1255,8 +1261,10 @@ func (f *framer) writeAuthResponseFrame(streamID int, data []byte) error {
 
 type queryValues struct {
 	value []byte
+
 	// optional name, will set With names for values flag
-	name string
+	name    string
+	isUnset bool
 }
 
 type queryParams struct {
@@ -1319,11 +1327,16 @@ func (f *framer) writeQueryParams(opts *queryParams) {
 
 	if n := len(opts.values); n > 0 {
 		f.writeShort(uint16(n))
+
 		for i := 0; i < n; i++ {
 			if names {
 				f.writeString(opts.values[i].name)
 			}
-			f.writeBytes(opts.values[i].value)
+			if opts.values[i].isUnset {
+				f.writeUnset()
+			} else {
+				f.writeBytes(opts.values[i].value)
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


frame.go
```diff
@@ -1319,11 +1327,16 @@ func (f *framer) writeQueryParams(opts *queryParams) {
 
 	if n := len(opts.values); n > 0 {
 		f.writeShort(uint16(n))
+
 		for i := 0; i < n; i++ {
 			if names {
 				f.writeString(opts.values[i].name)
 			}
-			f.writeBytes(opts.values[i].value)
+			if opts.values[i].isUnset {
```
Highstead: Not 100% this needs to be here.

marshal.go
```diff
@@ -1127,21 +1131,21 @@ func marshalDate(info TypeInfo, value interface{}) ([]byte, error) {
 		return v.MarshalCQL(info)
 	case int64:
 		timestamp = v
-		x := timestamp/86400000 + int64(1 << 31)
+		x := timestamp/86400000 + int64(1<<31)
```
Highstead: go fmt changes.

frame.go
```diff
@@ -16,6 +16,12 @@ import (
 	"time"
 )
 
+type UnsetColumn struct{}
+
+func UnsetValue() UnsetColumn {
```
Zariel: do we need this function? If we do can we not export UnsetColumn?

frame.go
```diff
@@ -1786,6 +1807,14 @@ func (f *framer) writeStringList(l []string) {
 	}
 }
 
+func (f *framer) writeUnset() {
+	// Protocol version 4 specifies that bind variables do not require having a
+	// value when executing a statement.   Bind variables without a value are
+	// called 'unset'. The 'unset' bind variable is serialized as the int
+	// value '-2' without following bytes.
+	f.writeInt(-2)
```
Zariel: here we need to write -1 if were on a protocol before version 4

marshal.go
```diff
@@ -57,6 +57,10 @@ func Marshal(info TypeInfo, value interface{}) ([]byte, error) {
 		}
 	}
 
+	if _, ok := value.(UnsetColumn); ok {
```
Zariel: why not have this in the type switch below?

frame.go
Highstead: Other drivers throw an exception here.  I opt'd to let the user hang themselves which potentially is not the right answer.  By inserting -1 we insert nulls which has the side effect of causing additional tombstones in cassandra.

Would we prefer i panic if protocol is < 4?

frame.go
Highstead: I'll expose a variable of type UnsetColumn instead.

marshal.go
Highstead: I can move it into the if statement below with the custom unmarshaller for readability/maintainability or alternatively it can exist in each of the Unmarshal functions.  (Some data types don't support unset, tuple, map and a couple others)

It wouldn't fit nicely into the switch statement below because those are the Type info that relates to the row's types.  So a column wouldn't be of type Uset.  I'll fire up a branch with the marshalling changes in their respective types, it just seemed gross at the time.

Highstead:
Sorry, could you elaborate on what you mean by "There should also be a check that it can't be used to unmarshal a value into."?

Zariel:
I guess its ok as it will error out regardless of handling the type. LGTM

PR #916, opened 2017-05-30
Add exponential backoff retry policy

rkuris:
To improve reliability in the event of a flaky network, we want to use this retry policy, which has a configurable backoff and uses some jitter to mitigate thundering herds when the network is misbehaving.


Zariel:
Thanks for the contribution, could you add yourself to the authors file and I will merge.

rkuris:
Done! Thank you!

Zariel:
Thanks very much for the contribution

PR #918, opened 2017-06-01
Fix iter Host method

furmmon:
# Purpose

The function below always return `nil` because the value it queries is never set.
```
// session.go

// Host returns the host which the query was sent to.
1050: func (iter *Iter) Host() *HostInfo {
1051:	return iter.host
1052:}
```

This is because the host is never set in the Iter structure.

# Implementation
Add the host to Iter when the query is executed with success on a host (query_executor.go:50)

Add a test to check if the host is well set on a successful query


Zariel:
Thanks!

gabsn:
I was waiting a lot, thanks for that!

PR #923, opened 2017-06-25
host: use broadcast_address in ConnectAddress

Zariel:
If the rpc_address is set to 0.0.0.0 then we should use
broadcast_address if it is set.


Zariel:
Should fix #923 

cc @CodingMinds

Issue #929, opened 2017-06-30
gocql failed to connect to Cassandra in Travis CI


hsluoyz:
https://travis-ci.org/casbin/cassandra-adapter/jobs/248833341

Zariel:
Could you try with the latest commit please.

hsluoyz:
Hi @Zariel ,

The error remains the same: https://travis-ci.org/casbin/cassandra-adapter/jobs/248833341

Zariel:
I was able to reproduce locally, this seems to only affect 2.0.x and I no longer remember what the changes to the system tables were between these versions so I hope this is now fixed. Can you try and reproduce and confirm. And FYI, I would update the version of Cassandra you test against :)

hsluoyz:
Hi @Zariel ,

The error has been fixed when I restart the build:)

My testbed is Travis CI. So the version of Cassandra is not controlled by me. I didn't see its version in the printed log of Travis page either. Thanks for your support!

PR #931, opened 2017-07-01
control: improve addr parsing

Zariel:
Check the parsed host from the remote addr when connecting to the
control conn is a valid parsed IP.

Short circuit the host lookup if the remote addr is a TCPAddr then use
its IP and port.


Zariel:
updates #929

PR #932, opened 2017-07-02
host_source: correctly get local hostinfo

Zariel:
The system.local table might not have the address info for the connected
host, if this is the case use the connect address from the control
connection.

Add more checks that the connect address is valid to not end up adding
invalid hosts.


Zariel:
updates #929

Issue #572, opened 2016-01-12
The current unmarshalling of tuples is incompatible with nested tuples e.g. map(tuple)


kristoiv:
Because of the way gocql unmarshal's tuples into additional columns, it fails when the tuple is nested. In the example below we create a test keyspace and a table with a map-nested tuple of two integers.

This thread is meant as a place to discuss alternative implementations.

If you run this example you'll see something like:

``` go
cannot unmarshal tuple into *[]interface {}
```

``` go
package main

import (
    "fmt"
    "log"

    "github.com/gocql/gocql"
)

func main() {
    cluster := gocql.NewCluster("127.0.0.1")
    cluster.Consistency = gocql.One
    cluster.ProtoVersion = 4

    session, err := cluster.CreateSession()
    if err != nil {
        log.Fatalln(err)
    }

    // Drop testing keyspace
    if err := session.Query(`DROP KEYSPACE gocqldebugging_test;`).Exec(); err != nil {
        log.Fatalln(err)
    }

    // Create testing keyspace
    if err := session.Query(`CREATE KEYSPACE gocqldebugging_test WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1};`).Exec(); err != nil {
        log.Fatalln(err)
    }

    // Create table with tuple in it
    if err := session.Query(`CREATE TABLE gocqldebugging_test.test (id uuid PRIMARY KEY, myNonCustom map<uuid,frozen<tuple<int, int>>>);`).Exec(); err != nil {
        log.Fatalln(err)
    }

    // Insert some data into our table
    if err := session.Query(`INSERT INTO gocqldebugging_test.test (id, myNonCustom) VALUES (uuid(), {uuid():(1, 2),uuid():(3,4)});`).Exec(); err != nil {
        log.Fatalln(err)
    }

    var id gocql.UUID
    m := make(map[gocql.UUID][]interface{})
    if err := session.Query(`select id, myNonCustom from gocqldebugging_test.test`).Scan(&id, &m); err != nil {
        log.Fatalln(err)
    }

    fmt.Println(id, m)
}
```

Zariel:
@kristoiv what sort of API makes sense for tuple unmarshalling? Type should be able to unmarshal into a single tuple column?

So far the best I can think of is `[]interface{}` or a list of anything, map would work as well.

kristoiv:
[...]

PR #937, opened 2017-07-08
marshal: support nested tuples

Zariel:
Add support for marshalling/unmarshalling nested tuples via slice, array
and structs.

Also improve tuple handling so they can be used with slices, arrays and
structs for values.


Zariel:
Fixes #572

PR #940, opened 2017-07-17
avoid SIGSEGV by not tracing unless query is trace enabled

kthommandra:
When 'nodetool settraceprobability' is used,
it's possible to get a response frame with trace information
even though client did not specifically mark the query
for tracing.

There are two places in conn.go where this can cause SIGSEGV, prepareStatement and executeQuery. 

The prepareStatement path was fixed in https://github.com/gocql/gocql/commit/990865929b372dba9393b0ba8e4967288594b59b


Zariel:
Thanks for the contribution!

PR #950, opened 2017-08-04
Changes to skip prepare statement

kaavee315:
Changes to allow execution of statements not following prepare and execute paradigm. It can be specified in session if the user wants to skip the prepare statement and directly give command as string and execute


Zariel:
This also needs a test to check the driver can correctly marshall and unmarshall queries

PR #952, opened 2017-08-05
hostsource: remove hosts no longer in peers

Zariel:
When refreshing the ring remove hosts from the drivers ring cache to
prevent it from trying to connect to removed hosts if it does not
receive a down host event.


Zariel:
updates #831

PR #963, opened 2017-08-17
HostInfo, remove reentrancy of read lock in Equal

gwik:
See the following discussion:
https://github.com/golang/go/issues/15418

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -123,16 +123,11 @@ type HostInfo struct {
 func (h *HostInfo) Equal(host *HostInfo) bool {
 	h.mu.RLock()
 	defer h.mu.RUnlock()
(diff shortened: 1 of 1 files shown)


host_source.go
```diff
@@ -175,13 +170,18 @@ func (h *HostInfo) connectAddressLocked() (net.IP, string) {
...
+// HostFilter to SetConnectAddress()
 func (h *HostInfo) ConnectAddress() net.IP {
 	h.mu.RLock()
 	defer h.mu.RUnlock()
 
+	return h.getConnectAddress()
+}
+
+// caller must hold mu
+func (h *HostInfo) getConnectAddress() net.IP {
```
Zariel: there is already an unlocked version of method, what is this adding?

PR #976, opened 2017-09-22
Add Go 1.8 as supported in README

dankinder:
Based on .travis.yml change it looks like 1.8 is supported right?


Zariel:
tbh this whole section needs rewriting (maybe the whole readme? or moving to another page...) as it was intended to convey that in CI we test using these versions of Go using these versions of Cassandra (and at this point I do not see much value in CI testing N and N-1 version of Go) and instead gets interpreted as what versions the driver supports, which is not accurate as it works with 2.0 and 3.11.

kirillDanshin:
I think this PR can be closed as #982 is now merged and now readme looks good

PR #980, opened 2017-09-25
Respect exponential retry Max

Gandem:
This PR makes sure that the sleep time for ExponentialBackoffRetryPolicy never exceeds Max.

Max was already defined in the struct but not used.


Zariel:
Great thanks for the contribution!

PR #983, opened 2017-09-28
Consistency no panic

mmatczuk:
This PR adds text marshaller and unmarshaller for `Consistency` and `SerialConsistency` it removes panic recover behavior while parsing. https://github.com/gocql/gocql/pull/983/commits/ab2b96ae82252983855b8291290d4b12fe8d3c75 changes the parsing method naming to more common ones.


eculver:
@Zariel @mmatczuk ab2b96a is an unnecessary breaking change. Can we revert and fix so that this doesn't break callers of `ParseConsistency`?

PR #986, opened 2017-10-05
fix DCAwareRoundRobinPolicy host distribution

bkrebsbach:
DCAwareRoundRobinPolicy currently picks the first host in a map
of all local and remote hosts. This has a tendency to pick the same
host more frequently than others, resulting in an uneven distribution
of operations within the cluster. Modify DCAwareRoundRobinPolicy to
more closely match the RoundRobinHostPolicy by always incrementing the
hosts in a DC.


Zariel:
Thanks very much!

PR #988, opened 2017-10-16
Fix race condition with tokenAwareHostPolicy SetPartitioner

inf-rno:
- it wasn't using the lock and causing data races


Zariel:

PR #991, opened 2017-10-23
added passing port to NewCluster

ror6ax:
Adressing https://github.com/gocql/gocql/issues/990


Zariel:
This won't compile as the splat had to the be the last argument and further this isn't necessary as this is the default value and can be overridden. The port for each host can be passed via the address supplied which is split and of it doesn't have a port it uses the default supplied in the configuration. In fact I don't even think having the port in the config makes much sense any longer.

ror6ax:
Let's remove port from config, then? @Zariel would you take that in as a PR?

Zariel:
No for the time being I would rather leave it to default to 9042.

Issue #992, opened 2017-10-23
Add ability to collect statistics


nyanshak:
I'd like to add the ability to send stats (to Datadog, statsd, or whatever) through an [xstats.XStater interface ](https://github.com/rs/xstats/blob/c67367528e160e557423e4d87059614d8c31a582/xstats.go#L41).

Right now, we're running several services using gocql connecting to a managed C* instance. This doesn't give us the visibility into certain stats that we'd like, and even if it did, it wouldn't be super-easy to break it down by WHICH service is having problems.

I'd like to add an option to send stats via an XStater interface. Right now, the biggest needs would be incrementing counters whenever connection states are changed (connection opened, connection closed, etc.). Lower priority would be adding stats around things such as latency (TTFB, etc).

@Zariel - I wanted to reach out with this issue to see if this idea is acceptable for the project and if any of the maintainers might be interested in taking this on. I'd be willing to try and help, but I'm not super experienced with the gocql code so I might need some pointers.

Thank you!
:)

nyanshak:
I went ahead and made https://github.com/gocql/gocql/pull/1010 as an example of what I'd like to do. Would love to get some feedback on the general approach as well as some agreement on good stat names.

jacobgreenleaf:
#1023 looks to be a more generic solution to this problem, and we like the general approach, but there are a lot of things that remain unmeasured after that change. Here are the metrics we think through experience will be useful for users of gocql:

**Connection pool instrumentation**. The logic around when the driver decides to start and stop connections in the pool ("fill" in the code) and how the driver chooses a particular connection in the connection pool is complex and needs more observability. We have observed numerous concerning bugs like #915 that cannot be analyzed without these metrics. Specifically:

* TCP Connection dialed (counter, result, timing)
* Protocol negotiated (counter, result, timing)
* Utilization of the in-use streams (gauge)
* Saturation (`IDGenerator.inuseStreams` gauge). In gocql case this pool cannot be saturated and in fact [emits an error](https://github.com/gocql/gocql/blob/58b3c7120a7ef9a077e4d089dd1badcb9ec756a8/internal/streams/streams.go#L75) rather than block for new stream. 

**Internal cache instrumentation**. The driver has a number of internal caches like the prepared statement LRU cache, routing key cache, and keyspace metadata. Specifically:

* Hit rate (counter, gauge)
* Cache size (gauge)
* Evictions (counter)

[...]

PR #1010, opened 2017-10-30
#992: Add ability to collect stats

nyanshak:
WIP, would like to get some feedback and refine the solution + stat names.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -8,6 +8,8 @@ import (
 	"errors"
 	"net"
 	"time"
+
+	"github.com/rs/xstats"
 )
 
 // PoolConfig configures the connection pool used by the driver, it defaults to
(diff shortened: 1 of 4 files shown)


Zariel:
Thanks for starting this, some high level comments about the approach:
* I don't think its a great idea to launch a goroutine for every stat
* We need a real noop stat collector

nyanshak:
Updated the PR with your feedback in mind.

nyanshak:
@Zariel can you give this another look?

Zariel:
Sorry for the delay. Fundamentally I think we need to add some kind of stat exporting to the driver, but I want to be careful to not tie us to specific ways of doing it. Take for example in this case, we can never change any of the metric names because someone might have alerts defined on them. I would rather have a mechanism to export info from the driver and have external code do the monitoring, for example by using #1023

PR #1012, opened 2017-10-31
Remove HostFilter for supplied hosts

wfanzju:
related: https://github.com/gocql/gocql/issues/1011

### DIFF
--- a/session.go
+++ b/session.go
@@ -160,12 +160,6 @@ func (s *Session) init() error {
 	if err != nil {
 		return err
 	}
-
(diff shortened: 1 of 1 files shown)


Zariel:
It looks like adding the check to the supplied host is a bug and can just be removed

wfanzju:
For hosts from system.peers table, we're able to run the HostFilter, do we want to do that?

Zariel:
The host filter will still need applying before adding to the pools after doing the host lookup

PR #1015, opened 2017-11-02
restrict refresh ring to heartBeat

### LINKED ISSUE #928: Deadlock during NewSession
Found on da4a50306e7bb46de0d077d7ac78c297de9aaac4

As part of NewSession, ring information is collected using some queries on system tables. If there is a query error then it leads to deadlock as shown in the stack trace below

[stack.txt](https://github.com/gocql/gocql/files/1110740/stack.txt)

kthommandra:
GetHosts uses queries on control connection. Any query error will
cause reconnection and refresh ring which calls GetHosts and thus
deadlocks.

Also it seems neater to have the ring refresh be done only by heartBeat.

Fixes #928

Verified that all tests that pass continue to pass with the fix.

Refer to issue #928 for details about this PR


kthommandra:
@Zariel - any comments on the PR?

PR #1022, opened 2017-11-20
Connect to multiple hosts behind a DNS name

sethrosenblum:
When using a DNS name as your connection point, this allows the client to attempt to connect to multiple nodes in the cluster when creating the control connection.

We've encountered an issue where our DNS will sometimes return a temporarily offline host first from DNS.  During this time, any client that attempts to bootstrap will return an error as it will only try to connect to that host.  Since we return many hosts in our DNS, the client should attempt to connect to several of them, as if we had passed in a list of IP addresses.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -732,11 +732,11 @@ func (srv *TestServer) session() (*Session, error) {
 }
 
 func (srv *TestServer) host() *HostInfo {
-	host, err := hostInfo(srv.Address, 9042)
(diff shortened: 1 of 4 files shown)


sethrosenblum:
@Zariel not sure if this will cause more problems WRT this doc: https://godoc.org/github.com/gocql/gocql#NewCluster

Zariel:
That doc can be updated as it is no longer correct, internally the driver maintains all hosts by the IP address and will deduplicate them.

Issue #357, opened 2015-04-09
CASSGO-23 Make Session immutable


Zariel:
Currently Session has some fields which can be changed, such as Cons, Tracing, and some others. This requires there to be a `RWMutex` which needs to be acquired when executing a query to copy the options.

I propose we make Session immutable and remove this lock.

0x6e6562:
It's always nice when a lock can be feasibly removed :-)

Zariel:
It would mean some things get moved to the config and become immutable session wide, such as the tracer.

tux21b:
I think it would be nice to have a separation between a `Session` and the `Cluster`. It might make sense to enable tracing (or the default consistency, etc.) only for one part of the application. The initial idea (I am not yet sure how far the current gocql API has diverged) was to have multiple sessions (with different session specific configs) that can use the same underlying cluster. So that each component of a larger application can configure the session differently (for logging / tracing purposes for example). I'm not sure about the popularity of this feature though.

But I agree that it would also be nice to have an immutable Session object. We could archive both by providing methods like `func (s *Session) WithTracing(t Tracer) *Session` that return a new copy of the Session.

Zariel:
Interesting, I've never come across that use case, in general I always assumed that an application would have a single session. Given that you can change tracing on a per query basis Im not sure what use cases there are to do it session wide (maybe set tracing and write the tracing output to a `http.ResposneWriter`?).

Given that #375 essentially made NewCluster a wrapper for `&ClusterConfig{Hosts: }` Im not sure what the current separation between Cluster and Session is used for, as now `NewSession` does all the setup.

0x6e6562:
I wonder how many apps in practice actually use more than one session instance (other than to control tracing in a more differentiated fashion)?

joao-r-reis:
I'm considering adding this to the scope of 2.0.0 but I'm a bit concerned about the impact of the breaking changes. Maybe we can just make these functions create a copy of `Session` like @tux21b mentioned

jameshartig:
I'm concerned about making them return a copy because the only time you call those methods is after the session is created. If it returns a copy then do you need to close the old session? it also means unnecessary connections if you only want 1 session.

Can we instead change `NewSession` to accept a `SessionConfig` that has a `ClusterConfig` inside of it? Then we can add fields to `SessionConfig` for whatever we need to set and remove the setter methods.

joao-r-reis:
[...]

PR #1023, opened 2017-11-23
adding observer to query

JavierZunzunegui:
Introducing `QueryObserver` and adding it to `Query` which gets called every time the query is executed (via `ExecutableQuery.attempt`, modified slightly).

Introduced to allow monitoring of gocql, should be generic enough to support all monitoring libraries.


session.go
```diff
@@ -1584,6 +1616,31 @@ func (t *traceWriter) Trace(traceId []byte) {
...
+}
+
+// TODO - BatchObservation, similar to QueryObservation but adapted to Batch
+
+// QueryObserver is the interface implemented by query observers / stat collectors.
+type QueryObserver interface {
+	// Observe gets called on every query to cassandra, including all queries in an iterator when paging is enabled.
+	// It doesn't get called if there is no query because the session is closed or there are no connections available.
+	// The error reported only shows query errors, i.e. if a SELECT is valid but finds no matches it will be nil.
+	Observe(QueryObservation)
```
Zariel: Call this ObserveQuery instead and change the signature to take a context as the first param

session.go
```diff
@@ -299,6 +300,14 @@ func (s *Session) SetTrace(trace Tracer) {
 	s.mu.Unlock()
 }
 
+// SetQueryObserver sets the default query-level observer for this session. This setting can also
+// be changed on a per-query basis.
+func (s *Session) SetQueryObserver(observer QueryObserver) {
```
Zariel: I eventually want to remove the option to set globals on the session to remove the lock on querying #357, please remove this part. Then the observer only need to live on the config.

session.go
```diff
@@ -720,6 +731,13 @@ func (q *Query) Trace(trace Tracer) *Query {
 	return q
 }
 
+// QueryObserver enables query-level observer on this query.
+// The provided observer will be called every time this query is executed.
+func (q *Query) Observer(observer QueryObserver) *Query {
```
Zariel: Should also add this to Batch

session.go
```diff
@@ -1584,6 +1616,31 @@ func (t *traceWriter) Trace(traceId []byte) {
...
 
+type QueryObservation struct {
+	ctx      context.Context // the query's context. Use it to pass arbitrary data to the collector
+	keyspace string
+	stmt     string
+
+	start time.Time // time immediately before the query was called
+	end   time.Time // time immediately after the query returned
+
+	rows int // the number of rows in the current iter. In multi-scans, rows from previous scans are not counted
```
Zariel: s/multi-scans/paginated queries/

cassandra_test.go
```diff
@@ -184,6 +185,97 @@ func TestTracing(t *testing.T) {
...
 
+func TestObserve(t *testing.T) {
+	session := createSession(t)
+	defer session.Close()
+
+	if err := createTable(session, `CREATE TABLE gocql_test.observe (id int primary key)`); err != nil {
+		t.Fatal("create:", err)
+	}
+
+	var observedErr error
```
Zariel: declare these in a block

session.go
```diff
@@ -770,10 +788,22 @@ func (q *Query) execute(conn *Conn) *Iter {
...
 	// TODO: track latencies per host and things as well instead of just total
+
+	if q.observer != nil {
+		q.observer.Observe(QueryObservation{
+			ctx:      q.context,
+			keyspace: q.session.pool.keyspace,
+			stmt:     q.stmt,
+			start:    start,
+			end:      end,
+			rows:     iter.numRows - iter.pos,
```
Zariel: iter.pos here should always be 0, is this what you intend?

JavierZunzunegui: Would it not be 0 in paginated queries?

session.go
JavierZunzunegui: Part of a slightly larger question - how to observe Batches? I'm not familiar with those, but from what I can see the `QueryObserver` may be called for each statement in the batch (with shared times and error) although I don't know if the rows have any meaning in batch queries.

session.go
JavierZunzunegui: yeah, always 0. Made TestObserve_Pagination to check it

JavierZunzunegui:
@Zariel any further feedback on this?

session.go
Zariel: Batch will have multiple statements but no return values unless BatchCAS is used.

JavierZunzunegui:
@Zariel added observe support for Batches

Zariel:
I would like somehow to have batches come through as a single observation as they resulted from a single query that went to a single host.

session.go
```diff
@@ -1584,6 +1628,32 @@ func (t *traceWriter) Trace(traceId []byte) {
 	}
 }
 
+type ObserveQuery struct {
```
Zariel: I had meant this the other way around, this is fine to be called `ObservedQuery`

session.go
```diff
@@ -1584,6 +1628,32 @@ func (t *traceWriter) Trace(traceId []byte) {
...
+	Err error
+}
+
+// QueryObserver is the interface implemented by query observers / stat collectors.
+type QueryObserver interface {
+	// Observe gets called on every query to cassandra, including all queries in an iterator when paging is enabled.
+	// It also gets called once for each query in a batch.
+	// It doesn't get called if there is no query because the session is closed or there are no connections available.
+	// The error reported only shows query errors, i.e. if a SELECT is valid but finds no matches it will be nil.
+	Observe(context.Context, ObserveQuery)
```
Zariel: But name this method `ObserveQuery`

session.go
```diff
@@ -1445,10 +1476,23 @@ func (b *Batch) WithTimestamp(timestamp int64) *Batch {
...
-func (b *Batch) attempt(d time.Duration) {
+func (b *Batch) attempt(keyspace string, end, start time.Time, iter *Iter) {
 	b.attempts++
-	b.totalLatency += d.Nanoseconds()
+	b.totalLatency += end.Sub(start).Nanoseconds()
 	// TODO: track latencies per host and things as well instead of just total
+
+	if b.observer != nil {
+		for _, entry := range b.Entries {
+			b.observer.Observe(b.context, ObserveQuery{
```
Zariel: I think we should here only get a single observation for a batch

session.go
```diff
@@ -1584,6 +1628,32 @@ func (t *traceWriter) Trace(traceId []byte) {
 	}
 }
 
+type ObserveQuery struct {
+	Keyspace string
+	Stmt     string
```
Zariel: call this `Statement`

Zariel:
Would It make more sense to do something akin to what `net/http` does? https://golang.org/pkg/net/http/httptrace/

zllak:
We've done something very similar on a fork. We are running this in production to export metrics to prometheus. It works very well and it would be awesome to have this integrated into gocql directly !

JavierZunzunegui:
@Zariel the monitoring approach in https://golang.org/pkg/net/http/httptrace/ relies in context. I don't think gocql should emulate that because:
- gocql already uses the non-context approach for tracing, it would be inconsistent to use content for monitoring.
- the context approach is, in my opinion, slightly lower level in that the caller is responsible for setting the monitoring methods in the context **for every request** as by default it is nil. The approach here sets a default value which is much more useful if you want all queries monitored by default and in a similar fashion. Of course you could do the same in the context approach wrapping the gocql session with something else that sets the observer in each given context, but that is extra complexity for no gain

Zariel:
Can you add that the interface is experimental then I’m happy to merge.
Thanks very much

Zariel:
@JavierZunzunegui something to note about tracing in Gocql is that is for tracing on the Cassandra side, which tells cassandra that we want to trace the query and allows us to read the results.

Zariel:
@zllak could you share a bit more about how you use it? Is the current interface sufficient in Gocql now?

zllak:
The interface provided here is what we (almost) exactly did. We use this to create some prometheus probe, to graph latencies of calls, count the number of select/update/delete (regexp on the statement tho :/) per table/keyspace. I'll update on our side to use this new interface and see, but it looks very good !

zllak:
@Zariel the only difference I can see is that it only report network errors, not "logical" errors like Notfound, or context.Canceled. We use those one. I'll try it tomorrow.

PR #1024, opened 2017-11-23
Introduce scan reporter

JavierZunzunegui:
work in progress. See https://github.com/gocql/gocql/pull/1023 first


Zariel:
As a comment, I would rather not have 2 ways of observing queries. Instead can we expose the missing info the layer below Scan?

JavierZunzunegui:
https://github.com/gocql/gocql/pull/1023 can be used to provide this functionality much more efficiently

Issue #1033, opened 2017-12-20
Murmur3H1 implementation doesn't match C* implementation


zviadm:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?
3.11.1

### What version of Gocql are you using?
master

### What did you do?
Strong suspicion TokenAware routing doesn't work because Murmur3H1 is incorrect and doesn't match C* implementation of murmur3
```
 10     // From Python Cassandra driver:                                                                
 11     // In [6]: murmur3.murmur3(codecs.decode("00104327529fb645dd00b883ec39ae448bb800000400066a6b00", "hex"))
 12     // Out[6]: -9223371632693506265                                                                 
 13     s, err := hex.DecodeString("00104327529fb645dd00b883ec39ae448bb800000400066a6b00")              
 14     if err != nil {                                                                                 
 15         t.Fatal("unexpected err: ", err)                                                            
 16     }                                                                                               
 17     v := int64(Murmur3H1(s))                                                                        
 18     if v != -9223371632693506265 {                                                                  
 19         t.Fatal("bad hash: ", v)                                                                    
 20     }                                                                                               
```

### What did you expect to see?
expected no error

### What did you see instead?
```
	murmur_test.go:62: bad hash:  8117980653506862255
```

---

If you are having connectivy related issues please share the following additional information

### Describe your Cassandra cluster
please provide the following information

- output of `nodetool status`
- output of `SELECT peer, rpc_address FROM system.peers`
- rebuild your application with the `gocql_debug` tag and post the output

zviadm:
The Murmur3 implementation in C* and in Python/Java clients is based on int64-s internally. This actually results into a different value, you can't just cast uint64 to int64 to get the same value. 

Python code here: https://github.com/datastax/python-driver/blob/master/cassandra/cmurmur3.c


Fyi, i think this is what the correct implementation should look like:
```
func Murmur3H1(data []byte) int64 {
        length := len(data)

        var h1, h2, k1, k2 int64

        const (
                c1 int64 = -8663945395140668459 // 0x87c37b91114253d5
                c2 int64 = 0x4cf5ad432745937f
        )

        // body
        nBlocks := length / 16
        for i := 0; i < nBlocks; i++ {
                block := (*[2]int64)(unsafe.Pointer(&data[i*16]))

                k1 = block[0]
                k2 = block[1]

                k1 *= c1
                k1 = (k1 << 31) | (k1 >> 33) // ROTL64(k1, 31)
[...]

PR #1034, opened 2017-12-20
Fix murmur3 library for murmur3partitioner

zviadm:
* Cassandra murmur3 needs to use int64s
* Removed use of unsafe package, because perf benefits are questionable,
  and creates unnecessary compatibility issues.


zviadm:
Fixes #1033

Zariel:
Thanks have fixed via #1035

zviadm:
Test case might still be helpful, just to make sure you are testing all 16 versions of the switch statement, and also to make sure you are testing with inputs that have bytes larger than 128

Zariel:
Agreed, would you like to submit he test cases by themselves from this?

PR #1036, opened 2017-12-22
Round-robin host selection for token aware policy based

hackworks:
The murmur3 partitioner in token aware policy does not take advantage of read replicas. It hits the same Cassandra node for a given token/partition. This PR attempts to pick other nodes having the same partition in a round-robin manner.

### DIFF
--- a/token.go
+++ b/token.go
@@ -12,6 +12,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync/atomic"
 
 	"github.com/gocql/gocql/internal/murmur"
 )
@@ -58,7 +59,7 @@ func (m murmur3Token) Less(token token) bool {
 
 // order preserving partitioner and token
 type orderedPartitioner struct{}
-type orderedToken []byte
+type orderedToken string
 
 func (p orderedPartitioner) Name() string {
 	return "OrderedPartitioner"
@@ -70,15 +71,15 @@ func (p orderedPartitioner) Hash(partitionKey []byte) token {
 }
 
 func (p orderedPartitioner) ParseString(str string) token {
-	return orderedToken([]byte(str))
+	return orderedToken(str)
 }
 
 func (o orderedToken) String() string {
-	return string([]byte(o))
+	return string(o)
 }
 
 func (o orderedToken) Less(token token) bool {
-	return -1 == bytes.Compare(o, token.(orderedToken))
+	return o < token.(orderedToken)
 }
 
 // random partitioner and token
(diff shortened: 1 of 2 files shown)


Zariel:
Thanks. I think one of the outstanding TODO in the host selection is that we should support replicas. Im not sure I follow how this solves the problem and is a bit non obvious and specific to the murmur hash partitioner. I think a better solution would be to some how be explicit about getting all the replicas for a key given a partitioner / token ring.

This could also do with tests to verify it works.

hackworks:
I have updated the PR to support other partitioners. The lookup map of token->hosts serves as a fast path for any of the partitioners. If it fails to find hosts for a given token, it now falls back into looking in the ordered collection of tokens. Once you get the index (based on existing logic), we pick the hosts from our lookup map and use the same round-robin logic.

token.go
```diff
@@ -58,7 +59,7 @@ func (m murmur3Token) Less(token token) bool {
 
 // order preserving partitioner and token
 type orderedPartitioner struct{}
-type orderedToken []byte
+type orderedToken string
```
hackworks: This lets me use orderedToken as a key in the look up map. Required to support token types from different partitioners in a common lookup map (in token ring)

Zariel: it's not necersarry to do this as doing a lookup via `myMap[string(someBytes)]` can be optimised by the compiler to avoid allocations

hackworks: I had to change it to string so that I can use it as a key in a map. If you use a slice of byte, you get an error: hash of unhashable type
Ex: https://play.golang.org/p/EEq2KABDHiV

Zariel:
I had a look into this and it appears to not be as simple as this. Cassandra  will not assign the same token to multiple nodes, instead it walks around the ring to find the next node until it meets the replication factor of the placement strategy. In the case of SimpleStrategy this is just the next node until replication_factor is met, but for NetworkTopologyStrategy it walks around racks in a DC until it has the RF needed.

This requires a significant amount of rewriting in the driver for the policies as they now require a session and to be notified when the schema changes. I have a patch pending that requires more testing but I think I should be able to implement this shortly.

hackworks:
If Cassandra does not assign the same token to different nodes during storing replicas, how can it access the read replicas when a query comes in? I would love to read the logic of you can point me to some documents or the code. Thank you for reviewing the change.

Zariel:
Implemented via #1039

Zariel:
@hackworks thanks for attempting to implement this. Each host has a token (or in the case of vnodes many but lets assume single tokens), these tokens have a defined min/max value per partitioner and are sortable. When we sort the list of tokens to host mapping by token we can treat it as a ring by wrapping when we index past the end of the list. For a token `i` if the replication factor is 2 then the next replica is the node which owns the token at `i+1 % len(tokens)`, for the simple strategy. For network topology it is the next node in a different rack within the same DC.

Issue #1030, opened 2017-12-08
Column types returned by KeyspaceMetadata are always of Type: (gocql.NativeType)


enjoylife:
I was trying to leverage the `UDTTypeInfo` for each of my user defined types, but to my surprise all columns types come across as the generic `NativeType`.
Is this expected? Anyone run across or solved something similar?

### What version of Cassandra are you using?
3.4.4

### What version of Gocql are you using?
master

### What did you do?
```
md, err := s.KeyspaceMetadata(keySpace)
	if err != nil {
		fmt.Errorf("keyspace metadata error", err)
		return
	}


	for _, table := range md.Tables {
		for _, col := range table.Columns {
			spew.Dump(col)
		}
	}
```

### What did you expect to see?
Expected to see column types vary based on their actual types

### What did you see instead?
`Type: (gocql.NativeType)` across all columns

carebdayrvis:
From looking through the source, each returned column type is wrapped in NativeType. That struct has a Type method to get the underlying type.

Zariel:
I've got a fix for this just need to write some tests to cover it.

PR #1052, opened 2018-01-23
metadata: fix all types being NativeType

Zariel:
Parse the correct TypeInfo from the cassandra string in the db. Fixing
representation to recursivly parse the nested types.


Zariel:
fixes #1030

PR #1060, opened 2018-02-14
Document unset value

vrischmann:
I tried to document `UnsetValue` based on https://datastax.github.io/python-driver/api/cassandra/query.html#cassandra.query.UNSET_VALUE and https://docs.datastax.com/en/developer/nodejs-driver/3.2/features/datatypes/nulls/


Zariel:
I really wish this was a function call not a var :(

Thanks!

PR #1062, opened 2018-02-15
send frame should not block

zhixinwen:
Issue: https://github.com/gocql/gocql/issues/896
Send the frame asynchronously, so a request can timeout properly when network is slow

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -148,7 +148,7 @@ type Conn struct {
 
 	timeouts int64
 
-	frameWriteArgs chan *frameWriteArgs
+	frameWriteArgChan chan *frameWriteArg
 }
 
 // Connect establishes a connection to a Cassandra node.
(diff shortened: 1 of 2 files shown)


zhixinwen:
Tested manually, the `context.Deadline()` would be respected during network failure.

conn.go
```diff
@@ -215,6 +218,25 @@ func (s *Session) dial(ip net.IP, port int, cfg *ConnConfig, errorHandler ConnEr
...
+			select {
+			case args := <- c.frameWriteArgChan:
+				if err := args.req.writeFrame(args.framer, args.stream); err != nil{
+					// I think this is the correct thing to do, im not entirely sure. It is not
+					// ideal as readers might still get some data, but they probably wont.
+					// Here we need to be careful as the stream is not available and if all
+					// writes just timeout or fail then the pool might use this connection to
+					// send a frame on, with all the streams used up and not returned.
+					c.closeWithError(err)
+				}
```
Zariel: should exit out of this loop on error

conn.go
```diff
@@ -215,6 +218,25 @@ func (s *Session) dial(ip net.IP, port int, cfg *ConnConfig, errorHandler ConnEr
 
 	frameTicker := make(chan struct{}, 1)
 	startupErr := make(chan error)
+	// need to call before we write to net.Conn
+	go func() {
```
Zariel: pull this into it's own function

conn.go
```diff
@@ -641,34 +655,70 @@ func (c *Conn) exec(ctx context.Context, req frameWriter, tracer Tracer) (*frame
...
+	case <-ctxDone:
+		close(call.timeout)
+		return ctx.Err()
+	case <-c.quit:
+		return ErrConnectionClosed
 	}
+}
 
-	return framer, nil
+func (c *Conn) resetTimeout(call *callReq) <-chan time.Time {
```
Zariel: it would make more sense to invert this and have it have the signature like

```go
func (c *callReq) resetTimeout(timeout time.Duration) <-chan time.Time
```

conn.go
```diff
@@ -594,39 +616,31 @@ func (c *Conn) exec(ctx context.Context, req frameWriter, tracer Tracer) (*frame
...
+
+	if v := framer.header.version.version(); v != c.version {
+		return nil, NewErrProtocol("unexpected protocol version in response: got %d expected %d", v, c.version)
 	}
 
+	return framer, nil
+}
+
+func (c *Conn) getResp(ctx context.Context, call *callReq) error {
+	timeoutCh := c.resetTimeout(call)
```
Zariel: I think it would make sense to have the current timeout cover both send + recv and later we can add different timeouts for send and recv.

conn.go
```diff
@@ -641,34 +655,70 @@ func (c *Conn) exec(ctx context.Context, req frameWriter, tracer Tracer) (*frame
...
+			}
+		}
+
+		call.timer.Reset(c.timeout)
+		timeoutCh = call.timer.C
+	}
+	return timeoutCh
+}
+
+type frameWriteArg struct {
```
Zariel: wouldn't it be easier to add the `frameWriter` to `callReq`?

conn.go
zhixinwen: done, although it should exit `case <-c.quit` even if it is not explicitly returned in the loop.

conn.go
zhixinwen: good point

zhixinwen:
Not sure why the build fails and whether it is due to the change

zhixinwen:
Retrigger the build

zhixinwen:
looks are tests are flappy, the failed tests are different in the two runs

Zariel:
Hm, I've not seen these tests flake recently but I agree they seem flaky. I'll do some checks later in the week and try to fix them.

PR #1069, opened 2018-02-24
fix memory leak issue 

zhixinwen:


zhixinwen:
https://github.com/gocql/gocql/issues/1068

conn.go
```diff
@@ -710,10 +710,12 @@ func (c *Conn) sendFrame(ctx context.Context, call *callReq, timeoutCh <-chan ti
 	case c.frameWriteArgChan <- call:
 		return nil
 	case <-timeoutCh:
+		c.releaseStream(call.streamID)
```
Zariel: I dont think it is safe to reuse this stream if we timeout, we may still receive a response on this stream which will go to the wrong caller.

zhixinwen: It is called in `sendFrame`, so the request is not really sent out. In this case, we would never receive a response and the memory needs to be recycled.

Issue #1079, opened 2018-03-19
New Feature: Reconnection Policy and Conviction Policy


changliuau:
Similar to Python/Java drivers, it will be good to have the following two configurable policies when connecting to a cluster.

Reconnection policy - determines how to reconnect to hosts before marking them as 'down'. 
Conviction policy - determines whether to mark a host as 'down' based on the returned error. 

For both of them, open an interface for users to implement their own policies.

changliuau:
PR: 
https://github.com/gocql/gocql/pull/1077

PR #1081, opened 2018-03-20
add conviction policy interface

changliuau:
#1079 

Conviction policy: 

decide when to mark the host as 'down' based on the error. Default:  `SimpleConvictionPolicy` - mark host as 'down' on any error.

Opened interface in policies.go for user to implement their own policies.

### DIFF
--- a/policies.go
+++ b/policies.go
@@ -706,3 +706,20 @@ func (d *dcAwareRR) Pick(q ExecutableQuery) NextHost {
 		return (*selectedHost)(host)
 	}
 }
+
+type ConvictionPolicy interface {
+	// Implementations should return `true` if the host should be convicted, `false` otherwise.
+	AddFailure(error error) bool
+	//Implementations should clear out any convictions or state regarding the host.
+	Reset()
+}
+
+//return true on any error
+type SimpleConvictionPolicy struct {
+}
+
+func (e *SimpleConvictionPolicy) AddFailure(error error) bool {
+	return true
+}
+
+func (e *SimpleConvictionPolicy) Reset() {}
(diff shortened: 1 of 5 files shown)


policies.go
```diff
@@ -706,3 +706,20 @@ func (d *dcAwareRR) Pick(q ExecutableQuery) NextHost {
 		return (*selectedHost)(host)
 	}
 }
+
+type ConvictionPolicy interface {
+	// Implementations should return `true` if the host should be convicted, `false` otherwise.
+	AddFailure(error error) bool
```
Zariel: Surely this should take the host info that caused the error?

policies.go
```diff
@@ -791,6 +791,22 @@ func (d *dcAwareRR) Pick(q ExecutableQuery) NextHost {
 	}
 }
 
+type ConvictionPolicy interface {
+	// Implementations should return `true` if the host should be convicted, `false` otherwise.
+	AddFailure(error error, host *HostInfo) bool
+	//Implementations should clear out any convictions or state regarding the host.
+	Reset()
```
Zariel: Should this not also take a host given the comment?

policies.go
```diff
@@ -791,6 +791,22 @@ func (d *dcAwareRR) Pick(q ExecutableQuery) NextHost {
...
+}
+
+//return true on any error
```
Zariel: Improve this comment, something like
`SimpleConvictionPolicy implements a ConvictionPolicy which convicts all hosts regardless of error`

policies.go
```diff
@@ -791,6 +791,22 @@ func (d *dcAwareRR) Pick(q ExecutableQuery) NextHost {
 	}
 }
 
+type ConvictionPolicy interface {
```
Zariel: Also add a comment explaining what this interface is for.

changliuau:
@Zariel  thanks - fixed the comments and Reset function as suggested

Issue #1070, opened 2018-03-01
Support of downgrading consistency retry policy


changliuau:
Hi there - I am new to the community and trying to contribute to the project. 

We are looking into adding 'downgrading consistency policy' as a new `RetryPolicy`. It is already implemented in the Python driver and Java driver for Cassandra. The core algorithm in Python driver's implementation looks like this: (http://datastax.github.io/python-driver/api/cassandra/policies.html#cassandra.policies.DowngradingConsistencyRetryPolicy)

```
    def _pick_consistency(self, num_responses):
        if num_responses >= 3:
            return self.RETRY, ConsistencyLevel.THREE
        elif num_responses >= 2:
            return self.RETRY, ConsistencyLevel.TWO
        elif num_responses >= 1:
            return self.RETRY, ConsistencyLevel.ONE
        else:
            return self.RETHROW, None
```

although they do slightly different things for `read_timeout`, `write_timeout` and `unavailable`, essentially it will get the number of replicas returned from the failed response, and adjust the consistency level for the retry based on this number.

Another way of doing it is to based on the current consistency level of the query, e.g., if the consistency level is `QUROM` and the query failed, try `LOCAL QUROM` in the next retry, and then `THREE` etc. The code change could be a lot smaller this way as there is already a function `GetConsistency()` in `RetryableQuery` interface.

So the questions I have are:

1. In the first way (ie the one similar to Python/Java driver), we will need to get the number of replicas returned from a query and therefore analyse the query failure response. Is there a easy way of doing that in `gocql`? Looks like it is not used in `policies.go` at this point. 
2. Which of the 2 ways do you guys think is better?

Note: it's a bit controversial whether the driver should be handling something like this. In Java driver it's deprecated and going to be removed in the next major version -- because people think this is something that should be handled by the application rather than the driver. The commit has just been made yesterday.  https://github.com/datastax/java-driver/commit/f5d44d3fea093a5c28d59ba0f587e9044e161813

Thanks in advance for any input.

Regards,
Chang

kgreav:
@robusto This always made more sense to me. I'm not surprised the java driver has deprecated it, because frankly the implementation is a bit useless. Especially when you consider cases of RF=5. I'd be in favour of the user being able to select what consistency they downgrade to, it also makes much more sense in the case of multiple DC's. 

[...]

PR #1082, opened 2018-03-23
DowngradingConsistencyPolicy and enhancements to retry policy behaviour

changliuau:
PR for #1070 

Tested against 3-node cluster with the new query execution and returned error types.

DowngradingConsistencyPolicy takes a slice of consistency levels and try them in order if the query fails. The rest of the retry logics are similar to the Python driver with minor tweaks:

http://datastax.github.io/python-driver/api/cassandra/policies.html#cassandra.policies.DowngradingConsistencyRetryPolicy

Looking for feedbacks.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -964,9 +964,18 @@ func (c *Conn) executeQuery(qry *Query) *Iter {
 		}
 
 		return &Iter{err: x, framer: framer}
+
+		// for DowngradeConsistencyPolicy
+	case *RequestErrUnavailable:
+		Logger.Printf("execute error, %s", qry)
+		return &Iter{err: resp.(*RequestErrUnavailable), framer: framer}
+
 	case error:
 		// DEBUG TO REMOVE
-		Logger.Printf("execute error, %s", qry)
+		//Logger.Printf("execute error, %s", qry)
+		err := fmt.Sprintf("%T", resp)
+		//Logger.Print(resp.(*RequestErrUnavailable).String())
+		Logger.Print(err)
 
 		return &Iter{err: x, framer: framer}
 	default:
--- a/conn.go
+++ b/conn.go
@@ -965,6 +965,9 @@ func (c *Conn) executeQuery(qry *Query) *Iter {
 
 		return &Iter{err: x, framer: framer}
 	case error:
+		// DEBUG TO REMOVE
+		Logger.Printf("execute error, %s", qry)
+
 		return &Iter{err: x, framer: framer}
 	default:
 		return &Iter{
--- a/conn.go
+++ b/conn.go
@@ -965,18 +965,15 @@ func (c *Conn) executeQuery(qry *Query) *Iter {
 
 		return &Iter{err: x, framer: framer}
 
-		// for DowngradeConsistencyPolicy
+	// for DowngradeConsistencyPolicy
 	case *RequestErrUnavailable:
-		Logger.Printf("execute error, %s", qry)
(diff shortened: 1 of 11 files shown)


Zariel:
please run gofmt on the code, this change will also require tests to verify each of the cases of retry work

changliuau:
@Zariel 

Thanks for your comment! 

Are you suggesting adding test cases for `query_executor` or `policies`? If it's `query_executor` I will probably add a new file just to test the `executeQuery()` function in there.

changliuau:
added unit tests in policies_test and ran `gofmt` on the code

conn.go
```diff
@@ -964,6 +964,15 @@ func (c *Conn) executeQuery(qry *Query) *Iter {
 		}
 
 		return &Iter{err: x, framer: framer}
+
+	// for DowngradeConsistencyPolicy
+	case *RequestErrUnavailable:
```
Zariel: I dont think we need these cases. The err interface will still hold the underlying value as `*RequestErrUnavailable`

policies.go
```diff
@@ -194,6 +209,69 @@ func getExponentialTime(min time.Duration, max time.Duration, attempts int) time
...
 }
 
+func (e *ExponentialBackoffRetryPolicy) GetRetryType(err error) RetryType {
+	return RetryNextHost
+}
+
+// DowngradingConsistencyRetryPolicy: Next retry will be with the next consistency level
+// provided in the slice
+/*
+* On a read timeout: the operation is retried with the next provided consistency
```
Zariel: refromat this comment to only use // style comments

policies.go
```diff
@@ -194,6 +209,69 @@ func getExponentialTime(min time.Duration, max time.Duration, attempts int) time
...
+	ConsistencyLevelsToTry []Consistency
+}
+
+func (d *DowngradingConsistencyRetryPolicy) Attempt(q RetryableQuery) bool {
+	currentAttempt := q.Attempts()
+
+	if currentAttempt > len(d.ConsistencyLevelsToTry) {
+		return false
+	}
+	if currentAttempt > 0 {
```
Zariel: make this an else if of the above condition

policies.go
```diff
@@ -194,6 +209,69 @@ func getExponentialTime(min time.Duration, max time.Duration, attempts int) time
...
+				d.ConsistencyLevelsToTry[currentAttempt-1])
+		}
+	}
+	return true
+}
+
+func (d *DowngradingConsistencyRetryPolicy) GetRetryType(err error) RetryType {
+	switch err.(type) {
+	case *RequestErrUnavailable:
+		if err.(*RequestErrUnavailable).Alive > 0 {
```
Zariel: this can be written like

```go
switch v := err.(type) {
case *RequestErrUnavailable:
    // v is now *RequestErrUnavailable
```

See https://golang.org/doc/effective_go.html#type_switch

query_executor.go
```diff
@@ -50,17 +50,41 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
...
 		// Update host
 		hostResponse.Mark(iter.err)
 
+		if rt == nil {
+			break
+		}
+
+		switch rt.GetRetryType(iter.err) {
+		case Retry:
+			for rt.Attempt(qry) {
```
Zariel: I would rather refactor the outerloop so that it takes care of this and there isn't as much branching.

Issue #1087, opened 2018-04-11
Implement Query Idempotancy


alourie:
In order to be able to approach the implementation of the #1083, we need to first implement [Query Idempotancy](https://docs.datastax.com/en/developer/java-driver/3.4/manual/idempotence/), so that the driver will know when it is allowed to perform Speculative queries.

This ticket is the placeholder for this work.

alourie:
I am working on this ticket and will update when there's any progress.

alourie:
@Zariel Just in case you're looking into this, here are my findings:

It is possible to add a field on Cluster and Query that would keep "idempotence" value. This would allow client applications to manually set the value during either query creation or cluster config.

In addition, we could try and write an automatic query parser that would try to detect whether the query is idempotent or not; unfortunately, that would be quite a hard of a task of parsing queries strings, so maybe try it out with super simple use-cases

Alternatively, gocqlx is a good candidate for utilising this work. The Java driver, for example, implements some heuristics in Query Builder, which gocql doesn't have, but gocqlx does. On that note, I was wondering whether the inclusion of gocqlx's Builders in the gocql code would make any sense. If not, and we only implement manual setting/simplistic automated parser here, and then make it work for #1083, I will be able to expand on this work in gocqlx and add the Java-like heuristics there.

Zariel:
I think the driver should just have a flag for idempotency then users can either set it or helper funcs can set it, we would need a query parser or builder to do it ourselves.

alourie:
@Zariel Thanks for the response. I'll then start with just having a flag that users can set. I'll create a separate issue for building query parser, as it's a bit more complex work.

I'll then proceed to work on #1083 to start using the flag itself for Speculative queries and then get back to helper parser. We might benefit from doing both the parser _and_ the Builder, but that's yet to be determined :-)

alourie:
This is already implemented and merged. Closing.

Issue #1083, opened 2018-03-28
Speculative query execution


rdelval:
Sorry if this has been discussed before, I looked through the code, previous issues, and PRs but didn't find anything related.

We're looking to leverage speculative query execution (https://docs.datastax.com/en/developer/java-driver/3.4/manual/speculative_execution/)

Is there a way of achieving the same behavior with with gocql?

alourie:
So to be able to implement this feature, we first need to implement [Query Idempotancy](https://docs.datastax.com/en/developer/java-driver/3.4/manual/idempotence/). I've created a new ticket for it at #1087 and started working on this feature.

Issue #1088, opened 2018-04-11
TestConnBlocked flake


Zariel:
```
--- FAIL: TestConnClosedBlocked (0.00s)
	conn_test.go:593: read tcp 127.0.0.1:36530->127.0.0.1:37878: use of closed network connection
	conn_test.go:614: expected to get use of closed networking connection error got: gocql: connection closed waiting for response
```

alourie:
@Zariel  yea, I've seen it few times by now. I can have a look at that.

martin-sucha:
TestConnClosedBlocked was removed in 271c061c7f16702ca550b0e9721320b1c92e00cb

PR #1090, opened 2018-04-12
Query idempotence 1087

alourie:
Implements #1087 

This work allows a client to set an Idempotence flag on a Cluster or a Query. Additionally, when building Batches, it's now would be possible to use Query{} as the first argument in batch.Query() and batch.Bind(), and not just a string. This would allow using query.GetIdempotence() call to help BatchEntry to figure out the idempotence of an entry in a batch. At the moment, this is not used anywhere else in the code; following this the #1083 would be somewhat possible.

Few unittest were added to test the new additions.

Also updated travis configuration to use a separate script to install dependencies and solve go fork issue by renaming the checked-out folder from <user>/gocql to gocql/gocql to allow the tests to run for other developers on TravisCI.

All unit tests/integration tests are currently passing (https://travis-ci.org/alourie/gocql/builds/365645537)

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -161,6 +161,16 @@ func NewCluster(hosts ...string) *ClusterConfig {
 	return cfg
 }
 
+
+// Manage the default idempotence values
+func (cfg *ClusterConfig) IsIdempotent() bool {
+	return cfg.defaultIdempotence
+}
+
+func (cfg *ClusterConfig) SetIdempotence(value bool) {
+	cfg.defaultIdempotence = value
+}
+
 // CreateSession initializes the cluster based on this config and returns a
 // session object that can be used to interact with the database.
 func (cfg *ClusterConfig) CreateSession() (*Session, error) {
--- a/cluster.go
+++ b/cluster.go
@@ -126,6 +126,9 @@ type ClusterConfig struct {
 
 	// internal config for testing
 	disableControlConn bool
+
+	// Default idempotence for queries
+	defaultIdempotence bool
 }
 
 // NewCluster generates a new config for the default cluster implementation.
@@ -153,6 +156,7 @@ func NewCluster(hosts ...string) *ClusterConfig {
 		MaxWaitSchemaAgreement: 60 * time.Second,
 		ReconnectInterval:      60 * time.Second,
 		ReconnectionPolicy:     &ConstantReconnectionPolicy{MaxRetries: 3, Interval: 1 * time.Second},
+		defaultIdempotence:     false,
 	}
 	return cfg
 }
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -2,9 +2,9 @@ package gocql
 
 import (
 	"net"
+	"reflect"
 	"testing"
 	"time"
-	"reflect"
 )
 
 func TestNewCluster_Defaults(t *testing.T) {
--- a/common_test.go
+++ b/common_test.go
@@ -182,6 +182,12 @@ func assertTrue(t *testing.T, description string, value bool) {
 	}
 }
 
+func assertFalse(t *testing.T, description string, value bool) {
+	if value {
+		t.Errorf("expected %s to be true", description)
+	}
+}
+
 func assertEqual(t *testing.T, description string, expected, actual interface{}) {
 	if expected != actual {
 		t.Errorf("expected %s to be (%+v) but was (%+v) instead", description, expected, actual)
--- a/common_test.go
+++ b/common_test.go
@@ -182,12 +182,6 @@ func assertTrue(t *testing.T, description string, value bool) {
 	}
 }
 
-func assertFalse(t *testing.T, description string, value bool) {
-	if value {
-		t.Errorf("expected %s to be true", description)
-	}
-}
-
 func assertEqual(t *testing.T, description string, expected, actual interface{}) {
 	if expected != actual {
 		t.Errorf("expected %s to be (%+v) but was (%+v) instead", description, expected, actual)
--- a/session.go
+++ b/session.go
@@ -673,6 +673,7 @@ type Query struct {
 	defaultTimestampValue int64
 	disableSkipMetadata   bool
 	context               context.Context
+	idempotent            bool
 
 	disableAutoPage bool
 }
@@ -1070,6 +1071,14 @@ func (q *Query) reset() {
 	q.disableSkipMetadata = false
 	q.disableAutoPage = false
 	q.context = nil
+}
+
+func (q *Query) IsIdempotent() bool {
+	return q.idempotent
+}
+
+func (q *Query) SetIdempotence(value bool) {
+	q.idempotent = value
 }
 
 // Iter represents an iterator that can be used to iterate over all rows that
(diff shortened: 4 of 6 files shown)


alourie:
There are 2 failed tests, one is #1088, not clear why the second one failed, all previous runs were ok.

session.go
```diff
@@ -1422,16 +1439,34 @@ func (b *Batch) GetConsistency() Consistency {
 	return b.Cons
 }
 
-// Query adds the query to the batch operation
-func (b *Batch) Query(stmt string, args ...interface{}) {
-	b.Entries = append(b.Entries, BatchEntry{Stmt: stmt, Args: args})
+// Query adds the query to the batch operation, supports simple string or Query
+func (b *Batch) Query(query interface{}, args ...interface{}) error {
```
Zariel: Please dont change this API

session.go
```diff
@@ -1422,16 +1439,34 @@ func (b *Batch) GetConsistency() Consistency {
...
+		return fmt.Errorf("only 'string' and 'Query' types are accepted as the first argument to batch.Query()")
+	}
 }
 
 // Bind adds the query to the batch operation and correlates it with a binding callback
 // that will be invoked when the batch is executed. The binding callback allows the application
 // to define which query argument values will be marshalled as part of the batch execution.
-func (b *Batch) Bind(stmt string, bind func(q *QueryInfo) ([]interface{}, error)) {
-	b.Entries = append(b.Entries, BatchEntry{Stmt: stmt, binding: bind})
+func (b *Batch) Bind(query interface{}, bind func(q *QueryInfo) ([]interface{}, error)) error {
```
Zariel: Likewise please dont change this API

session.go
```diff
@@ -1064,6 +1065,22 @@ func (q *Query) reset() {
 	q.context = nil
 }
 
+func (q *Query) GetIdempotence() bool {
```
Zariel: rename this is `IsIdempotent`

session.go
```diff
@@ -1520,6 +1555,16 @@ func (b *Batch) GetRoutingKey() ([]byte, error) {
 	return nil, nil
 }
 
+func (b *Batch) isBatchIdempotent() bool {
```
Zariel: Shouldn't a batch be idempotent or not as a whole?

cluster.go
```diff
@@ -126,6 +126,9 @@ type ClusterConfig struct {
 
 	// internal config for testing
 	disableControlConn bool
+
+	// Default idempotence for queries
+	DefaultIdempotence bool
```
Zariel: I dont see why having a global default is useful?

cluster_test.go
```diff
@@ -54,3 +54,11 @@ func TestClusterConfig_translateAddressAndPort_Success(t *testing.T) {
 	assertTrue(t, "translated address", net.ParseIP("10.10.10.10").Equal(newAddr))
 	assertEqual(t, "translated port", 5432, newPort)
 }
+
+func TestClusterConfig_testIdempotancy(t *testing.T) {
+	cfg := NewCluster()
+	assertFalse(t, "cluster config default idempotancy", cfg.DefaultIdempotence)
+	cfg.DefaultIdempotence = true
+	assertTrue(t, "cluster config changed default idempotancy", cfg.DefaultIdempotence)
```
Zariel: This test is really just checking that setting a field to true works and nothing else

common_test.go
```diff
@@ -182,6 +182,12 @@ func assertTrue(t *testing.T, description string, value bool) {
 	}
 }
 
+func assertFalse(t *testing.T, description string, value bool) {
```
Zariel: This isn't necessary and hides the location of the error in the tests, unless there is a lot of repetition to warrant a function

helpers.go
```diff
@@ -19,6 +19,12 @@ type RowData struct {
 	Values  []interface{}
 }
 
+// Idempotence. It can be either true or false, and set or not set, hence bool won't do
+type Idempotence struct {
```
Zariel: Why do we care if it is set or not? If we default to false which should be a safe value then why does it matter if a user set it explicitly to false?

session.go
```diff
@@ -673,6 +673,7 @@ type Query struct {
 	defaultTimestampValue int64
 	disableSkipMetadata   bool
 	context               context.Context
+	idempotence           Idempotence
```
Zariel: as above this should just be a boolean

install_test_deps.sh
```diff
@@ -0,0 +1,16 @@
...
+git clone https://github.com/pcmanus/ccm.git
+pushd ccm
+./setup.py install --user
+popd
+
+if [ "$1" != "gocql/gocql" ]; then
+    USER=$(echo $1 | cut -f1 -d'/')
+    cd ../..
+    mv ${USER} gocql
+fi
```
Zariel: This change is untreated please remove and submit separately

session.go
```diff
@@ -1064,6 +1065,22 @@ func (q *Query) reset() {
...
+	} else {
+		return q.session.cfg.DefaultIdempotence
+	}
+}
+
+func (q *Query) SetIdempotence(value bool) {
+	q.idempotence = Idempotence{set: true, value: value}
+}
+
+func (q *Query) detectIdempotence() {
```
Zariel: this doesn't do anything please remove

helpers.go
alourie: @Zariel  it's a very good question. Basically, I copied this behaviour from the Java driver which is as follows (and bool in Java can be either true/false/null):

1) If the value is either true or false, use it in the following code (in most cases in speculative queries)
2) If the value is null (that is not set), fetch the default for the cluster (which can be either true or false).

The issue we have here is that if we keep it as bool, we have no way to "initialise" it to the default value for the cluster. If a client uses Query{stmt: "something"}, idempotence will be set to false, and we have no way to say whether it's been set this way by the client or is it a default, especially as there's no action in Query value to sync with cluster's default.

Of course we can do this in the following manner: for the first release we only create the flag on the query, allow cluents to set it and default to false. A following release would add a "constructor" call for query (something like "NewQuery" ), potentially chanding/breaking the existing API, which will then allow using a cluster-wide default by querying it and setting the Query value accordingly. Then the next following release would bring "breaking" changes to Batch.Query() and Batch.Bind() APIs allowing to also use idempotence in the Batch statements. Would that make any sense? If so, I'll simplify this one and create separate tickets for the follow-ups.

install_test_deps.sh
alourie: will do

session.go
alourie: yes, of course.  That's exactly what this call is checking. But at the moment there's no use for it, so will remove with other Batch api changes.

common_test.go
alourie: It's just a simple counterpart for assertFalse for asserting a false value. Of course, it's possible to use `assert.Equal(t, description, value, false)` instead, but I feel that assertFalse() better expresses the intent.

I can remove it of course.

alourie:
Sorry for the force-push. My initial branch got broken. But all comments were taken care of.

Zariel:
merged in via a7b9f75f1028a7484497d8d3d9cc674d3f238259 thanks for the contribution, I wanted to make some slight tweaks to the API before merging.

PR #1092, opened 2018-04-15
marshaler: support encoding.TextMarshaler

nightlyone:
quite a few users store semantically rich data in cassandra. 

In order to avoid making all these data structures cassandra aware,
support the encoding.TextMarshaler and encoding.TextUnmarshaler interfaces
from the Go standard library.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -6,6 +6,7 @@ package gocql
 
 import (
 	"bytes"
+	"encoding"
 	"encoding/binary"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


nightlyone:
Note: I could provide support for converting between TypeBlob and encoding.BinaryMarshaler too, but wasn't sure whether Cassandra would need hex-encoding of those.

Zariel:
I'm not sure this is a great idea and could potentially lead to surprising results if users don't understand this happens. It's not a huge deal for uses which have a type which implements `Textmarshaler`  to call the methods to convert to a byte slice before passing down.

nightlyone:
After thinking about it a bit longer, I agree with you @Zariel , that we better do this in some higher level layer.

PR #1097, opened 2018-04-19
Potentially fixing the flakiness in TestConnClosedBlocked

### LINKED ISSUE #1088: TestConnBlocked flake

alourie:
Fixes #1088 

It seems this change fixes the most of the "connection closed waiting for response" errors as in the original report.

Running the single test with high `-count` seems to lead to " no response received from cassandra within timeout period", possibly due to a rapid execution in fast succession. It seems that running with `-cpu=1` solves this problem.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -403,9 +403,14 @@ func (c *Conn) closeWithError(err error) {
 	}
 
 	// if error was nil then unblock the quit channel
-	close(c.quit)
+	if err == nil {
+		close(c.quit)
+	}
+
+	// Now close the connection and save the error
 	cerr := c.close()
 
+	// Handle the original error or the channel-closure error
 	if err != nil {
 		c.errorHandler.HandleError(c, err, true)
 	} else if cerr != nil {


bjmb:
I've opened https://github.com/gocql/gocql/pull/1107 with an alternative solution, with this I have the problem that it fails occasionally but with a different error: `conn_test.go:616: expected to get use of closed networking connection error got: gocql: no response received from cassandra within timeout period`. I can reproduce this consistently if I add `time.Sleep(100 * time.Millisecond)` before https://github.com/gocql/gocql/blob/181004e14a3fb735efcc826a4256369d0c96747b/conn_test.go#L611

alourie:
@bjmb I checked your branch and it looks good, I've also run unit tests and they also look good. That being said, the fix in my branch is also a good idea, as it unblocks the channel only when there's no error, otherwise, the channel can get closed too soon and still cause problems. Would you think that merging our branches would be a good idea for solving this?

bjmb:
I've tried to PR your branch but I haven't been able, however I can PR your branch to mine here: https://github.com/bjmb/gocql/pull/1 , would you be OK with merging that and closing this?

alourie:
@bjmb Yep, that'd work. If you merge mine into yours, and update your PR #1107 I'd be happy to close off this one.

bjmb:
:+1: done

alourie:
Closing in favour of #1107

PR #1101, opened 2018-04-26
ObservedQuery,ObservedBatch: pass the address of the node

zllak:
By passing the Conn, we can now have the address of the cassandra node which
made the query/batch.
This is very useful as we can now graph our latencies per nodes.

### DIFF
--- a/query_executor.go
+++ b/query_executor.go
@@ -6,7 +6,7 @@ import (
 
 type ExecutableQuery interface {
 	execute(conn *Conn) *Iter
-	attempt(keyspace string, end, start time.Time, iter *Iter)
+	attempt(keyspace string, end, start time.Time, iter *Iter, address string)
 	retryPolicy() RetryPolicy
 	GetRoutingKey() ([]byte, error)
 	Keyspace() string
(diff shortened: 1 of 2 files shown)


Zariel:
Can we instead not use the HostInfo? I would rather not be exposing the Conn as really that should be internal

zllak:
@Zariel done, removed the use of the Conn, pass directly the address as a string.

query_executor.go
```diff
@@ -23,7 +23,7 @@ func (q *queryExecutor) attemptQuery(qry ExecutableQuery, conn *Conn) *Iter {
 	iter := qry.execute(conn)
 	end := time.Now()
 
-	qry.attempt(q.pool.keyspace, end, start, iter)
+	qry.attempt(q.pool.keyspace, end, start, iter, conn.Address())
```
Zariel: can we just use the host info instead here?

zllak: unless I'm missing something, I can access the `HostInfo` from the `Conn` only by using the `localHostInfo()` method no ?
Method that can return an error, not ideal

Zariel: the *HostInfo is now accessible on *conn

PR #1104, opened 2018-05-05
Improve query pool

nightlyone:
So query creation is more efficient as well as as identical as possible in Session.Bind and Session.Query.

### DIFF
--- a/session.go
+++ b/session.go
@@ -311,12 +311,14 @@ func (s *Session) SetTrace(trace Tracer) {
 // value before the query is executed. Query is automatically prepared
 // if it has not previously been executed.
 func (s *Session) Query(stmt string, values ...interface{}) *Query {
-	s.mu.RLock()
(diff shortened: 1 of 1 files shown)


nightlyone:
Hi all,
not sure about using queryPool also for Session.Bind, so I allowed maintainer edits for you.

session.go
```diff
@@ -1051,25 +1058,7 @@ func (q *Query) Release() {
 
 // reset zeroes out all fields of a query so that it can be safely pooled.
 func (q *Query) reset() {
-	q.stmt = ""
-	q.values = nil
-	q.cons = 0
-	q.pageSize = 0
-	q.routingKey = nil
-	q.routingKeyBuffer = nil
```
Zariel: looking at this, we could probably keep these buffers between pooling but would need to verify that they are not kept around anywhere after being used.

PR #1105, opened 2018-05-05
reduce Conn.mu hold times

nightlyone:
The streamPool is already internally serialized by the Go runtime and
designed for high scalability. But currently retrieval from that pool
is unnecessarily serialized by the Conn.mu instead.

Additionally when the pool is empty, the hold times are even higher,
also this channel allocation for the timeout channel eas in that lock scope.

All of this amounted to enough overhead to make the lock appear in
internal profiles.

Technically the only thing required to be serialized are the accesses to
the Conn.calls map, so limit lock scope to exactly these operations.


Zariel:
Thanks!

PR #1107, opened 2018-05-08
ErrConnectionClosed means the connection was closed, not only when re…

beltran:
…ceiving

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -644,7 +644,6 @@ func (c *Conn) exec(ctx context.Context, req frameWriter, tracer Tracer) (*frame
 	if tracer != nil {
 		framer.trace()
 	}
-
 	timeoutCh := call.resetTimeout(c.timeout)
 	if err := c.sendFrame(ctx, call, timeoutCh); err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -1231,6 +1231,6 @@ var (
 	ErrQueryArgLength    = errors.New("gocql: query argument length mismatch")
 	ErrTimeoutNoResponse = errors.New("gocql: no response received from cassandra within timeout period")
 	ErrTooManyTimeouts   = errors.New("gocql: too many query timeouts on the connection")
-	ErrConnectionClosed  = errors.New("gocql: connection closed waiting for response")
+	ErrConnectionClosed  = errors.New("gocql: connection was closed")
```
bjmb: This is used returned when sending, not only when receiving

conn.go
```diff
@@ -403,9 +403,14 @@ func (c *Conn) closeWithError(err error) {
 	}
 
 	// if error was nil then unblock the quit channel
-	close(c.quit)
+	if err == nil {
+		close(c.quit)
```
Zariel: Im not sure this is correct, as were stopping this will cause the `writeToConn` goroutine to leak

conn.go
```diff
@@ -685,8 +689,9 @@ func (c *Conn) getResp(ctx context.Context, call *callReq, timeoutCh <-chan time
 				// been handed another error from another stream which caused the
 				// connection to close.
 				c.releaseStream(call.streamID)
+				return err
 			}
-			return err
+			return ErrConnectionClosed
```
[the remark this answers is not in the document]
bjmb: `c.Closed()` is `true` at that point so I figured it was the correct value to return.

PR #1115, opened 2018-05-18
Add frame header observer for per-frame instrumentation

jacobgreenleaf:
Continuing with #1023, add per-frame instrumentation to allow client libraries to instrument e.g. frame body lengths, to tune query page size and observe effects of compression, as well as monitor the authentication handshake frames. 

This PR doesn't include instrumentation for the authentication handshake but introduces the interface and adds the observer hook into the query response path. I wasn't sure whether I should put the hook into `exec` or `recv` or both since `startup` uses `exec` but queries seem to go through `recv` and they seem to be doing the same thing.


Zariel:
cherry picked into master cb53b83ccd36ef7fedd734e87816c7ea3db8f7b3 thanks for the contribution!

PR #1116, opened 2018-05-20
Prevent race condition when writing

bjmb:
This fixes:
https://github.com/gocql/gocql/issues/1103

I've been able to reproduce it with the simpler script:

```python
package gocql

import (
	"fmt"
	"testing"
	"time"
	"github.com/stretchr/testify/require"
)

func getCassandraSession() *Session {

	cluster := NewCluster("127.0.0.1")
	cluster.Port = 9042
	cluster.Timeout = 40 * time.Second
	cluster.ConnectTimeout = 40 * time.Second
	cluster.NumConns = 100

	dbSession, err := cluster.CreateSession()
	if err != nil {
		fmt.Errorf("failed to get cassandra session: %v", err)
		return nil
	}
	return dbSession
}

func TestGocqlMViewDataRace(t *testing.T) {
	// keyspace := `keyspace`
	// table := `table`
	// view := `view`
	id1 := `c818ce88-78b5-46a3-8370-18fdbc330828`

	dropKeyspaceStmt := `DROP KEYSPACE IF EXISTS "keyspace"`
	createKeyspaceStmt := `CREATE KEYSPACE IF NOT EXISTS "keyspace" WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1 }`

	createTableStmt := `CREATE TABLE IF NOT EXISTS "keyspace"."table" ("ID" uuid, "Name" varchar, PRIMARY KEY (("ID")));`

	insertStmt := `INSERT INTO "keyspace"."table" ("ID", "Name") VALUES (?, ?)`

	s := getCassandraSession()
	require.NotNil(t, s)

	err := s.Query(dropKeyspaceStmt).Exec()
	require.NoError(t, err)

	err = s.Query(createKeyspaceStmt).Exec()
	require.NoError(t, err)

	err = s.Query(createTableStmt).Exec()
	require.NoError(t, err)

	// increase the number of iterations OR run the test multiple times.
	for i := 0; i < 100000; i++ {
		err = s.Query(insertStmt, id1, "name").Exec()
	}
}
```

I don't think this is really a bug but it makes the data race detector get off.

What is happening is the following:
* At some point one of the iterations writes to the socket [here but doesn't exit the call](https://github.com/gocql/gocql/blob/ed13cda9d60fb18fdb6030aa07c9f058273de493/frame.go#L720). The write has happened but [this if](https://github.com/golang/sys/blob/master/unix/syscall_unix.go#L164) hasn't been executed.
* Between the write and the `if` the package goes to Cassandra, returns, the framer is returned to the pool.
* Still between the write and the `if` the next iteration is started and from the pool we get the same framer of the previous iteration
* Still between the write and the `if` gocql sends a message using the framer, a write happens and the race detector goes off

Since the write has happened it's not dangerous to reuse the framer but the race detector thinks so. I'm not convinced this is the best solution but it fixes the issue. This PR waits for the write to come back before returning the framer to the pool.

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -346,12 +346,27 @@ func (f frameHeader) Header() frameHeader {
 
 const defaultBufSize = 128
 
-var framerPool = sync.Pool{
-	New: func() interface{} {
-		return &framer{
-			wbuf:       make([]byte, defaultBufSize),
-			readBuffer: make([]byte, defaultBufSize),
-		}
+type framerPoolType struct {
+	pool		sync.Pool
+}
+
+func (fp *framerPoolType) Get() *framer{
+	return fp.pool.Get().(*framer)
+}
+
+func (fp *framerPoolType) Put(f *framer){
+	<-f.writingDone
+	fp.pool.Put(f)
+}
+
+var framerPool = framerPoolType{
+	pool:  sync.Pool{
+		New: func() interface{} {
+			return &framer{
+				wbuf:       make([]byte, defaultBufSize),
+				readBuffer: make([]byte, defaultBufSize),
+			}
+		},
 	},
 }
 
@@ -377,10 +392,11 @@ type framer struct {
 
 	rbuf []byte
 	wbuf []byte
+	writingDone chan struct{}
 }
 
 func newFramer(r io.Reader, w io.Writer, compressor Compressor, version byte) *framer {
-	f := framerPool.Get().(*framer)
+	f := framerPool.Get()
 	var flags byte
 	if compressor != nil {
 		flags |= flagCompress
@@ -392,6 +408,9 @@ func newFramer(r io.Reader, w io.Writer, compressor Compressor, version byte) *f
 	if version > protoVersion2 {
 		headSize = 9
 	}
+
+	f.writingDone = make(chan struct{}, 1)
+	f.writingDone <- struct{}{}
 
 	f.compres = compressor
 	f.proto = version
@@ -655,6 +674,7 @@ func (f *framer) parseErrorFrame() frame {
 }
 
 func (f *framer) writeHeader(flags byte, op frameOp, stream int) {
+	<-f.writingDone
 	f.wbuf = f.wbuf[:0]
 	f.wbuf = append(f.wbuf,
 		f.proto,
@@ -695,6 +715,7 @@ func (f *framer) setLength(length int) {
 }
 
 func (f *framer) finishWrite() error {
+	defer func() { f.writingDone <- struct{}{} }()
 	if len(f.wbuf) > maxFrameSize {
 		// huge app frame, lets remove it so it doesn't bloat the heap
 		f.wbuf = make([]byte, defaultBufSize)


Zariel:
Thanks for looking into this. I'm not convinced this is a false positive, the race detecting `Write` should return before we execute any code which can release the framer. I can reproduce this without sending any queries by setting `f.wbuf = f.wbuf[:0]` immediately after the `Write` returns in framer.

Bisecting points to acd6671d04577df9e6cd95fa1bd0aa70f98122af but I don't think that introduced the underlying race but just exposed it.

I'd rather not band aid over the issue with the extra synchronisation on write without understanding what is going on.

Zariel:
This testcase replicates for me

```go
func TestGocqlMViewDataRace(t *testing.T) {
	const table = "TestGocqlMViewDataRace"
	const id1 = "c818ce88-78b5-46a3-8370-18fdbc330828"

	s := createSession(t)
	defer s.Close()

	if err := createTable(s, "CREATE TABLE IF NOT EXISTS "+table+" (ID uuid, Name varchar, PRIMARY KEY (ID))"); err != nil {
		t.Fatal(err)
	}
}
```

with the following patch to `frame.go`
```patch
@@ -721,6 +762,7 @@ func (f *framer) finishWrite() error {
        if err != nil {
                return err
        }
+       f.wbuf = f.wbuf[:0]

        return nil
 }
```

Zariel:
The framer shouldn't be put back into the pool until it has been used, I've checked the error cases and the first framer that gets released is a `resultRows` which means an `Iter` is returned out which has a guard to release the framer once on the call to `Close`.

Once `Conn.exec` has returned the frame will have been been written and the response read.

bjmb:
> Once `Conn.exec` has returned the frame will have been been written and the response read.

This is true, but `finishWrite` or the `Write`from the goroutine `writeToConn` maybe haven't returned yet. Even `Close` on `Iter` can be called **before**  `finishWrite` or the `Write` return.

Zariel:
For that happen then the following must happen which I've confirmed isn't.

1. `sendFrame` sends on `frameWriteArgChan` and returns
2. `getResp` either gets a frame, which means that the frame has been written and Cassandra sent a response back, or it gets a timeout/ctx done or quit which it returns an error.

Removing the pool put in iter close still triggers the race.

There should be no race triggered by the reset of the `wbuf` because no 2 goroutines should ever reuse the same framer. Still at a bit of a loss to what is causing this.

Zariel:
At this point I'm inclined to do one of the following

1. Completely disable framer pooling
2. Revert acd6671

1 seems the most necessary at this point as I've been starting at this race and code for 4 hours and haven't got any closer.

Zariel:
Adding a mutex to newFramer and to finishWrite fixes the race which is the same as this change.

Zariel:
Though also adding this block to the top of both functions fixes it too which I don't expect.

```go
	if atomic.AddInt64(&f.inUse, 1) != 1 {
		panic("framer already in use")
	}
	defer atomic.AddInt64(&f.inUse, -1)
```

bjmb:
I'd expect that to panic at some point

Zariel:
Likewise, no panics even ran in stress 5000 times.

Zariel:
#1121 points to it not being the framer reuse

PR #1121, opened 2018-05-20
Remove framer pooling

Zariel:


bjmb:
Can't trigger the race condition here with either of the two tests in https://github.com/gocql/gocql/pull/1116, no matter if I add `f.wbuf = f.wbuf[:0]` in `finishWrite` :confused:

PR #1151, opened 2018-07-27
Provide a way to limit per-query-attempt deadlines

annismckenzie:
This adds the ability to constrain the runtime of individual query attempts. Right now we have the situation that there are no retries if the coordinator times out responding as there's only the possibility to time out a query via the context. This is also somewhat related to https://github.com/gocql/gocql/issues/812#issuecomment-257107273.

In working with this we found and fixed a couple of related problems:
1. Contexts expiring or canceled while the query executor was in the retry loop were not respected and always resulted in the `conn.exec` to still write the query to the connection because the context was checked too late.
2. The query executor's attempt loop was simplified to bail earlier.

The test we implemented to check whether retries are done and respect the per-attempt timeout is rather crude. Pointers on how to go about testing this are appreciated.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -585,6 +585,10 @@ type callReq struct {
 }
 
 func (c *Conn) exec(ctx context.Context, req frameWriter, tracer Tracer) (*framer, error) {
+	if ctx != nil && ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
+
 	// TODO: move tracer onto conn
 	stream, ok := c.streams.GetStream()
 	if !ok {
--- a/conn.go
+++ b/conn.go
@@ -790,6 +790,16 @@ func marshalQueryValue(typ TypeInfo, value interface{}, dst *queryValues) error 
 }
 
 func (c *Conn) executeQuery(qry *Query) *Iter {
+	ctx := qry.context
+	if qry.attemptTimeout > 0 {
+		if ctx == nil {
+			ctx = context.Background()
+		}
+		var cancel func()
+		ctx, cancel = context.WithTimeout(ctx, qry.attemptTimeout)
+		defer cancel()
+	}
+
 	params := queryParams{
 		consistency: qry.cons,
 	}
@@ -814,7 +824,7 @@ func (c *Conn) executeQuery(qry *Query) *Iter {
 	if qry.shouldPrepare() {
 		// Prepare all DML queries. Other queries can not be prepared.
 		var err error
-		info, err = c.prepareStatement(qry.context, qry.stmt, qry.trace)
+		info, err = c.prepareStatement(ctx, qry.stmt, qry.trace)
 		if err != nil {
 			return &Iter{err: err}
 		}
@@ -863,7 +873,7 @@ func (c *Conn) executeQuery(qry *Query) *Iter {
 		}
 	}
 
-	framer, err := c.exec(qry.context, frame, qry.trace)
+	framer, err := c.exec(ctx, frame, qry.trace)
 	if err != nil {
 		return &Iter{err: err}
 	}
--- a/conn.go
+++ b/conn.go
@@ -800,8 +800,33 @@ func (c *Conn) executeQuery(qry *Query) *Iter {
 			ctx = context.Background()
 		}
 		var cancel func()
-		ctx, cancel = context.WithTimeout(ctx, rt.AttemptTimeout())
+		ctx, cancel = context.WithCancel(ctx)
 		defer cancel()
+		if qry.attemptTimeoutTimer == nil {
+			qry.attemptTimeoutTimer = time.NewTimer(0)
+			<-qry.attemptTimeoutTimer.C
+		} else {
+			if !qry.attemptTimeoutTimer.Stop() {
+				select {
+				case <-qry.attemptTimeoutTimer.C:
+				default:
+				}
+			}
+		}
+
+		qry.attemptTimeoutTimer.Reset(rt.AttemptTimeout())
+		timeoutCh := qry.attemptTimeoutTimer.C
+
+		go func() {
+			select {
+			case <-ctx.Done():
+				qry.attemptTimeoutTimer.Stop()
+				break
+			case <-timeoutCh:
+				break
+			}
+			cancel()
+		}()
 	}
 
 	params := queryParams{
--- a/conn.go
+++ b/conn.go
@@ -795,12 +795,12 @@ func marshalQueryValue(typ TypeInfo, value interface{}, dst *queryValues) error 
 
 func (c *Conn) executeQuery(qry *Query) *Iter {
 	ctx := qry.context
-	if qry.attemptTimeout > 0 {
+	if rt, ok := qry.rt.(RetryPolicyWithAttemptTimeout); ok && rt.AttemptTimeout() > 0 {
 		if ctx == nil {
 			ctx = context.Background()
 		}
 		var cancel func()
-		ctx, cancel = context.WithTimeout(ctx, qry.attemptTimeout)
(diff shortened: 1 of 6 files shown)


conn_test.go
```diff
@@ -314,9 +345,8 @@ func TestQueryRetry(t *testing.T) {
 		t.Fatalf("expected requests %v to match query attempts %v", requests, attempts)
 	}
 
-	// the query will only be attempted once, but is being retried
-	if requests != int64(rt.NumRetries) {
-		t.Fatalf("failed to retry the query %v time(s). Query executed %v times", rt.NumRetries, requests-1)
+	if requests > 8 {
```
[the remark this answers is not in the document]
ekle: this will never reach 9 retries because the `QueryTimeout` of `200ms` will stop the query before.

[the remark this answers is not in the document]
ekle: i have added a comment about it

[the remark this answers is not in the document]
alourie: yep, way more clear. Thanks!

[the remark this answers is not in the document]
annismckenzie: Yeah, this is a bit of a mess. But we couldn't figure out how to make it any less messy because there's no way in the test to pass the attempt timeout to the framer (tbh, I really hate that test code). Also, if we do keep it we should fail the test if there aren't any retries because right now it only fails if there's too many. Ideally, the reasoning should be improved by asserting those retries at the framer level but as I said, that's currently not possible. :/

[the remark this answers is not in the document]
annismckenzie: I really really like that suggestion and implemented it right away! This feels so much better. I didn't add the `AttemptTimeout` method to the RetryPolicy interface in order to not break existing code. Take a look?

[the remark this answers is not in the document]
annismckenzie: I also added a guard assertion to make sure the query is retried.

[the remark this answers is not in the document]
alourie: It looks way way better now. I even like how you implemented the RetryPolicyWithTimeout separately. It's a neat trick, will def bookmark it for myself.

Everything is so much cleaner now and totally makes sense.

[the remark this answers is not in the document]
alourie: just let's make sure it builds cleanly. I saw the error in the build, totally not your fault :-)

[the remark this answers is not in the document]
annismckenzie: I'll give it another shot now.

alourie:
@Zariel would it be possible for you to have a look?
Thanks.

conn.go
```diff
@@ -791,6 +795,16 @@ func marshalQueryValue(typ TypeInfo, value interface{}, dst *queryValues) error
 }
 
 func (c *Conn) executeQuery(qry *Query) *Iter {
+	ctx := qry.context
+	if rt, ok := qry.rt.(RetryPolicyWithAttemptTimeout); ok && rt.AttemptTimeout() > 0 {
+		if ctx == nil {
+			ctx = context.Background()
```
Zariel: Tbh we should just move this all the way up to default on Query, when we do an API change I want to change all the frontend functions to take context as their first param. Ill make an issue for this, dont worry about changing it now

conn.go
```diff
@@ -791,6 +795,16 @@ func marshalQueryValue(typ TypeInfo, value interface{}, dst *queryValues) error
...
+		}
+		var cancel func()
+		ctx, cancel = context.WithTimeout(ctx, rt.AttemptTimeout())
```
Zariel: Timer allocation has shown to cause significant impact on query performance see #660

policies.go
```diff
@@ -153,6 +155,14 @@ type RetryPolicy interface {
 	GetRetryType(error) RetryType
 }
 
+// RetryPolicyWithAttemptTimeout is an optional interface retry policies can implement
+// in order to control the duration to use before a query attempt is considered
+// as a timeout and will potentially be retried.
+// It's not part of the RetryPolicy interface to remain backwards compatible.
+type RetryPolicyWithAttemptTimeout interface {
+	AttemptTimeout() time.Duration
```
Zariel: Should this interface embed RetryPolicy to make it explicit?

session.go
```diff
@@ -1484,6 +1488,11 @@ func (b *Batch) WithContext(ctx context.Context) *Batch {
 	return b
 }
 
+// GetContext satisfies the ExecutableQuery interface.
+func (b *Batch) GetContext() context.Context {
```
Zariel: Call this `Context()` along with the one on Query to me more inline with go style.

conn_test.go
```diff
@@ -896,6 +937,14 @@ func (srv *TestServer) process(f *framer) {
 			f.writeHeader(0, opError, head.stream)
 			f.writeInt(0x1001)
 			f.writeString("query killed")
+		case "killaftertimeout":
+			atomic.AddInt64(&srv.nKillReq, 1)
+			f.writeHeader(0, opError, head.stream)
+			f.writeInt(0x1001)
+			f.writeString("query killed")
+			rand.Seed(time.Now().UnixNano())
+			sleepFor := time.Duration(time.Millisecond * time.Duration((rand.Intn(20) + 25)))
```
Zariel: can you run this test with stress to ensure it does not flake https://godoc.org/golang.org/x/tools/cmd/stress

annismckenzie: Yeah, I dropped that addition and went with the existing `slow` query. There's really no need for this randomized timing out of the query here because it isn't successful in either case so just using `slow` and checking that the attempt timeout is used is fine.

annismckenzie: Oh, and thanks for pointing me in the direction of the `stress` tool. I still let it run over the whole suite and saw that the date marshaling tests are flaky. After disabling that it ran:
```
226 runs so far, 0 failures
^C

real	5m24.240s
user	10m25.325s
sys	11m45.440s
```

I also let the timeout test run under stress on its own and it's all good.

policies.go
annismckenzie: Good thinking, done.

session.go
annismckenzie: all good now

conn.go
annismckenzie: Please see 81c97e428ca091d87293fa4f032d7e55a8db98d2 for my changes and some explanations. I really didn't think it would have such a noticable impact on the timing of the test but on a 25ms attempt timeout in the test each attempt is now measured as taking ~25.2ms instead of the 27-28ms previously. I took the code you wrote in #661 and copied it here because there's no good way to extract it. It works correctly (at least according to the tests) but it is somewhat convoluted.

annismckenzie:
I addressed all of your comments, PTAL.

I did introduce one more change in c65250d26163b5442348f589dfad1ccca24581ef to abandon parsing frames if the context is already canceled. I can drop that commit if you don't want it in here. :)

conn_test.go
annismckenzie: Isn't that weird, I can't reproduce it right now with the stress tool. Here's what I did:
- `go test -tags=unit -c`
- `time stress ./gocql.test -test.cpu=8`

I swear it was flaking on the `TestMarshalDate` where sometimes the result would be off by 1 millisecond.

Zariel: stress is mostly useful for running a single test that can execute pretty quickly, found lots of flakes it with.

conn.go
Zariel: on query/batch

Zariel:
I don't think it's worth the added complexity of doing that, parsing the frame is very trivial part of the request flow, really it should be for stopping the loops/select's. Can you remove that last commit and ill merge.

Thanks

annismckenzie:
Sure thing, you're right. Dropped the commit, rebased onto the current master and Travis is also happy.

Thank you for accepting our contributions! 🎉

Zariel:

PR #1160, opened 2018-08-21
Shorten runtime for long tests

alourie:
`skipLong` option allows skipping DNS and StartupTimeout tests, making unittests run in under 5s. 

Considering that we run unittests a lot to make sure stuff is not overly broken, this saves a lot of time during dev.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -152,6 +152,10 @@ func TestDNSLookupConnected(t *testing.T) {
 		Logger = &defaultLogger{}
 	}()
 
+	// Override the defaul DNS resolver and restore at the end
+	failDNS = true
+	defer func() { failDNS = false }()
+
 	srv := NewTestServer(t, defaultProto, context.Background())
 	defer srv.Stop()
 
@@ -178,8 +182,9 @@ func TestDNSLookupError(t *testing.T) {
 		Logger = &defaultLogger{}
 	}()
 
-	srv := NewTestServer(t, defaultProto, context.Background())
-	defer srv.Stop()
+	// Override the defaul DNS resolver and restore at the end
+	failDNS = true
+	defer func() { failDNS = false }()
 
 	cluster := NewCluster("cassandra1.invalid", "cassandra2.invalid")
 	cluster.ProtoVersion = int(defaultProto)
--- a/control.go
+++ b/control.go
@@ -121,7 +121,7 @@ func hostInfo(addr string, defaultPort int) ([]*HostInfo, error) {
 	}
 
 	// Look up host in DNS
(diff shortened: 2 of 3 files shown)


common_test.go
```diff
@@ -23,6 +23,7 @@ var (
 	flagRunAuthTest  = flag.Bool("runauth", false, "Set to true to run authentication test")
 	flagCompressTest = flag.String("compressor", "", "compressor to use")
 	flagTimeout      = flag.Duration("gocql.timeout", 5*time.Second, "sets the connection `timeout` for all operations")
+	flagSkipLong     = flag.Bool("skipLong", false, "Don't run long tests")
```
Zariel: go's testing framework already includes support for this, `go test -short` and then in the test `testing.Short()`, see https://golang.org/pkg/testing/#Short

conn_test.go
```diff
@@ -145,6 +145,9 @@ func newTestSession(addr string, proto protoVersion) (*Session, error) {
 }
 
 func TestDNSLookupConnected(t *testing.T) {
+	if *flagSkipLong {
+		t.Skip("Skipping long test")
+	}
```
Zariel: Can we just reduce some timeouts for this test? How long does it take to complete?

common_test.go
alourie: Removed.

conn_test.go
alourie: Now using testing.Short() to test for the option.

alourie:
@Zariel I think it's in a better shape now. Supports skipping these long tests with `-short` option; also reworked a bit a code around using centralised Logger. This, in turn, also allowed parallelising the execution of the longer tests, bringing down the whole execution to about 20s....which I still long IMHO.

As I mentioned, we can lower the timeout in the TestStartupTimeout as we control it. But it doesn't seem to be possible to change the timeout in the DNS tests as eventually, they execute the `net.LookupIP()` call, which we have no control over. Maybe we can just mock it? Or, alternatively, move these tests from unit tests into integration tests?

If you like the logger refactoring, I can do it for all the codebase, or if you have other ideas on how to implement it, it would be great.

Thanks!

alourie:
@Zariel sorry for all the fuss. Found a much better approach.

I've created a helper `LookupIP` function that checks a global test value `failDNS`; if it's false it runs the usual `net.LookupIP()`, if it's true it just returns an error straight away. Thus, almost no code changes are needed; only DNS tests set that flag and reset it when they're done, it takes almost no time to run these tests now, and the whole unittest run is now cut from about 50s to about 9s, which is much more tolerable.

PR #1162, opened 2018-08-24
WIP: Stop retrying errors that are fatal

### LINKED ISSUE #1161: If using RetryPolicy server errors are not propagated to user

Any.

e48272f

### What did you do?

```go
package gocql

import "testing"

func TestErrorNotPropagated(t *testing.T) {
	cluster := createCluster()
	t.Logf("Using %T", cluster.RetryPolicy)

	session, err := cluster.CreateSession()
	if err != nil {
		t.Fatal(err)
	}
	defer session.Close()

	q := session.Query("BLA BLA BLA")
	if err := q.Exec(); err != nil {
		t.Log(err)
	}
	// e48272f
	// grr_test.go:7: Using *gocql.SimpleRetryPolicy
	// grr_test.go:17: gocql: no hosts available in the pool

	// 153a728
	// grr_test.go:7: Using *gocql.SimpleRetryPolicy
	// grr_test.go:17: line 1:0 no viable alternative at input 'BLA' ([BLA]...)
}
```

Proper error form server:

```
line 1:0 no viable alternative at input 'BLA' ([BLA]...)
```

### What did you see instead?

Wrong information that the pool is empty

```
```

With the changes introduced in e48272ff, if using RetryPolicy user is not notified about server (or any other) errors.

cc @Zariel @annismckenzie @alourie

annismckenzie:
This will fix #1161. The code that's there now fixes that specific issue but there's more to come.

So, this was fun. The good news is that we're retrying too many query failures after merging #1151 but masking the real error in case it's fatal and should not be retried wasn't very good.

This isn't finished yet but the basics are there. I scoured the codebase for more errors that could be returned by the various parts that run when executing a query and already got a bunch of those that shouldn't be retried.

The next step is grouping all the errors defined in errors.go where most are also already wrapped in custom error structs that can be type switched against (also by retry policies but more on that later). I looked at DataStax's Java driver implementation – specifically at https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/policies/RetryPolicy.java – that was very enlightening. I also want to group those errors into the four categories read timeout/write timeout/unavailable/request failure which in the Java implementation also each subsume a bunch of errors returned by Cassandra. That'll make it easier to decide what's retryable and what's not. What might be interesting to pursue is making those groupings available to the retry policy – I'll come up with something.

I know now that even for a given error it's possible to decide in a multitude of ways with how to handle the retry or return it to the caller and I don't want to lose that. I'll also be looking at the more sophisticated retry policies available in the codebase and decide whether it makes sense to move some of that logic up to here. A good retry policy is a pretty hard thing to write.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -876,7 +876,7 @@ func (c *Conn) executeQuery(qry *Query) *Iter {
 		}
 
 		if len(values) != info.request.actualColCount {
-			return &Iter{err: fmt.Errorf("gocql: expected %d values send got %d", info.request.actualColCount, len(values))}
+			return &Iter{err: NewErrProtocol("gocql: expected %d values send got %d", info.request.actualColCount, len(values))}
 		}
 
 		params.values = make([]queryValues, len(values))
--- a/query_executor.go
+++ b/query_executor.go
@@ -37,8 +37,18 @@ func (q *queryExecutor) attemptQuery(qry ExecutableQuery, conn *Conn) *Iter {
 // In cases that are ambiguous or should be decided on a case-by-case basis it errs on
 // the side of returning true and pushing the decision onto the retry policy.
 func (q *queryExecutor) checkErrorIsRetryable(err error) bool {
+	switch err {
+	// skip ErrNotFound: evaluated after a query succeeds (just with 0 rows)
+	// skip ErrNoConnections because that's actually only returned below when the pool runs out of hosts to retry on
+	case ErrUnsupported, ErrTooManyStmts, ErrUseStmt, ErrSessionClosed, ErrNoKeyspace,
+		ErrKeyspaceDoesNotExist, ErrNoMetadata, ErrorUDTUnavailable:
+		return false
+	}
+
 	switch err.(type) {
 	case *RequestErrSyntax:
+		return false
+	case *ErrProtocol:
 		return false
 	}
 
--- a/query_executor.go
+++ b/query_executor.go
@@ -36,10 +36,8 @@ func (q *queryExecutor) attemptQuery(qry ExecutableQuery, conn *Conn) *Iter {
 // checkRetryPolicy is used by the query executor to determine how a failed query should be handled.
 // It consults the query context and the query's retry policy.
 func (q *queryExecutor) checkRetryPolicy(rq ExecutableQuery, err error) (RetryType, error) {
-	if ctx := rq.Context(); ctx != nil {
-		if ctx.Err() != nil {
-			return Rethrow, ctx.Err()
-		}
+	if ctx := rq.Context(); ctx != nil && ctx.Err() != nil {
+		return Rethrow, ctx.Err()
 	}
 	p := rq.retryPolicy()
 	if p == nil {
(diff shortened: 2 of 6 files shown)


conn.go
```diff
@@ -876,7 +876,7 @@ func (c *Conn) executeQuery(qry *Query) *Iter {
 		}
 
 		if len(values) != info.request.actualColCount {
-			return &Iter{err: fmt.Errorf("gocql: expected %d values send got %d", info.request.actualColCount, len(values))}
+			return &Iter{err: NewErrProtocol("gocql: expected %d values send got %d", info.request.actualColCount, len(values))}
```
alourie: maybe print the `qry.stmt` here too? would be easier to understand which query didn't work.

query_executor.go
```diff
@@ -33,13 +33,36 @@ func (q *queryExecutor) attemptQuery(qry ExecutableQuery, conn *Conn) *Iter {
 	return iter
 }
 
+// checkErrorIsRetryable evaluates whether a particular error is retryable.
+// In cases that are ambiguous or should be decided on a case-by-case basis it errs on
+// the side of returning true and pushing the decision onto the retry policy.
+func (q *queryExecutor) checkErrorIsRetryable(err error) bool {
```
alourie: And, +1000

The next logical step would be to add the flag needed to take the idempotence of the query and act accordingly, for instance, if the query is not idempotent, some errors wouldn't be retried; whereas if it's idempotent, then the query would retry. More on that can be found at https://docs.datastax.com/en/developer/java-driver/3.4/manual/retries/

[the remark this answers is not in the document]
alourie: or `retriableErr`...or even simply `retriable`. Check this out:

```if retriableErr(err) ... ``` 
or 
```if retriable(err) ... ```

Feels better than **check** :-)

alourie:
@gwik I do agree that implementing a common interface in all error wrappers is a good idea, I have to disagree on the idea of making an error decide if it's a retryable or not. 

The decision to retry a query or not should solely be under the retry policy control. You may want to repeat a query under certain conditions if you get a specific error, but I may **not** want to repeat the query after getting the same error because I might be under different conditions. You may want to retry queries on the same host, I may want to choose to try other hosts instead. All this should be implemented with different RetryPolicies.

gwik:
> disagree on the idea of making an error decide.

My point was to simplify the writing of `checkErrorIsRetryable` and make it future proof, the error can tell whether it is retryable. Implementation can decide if retryable or not and the `retryable` method/interface  could be private.
For example a query syntax error is always non-retryable. Some other error type can sometimes be or not depending on the specific implementation. But it's easier to manage when you generate the error than to update every location where handling of this error might make sense.

For errors that could have a different outcome under other conditions of course this has to be the responsibility of the retry policy and the interface should help it to make a decision without having to switch over every possible error types and compare to every possible error value. What should help with that I have not the answer for the moment, but I'll think about it.

alourie:
@gwik I see where you going with this. Maybe `checkErrorIsRetryable` can be error's own interface method, then the policy can additionally check it.

I guess it's up with you @annismckenzie at the moment, depending on how you'd like to approach it. Maybe have a separate ticket to wrap all errors into a typed interface, and then you could use it here?

PR #1163, opened 2018-08-24
Add CDC error

beltran:
This error can be [thrown](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/exceptions/ExceptionCode.java#L45])


Zariel:
What is a CDC Error?

beltran:
@Zariel http://cassandra.apache.org/doc/latest/operating/cdc.html. Error is returned when no more CDC data can be written

PR #1164, opened 2018-08-25
Fix bug in query executor that prevented retries from ceasing

annismckenzie:
If a retry policy is defined for the session/query, retries wouldn't stop even if the retry policy returned false in the call to `Attempt()`. The bug was introduced in #1151. As long as queries keep failing and no other condition applies the query would be retried. One such condition being the exhaustion of the host pool but even that might not happen in the case where the retry policy always returns `Retry` and retries are executed on the same connection. This might lead to any number of unexpected behaviors, even going as far as looping endlessly if the session/query doesn't have a timeout defined and no context is used that would eventually be canceled or time out.
The unit test that checks the retries being executed now checks for an exact number of query retries, making it a regression test as well. I shouldn't have gone for testing these things in a fuzzy manner which would likely have prevented it from sneaking in.

I'm sorry for introducing this. As it's quite urgent, may I ask @Zariel to give it a look and merge soon-ish?

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -336,7 +336,7 @@ func TestQueryRetry(t *testing.T) {
 		}
 	}()
 
-	rt := &testRetryPolicy{numRetries: 10, t: t, attemptTimeout: time.Millisecond * 25}
+	rt := &testRetryPolicy{numRetries: 2, t: t, attemptTimeout: time.Millisecond * 25}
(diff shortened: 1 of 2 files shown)


Zariel:
Thanks for the quick fix

annismckenzie:
Thanks for the ⚡️quick turnaround! 🎉

PR #1166, opened 2018-08-27
Revert #1164 and #1151

annismckenzie:
This reverts #1151 (and #1164 by association because it builds on top of it) because we introduced a serious issue reported in #1161.

I'll open a new PR with better tests, do a more rigorous review and also use that version in our staging services first (/cc @ekle to keep him in the loop).


annismckenzie:
@Zariel see the comments on #1161. I think we should revert.

Issue #1173, opened 2018-09-06
Expose query cancel() call


alourie:
In speculative query executions, one execution may succeed, which should lead to others dropping their work. For that, the finishing goroutine should have the ability to call the `cancel()` on the query's context.

This is required for #1083

PR #1174, opened 2018-09-06
Allow to Cancel() query at any time

alourie:
Solving #1173

* Add a Cancel() call to the ExecutableQuery interface
* Add cancelQuery and cancelBatch fields
* Initiate Query/Batch context/cancel functions
* Implement Cancel() call by calling the cancelQuery/cancelBatch
functions
* Add TestCancel to verify that the query is being cancelled by the
cancel call before it's finished.


alourie:
This will be used in Speculative Query executions (#1083) to cancel the outstanding/other running queries when one has succeeded.

PR #1175, opened 2018-09-06
conn: implement write coalescing

### LINKED ISSUE #383: Allow setting Nagle's algorithm on server connections
For workloads with lots of small inserts, we've seen a small performance boost when enabling Nagle's algorithm (net.TcpConn.SetNoDelay) on the Cassandra socket. It's probably not a good idea to have this enabled by default, however we could add an option in the ConnConfig struct for that. If you think it makes sense, I'll send a PR.

### LINKED ISSUE #458: Implement write coalescing
http://www.datastax.com/dev/blog/performance-doubling-with-message-coalescing details how coalescing socket writes on client side can reduce the pressure on the network, and lead to performance improvement.

This issue is somehow related to issue #383 , except that coalescing in user space also helps reducing the syscall load (on loads with lots of small inserts, we see `syscall.Syscall` taking a significant share of CPU time).

I have some prototype code in the `perf` branch of my `abustany/gocql` fork, specifically the first two commits, this issue is more to spark discussion on whether there is interest, and how the API should look like (with my prototype, coalescing is hardcoded, but we probably want to give the users some knobs to tweak).

Zariel:
Add an option to switch on write coalescing outgoing requests using
net.Buffers. This can be switched off via cluster config and the
interval to buffer for can be adjusted.

fixes #458
fixes #383

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -637,15 +637,13 @@ func (w *writeCoalescer) flush() {
 		w.w.(deadliner).SetWriteDeadline(time.Now().Add(w.timeout))
 	}
 
-	defer w.mu.Unlock()
 	w.mu.Lock()
-
+	defer w.mu.Unlock()
 	if len(w.buffers) == 0 {
 		return
 	}
 
-	// TODO: do we care about the number of bytes written? Given
-	// we are going to do a fanout n is useless and according to
+	// Given we are going to do a fanout n is useless and according to
 	// the docs WriteTo should return 0 and err or bytes written and
 	// no error.
 	_, w.err = w.buffers.WriteTo(w.w)
@@ -658,10 +656,10 @@ func (w *writeCoalescer) flush() {
 func (w *writeCoalescer) write(p []byte) (int, error) {
 	w.mu.Lock()
 	w.buffers = append(w.buffers, p)
-
 	for len(w.buffers) != 0 {
 		w.cond.Wait()
 	}
+
 	err := w.err
 	w.mu.Unlock()
 
--- a/conn.go
+++ b/conn.go
@@ -124,8 +124,11 @@ var TimeoutLimit int64 = 0
 // queries, but users are usually advised to use a more reliable, higher
 // level API.
 type Conn struct {
-	conn          net.Conn
-	r             *bufio.Reader
+	conn net.Conn
+	r    *bufio.Reader
+
+	w *writeCoalescer
+
 	timeout       time.Duration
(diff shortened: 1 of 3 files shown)


Zariel:
want to take a look @alourie ?

alourie:
@Zariel will do

conn.go
```diff
@@ -584,6 +617,60 @@ type callReq struct {
...
+	// result of the write
+	err error
+}
+
+func (w *writeCoalescer) flush() {
+	if w.timeout > 0 {
+		type deadliner interface {
+			SetWriteDeadline(time.Time) error
+		}
+		w.w.(deadliner).SetWriteDeadline(time.Now().Add(w.timeout))
```
alourie: Do we have to create a custom interface here? Can we just do away with `w.w.(net.Conn)...` ? We don't really use it anywhere else, and this is all internal anyway.

conn.go
```diff
@@ -584,6 +617,60 @@ type callReq struct {
...
+	}
+
+	defer w.mu.Unlock()
```
alourie: very small nitpick: would you mind swapping the order of Unlock() and Lock() here? I understand it's doing the right thing, it just looks weird ;-)

conn.go
```diff
@@ -584,6 +617,60 @@ type callReq struct {
...
+	w.mu.Lock()
+
+	if len(w.buffers) == 0 {
+		return
+	}
+
+	// TODO: do we care about the number of bytes written? Given
+	// we are going to do a fanout n is useless and according to
+	// the docs WriteTo should return 0 and err or bytes written and
+	// no error.
```
alourie: Considering the `WriteTo` seems to be pretty clear in it's writing the whole buffer until drained or erroring out, I think we really don't care about the number of bytes written. We can and do deduct that from the buffer's length. So maybe just leave the comment why we don't store that without the TODO mark.

conn_test.go
```diff
@@ -751,6 +751,39 @@ func TestContext_Timeout(t *testing.T) {
...
+	wg.Add(1)
+	go func() {
+		wg.Done()
+		if _, err := w.write([]byte("two")); err != nil {
+			t.Error(err)
+		}
+	}()
+	wg.Wait()
+
+	w.flush()
```
alourie: Considering we test that the write() won't finish until we flash, maybe check that pre-flash the `len(buf)` is still 0?

PR #1178, opened 2018-09-07
Speculative query execution

### LINKED ISSUE #1083: Speculative query execution

alourie:
fixes #1083 

This is an attempt to implement speculative query execution. I think it's more or less covers the feature, but I clearly could have missed some things.

This is still a bit of WIP, so any feedback is welcome @Zariel ( and @annismckenzie if you have the time).

Thanks!

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -316,7 +316,7 @@ func TestCancel(t *testing.T) {
 }
 
 type testQueryObserver struct {
-	metrics map[string]*queryMetrics
+	metrics map[string]*hostMetrics
 	verbose bool
 }
 
@@ -329,7 +329,7 @@ func (o *testQueryObserver) ObserveQuery(ctx context.Context, q ObservedQuery) {
 	}
 }
 
-func (o *testQueryObserver) GetMetrics(host *HostInfo) *queryMetrics {
+func (o *testQueryObserver) GetMetrics(host *HostInfo) *hostMetrics {
 	return o.metrics[host.ConnectAddress().String()]
 }
 
@@ -401,7 +401,7 @@ func TestQueryMultinodeWithMetrics(t *testing.T) {
 
 	// 1 retry per host
 	rt := &SimpleRetryPolicy{NumRetries: 3}
-	observer := &testQueryObserver{metrics: make(map[string]*queryMetrics), verbose: false}
+	observer := &testQueryObserver{metrics: make(map[string]*hostMetrics), verbose: false}
 	qry := db.Query("kill").RetryPolicy(rt).Observer(observer)
 	if err := qry.Exec(); err == nil {
 		t.Fatalf("expected error")
@@ -409,10 +409,11 @@ func TestQueryMultinodeWithMetrics(t *testing.T) {
 
 	for i, ip := range addresses {
 		host := &HostInfo{connectAddress: net.ParseIP(ip)}
+		queryMetric := qry.getHostMetrics(host)
 		observedMetrics := observer.GetMetrics(host)
 
 		requests := int(atomic.LoadInt64(&nodes[i].nKillReq))
-		hostAttempts := qry.metrics[ip].Attempts
+		hostAttempts := queryMetric.Attempts
 		if requests != hostAttempts {
 			t.Fatalf("expected requests %v to match query attempts %v", requests, hostAttempts)
 		}
@@ -421,7 +422,7 @@ func TestQueryMultinodeWithMetrics(t *testing.T) {
 			t.Fatalf("expected observed attempts %v to match query attempts %v on host %v", observedMetrics.Attempts, hostAttempts, ip)
 		}
 
-		hostLatency := qry.metrics[ip].TotalLatency
+		hostLatency := queryMetric.TotalLatency
 		observedLatency := observedMetrics.TotalLatency
 		if hostLatency != observedLatency {
 			t.Fatalf("expected observed latency %v to match query latency %v on host %v", observedLatency, hostLatency, ip)
--- a/conn_test.go
+++ b/conn_test.go
@@ -14,7 +14,9 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"net"
+	"os"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -44,8 +46,8 @@ func TestApprove(t *testing.T) {
 
 func TestJoinHostPort(t *testing.T) {
 	tests := map[string]string{
-		"127.0.0.1:0":                                 JoinHostPort("127.0.0.1", 0),
-		"127.0.0.1:1":                                 JoinHostPort("127.0.0.1:1", 9142),
+		"127.0.0.1:0": JoinHostPort("127.0.0.1", 0),
+		"127.0.0.1:1": JoinHostPort("127.0.0.1:1", 9142),
 		"[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:0": JoinHostPort("2001:0db8:85a3:0000:0000:8a2e:0370:7334", 0),
 		"[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:1": JoinHostPort("[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:1", 9142),
 	}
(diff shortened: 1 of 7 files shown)


alourie:
Fixed locking mechanism and Travis is now happy.

alourie:
Restructured the work. Now it's in 4 specific commits, each with self-descriptive commit message and can be looked at separately.

query_executor.go
```diff
@@ -50,51 +70,148 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
...
+		}
+
+		conn := pool.Pick()
+		if conn == nil {
+			continue
+		}
+
+		// if it's not the first attempt, pause
+		if specExecCounter > 0 {
+			<-time.After(sp.Delay())
```
Zariel: reuse a timer instead of using time.After. This should also have a way to cancel, for now just doing a select on the ctx.Done() should be enough. Though that would need exposing from the query

query_executor.go
```diff
@@ -50,51 +70,148 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
...
+	}
+
+	return res.iter, nil
+}
+
+func (q *queryExecutor) executeSpeculativeQuery(qry ExecutableQuery) (*Iter, error) {
+	hostIter := q.policy.Pick(qry)
+	sp := qry.speculativeExecutionPolicy()
+
+	results := make(chan queryResponse)
```
Zariel: buffer this to the number of possible executions so goroutines dont leak

query_executor.go
```diff
@@ -1,22 +1,26 @@
...
 	Keyspace() string
 	Cancel()
+	IsIdempotent() bool
 	RetryableQuery
 }
 
 type queryExecutor struct {
 	pool   *policyConnPool
 	policy HostSelectionPolicy
+	specWG sync.WaitGroup
```
Zariel: we need one of these per query, define it local to the execution loop

query_executor.go
```diff
@@ -50,51 +70,148 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
...
+	results := make(chan queryResponse)
+
+	stopExecutions := make(chan struct{})
+	defer close(stopExecutions)
+
+	var (
+		res             queryResponse
+		specExecCounter int
+	)
+	for selectedHost := hostIter(); selectedHost != nil && specExecCounter < sp.Executions(); selectedHost = hostIter() {
```
Zariel: I'm not sure I like this logic very much. I was thinking this would look something more like

```go
timer := time.NewTimer(speculation.After())
defer timer.Stop()
result := make(chan *Iter, speculation.Attempts())
go queryAttempt(ctx, hostIter, result)
for i := 0; i < speculation.Attempts(); i++ {
    select {
    case <-timer.C:
         go queryAttempt(ctx, hostIter, result)
         timer.Result(speculation.After())
    case result := <-result:
        return result
    case ctx.Done():
        return ctx.Err()
    }
}
```

And have all retrying logic handled within `queryAttempt` somehow.

This may require adjustments to the hostIter interface to say if we have `more` hosts to try

query_executor.go
```diff
@@ -30,12 +34,28 @@ func (q *queryExecutor) attemptQuery(qry ExecutableQuery, conn *Conn) *Iter {
...
+	// run normal, otherwise, runs speculative.
+	_, nonSpeculativeExecution := qry.speculativeExecutionPolicy().(NonSpeculativeExecution)
+	if nonSpeculativeExecution || !qry.IsIdempotent() {
+		return q.executeNormalQuery(qry)
+	} else {
+		return q.executeSpeculativeQuery(qry)
+	}
+}
+
+func (q *queryExecutor) executeNormalQuery(qry ExecutableQuery) (*Iter, error) {
```
Zariel: I'm not keen on how this requires having 2 implementations of a very similar function to execute a query if it is speculatable.

query_executor.go
Zariel: Should query speculation use the same host iterator for each query or a new one?

alourie: Considering that each speculative "execution" is defined per node, I think we need the same host iterator in order to go only once over all the endpoints. We don't want a different query to try a node that has already been tried.

query_executor.go
alourie: Will do.

query_executor.go
alourie: Thanks, that's actually a great idea.

query_executor.go
alourie: will do.

query_executor.go
alourie: I totally agree. Just haven't found an elegant way to do it yet.

alourie:
@Zariel I think it's much better, more clear and kinda elegant now. Would love your feedback. Thanks.

session.go
```diff
@@ -735,18 +744,29 @@ func (q Query) String() string {
 
 //Attempts returns the number of times the query was executed.
 func (q *Query) Attempts() int {
-	attempts := 0
-	for _, metric := range q.metrics {
+	q.metrics.l.Lock()
+	defer q.metrics.l.Unlock()
```
annismckenzie: I'd handle this the same way with the unlock as you did in the other functions below – and not take the defer hit and unlock after the for loop. There's no other code path executed so the defer isn't necessary.

session.go
```diff
@@ -704,23 +710,26 @@ func (q *Query) defaultsFromSession() {
...
 
+	q.spec = &NonSpeculativeExecution{}
 	s.mu.RUnlock()
 }
 
-func (q *Query) getHostMetrics(host *HostInfo) *queryMetrics {
-	hostMetrics, exists := q.metrics[host.ConnectAddress().String()]
+func (q *Query) getHostMetrics(host *HostInfo) *hostMetrics {
+	q.metrics.l.Lock()
+	defer q.metrics.l.Unlock()
```
annismckenzie: See my comment below.

policies.go
```diff
@@ -852,3 +854,21 @@ func (e *ExponentialReconnectionPolicy) GetInterval(currentRetry int) time.Durat
...
+
+type SpeculativeExecutionPolicy interface {
+	Attempts() int
+	Delay() time.Duration
+}
+
+type NonSpeculativeExecution struct{}
+
+func (sp NonSpeculativeExecution) Attempts() int        { return 0 }
+func (sp NonSpeculativeExecution) Delay() time.Duration { return 1 }
```
annismckenzie: Why is the delay 1 in the non-speculative execution policy? 🤔

query_executor.go
```diff
@@ -50,51 +121,46 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
...
-		}
+			// Run the query
+			iter = q.attemptQuery(qry, conn)
+			iter.host = selectedHost.Info()
+			// Update host
+			selectedHost.Mark(iter.err)
+
+			// Exit if the query was successful
+			// or no retry policy defined or retry attempts were reached
+			if rt == nil || iter.err == nil || !rt.Attempt(qry) {
```
annismckenzie: Reorder the statements according to the comment (which is also the logical order to check, even though, yes, it doesn't really matter – a nit is a nit is a nit):
```go
if iter.err == nil || rt == nil || !rt.Attempt(qry) {
```

annismckenzie: Right, came back after I read your comment over in the [other PR](https://github.com/gocql/gocql/pull/1159/files#r218965827). I think we should take another stab at pulling #1151, #1164, and #1162 together into a coherent whole so that we can put speculative on top. I'm not harking on your implementation in here – it reads really fluidly. What it doesn't do right now is handle the case if non-idempotent query being executing speculatively and we could do that if #1162 were implemented by explicitly stopping retries in those cases like write errors as well. On the other hand… one has to specifically turn on speculative execution on a query, right? 🤔 I may be overthinkingn this in that case.

query_executor.go
```diff
@@ -50,51 +121,46 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
...
+			case Ignore:
+				results <- queryResponse{iter: iter}
+				return
+			case RetryNextHost:
+				// retry on the next host
+				selectedHost = hostIter()
+				continue
+			default:
+				// Undefined?
+				results <- queryResponse{iter: iter, err: iter.err}
```
annismckenzie: We had the same problem over in #1151. Ideally, we'd just want to panic here.

query_executor.go
```diff
@@ -30,12 +38,75 @@ func (q *queryExecutor) attemptQuery(qry ExecutableQuery, conn *Conn) *Iter {
...
+	res := <-results
+	if res.iter == nil && res.err == nil {
+		// if we're here, the results channel was closed, so no more hosts
+		return nil, ErrNoConnections
+	}
+	return res.iter, res.err
+}
+
+func (q *queryExecutor) run(qry ExecutableQuery, specWG *sync.WaitGroup, results chan queryResponse, stop chan struct{}) {
+
```
annismckenzie: drop this blank line

query_executor.go
alourie: @annismckenzie Yea, I was thinking to myself whether you're getting back to this :-) So indeed, it is possible that non-idempotent query will be retried, but definitely not in a speculative execution. `Speculative` in this case means that there are multiple executions running at the same time, and the first one to get a response wins. We take care of that in the entrance to the `executeQuery` by forcing the speculative execution policy to NonSpeculative if the policy is not idempotent.

Back to the issue. I'm all for fixing the retries to ack the idempotence, and definitely would be for taking another stab at error handling. I'm not overly sure these depend on this work, they can be layered in any order. It can be pretty decent in size, especially if we include the retries/idempotence handling and actually wrapping errors in their own types. Do you need a hand with that one?

annismckenzie: I would really, really, really appreciate that! 🤝 I'll text you via other channels.

alourie: :+1:

query_executor.go
```diff
@@ -30,12 +38,74 @@ func (q *queryExecutor) attemptQuery(qry ExecutableQuery, conn *Conn) *Iter {
...
+		for i := 0; i < sp.Attempts(); i++ {
+			select {
+			case <-ticker.C:
+				// Launch the additional execution
+				specWG.Add(1)
+				go q.run(qry, &specWG, results, stop)
+			case <-qry.GetContext().Done():
+				// not starting additional executions
+				return
```
beltran: Is it possible that return happens here and `specWG.Wait()` waits forever because the `for` hasn't done all the loops?

alourie: No, it won't. We actually already have at least 1 execution added to the waiting group in line 55 (the main execution).

beltran: Yep, I see

conn_test.go
```diff
@@ -435,6 +444,79 @@ func TestQueryMultinodeWithMetrics(t *testing.T) {
...
+	NumRetries int
+}
+
+func (t *testRetryPolicy) Attempt(qry RetryableQuery) bool {
+	return qry.Attempts() <= t.NumRetries
+}
+func (t *testRetryPolicy) GetRetryType(err error) RetryType {
+	return Retry
+}
+
```
beltran: Test looks good. It'd be great if we could have:
* A test where the context gets done
* A test with delay zero (I'd like this to run with `-race`),which I believe is valid use case

alourie: Sure, I'll have a go at adding that.

PR #1180, opened 2018-09-08
Return routing key for one the elements in the IN clause if they are …

beltran:
…part of the key

See https://github.com/gocql/gocql/issues/1177


Zariel:
please see https://github.com/gocql/gocql/issues/1177#issuecomment-419672095

Zariel:
Also this does not have tests and I doubt it would work, fundamentally this is not a good idea.

beltran:
Why wouldn't it work @Zariel ?

Zariel:
`SELECT * FROM table WHERE key IN (1, 2, 3, 4, 5)`

beltran:
I don't think I understand @Zariel

PR #1182, opened 2018-09-12
Allow to set the custom payload

beltran:
* Moved the customPayload variable from the header to the framer because it's not really part of the header. It's part of the body and can be compressed.
* This PR theoretically allows to set the custom payload in batch and prepare messages but I haven't found a straightforward way for an API to allow this.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -2422,6 +2422,64 @@ func TestTokenAwareConnPool(t *testing.T) {
 	// TODO add verification that the query went to the correct host
 }
 
+func TestCustomPayloadMessages(t *testing.T) {
+	cluster := createCluster()
+	session := createSessionFromCluster(cluster, t)
+	defer session.Close()
+
+	if err := createTable(session, "CREATE TABLE gocql_test.testCustomPayloadMessages (id int, value int, PRIMARY KEY (id))"); err != nil {
+		t.Fatal(err)
+	}
+
+	// QueryMessage
+	var customPayload = map[string][]byte{"a": []byte{10, 20}, "b": []byte{20, 30}}
+	query := session.Query("SELECT id FROM testCustomPayloadMessages where id = ?", 42).Consistency(One).CustomPayload(customPayload)
+	iter := query.Iter()
+	rCustomPayload := iter.GetCustomPayload()
+	if !reflect.DeepEqual(customPayload, rCustomPayload) {
+		t.Fatal("The received custom payload should match the sent")
+	}
+	iter.Close()
+
+	// Insert query
+	query = session.Query("INSERT INTO testCustomPayloadMessages(id,value) VALUES(1, 1)").Consistency(One).CustomPayload(customPayload)
+	iter = query.Iter()
+	rCustomPayload = iter.GetCustomPayload()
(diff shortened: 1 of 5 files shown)


cassandra_test.go
```diff
@@ -2422,6 +2422,64 @@ func TestTokenAwareConnPool(t *testing.T) {
...
+	session := createSessionFromCluster(cluster, t)
+	defer session.Close()
+
+	if err := createTable(session, "CREATE TABLE gocql_test.testCustomPayloadValues (id int, value int, PRIMARY KEY (id))"); err != nil {
+		t.Fatal(err)
+	}
+
+	values := []map[string][]byte{map[string][]byte{"a": []byte{10, 20}, "b": []byte{20, 30}}, nil, map[string][]byte{"a": []byte{10, 20}, "b": nil}}
+
+	for _, customPayload := range values {
```
alourie: Really small nit: is this test doing the same thing as the previous one? Looks kinda the same....

alourie:
> this PR theoretically allows to set the custom payload in batch and prepare messages but I haven't found a straightforward way for an API to allow this.

You could do it similarly to how it's done in `Query` with a setter, maybe for consistency? That's what Java driver does:
**Statement**: `statement.setOutgoingPayload(myCustomPayload);`
**QueryBuilder**: `Statement statement = QueryBuilder.select("c2").from("t1").where(eq("c1", 1)).setOutgoingPayload(myCustomPayload);`
**Batch**: `batch.setOutgoingPayload(myCustomPayload);`

beltran:
Thank you for you review. This PR allows to set the payload for `Query` with the method `CustomPayoad` and for `Batch` with the public field `CustomPayload `. 

However it can't read the payload if it's returned in response to a `writeBatchFrame` because the user doesn't get a public object when executing a batch to which we could attach the payload:
```
err := session.ExecuteBatch(b)
```
This is happening in [this test](https://github.com/gocql/gocql/pull/1182/files#diff-5534e2a9a30aea730ccd0c876cad7117R2455) where the payload is sent to Cassandra, received back, but the user doesn't have access to it.

frame.go
```diff
@@ -1962,6 +1994,16 @@ func appendLong(p []byte, n int64) []byte {
 	)
 }
 
+func (f *framer) writeCustomPayload(customPayload *map[string][]byte) {
+	if len(*customPayload) > 0 {
+		if f.proto < protoVersion4 {
+			panic("Custom payload is not supported with version V3 or less")
+		}
+		f.payload()
```
Zariel: Doing this here has no impact as the header flags are already written

Zariel:
I want to get rid of the Batch execute on Session and make it behave the same as Query

frame.go
beltran: Good catch, I realized that and moved it to [writeFrame](https://github.com/gocql/gocql/pull/1182/files#diff-38a0fa12b0c511105c3a411f0df3e318R842) but forgot to remove it from here

alourie:
> I want to get rid of the Batch execute on Session and make it behave the same as Query

Totally agree, great idea.

Zariel:
#1187

PR #1185, opened 2018-09-13
Fix metadata parser for composite collections

beltran:
Fixes https://github.com/gocql/gocql/issues/1184


nejordan15:
this seems to only solve maps, partially, and leaves a problem with tuples
I am glad there is no panic anymore though :)

PR #1190, opened 2018-09-14
Fix metadata tuple parsing

beltran:
Fixes the tests also https://github.com/gocql/gocql/pull/1189


nejordan15:
Can you add this test:
https://github.com/gocql/gocql/pull/1189/files#diff-33a78ff37088f12c18f6ef27128b87dfR106
Difference being when the map has a tuple as the first item, which fails on master vs. having a tuple as the second item which doesn't.

helpers.go
```diff
@@ -132,14 +132,14 @@ func getCassandraType(name string) TypeInfo {
 			Elem:       getCassandraType(strings.TrimPrefix(name[:len(name)-1], "list<")),
 		}
 	} else if strings.HasPrefix(name, "map<") {
-		names := strings.SplitN(strings.TrimPrefix(name[:len(name)-1], "map<"), ", ", 2)
+		names := splitCompositeTypes(strings.TrimPrefix(name[:len(name)-1], "map<"))
```
nejordan15: though it shouldn't happen it would be good to make sure there is 2 items here to avoid a seg fault below.

helpers.go
```diff
@@ -157,6 +157,41 @@ func getCassandraType(name string) TypeInfo {
...
+			continue
+		}
+		segment += string(char)
+		if char == '<' {
+			lessCount++
+		} else if char == '>' {
+			lessCount--
+			foundMore = true
+		}
+		if lessCount == 0 && foundMore {
```
nejordan15: I dont see the need for this and `foundMore`. The rest of the logic should handle this case.

beltran: `foundMore` is necessary because `lessCount` will be zero when maybe an `<` hasn't been found and it shouldn't get in at that point

helpers.go
beltran: What to do then if not panic?

beltran: I'll print a log at the moment.

helpers.go
nejordan15: If less count is zero, it will hit a , before a <

beltran: Not in the case of a regular type for example `tuple<frozen<tuple<int, int>>, int, frozen<tuple<int, int>>>`, after reading `frozen<tuple<int, int>>` when it gets to `int`, `lessCount` will be zero but it shouldn't get in there. Also the tests fail if we remove `foundMore`.

PR #1194, opened 2018-09-20
conn: avoid spinning in write coalesce with no work

Zariel:
Spinning in the write coalescer when there is no work causes high CPU
load for no reason. Use a second cond to start the flush loop once per
coalesce.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -266,8 +266,9 @@ func (s *Session) dial(host *HostInfo, cfg *ConnConfig, errorHandler ConnErrorHa
 	// dont coalesce startup frames
 	if s.cfg.WriteCoalesceWaitTime > 0 {
 		w := &writeCoalescer{
-			cond: sync.NewCond(&sync.Mutex{}),
-			w:    c.w,
+			fcond: sync.NewCond(&sync.Mutex{}),
+			cond:  sync.NewCond(&sync.Mutex{}),
+			w:     c.w,
 		}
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -640,6 +646,14 @@ func (w *writeCoalescer) flush() {
 }
 
 func (w *writeCoalescer) Write(p []byte) (int, error) {
+	// TODO: use atomics for this?
+	w.fcond.L.Lock()
```
alourie: I think it would be pretty safe here to just mark it as running and broadcasting without checking the current status.

conn.go
```diff
@@ -656,18 +670,34 @@ func (w *writeCoalescer) Write(p []byte) (int, error) {
...
 
+	if !timer.Stop() {
+		<-timer.C
+	}
+
 	for {
+		// wait for a write to start the flush loop
+		w.fcond.L.Lock()
+		for !w.running {
+			w.fcond.Wait()
```
AlexisMontagne: if the connection has been closed without any `Write` this goroutine will be waiting here forever

Zariel: Good spot, do you want to submit a PR? Think it should be sufficient to broadcast if not running?

AlexisMontagne: Sure, i can do it. 

I also experienced an issue while i was running this version of gocql, with a deadlock always waiting for this line and the `Write` calls stuck in L660. I haven't been able to reproduce it in a test in a consistent way.

alourie: and there's now #1200 for the same issue, so clearly reproduced.

PR #1198, opened 2018-09-23
conn: Avoid goroutine leak if the connection closes and there is nothing to write

AlexisMontagne:
Related to https://github.com/gocql/gocql/pull/1194/files/d845a0e2eba7b867634c207bb8f7b84343540fd6#r219705159

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -678,6 +678,11 @@ func (w *writeCoalescer) writeFlusher(interval time.Duration, quit chan struct{}
 		<-timer.C
 	}
 
+	go func() {
+		<-quit
+		w.fcond.Broadcast()
+	}()
+
 	for {
 		// wait for a write to start the flush loop
 		w.fcond.L.Lock()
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -678,6 +678,11 @@ func (w *writeCoalescer) writeFlusher(interval time.Duration, quit chan struct{}
 		<-timer.C
 	}
 
+	go func() {
+		<-quit
+		w.fcond.Broadcast()
```
Zariel: Create a method called Close on writeCoalescer like

 ```go
func (w *writeCoalescer) Close() {
    w.fcond.L.Lock()
    defer w.fcond.L.Unlock()
    if !w.running {
        w.fcond.Broadcast()
     }
}
```

alourie: Pretty sure locking here won't work. Report #1200 shows that there's another goroutine waiting in `writeFlusher` in conn.go#685, which means the w.fcond.L lock has been acquired and will not allow another one. So, if the connection is closed, pretty sure just broadcasting would be enough.

AlexisMontagne: @Zariel having two closing mechanism (quit channel, and closer implementation) sounds pretty confusing two me. Should we really keep both?

alourie: @AlexisMontagne pretty sure it's covered now by https://github.com/gocql/gocql/blob/master/conn.go#L708

PR #1199, opened 2018-09-25
dont trace for internal queries

beltran:
Should fix https://github.com/gocql/gocql/issues/1197


Zariel:
I think this is bit of a bandaid solution, I would rather fix the reason why there are races, either in the test or in the actual tracing code.

beltran:
hmm I see, I agree the bug in https://github.com/gocql/gocql/issues/1197 may still be present but I think this PR still has has value independently @Zariel

PR #1201, opened 2018-09-28
Add test for writeFailure

beltran:


cassandra_test.go
```diff
@@ -1564,6 +1564,51 @@ func TestPrepare_PreparedCacheEviction(t *testing.T) {
...
+	}
+
+	if err := session.Query(`INSERT INTO test.test (id, value)
+	VALUES (1, 1)`).Exec(); err != nil {
+		errWrite, ok := err.(*RequestErrWriteFailure)
+		if ok {
+			if session.cfg.ProtoVersion >= 5 {
+				// Map should get filled and the hosts should've errored
+				if len(errWrite.ErrorMap) > len(cluster.Hosts) {
+					t.Fatal("errWrite.ErrorMap should have length 2, it's: ", len(errWrite.ErrorMap))
```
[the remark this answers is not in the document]
beltran: I'd expect you to be right, but the map gets filled with 2 or 3 hosts depending on the run. This is what Cassandra is sending back.

beltran:
Thank you for the review @alourie

integration.sh
```diff
@@ -50,11 +50,11 @@ function run_tests() {
 	elif [[ $version == 2.2.* || $version == 3.0.* ]]; then
 		proto=4
 		ccm updateconf 'enable_user_defined_functions: true'
-		export JVM_EXTRA_OPTS=" -Dcassandra.custom_query_handler_class=org.apache.cassandra.cql3.CustomPayloadMirroringQueryHandler"
+		export JVM_EXTRA_OPTS=" -Dcassandra.test.fail_writes_ks=test -Dcassandra.custom_query_handler_class=org.apache.cassandra.cql3.CustomPayloadMirroringQueryHandler"
```
alourie: actually, not sure if you need this here as the test will skip if the version is < 3.11.x

beltran: I thought I had removed that `Skip`. I realized Cassandra prior to 3.11 actually had support for that when the tests pass when the `Before` was inverted.

PR #1203, opened 2018-09-29
run TestWriteFailure against all versions

beltran:


Zariel:
can we put this behind a flag or detect if cassandra is configured to run like this? Or maybe put this behind a custom build tag that is used in CI?

alourie:
>  Or maybe put this behind a custom build tag that is used in CI?

Yep, agreed.

alourie:
> This would imply moving this test, and any other that needs Cassandra to be set up with a special environment variable to a different file?

Right. Using a flag is simpler in this case, so I'd say just separate it into a separate run of one test, and that's where you'd put the flag and the java options in integration.sh

beltran:
Since it was two tests and not just one I've finally moved them to a different file.

Zariel:
I was thinking of removing `unit`, having `cassandra` build tag for all tests which need a running cassandra which is possible with the default flags and `integration` which is require tests to be setup as they are in integration.

integration_test.go
```diff
@@ -0,0 +1,189 @@
...
+	if *clusterSize != size {
+		for p, pool := range session.pool.hostConnPools {
+			t.Logf("p=%q host=%v ips=%s", p, pool.host, pool.host.ConnectAddress().String())
+
+		}
+		t.Errorf("Expected a cluster size of %d, but actual size was %d", *clusterSize, size)
+	}
+}
+
+func TestWriteFailure(t *testing.T) {
```
alourie: So, for the versions where we do not set the appropriate JVM extra option this fails. So we either need to check a flag here (i.e., maybe check that the JVM option is set?) or set the JVM flag for all versions.

alourie:
@beltran I think we got the point where parallel Logger access is unavoidable. We should just lock it:

```
diff --git a/logger.go b/logger.go
index bd16d41..3508db5 100644
--- a/logger.go
+++ b/logger.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"log"
+	"sync"
 )
 
 type StdLogger interface {
@@ -13,13 +14,30 @@ type StdLogger interface {
 }
 
 type testLogger struct {
+	mu      sync.RWMutex
 	capture bytes.Buffer
 }
 
-func (l *testLogger) Print(v ...interface{})                 { fmt.Fprint(&l.capture, v...) }
-func (l *testLogger) Printf(format string, v ...interface{}) { fmt.Fprintf(&l.capture, format, v...) }
-func (l *testLogger) Println(v ...interface{})               { fmt.Fprintln(&l.capture, v...) }
-func (l *testLogger) String() string                         { return l.capture.String() }
+func (l *testLogger) Print(v ...interface{}) {
+	l.mu.Lock()
+	fmt.Fprint(&l.capture, v...)
+	l.mu.Unlock()
+}
+func (l *testLogger) Printf(format string, v ...interface{}) {
+	l.mu.Lock()
+	fmt.Fprintf(&l.capture, format, v...)
+	l.mu.Unlock()
+}
+func (l *testLogger) Println(v ...interface{}) {
+	l.mu.Lock()
+	fmt.Fprintln(&l.capture, v...)
+	l.mu.Unlock()
+}
+func (l *testLogger) String() string {
+	l.mu.RLock()
+	defer l.mu.RUnlock()
+	return l.capture.String()
+}
 
 type defaultLogger struct{}
```

This seems to solve the race in the TestStartupTimeout, and frankly I've had this issue plenty of times before.

beltran:
Yes, that's what causing this. My concern if this only happens in the tests we'd be making everything less performant by adding the locks just so the test pass.

alourie:
> My concern if this only happens in the tests we'd be making everything less performant 
Well, the alternative is 3rd party client using the `Logger` directly (as these tests do) and crashing their app with potential races. Additionally, if we see serious performance degradation with `Logger`'s locking, we might want to rethink the approach to logging we now have.

beltran:
We can set `gocqlDebug` to false for the unit tests like the tests were running before.

alourie:
Yep, we can, but there are still areas of code that use Logger without checking the debug flag.

beltran:
But my understanding is that that won't cause a race condition if we are just logging in two different places simultaneously, the problem is when we log (a read occurs) and we assign logger to a new value (a write occurs).
What I'd do for this case is leave the `unit` tag. Rename the current `integration` to `cassandra` and use `integration` for the travis tests.

alourie:
how would Travis tests differ from cassandra?

beltran:
`cassandra` tests would run against any Cassandra and `integration` tests would run against the specially configured Cassandra in travis (certain confs and environment variables)

alourie:
sure, that would work. Would you then disable the debug for integration tests?

beltran:
I think it's only giving problems in the unit tests

alourie:
okay, I'll just wait for the code update then :-) and hopefully understand.

beltran:
This is how I've tried to group the tests:
`unit`: they remain what they were
`cassandra`: they run against any locally running Cassandra cluster
`integration`: some assumptions are made against the Cassandra cluster(they are under `integration_test.sh`), these could be:
 * Number of nodes of the cluster
 * Some configuration option like `enable_user_defined_functions: true`
 * Some environment variable was set to start it like `-Dcassandra.test.fail_writes_ks=test`

PR #1206, opened 2018-10-05
Shard awareness extension for Scylla

dahankzter:
This PR brings shard awareness to the token aware policy for Scylla servers version 2.3 or later.

This is done by implementing the Scylla protocol extension mechanism as as defined by: 
https://github.com/scylladb/scylla/blob/master/docs/protocol-extensions.md
The impact for Scylla users is significant improvement in performance while still maintaining the original performance for Cassandra users.
Both latency and throughput are greatly improved but especially latency can see an order of magnitude improvement.

No config is needed for the user to benefit from this patch, the driver will discover the sharding support on its own.


conn_test.go
```diff
@@ -828,14 +828,19 @@ func TestFrameHeaderObserver(t *testing.T) {
 
 	frames := observer.getFrames()
 
-	if len(frames) != 2 {
+	if len(frames) != 3 {
 		t.Fatalf("Expected to receive 2 frames, instead received %d", len(frames))
```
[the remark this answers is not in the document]
dahankzter: Ouch @JvGinkel I read your message too loosely it seems.
I will correct the message. Good catch!

connpicker.go
```diff
@@ -0,0 +1,87 @@
...
+	Put(*Conn)
+	Remove(conn *Conn)
+	Size() (int, int)
+	Close()
+}
+
+type DefaultConnPicker struct {
+	conns []*Conn
+	pos   uint32
+	size  int
```
alourie: I'd use an unsigned int variant for size.

connpicker.go
```diff
@@ -0,0 +1,87 @@
...
+	mu    sync.RWMutex
+}
+
+func NewDefaultConnPicker(size int) *DefaultConnPicker {
+	return &DefaultConnPicker{
```
alourie: check for size to be non-zero to avoid errors?

connpicker.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func (p *DefaultConnPicker) Remove(conn *Conn) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	for i, candidate := range p.conns {
+		if candidate == conn {
+			p.conns[i] = conn
```
alourie: `p.conns[i] == nil`?

connpicker.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func (p *DefaultConnPicker) Size() (int, int) {
+	sz := len(p.conns)
+	return sz, p.size - sz
+}
+
+func (p *DefaultConnPicker) Pick(token) *Conn {
+	pos := int(atomic.AddUint32(&p.pos, 1) - 1)
```
alourie: Not sure the initial reason for this, but would you mind just reading atomically directly?

policies.go
```diff
@@ -305,16 +305,24 @@ type HostSelectionPolicy interface {
...
 }
 
-type selectedHost HostInfo
+type selectedHost struct {
+	info  *HostInfo
+	token token
+}
+
+func (host selectedHost) Info() *HostInfo {
+	return (*HostInfo)(host.info)
```
alourie: probably no need to cast here

connpicker.go
dahankzter: This keeps track of the current position in the conns list. I think it was always like this to step to the next connection in one atomic go.

dahankzter:
About your major comment @alourie I would also want to isolate and modularize it more but when we implemented this we saw now real option without major refactoring which felt very invasive for such a relatively small patch. It depends of course on what you mean with this backends module? A factory struct that knows how to instantiate backends? 

I will try to address your minor issues asap, thanks for noticing. Nitpicks accumulate so it's good to catch them early!

connpicker.go
dahankzter: It's very viral. This property is threaded through the code and it makes sense to have them as `uint` but perhaps not in this patch?

connpicker.go
dahankzter: I can check it here and return an error. If so what do we do then, panic or default to something?

dahankzter: There is no real infrastructure available to handle errors in this. It all relies on the config defaulting to 2. We can change the assumptions but I'd rather not in a patch that is about something else.

What do you think?

connpicker.go
dahankzter: `p.conns[i] = nil` perhaps? I think we should maybe eve resize it what do you think?

dahankzter: I have set it to nil now. We can change it if you prefer but the rest of the code works guards against nil.

policies.go
dahankzter: Ah, thx!

connpicker.go
alourie: I think panic here is fine.

connpicker.go
alourie: right, ok

connpicker.go
alourie: nah, nil is fine. That's exactly what I meant :-)

alourie:
@dahankzter I don't know what's the best way to implement backends, for now, would need to wait until @Zariel has an opinion. Maybe in the next major API change.

But splitting this into 2 PRs should not be complicated, no? The connection picker and the Scylla's implementation are quite separate, so I'd assume it would be possible to do.

dahankzter:
We can probably split them up but the assignment of the picker to use is dynamic so at some point the code will blend.

There are basically 2 reasons for the code being as is today.

1. Make as small an impact as possible on the core code.
2. Make the usage for users as small as possible. They only need to update the driver now for it to kick in.

That's it really we can make any sort of bootstrapping otherwise if we don't want 2.
Maybe there even is a way to implement it more isolated that we haven't thought of and while we prefer no-config for end users we are not married to the idea. We want to be pragmatic about it and a config option can work fine as well. We just need to tell users to enable it.

mmatczuk:
Hi @alourie,
I have fixed the code to limit scope of Scylla related stuff as much possible. It's split into multiple commits and two PRs. Please find the gocql generic changes in #1210 and Scylla specific changes in #1211 (note the dependency).

cc @dahankzter

dahankzter:
Lets work off of #1210 and #1211 instead.

alourie:
:+1:

PR #1209, opened 2018-10-08
Simplify the flush loop

alourie:
Hopefully, this works.


Zariel:
It will work, but I think it's better to seperate the 2 cases of waiting for a write to start the flush timer then flushing, then waiting for another write

PR #1213, opened 2018-10-11
error if can't retry on more hosts

beltran:
This is breaking tests in master. Can you review @alourie ?

### DIFF
--- a/query_executor.go
+++ b/query_executor.go
@@ -152,6 +152,9 @@ func (q *queryExecutor) run(qry ExecutableQuery, specWG *sync.WaitGroup, results
 			case RetryNextHost:
 				// retry on the next host
 				selectedHost = hostIter()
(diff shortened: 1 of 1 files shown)


query_executor.go
```diff
@@ -152,6 +152,9 @@ func (q *queryExecutor) run(qry ExecutableQuery, specWG *sync.WaitGroup, results
 			case RetryNextHost:
 				// retry on the next host
 				selectedHost = hostIter()
+				if selectedHost == nil {
+					results <- queryResponse{iter: iter}
+				}
```
alourie: right, I thought it's covered but the main loop....so to avoid confusion, let's return after this. Alternatively, we can move the channel loading to after the loop, and check if the iter pointer is not nil, then push it to the results channel. I actually have no preference here on one way or the other.

PR #1217, opened 2018-10-17
use ConnectTimeout for options message

beltran:
Fixes https://github.com/gocql/gocql/issues/1216

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -658,7 +658,7 @@ func TestQueryTimeout(t *testing.T) {
 		if err != ErrTimeoutNoResponse {
 			t.Fatalf("expected to get %v for timeout got %v", ErrTimeoutNoResponse, err)
 		}
-	case <-time.After(10*time.Millisecond + db.cfg.Timeout):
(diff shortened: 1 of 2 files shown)


conn_test.go
```diff
@@ -658,7 +658,7 @@ func TestQueryTimeout(t *testing.T) {
 		if err != ErrTimeoutNoResponse {
 			t.Fatalf("expected to get %v for timeout got %v", ErrTimeoutNoResponse, err)
 		}
-	case <-time.After(10*time.Millisecond + db.cfg.Timeout):
+	case <-time.After(40*time.Millisecond + db.cfg.Timeout):
```
alourie: any specific reason for 40 over 10?

beltran: When running the `stress` I go a failure out of 10000 without that. I assume this was happening before as well. I don't think it will make a difference in the travis tests

alourie: Yea, agree. Just was wondering what's happening that warrants the increase in the waiting there.

PR #1226, opened 2018-10-23
query: dont leak contexts with cancels

### LINKED ISSUE #1224: Memory leak in context.WithCancel
2.2.12

https://github.com/gocql/gocql/commit/22229812a83e

Commit https://github.com/gocql/gocql/commit/08a3a27c42ccc63167b9d470472b85983e17e0e2 introduced memory leak.
image
Rollback to previous version fixes issue.

From https://golang.org/pkg/context/#WithCancel:
_Canceling this context releases resources associated with it, so code should call cancel as soon as the operations running in this Context complete._
So `cancelQuery` should be called after query execution?

Zariel:
Use context to propagate failure of queries when running query
speculation and remove the ability to explicitly cancel queries, instead
relying on the caller to cancel the context provided.

fixes #1224

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -299,7 +299,7 @@ func TestCancel(t *testing.T) {
 	}
 	defer db.Close()
 
-	qry := db.Query("timeout")
+	qry := db.Query("timeout").WithContext(ctx)
 
 	// Make sure we finish the query without leftovers
 	var wg sync.WaitGroup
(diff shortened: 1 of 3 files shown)


Zariel:
cc @alourie

Zariel:
failure is due to context canceled I think, doesnt replicate locally though.

beltran:
I can repro locally @Zariel, gets fixed if I add `results <- queryResponse{err: ctx.Err()}` before [this return](https://github.com/Zariel/gocql/blob/aba4194327a07baa1d682f05aa5cddbd6c00903d/query_executor.go#L125).

Zariel:
closing in favour of #1228 which is more robust

PR #1234, opened 2018-10-30
Enable SetReadDeadline in recv

MichaelS11:


Zariel:
-1 this is just going to cause connections to close themselves when they are idle and have the pool spin dialing them, that goroutine is intended to be blocked reading a header from cassandra for instance the control conn waiting for events.

PR #1240, opened 2018-11-07
doc(session): document Session.ScanCAS nil dest params

### LINKED ISSUE #1235: Support zero parameters to ScanCAS
I am executing

    applied, err := session.Query(`INSERT INTO my_table (key, data) VALUES (?, ?) IF NOT EXISTS`, key, data).ScanCAS(&keyCAS, &dataCAS)

but really I'm just interested in `err != nil` and `applied == true`, not the value of `keyCAS` nor `dataCAS`.

**Expected:** That I can skip the parameters to `ScanCAS` and simply execute

    applied, err := session.Query(`INSERT INTO my_table (key, data) VALUES (?, ?) IF NOT EXISTS`, key, data).ScanCAS()

**What happens:** If I skip the parameters I get an error something like:

    gocql: not enough columns to scan into: have 1 want 3

JensRantil:
This fixes #1235.

### DIFF
--- a/session.go
+++ b/session.go
@@ -1133,6 +1133,10 @@ func (q *Query) Scan(dest ...interface{}) error {
 // statement containing an IF clause). If the transaction fails because
 // the existing values did not match, the previous values will be stored
 // in dest.
+//
+// You must always supply the same number of dest parameters as the number of
+// fields being modified. If there is a count mismatch, an error is returned.
+// Supplying nil for dest parameters not of interest is supported.
 func (q *Query) ScanCAS(dest ...interface{}) (applied bool, err error) {
 	q.disableSkipMetadata = true
 	iter := q.Iter()


session.go
```diff
@@ -1133,6 +1133,10 @@ func (q *Query) Scan(dest ...interface{}) error {
 // statement containing an IF clause). If the transaction fails because
 // the existing values did not match, the previous values will be stored
 // in dest.
+//
+// You must always supply the same number of dest parameters as the number of
+// fields being modified. If there is a count mismatch, an error is returned.
+// Supplying nil for dest parameters not of interest is supported.
```
alourie: This is a bit cryptic. Just say: `Use nil for parameters you don't care for`

JensRantil:
Probably easier you file a PR yourself.

PR #1242, opened 2018-11-15
Allow setting a custom Authenticator

### LINKED ISSUE #1109: approvedAuthenticators - hard coded list blocks use of custom authenticator functions.

### What did you do? I'm trying to utilize a different authenticator routine and am getting an error - my authenticator is not one of the following:
	approvedAuthenticators = [...]string{
		"org.apache.cassandra.auth.PasswordAuthenticator",
		"com.instaclustr.cassandra.auth.SharedSecretAuthenticator",
		"com.datastax.bdp.cassandra.auth.DseAuthenticator",

### What did you see instead?  Error - unexpected authenticator

---

- output of `nodetool status`

### LINKED ISSUE #1238: Authenticator per Host

v2.2.1

latest master

### What did you do?
Attempt to connect to a series of hosts in primary DC. The default cluster settings, `NumConns = 2` and `DisableInitialHostLookup = false` means that each one of the 3 hosts in this DC attempt to authorize my connection. The `Authenticator` interface only allows for setting one global authenticator per cluster while I need a separate authenticator per host since each host's challenge + response will depend on the given hostname (using PKI authentication).

I'd expect to be able to authenticate per host in the cluster with a unique challenge response.

### What did you see instead?
I'm getting WRONG_TARGET error when responding to the second host with the same response I was successfully able to connect to with the first host. 

I'm was currently able to bypass this problem by setting `NumConns = 1` and `DisableInitialHostLookup = true`.. however this is not a long-term solution as it limits the cluster pool connections. 

I see 2 potential solutions to this issue:
1. Don't change the `gocql` library and build an `Authenticator` that is also implements the `HostSelectionPolicy` interface so I'm able to extract the context of which host is currently active or attempting to authorize.
2. Change the `gocql` library to allow setting an `AuthProvider` with the following spec:
```
AuthProvider func(hostname string) (Authenticator, error)
```
This would possibly replace or augment the default authenticator and be called internally when each host is attempting to authenticate. I'm following examples from the JAVA CQL implementation.

alourie:
Fixes #1238 
Fixes #1109 

Signed-off-by: Alex Lourie <djay.il@gmail.com>


alourie:
@zariel, @magaldima something along these lines? If this looks the right direction, I'll add the failure when can't create the `Authenticator` and will add tests.

magaldima:
I'm specifically worried now that the input `host` to the `AuthProvider` is hardcoded as `host.ConnectAddress().String()`. This doesn't solve my use case since I have to authenticate against a domain hostname such as `devxxx12.company.com` (which I include as part of the initialConnectHosts) instead of the IPV4 connect address hostname. It's possible to get around this by building an explicit mapping from IPV4 address to name myself in my custom `Authenticator`.

cluster.go
```diff
@@ -45,22 +45,23 @@ type ClusterConfig struct {
...
-	SerialConsistency  SerialConsistency  // Sets the consistency for the serial part of queries, values can be either SERIAL or LOCAL_SERIAL (default: unset)
+	Timeout            time.Duration                         // connection timeout (default: 600ms)
+	ConnectTimeout     time.Duration                         // initial connection timeout, used during initial dial to server (default: 600ms)
+	Port               int                                   // port (default: 9042)
+	Keyspace           string                                // initial keyspace (optional)
+	NumConns           int                                   // number of connections per host (default: 2)
+	Consistency        Consistency                           // default consistency level (default: Quorum)
+	Compressor         Compressor                            // compression algorithm (default: nil)
+	Authenticator      Authenticator                         // authenticator (default: nil)
+	AuthProvider       func(h string) (Authenticator, error) // an authenticator factory. Can be used to create alternative authenticators (default: nil)
```
Zariel: I think this should take a HostInfo

conn.go
```diff
@@ -215,6 +215,15 @@ func (s *Session) dial(host *HostInfo, cfg *ConnConfig, errorHandler ConnErrorHa
 		},
 	}
 
+	if cfg.AuthProvider != nil {
+		c.auth, err = cfg.AuthProvider(host.ConnectAddress().String())
+		if err != nil {
+			// Fail?
```
Zariel: yes

PR #1245, opened 2018-11-18
repair write coalescing

nightlyone:
Current write coalescing depends on the idea that the passed in
connection can do the writev syscall.

In Go only net.Conns implemented by the net package have this property,
since that the mechanism requires an implementation of a private method,
checked against an interface private to the net package.

So in practise only real TCP connections implement that.

TLS connections cannot do that either, but that is not handled here.

To get a real tcp connection, I broke up the abstraction of the
deadlineWriter and put the effect of it into the flush call.

To make it clearer that we really need a network connection, I changed
the interface passed in to the clearer net.Conn.

This is just the minimal fix to get writev back. A little cleanup work
is needed too.

Without this fix, we get only delayed writes which later happen
in quick succession instead of coalesced writes,
which adds only latency without any beneficial effect.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -252,7 +252,7 @@ func (s *Session) dial(host *HostInfo, cfg *ConnConfig, errorHandler ConnErrorHa
 
 	// dont coalesce startup frames
 	if s.cfg.WriteCoalesceWaitTime > 0 && !cfg.disableCoalesce {
-		c.w = newWriteCoalescer(c.w, s.cfg.WriteCoalesceWaitTime, c.quit)
+		c.w = newWriteCoalescer(conn, c.timeout, s.cfg.WriteCoalesceWaitTime, c.quit)
 	}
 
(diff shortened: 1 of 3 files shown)


alourie:
I'll review this next week.

alourie:
@nightlyone While the code looks nice, there's a race condition in the new `TestWriteCoalescing` test. Would you mind looking into it? Thanks!

nightlyone:
@alourie fixed the test race 

Also added extra commit to disabled write coalescing for connections via TLS, since that won't work anyway and will lead to the same problem of just delaying small writes instead of triggering writev.

alourie:
@nightlyone :+1:

nightlyone:
@alourie looks like the previous test failure was just a flake. With adding a comment it just works :joy: 

So please do a final round of review and tell me anything that prevents merging.

PR #1246, opened 2018-11-30
marshaler float from int64, float64, timestamp from string support

adranwit:
marshaler float from int64, float64, timestamp from string support to integrate with e2e testing framework (viant/endly) via dependencies https://github.com/viant/dsunit/tree/casandra and https://github.com/viant/dsc/tree/casandra


adranwit:
Your call, my use case is to use native JSON types (float64, int64) as a population source for casandra,

PR #1254, opened 2019-01-18
Add Aiven authenticator

relistan:
Hosted Cassandra provider [Aiven](http://aiven.io) uses their own authenticator and it causes `gocql` to fail to connect. This just adds it to the approve list with the others.


alourie:
LGTM

PR #1263, opened 2019-02-05
query executor: better handling of errors for host selection

zllak:
We cannot mark the host as failed in case of logical errors, whether it is from
gocql itself or from the context package.
Ideally, ErrNotFound should not be the only error to be handled, all errors defined in `session.go` should be used.

### DIFF
--- a/query_executor.go
+++ b/query_executor.go
@@ -112,7 +112,15 @@ func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery) *Iter {
 		iter = q.attemptQuery(ctx, qry, conn)
 		iter.host = selectedHost.Info()
 		// Update host
-		selectedHost.Mark(iter.err)
+		switch iter.err {
+		case context.Canceled, context.DeadlineExceeded, ErrNotFound:
+			// those errors represents logical errors, they should not count
+			// toward removing a node from the pool
+			selectedHost.Mark(nil)
+			return iter
+		default:
+			selectedHost.Mark(iter.err)
+		}
 
 		// Exit if the query was successful
 		// or no retry policy defined or retry attempts were reached
--- a/AUTHORS
+++ b/AUTHORS
@@ -109,3 +109,4 @@ Jacob Greenleaf <jacob@jacobgreenleaf.com>
 Alex Lourie <alex@instaclustr.com>; <djay.il@gmail.com>
 Marco Cadetg <cadetg@gmail.com>
 Karl Matthias <karl@matthias.org>
+Thomas Meson <zllak@hycik.org>


query_executor.go
```diff
@@ -112,7 +112,15 @@ func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery) *Iter {
 		iter = q.attemptQuery(ctx, qry, conn)
 		iter.host = selectedHost.Info()
 		// Update host
-		selectedHost.Mark(iter.err)
+		switch iter.err {
+		case context.Canceled, context.DeadlineExceeded, ErrNotFound:
+			// those errors represents logical errors, they should not count
```
alourie: `Deadline exceeded` and `ErrNotFound` do not sound too logical to me. They are errors that mean that the node cannot be contacted, so it seems legit to remove them from the pool.

As for the context.Canceled, if we don't know the reason, probably safer to remove as well.

So I'm not yet convinced :-)

zllak: We had a lot of trouble when runs tests locally where we hit a lot of context canceled and deadline exceeded and not found, and quickly there was no host available in the pool, because those errors was marking the node as not responding correctly.

zllak: Why do we need it:
- We set a 10s timeout on our "main" context for one query (gRPC query).
- We do a lot of scylla calls within this 10s period
- We do a lot of cancellation of queries "mid execution" because the client decided to close its app, making us cancel the context, making the server-side context for the query cancelled as well. That is why we can have a lot of `context.Canceled` errors in gocql.
- We have some queries that does a lot of work, and when there is a bit of lag, whether it's on scylla part or another part of the query, we can see `context.DeadlineExceeded` returned by gocql.
We use gocql in production for a long time now with this patch.

If we did not had this patch in our fork and running in production, anytime scylla experiences a bit of lag, latencies goes up, and we will end up very very quickly removing node from the pool, leaving us with even less available nodes, more stress on them, etc...
We have more than 60k requests per second on our main cluster, a spike of latencies will gocql discard most of the nodes of the cluster. We obviously cannot have that.

For the `ErrNotFound`, with do you consider this as an error as the node not responding correctly ? We use scylla has our main database, and have a lot of use cases where getting an `ErrNotFound` simply means that we tried to get an element that was simply not there, it does not mean that the node did not responded correctly.

Another example:
- We run a lot of local tests, and we do test a lot the "missing" element of our queries. So very quickly, our tests will run out of scylla nodes, because we are running on only one node when testing.

What can we do about this PR ? We currently have a fork, and we do want to upstream all that can be, and this is a good candidate as we cannot be the only one with this kind of problems.

steeve: @alourie you are right that `context.DeadlineExceeded` can mean that the whole node is indeed dead.

That said, it may also be super localised and/or normal error: say we do a particular query with a very aggressive timeout or, as @zllak said, a latency spike on a particular shard (which as you may know is per core on ScyllaDB, not per host as in Cassandra).

Should a particular core be under stress at a given moment (which can happen, for instance, if the payload is large), then the whole host will get kicked out and we're not sure when it comes back (meaning re-registered in the pool). Also, this may mean that all replicas will get kicked too, provoking a cascading failure.
The whole reason we started investigating was because we've seen happen in our tests and in prod (although less), ending up with an empty pool.

As for `ErrNotFound`, if this is an error we get when a host is down, something should be done in order to differentiate that from a real query that was responded with `not_found`, because to me it is a completely legitimate thing to try and query keys that don't exist. Which is why Cassandra and Scylla have bloom filters to fast check that.

For `context.Canceled`, I don't see why this is not a logical error. It's the preferred way to cancel queries in a normal setting (for instance when bound to a gPRC request context).

alourie: @zllak @steeve fair enough.

I'll give session.go errors another look as it's been awhile.

zllak:
@alourie Thanks for the approval !
I was looking at the errors in `session.go` and beside the `ErrNotFound`, the rest seems pretty "fatal" to me. `Unavaiable`, `Unsupported`, `TooManyStmts`, ... all seems to point to errors that cannot be recovered if retried.
Maybe I'm missing something, what do you think ?

alourie:
@zllak thank you for the contribution :-)

I haven't looked too closely in the `session.go`, I'll give it another go.

rkuska:
I am just bumping this as we too use context for request cancellation and I don't think we should remove nodes from pool based on this error.

steeve:
I think this PR can go in before we take a broader look at `session.go` perhaps ?

alourie:
@steeve agreed. Merging in.

PR #1264, opened 2019-02-05
gocql is now apache-2.0 licensed

Zariel:


rkuris:
Can't we keep AUTHORS?

alourie:
Not diving too deep into the legality of this, but how is the change of license managed? Should "we" ask all contributors for this? Clearly all authors can be figured from the commits, but I just want to make sure we do it properly.

If there's no need for an official process, I don't mind dropping authors, and then LGTM.

penberg:
Yeah, the BSD license gocql now uses does not allow relicensing.

It's the part in the license that says:

```
* Redistributions of source code must retain the above copyright notice, this
  list of conditions and the following disclaimer.
```

The safest option for relicensing is to contact all contributors (that are copyright holders), ask for permission to relicense, and also mention everyone who agreed in the commit for audit trail.

alourie:
@penberg so what if not all contributor can be contacted? Cause this can be a major pain in the arse to arrange. Is there a "majority" rule to make it work somehow?

penberg:
@alourie I am not a lawyer so please don't listen to my advise ;-).

I am not aware of such "majority rule". What I think other projects have done for such cases have been to _rewrite_ the parts that could not be relicensed, as discussed in this thread, for example:

https://opensource.stackexchange.com/questions/33/how-can-a-project-be-relicensed

Yes, relicensing can be a major pain, but it's the kind of pain you want to take to ensure there are no copyright violations, which could cause problems in the future.

It seems to me writing a script that sends an email to every author in `git log` is not too much work. You can ask folks to, for example, comment on this pull request to give their explicit acknowledgement for relicensing. Once you have a clear understanding of what parts of the code folks agreed for relicensing, you can make a better decision on how to address the rest.

Zariel:
Thanks for the feedback, my apologies this was the wrong way to do what was intended.

For a bit of background, I intended to contribute Gocql to the Apache Cassandra project and as part of that we will reach out to copyright holders to get permission to contribute their changes to the Apache Foundation.

The license of existing code will remain 3 clause BSD but I want new contributions to contribute the code to Gocql without retaining copyright , this is to make the transition to Apache easier. The current license setup was from before my time as a maintainer of the project, this project pre dates Go1 releases and from what I can tell it uses the same license setup that the Go project uses. The difference between Gocql and the Go project is that Go is owned by Google and all contributors must sign a CLA granting Google various rights to do things with the contributed IP [0].

[0] https://cla.developers.google.com/about/google-individual

PR #1272, opened 2019-03-01
Sort AUTHORS file

martin-sucha:
Adding new entries at the end creates unnecessary merge conflicts,
which could be mostly avoided if the list of authors was sorted.

Added a helper script to manage the file so that there is a single
canonical order.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,7 @@ script:
   - set -e
   - PATH=$PATH:$HOME/.local/bin bash integration.sh $CASS $AUTH
   - go vet .
+  - ./authors.py --check
 
 notifications:
   - email: false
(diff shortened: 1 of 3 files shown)


alourie:
@martin-sucha Not trying to play this down, but would you mind adding a check option to the authors.py and then adding the check to the CI so that if it's not sorted it fails? This way it's going to be followed upon, otherwise we'd have to keep maintaining AUTHORS file _and_ the helper script.

Thanks!

Zariel:
this is a -1 from me, I want to get rid of new contributors adding themselves to the authors file and change the license so that new contributions copyright is assigned to the gocql project.

alourie:
Fair enough @Zariel

PR #1278, opened 2019-03-15
Removing redundant connection + dependant goroutines

alourie:
Fixes #1273

Signed-off-by: Alex Lourie <djay.il@gmail.com>

PR #1279, opened 2019-03-19
Don't dial control connection twice

martin-sucha:
The dial was added in 9a2217868665d0540eaf7de957d71de90274d10b to
disable write coalescing for the control connection, but the
return value from dial was not used. It was probably intended
to change the connect to dial.


martin-sucha:
Oh, this is a duplicate of #1278

PR #1280, opened 2019-03-19
Make sure all connections are observed

martin-sucha:
Currently, control connections used for example to discover protocol
version are not observable using connection observer. This change
aims to make sure all connections created by the driver are seen
by the connection observer.

The `dialWithoutObserver` name is intentionally longer than `dial`
so that one must explicitly acknowledge the connection will not be
observed when using the function.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -159,8 +159,34 @@ type Conn struct {
 	timeouts int64
 }
 
-// Connect establishes a connection to a Cassandra node.
-func (s *Session) dial(host *HostInfo, cfg *ConnConfig, errorHandler ConnErrorHandler) (*Conn, error) {
+// connect establishes a connection to a Cassandra node using session's connection config.
+func (s *Session) connect(host *HostInfo, errorHandler ConnErrorHandler) (*Conn, error) {
+	return s.dial(host, s.connCfg, errorHandler)
+}
+
+// dial establishes a connection to a Cassandra node and notifies the session's connectObserver.
+func (s *Session) dial(host *HostInfo, connConfig *ConnConfig, errorHandler ConnErrorHandler) (*Conn, error) {
+	var obs ObservedConnect
+	if s.connectObserver != nil {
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -159,8 +159,35 @@ type Conn struct {
...
-// Connect establishes a connection to a Cassandra node.
-func (s *Session) dial(host *HostInfo, cfg *ConnConfig, errorHandler ConnErrorHandler) (*Conn, error) {
+// connect establishes a connection to a Cassandra node using session's connection config.
+func (s *Session) connect(host *HostInfo, errorHandler ConnErrorHandler) (*Conn, error) {
+	return s.dial(host, s.connCfg, errorHandler)
+}
+
+// dial establishes a connection to a Cassandra node and notifies the session's connectObserver.
+func (s *Session) dial(host *HostInfo, connConfig *ConnConfig,
+		errorHandler ConnErrorHandler) (*Conn, error) {
```
alourie: @martin-sucha would you mind putting this all in one line? No need to line break

martin-sucha: Fixed.

alourie:
@martin-sucha so this in fact makes _all_ the connections (internally) to be observable now? And if we don't want that, we'd need to use the longer `dialWithoutObserver` specifically?

martin-sucha:
> so this in fact makes all the connections (internally) to be observable now?

@alourie Yes, it passes _all_ connections to the observer. I don't see why any connection should be hidden from the observer. If you want to have metrics or logs of connections from the driver, including control connections makes debugging easier as it can be correlated with other graphs/logs/network capture.

> And if we don't want that, we'd need to use the longer dialWithoutObserver specifically?

Yes. I don't think we'd want to call `dialWithoutObserver` elsewhere than in `dial` though, see above.

alourie:
@martin-sucha fair enough. Merging.

PR #1283, opened 2019-03-27
Helpers function goType supports TypeTime

dahankzter:
TypeTime is handled like TypeTimestamp elsewhere but was missing in the
helpers function goType.

The panic that led to finding this:

```
panic: reflect: New(nil)

goroutine 105 [running]:
reflect.New(0x0, 0x0, 0x0, 0x0, 0x811f40)
	/usr/lib/go/src/reflect/value.go:2287 +0xc3
github.com/gocql/gocql.NativeType.New(0xc00023b504, 0x12, 0x0, 0x0, 0x811f40, 0xc000288380)
	/home/henrik/go/pkg/mod/github.com/scylladb/gocql@v1.0.1/marshal.go:2132 +0x92
github.com/gocql/gocql.(*Iter).RowData(0xc0000f5200, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc00006e000, 0xc00023b850)
	/home/henrik/go/pkg/mod/github.com/scylladb/gocql@v1.0.1/helpers.go:300 +0x66b
github.com/gocql/gocql.(*Iter).MapScan(0xc0000f5200, 0xc0002860c0, 0x0)
	/home/henrik/go/pkg/mod/github.com/scylladb/gocql@v1.0.1/helpers.go:397 +0x4a
github.com/scylladb/gemini.loadSet(0xc0000f5200, 0x831570, 0xc0000f5200, 0xc00027a120)
	/home/henrik/dev/gemini/session.go:143 +0x62
github.com/scylladb/gemini.(*Session).Check(0xc000099f00, 0x819bce, 0x6, 0xc000098e80, 0x1, 0x1, 0xc0000b8900, 0x3, 0x4, 0xc0000c8d00, ...)
	/home/henrik/dev/gemini/session.go:71 +0x161
main.validationJob(0xc000090ba0, 0x819bce, 0x6, 0xc000098e80, 0x1, 0x1, 0xc0000b8900, 0x3, 0x4, 0xc0000c8d00, ...)
	/home/henrik/dev/gemini/cmd/gemini/root.go:232 +0x134
main.Job(0x8a52a0, 0xc000022c80, 0xc000018500, 0xc000090ba0, 0x819bce, 0x6, 0xc000098e80, 0x1, 0x1, 0xc0000b8900, ...)
	/home/henrik/dev/gemini/cmd/gemini/root.go:264 +0x33e
created by main.runJob
	/home/henrik/dev/gemini/cmd/gemini/root.go:163 +0x1ed
```

Maybe the error lies elsewhere but this should go in regardless I believe.

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -26,7 +26,7 @@ func goType(t TypeInfo) reflect.Type {
 		return reflect.TypeOf(*new(string))
 	case TypeBigInt, TypeCounter:
 		return reflect.TypeOf(*new(int64))
-	case TypeTimestamp:
+	case TypeTimestamp, TypeTime:
 		return reflect.TypeOf(*new(time.Time))
 	case TypeBlob:
 		return reflect.TypeOf(*new([]byte))


penberg:
First of all, I'm in favor of fixing the panic with this pull request to be consistent with current implementation.

However, the `time` type in CQL is 64-bit integer that represents nanoseconds since midnight. You'd use it, for example, with time series data where you already know the date and want nanosecond accuracy within that day. It's not clear to me we should be representing this as a `time.Time´. Instead, `u64` seems more appropriate.

Also, quoting the CQL binary protocol specification:

```
6.16 time

  An 8 byte two's complement long representing nanoseconds since midnight.
  Valid values are in the range 0 to 86399999999999

6.17 timestamp

  An 8 byte two's complement integer representing a millisecond-precision
  offset from the unix epoch (00:00:00, January 1st, 1970).  Negative values
  represent a negative offset from the epoch.
```

suggests that `marshalTimestamp` and `unmarshalTimestamp` are not what we want for `time`...

dahankzter:
Any thoughts @Zariel and @alourie perhaps?

alourie:
I can look more into this, but looks as a duplicate of #1276

penberg:
Yeah, the fix to `goType` also exists in #1276, so if that pull request is merged, this becomes obsolete.

dahankzter:
I don't know @alourie it has what seems like an accidental commit wrt the helpers function that concerns the time type but the rest is about the date type right?

Zariel:
tbh these functions should be auto generated from a list of types

dahankzter:
In the meantime perhaps we can merge this one but yes your suggestion to generate these makes sense.

alourie:
@Zariel @dahankzter would you guys mind elaborating on autogeneration? I'm not too familiar with it in golang's context.

dahankzter:
@alourie I just thought it was a good idea in general but I can't elaborate on the complexities for this particular case.

Zariel:
I agree with @penberg it looks like it should be a `time.Duration` on the Go side not `time.Time`

dahankzter:
That would probably make sense @Zariel but afaik the Java driver uses `java.sql.Time` although with Java 8 that could change to `java.time.Duration`. Don't know of any issues than the cognitive mismatch between the cql type `time` and Go type `time.Duration`. Maybe it's not a real problem.

penberg:
I think `time.Duration` is fine. Semantically it makes sense, it has nanosecond resolution, and is large enough to fit the CQL type specification:

```6.16 time

  An 8 byte two's complement long representing nanoseconds since midnight.
  Valid values are in the range 0 to 86399999999999
```

So 👍 from me.

alourie:
:+1:  from me to use `time.Duration` on the Go side, don't think it creates a too big of a cognitive mismatch.

dahankzter:
I have a new PR https://github.com/gocql/gocql/pull/1286 that implements this. It may be missing some for me unknown parts but all the old tests pass and the new I added of course also pass. What do you think @penberg @alourie @Zariel ?

penberg:
The new pull request looks good to me. Let's close this one as obsolete.

Issue #1281, opened 2019-03-20
Data race in tests that change logger


martin-sucha:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?
Go: 1.11 CASS=3.0.17 AUTH=false

### What version of Gocql are you using?
59996b52a6da045fb595758fbcbedd8219bd0914

### What did you do?
There was a data race in tests in Travis CI: https://travis-ci.org/gocql/gocql/jobs/508332390

It looks to be caused by changing global `Logger` variable. The test is accessing the logger like this:

```go
log := &testLogger{}
Logger = log
defer func() {
	Logger = &defaultLogger{}
}()
```

The following tests do this and are
susceptible to the race:

- TestDNSLookupConnected
- TestDNSLookupError
- TestQueryMultinodeWithMetrics
- TestSpeculativeExecution
- TestStartupTimeout

### What did you expect to see?
Tests pass without data race

### What did you see instead?

```
=== RUN   TestDNSLookupConnected
==================
WARNING: DATA RACE
Write at 0x000000b7a8c0 by goroutine 90:
  github.com/gocql/gocql.TestDNSLookupConnected()
      /home/travis/gopath/src/github.com/gocql/gocql/conn_test.go:152 +0xb5
  testing.tRunner()
      /home/travis/.gimme/versions/go1.11.linux.amd64/src/testing/testing.go:827 +0x162
Previous read at 0x000000b7a8c0 by goroutine 85:
  github.com/gocql/gocql.(*hostConnPool).logConnectErr()
      /home/travis/gopath/src/github.com/gocql/gocql/connectionpool.go:454 +0x124
  github.com/gocql/gocql.(*hostConnPool).connectMany.func1()
      /home/travis/gopath/src/github.com/gocql/gocql/connectionpool.go:487 +0x88
Goroutine 90 (running) created at:
  testing.(*T).Run()
      /home/travis/.gimme/versions/go1.11.linux.amd64/src/testing/testing.go:878 +0x650
  testing.runTests.func1()
      /home/travis/.gimme/versions/go1.11.linux.amd64/src/testing/testing.go:1119 +0xa8
  testing.tRunner()
      /home/travis/.gimme/versions/go1.11.linux.amd64/src/testing/testing.go:827 +0x162
  testing.runTests()
      /home/travis/.gimme/versions/go1.11.linux.amd64/src/testing/testing.go:1117 +0x4ee
  testing.(*M).Run()
[...]

Issue #1257, opened 2019-01-24
Add session-specific logger


NWilson:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?

N/A

### What version of Gocql are you using?

Commit 70385f88b28b (released 2018-11-24 15:14:48)

### What did you do?

Gocql exposes the `gocql.StdLogger` object, which is a package-global logger.

### What did you expect to see?

I expected to be able to set the logger on the `Cluster` or `Session` objects. I would like to be able to have multiple Cassandra sessions in my application, and handle the logging separately for each session.

Gocql should fall back to the using the package-global StdLogger only if a session-specific logger is not provided.

### What did you see instead?

Gocql has its logging configuring using a global :(

alourie:
I've looked at this once, could give it another go (pun intended)

martin-sucha:
Fixed with #1553

Issue #901, opened 2017-05-02
A more generic logging interface


Xe:
Hey all,

In Go logging is not simple. Every team and project ends up having their own idea and theory on what logs are and the purpose they serve.

I propose that instead of the current logging interface (which is really low level and wraps a subset of stdlib log.Logger) gocql should use a logging interface like this:

```go
type BetterLogger interface {
  // some logging libraries want the underlying error instance
  Error(ctx context.Context, err error, args ...interface{})
  Errotf(ctx context.Context, err error, format string, args ...interface{})
  Info(ctx context.Context, args ...interface{})
  Infof(ctx context.Context, format string, args ...interface{})
}
```

This better encapsulates the idea of what log information would actually be presented. It also leaves implementation for something other than the standard library as a trivial effort.

Ex impl for stdlib log.Logger:

```go
type stdLogger struct{}

func (s stdLogger) Error(ctx context.Context, err error, args ...interface{}) {
  result := fmt.Sprintf("error: %v", err)
  args = append([]interface{}{result}, args)
  log.Println(args)
}

func (s stdLogger) Errorf(ctx context.Context, err error, format string, args ...interface{}) {
  args = append([]interface{}{error}, args)
  log.Printf(format, args)
}

func (s stdLogger) Info(ctx context.Context, args ...interface{}) {
  log.Println(args)
}

func (s stdLogger) Infof(ctx context.Context, format string, args ...interface{}) {
  log.Printf(format, args)
}
```

OleksiienkoMykyta:
Here is the PR from Martin which makes it better (https://github.com/apache/cassandra-gocql-driver/pull/1553).
I believe that @joao-r-reis worked on refactor for it, but I can't find the source.
Anyway, I think we can close it.

joao-r-reis:
This is related to another issue that is being tracked on JIRA (linked here: https://github.com/apache/cassandra-gocql-driver/pull/1755)

PR #1285, opened 2019-03-29
Logger rework

alourie:
A draft for replacing the global Logger object with Session/Cluster level Loggers.

Addresses #1281 and #1257

Also can be a base for addressing #901 

Signed-off-by: Alex Lourie <djay.il@gmail.com>

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -26,5 +26,3 @@ type defaultLogger struct{}
 func (l *defaultLogger) Print(v ...interface{})                 { log.Print(v...) }
 func (l *defaultLogger) Printf(format string, v ...interface{}) { log.Printf(format, v...) }
 func (l *defaultLogger) Println(v ...interface{})               { log.Println(v...) }
-
-var Logger StdLogger = &defaultLogger{}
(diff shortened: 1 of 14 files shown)


logger.go
```diff
@@ -26,5 +26,3 @@ type defaultLogger struct{}
 func (l *defaultLogger) Print(v ...interface{})                 { log.Print(v...) }
 func (l *defaultLogger) Printf(format string, v ...interface{}) { log.Printf(format, v...) }
 func (l *defaultLogger) Println(v ...interface{})               { log.Println(v...) }
-
-var Logger StdLogger = &defaultLogger{}
```
NWilson: Presumably, for backwards compat you'll have to keep the global Logger object, and use it as the default (fallback) if the cluster/session logger is nil? That is, inside NewCluster/NewSession you'd set the cluster's logger variable based on the config, falling back to the global.

Otherwise it would be an API-breaking change.

alourie: ugh....right. Thanks for pointing out!

I'll look into this on Friday.

martin-sucha:
Closed in favor of #1553

PR #1286, opened 2019-04-01
marshalling: treat cql time type as time.Duration

dahankzter:
After discussions it was proposed to treat time as Go time.Duration.
This makes it consistent with what time is in the spec despite it's name.

The discussion took place in https://github.com/gocql/gocql/pull/1283 which was about a panic from incorrectly handling of the cql type time.

The cassandra spec (https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v5.spec#L986) clearly frames this type as a "duration" although the name hints at a "time". It was agreed that using a `time.Duration` instead makes much sense and any cognitive mismatch was negligible.

### DIFF
--- a/marshal_test.go
+++ b/marshal_test.go
@@ -315,6 +315,20 @@ var marshalTests = []struct {
 		nil,
 	},
 	{
+		NativeType{proto: 4, typ: TypeTime},
+		[]byte("\x00\x00\x01\x40\x77\x16\xe1\xb8"),
+		time.Duration(int64(1376387523000)),
+		nil,
+		nil,
+	},
+	{
+		NativeType{proto: 4, typ: TypeTime},
+		[]byte("\x00\x00\x01\x40\x77\x16\xe1\xb8"),
+		int64(1376387523000),
+		nil,
+		nil,
+	},
+	{
 		NativeType{proto: 2, typ: TypeTimestamp},
 		[]byte("\x00\x00\x01\x40\x77\x16\xe1\xb8"),
 		time.Date(2013, time.August, 13, 9, 52, 3, 0, time.UTC),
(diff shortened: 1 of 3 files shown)


marshal_test.go
```diff
@@ -1217,6 +1231,46 @@ func TestMarshalPointer(t *testing.T) {
...
+	durationS := "1h10m10s"
+	duration, _ := time.ParseDuration(durationS)
+	expectedData := encBigInt(duration.Nanoseconds())
+	var marshalTimeTests = []struct {
+		Info  TypeInfo
+		Data  []byte
+		Value interface{}
+	}{
+		{
+			NativeType{proto: 5, typ: TypeTime},
```
penberg: The `time` and `date` types were added in CQL protocol v4.

marshal_test.go
```diff
@@ -314,6 +314,20 @@ var marshalTests = []struct {
 		nil,
 		nil,
 	},
+	{
+		NativeType{proto: 2, typ: TypeTime},
```
penberg: I think `proto` needs to be `4` here (see the comment below).

marshal_test.go
```diff
@@ -314,6 +314,20 @@ var marshalTests = []struct {
...
+		[]byte("\x00\x00\x01\x40\x77\x16\xe1\xb8"),
+		time.Duration(int64(1376387523000)),
+		nil,
+		nil,
+	},
+	{
+		NativeType{proto: 2, typ: TypeTime},
```
penberg: And here.

marshal_test.go
dahankzter: Fixing!

penberg:
I had some minor comments, but overall looks good to me. 👍

dahankzter:
The protocol version in the tests are fixed and pushed @penberg thanks!

penberg:
It's worth discussing the background and motivation for this change in the pull request description (including the panic) so that we leave a trace in the git logs.

Issue #1287, opened 2019-04-01
Streams use of sync.Pool contended lock


dahankzter:
Please answer these questions before submitting your issue. Thanks!
### What version of Gocql are you using?
fc3925ac2cbd49839569d6a454bf5b66c89f347e

### What did you do?
Profiling mutexes and heap. It turns out that sync.Pools internal lock is fairly contended. 

![Screenshot 2019-04-01 at 15 34 50](https://user-images.githubusercontent.com/171091/55338340-d06fd000-54a0-11e9-8f59-eff11d49e6cb.png)

The attached svg is from another program that exercise the driver even more than a normal program would.

This is of course a discussion. I have no background other than v3 stream bump caused issues for some. 

A standard mutex profile shows similar results:

![Screenshot from 2019-04-01 17-18-34](https://user-images.githubusercontent.com/171091/55339102-4fb1d380-54a2-11e9-96b2-7b5e34ad451a.png)

Zariel:
that suggests that when pooling streams we are not hitting this case https://github.com/golang/go/blob/d6b2b35e641eeac9f764d21dcaed46973b3e2720/src/sync/pool.go#L101 which means we are generally sharing streams across P's. Can you experiment with replacing the pool with a fixed sized buffered chan and having a default select on the Get/Put steps?

dahankzter:
A simple fixed size channel removes this hot spot completely. The `conn::mu` lock now takes over the hot path here but at much lower fractions. The program mutex profile is now dominated by the writeCoalescer::flush().

With sync.Pool:
```
(pprof) list releaseStream
Total: 1.14mins
ROUTINE ======================== github.com/gocql/gocql.(*Conn).releaseStream in /home/henrik/dev/gocql/conn.go
         0     30.94s (flat, cum) 45.35% of Total
         .          .    668:}
         .          .    669:
         .          .    670:func (c *Conn) releaseStream(stream int) {
         .          .    671:	c.mu.RLock()
         .          .    672:	call := c.calls[stream]
         .   102.58ms    673:	c.mu.RUnlock()
         .          .    674:	if call != nil && stream != call.streamID {
         .          .    675:		panic(fmt.Sprintf("attempt to release streamID with invalid stream: %d -> %+v\n", stream, call))
         .          .    676:	} else if call == nil {
         .          .    677:		panic(fmt.Sprintf("releasing a stream not in use: %d", stream))
         .          .    678:	}
         .          .    679:	c.mu.Lock()
[...]

PR #1288, opened 2019-04-02
connection: Replaced sync.Pool with a channel

dahankzter:
The internal sync.Pool lock was subject to contention.
This patch replaces the sync.Pool with a pool based on
a simple channel. The pool size can be configured by
the user through ClusterConfig::StreamPoolSize.

Fixes: #1287


Zariel:
superseded by #1289

PR #1290, opened 2019-04-02
make session immuatable

### LINKED ISSUE #357: CASSGO-23 Make Session immutable

Zariel:
This removes the ability to change the consistency, tracer, page
prefetch and page size after the session has been created globally.
Instead they can continue to be change on a per query level. This allows
us to remove the RWMutex in query creation.

fixes #357

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -124,17 +124,6 @@ func TestTracing(t *testing.T) {
 		t.Fatal("select: failed to obtain any tracing")
 	}
 
-	// also works from session tracer
-	session.SetTrace(trace)
-	trace.mu.Lock()
-	buf.Reset()
-	trace.mu.Unlock()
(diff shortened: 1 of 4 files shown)


rkuska:
@alourie Does it matter if there is no versioning?

alourie:
@rkuska Clearly it might break someone's code. I did approve the change, so I don't think it's critical in this case, but just wanted to make sure we didn't miss it if important.

Zariel:
We can reimplement the methods and use an atomic value to still remove the
lock and deprecate the methods

alourie:
@Zariel yes, I think it would be better. We can keep a note/label here maybe for the future API change to incorporate this?

mmatczuk:
Maybe it's time for v2 of this work?

Issue #1291, opened 2019-04-03
Contention under higher load


rkuska:
### What version of Gocql are you using?
https://github.com/gocql/gocql/commit/0e1d5de854dfd269e343440b9807f43a5f2e62cf

### What did you do?
Running a profiler on my service in production on some percentage of traffic.

### What did you expect to see?
Service should be able to scale with load.

### What did you see instead?
Contention with higher CPU count and increased load. See svg:

<img width="671" alt="Screenshot 2019-04-03 at 11 14 29" src="https://user-images.githubusercontent.com/1709417/55467675-155f4800-5602-11e9-9c69-5b5f98a443a8.png">

Output of peek in pprof tool:

```
      flat  flat%   sum%        cum   cum%   calls calls% + context 	 	 
----------------------------------------------------------+-------------
                                          134.28ms 47.07% |   sync.(*Pool).getSlow /usr/local/go/src/sync/pool.go:173
                                           55.30ms 19.38% |   github.com/gocql/gocql.(*preparedLRU).execIfMissing /go/pkg/mod/github.com/scylladb/gocql@v1.0.4/prepared_cache.go:55
                                           32.52ms 11.40% |   github.com/gocql/gocql.(*Session).routingKeyInfo /go/pkg/mod/github.com/scylladb/gocql@v1.0.4/session.go:459
                                           27.03ms  9.48% |   sync.(*Pool).pinSlow /usr/local/go/src/sync/pool.go:205
                                           22.67ms  7.95% |   context.(*cancelCtx).Done /usr/local/go/src/context/context.go:330
                                            8.30ms  2.91% |   sync.(*Pool).Get /usr/local/go/src/sync/pool.go:139
                                            3.89ms  1.36% |   sync.(*Pool).getSlow /usr/local/go/src/sync/pool.go:170
                                            0.97ms  0.34% |   sync.(*Pool).Put /usr/local/go/src/sync/pool.go:109
                                            0.35ms  0.12% |   sync.(*Pool).pinSlow /usr/local/go/src/sync/pool.go:215
  285.30ms 94.16% 94.16%   285.30ms 94.16%                | sync.(*Mutex).Unlock /usr/local/go/src/sync/mutex.go:201
```

[...]

PR #1293, opened 2019-04-03
prepared statements: reduce contentions

dahankzter:
A sharded layer on top of the existing LRU cache dispatches the keys across 256 buckets.
The hashing used is Highwayhash which should provide accurate dispersion and sufficient speed.

This is a work in progress. It needs some cleanup and an analysis as to the hash key generation.

Fixes: #1291

### DIFF
--- a/prepared_cache.go
+++ b/prepared_cache.go
@@ -1,64 +1,121 @@
 package gocql
 
 import (
-	"github.com/gocql/gocql/internal/lru"
+	"encoding/hex"
 	"sync"
+
+	"github.com/gocql/gocql/internal/lru"
+	"github.com/minio/highwayhash"
 )
 
-const defaultMaxPreparedStmts = 1000
+const (
+	defaultMaxPreparedStmts = 1000
+	shardCount              = 256
+)
 
 // preparedLRU is the prepared statement cache
 type preparedLRU struct {
-	mu  sync.Mutex
-	lru *lru.Cache
+	shards [shardCount]*shard
+	key    []byte
+}
+
+type shard struct {
+	items *lru.Cache
+	lock  *sync.RWMutex
+}
+
+// newPreparedLRU creates and initializes a * preparedLRU instance.
+func newPreparedLRU(size int) *preparedLRU {
+	key, err := hex.DecodeString("000102030405060708090A0B0C0D0E0FF0E0D0C0B0A090807060504030201000")
+	if err != nil {
+		panic(err)
+	}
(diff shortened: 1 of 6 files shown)


penberg:
Oh my, looks like we are taking a global lock per each `prepareStatement` call.

Did you consider using one of the lock-free hash map implementations (e.g., https://github.com/cornelk/hashmap)? Alternatively, I'd assume RWMutex is less expensive for the fast-path case where query is in the prepared statement cache.

dahankzter:
Where are we taking a global lock? The main idea here is to shard the locks
over a number of buckets.

dahankzter:
But yes @penberg we may get away with another hashmap unless there is a real reason for using the LRU functionality. What do you think @Zariel ?

dahankzter:
Actually, if we assume the number of prepared statements are relatively small then using the lock free cache  https://github.com/cornelk/hashmap seems to be a good idea. Since it supports proper CAS we can avoid the closure bit in execIfMissing without much problems.

prepared_cache.go
```diff
@@ -1,64 +1,121 @@
...
 
 func (p *preparedLRU) clear() {
-	p.mu.Lock()
-	defer p.mu.Unlock()
-
-	for p.lru.Len() > 0 {
-		p.lru.RemoveOldest()
+	for _, shard := range p.shards {
+		shard.lock.Lock()
+		if shard.items.Len() > 0 {
```
alourie: Should this be a loop over all items?

prepared_cache.go
```diff
@@ -1,64 +1,121 @@
...
-	return fn(p.lru), false
+	return fn(shard.items), false
 }
 
 func (p *preparedLRU) keyFor(addr, keyspace, statement string) string {
 	// TODO: maybe use []byte for keys?
 	return addr + keyspace + statement
 }
+
+func (c *preparedLRU) get(key string) *inflightPrepare {
```
alourie: Can we also keep it consistent as either `c *preparedLRU` or `p *preparedLRU`

dahankzter: Sure we can but I think "alt 2" is currently the most promising solution. Simple and should also work now ;)

dahankzter: Sorry I meant "alt 3", the one with a simple read/write lock.

alourie: :+1:

dahankzter:
Closing, superseded by #1295 .

PR #1294, opened 2019-04-03
prepared statements: reduce contentions, alt 2

dahankzter:
A lockfree hashmap is used in place of the current LRU cache. If this tradeoff is worth it
it gives a great boon to scalability.

### DIFF
--- a/prepared_cache.go
+++ b/prepared_cache.go
@@ -1,61 +1,43 @@
 package gocql
 
 import (
-	"github.com/gocql/gocql/internal/lru"
-	"sync"
+	"github.com/cornelk/hashmap"
 )
 
 const defaultMaxPreparedStmts = 1000
 
 // preparedLRU is the prepared statement cache
 type preparedLRU struct {
-	mu  sync.Mutex
-	lru *lru.Cache
+	cache *hashmap.HashMap
 }
 
 // Max adjusts the maximum size of the cache and cleans up the oldest records if
 // the new max is lower than the previous value. Not concurrency safe.
 func (p *preparedLRU) max(max int) {
-	p.mu.Lock()
-	defer p.mu.Unlock()
-
-	for p.lru.Len() > max {
-		p.lru.RemoveOldest()
+	if p.cache.Len() > max {
+		p.cache.Grow(uintptr(max))
 	}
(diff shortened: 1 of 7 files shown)


prepared_cache.go
```diff
@@ -1,61 +1,43 @@
...
 
 // Max adjusts the maximum size of the cache and cleans up the oldest records if
 // the new max is lower than the previous value. Not concurrency safe.
 func (p *preparedLRU) max(max int) {
-	p.mu.Lock()
-	defer p.mu.Unlock()
-
-	for p.lru.Len() > max {
-		p.lru.RemoveOldest()
+	if p.cache.Len() > max {
```
alourie: So now this is not matching the description, as it grows the current cache instead of clearing the old entries. If the implementation is beneficial, I'd update the description to match ;-)

dahankzter: Yes I just hacked away. It probably needs more docs updates. :)

dahankzter:
It's simpler because all the nitty gritty is under the hood but yes I like it's properties a lot. So if we can either ignore the "lru" features of the current cache or introduce another object:
```go
type elem struct {
    inflight *inflightPrepare
    touched int64
}
```

Where we update the objects timestamp when used and then in a new separate goroutine we launch a ticker to periodically evict the least recently used items. We can also piggyback on the existing calls and rescale as needed but perhaps that kind of precision isn't needed. What do you think @alourie @Zariel @penberg ?

prepared_cache.go
penberg: But won't this mean the cache can grow without limits now? We had a similar issue on Scylla side and discovered that there are workloads that will eventually make your application run out of memory...

dahankzter:
If that's the case then we may need the wrapper object to track the time.
Or simply just remove random entries if the user just keeps generating
unique queries.

alourie:
Well, by adding this we complicate things yet again. As for the time, you can have a ticker in the separate goroutine that handles the eviction of the last recently used items, so that's not a real concern.

As I'm not familiar with the initial considerations for the LRU that we have, and I have no personal preference here, I'll wait for @Zariel to weigh in.

dahankzter:
Closing, superseded by #1295 .

PR #1295, opened 2019-04-04
prepared statements: reduce contentions, alt 3

dahankzter:
A simple readlock in the fast path of prepared statements cache.
Protected by an additional read to ensure atomic creation of inflights.

Simplest possible fix. Solution thanks to @penberg and @rkuska and this has very good contention characteristics. It is also infinitely smaller in scope than the previous alternatives.

Fixes: #1291

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -1484,9 +1484,6 @@ func TestPrepare_PreparedCacheEviction(t *testing.T) {
 		t.Fatalf("insert into prepcachetest failed, error '%v'", err)
 	}
 
-	session.stmtsLRU.mu.Lock()
-	defer session.stmtsLRU.mu.Unlock()
-
 	//Make sure the cache size is maintained
 	if session.stmtsLRU.lru.Len() != session.stmtsLRU.lru.MaxEntries {
 		t.Fatalf("expected cache size of %v, got %v", session.stmtsLRU.lru.MaxEntries, session.stmtsLRU.lru.Len())
--- a/internal/lru/lru.go
+++ b/internal/lru/lru.go
@@ -18,9 +18,13 @@ limitations under the License.
 // Package lru implements an LRU cache.
 package lru
 
-import "container/list"
+import (
+	"container/list"
+	"sync"
+	"time"
+)
 
-// Cache is an LRU cache. It is not safe for concurrent access.
+// Cache is an LRU cache. It is safe for concurrent access.
 //
 // This cache has been forked from github.com/golang/groupcache/lru, but
 // specialized with string keys to avoid the allocations caused by wrapping them
@@ -34,64 +38,130 @@ type Cache struct {
 	// executed when an entry is purged from the cache.
 	OnEvicted func(key string, value interface{})
 
-	ll    *list.List
-	cache map[string]*list.Element
+	window time.Duration
+	ll     *list.List
+	cache  map[string]*list.Element
+	mu     sync.RWMutex
 }
 
 type entry struct {
-	key   string
-	value interface{}
+	key       string
+	value     interface{}
+	timestamp time.Time
 }
 
 // New creates a new Cache.
 // If maxEntries is zero, the cache has no limit and it's assumed
 // that eviction is done by the caller.
-func New(maxEntries int) *Cache {
+// If window is zero the "windowed updates" of the LRU list will be disabled.
+// For further information on this technique see: https://www.openmymind.net/High-Concurrency-LRU-Caching/
+func New(maxEntries int, window time.Duration) *Cache {
 	return &Cache{
 		MaxEntries: maxEntries,
 		ll:         list.New(),
 		cache:      make(map[string]*list.Element),
+		window:     window,
 	}
 }
 
 // Add adds a value to the cache.
-func (c *Cache) Add(key string, value interface{}) {
-	if c.cache == nil {
-		c.cache = make(map[string]*list.Element)
-		c.ll = list.New()
-	}
-	if ee, ok := c.cache[key]; ok {
-		c.ll.MoveToFront(ee)
-		ee.Value.(*entry).value = value
+func (c *Cache) Add(key string, val interface{}) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if ele, ok := c.cache[key]; ok {
+		c.ll.MoveToFront(ele)
+		ele.Value.(*entry).value = val
 		return
 	}
-	ele := c.ll.PushFront(&entry{key, value})
+	c.addLocked(key, val)
+}
+
+func (c *Cache) addLocked(key string, val interface{}) {
+	timestamp := time.Now()
+	ent := &entry{key, val, timestamp}
+	ele := c.ll.PushFront(ent)
 	c.cache[key] = ele
 	if c.MaxEntries != 0 && c.ll.Len() > c.MaxEntries {
-		c.RemoveOldest()
+		c.removeOldestLocked()
 	}
 }
 
 // Get looks up a key's value from the cache.
 func (c *Cache) Get(key string) (value interface{}, ok bool) {
-	if c.cache == nil {
-		return
+	timestamp := time.Now()
+
+	if c.window == 0 {
+		return c.getWithLock(key, timestamp)
 	}
-	if ele, hit := c.cache[key]; hit {
+
+	c.mu.RLock()
+	if ele, ok := c.cache[key]; ok {
+		ent := ele.Value.(*entry)
+		if timestamp.After(ent.timestamp.Add(c.window)) {
+			c.mu.RUnlock()
+			// We dropped the lock so we need to perform the lookup again.
+			return c.getWithLock(key, timestamp)
+		} else {
+			v := ent.value
+			c.mu.RUnlock()
+			return v, true
+		}
+	}
+	c.mu.RUnlock()
+	return
+}
+
+func (c *Cache) getWithLock(key string, timestamp time.Time) (interface{}, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if ele, ok := c.cache[key]; ok {
+		ent := ele.Value.(*entry)
+		ent.timestamp = timestamp
 		c.ll.MoveToFront(ele)
+		return ent.value, true
+	} else {
+		return nil, false
+	}
+}
+
+// GetOrInsert either retrieves the current entry for the key or inserts the
+// provided value for the key.
+// It returns true if the value was loaded and false if the provided value
+// was stored.
+func (c *Cache) GetOrInsert(key string, val interface{}) (interface{}, bool) {
+	// Optimistically try a Get first
+	value, ok := c.Get(key)
+	if ok {
+		return value, ok
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// We need to check again
+	if ele, ok := c.cache[key]; ok {
 		return ele.Value.(*entry).value, true
 	}
-	return
+
+	c.addLocked(key, val)
+	return val, false
+}
+
+// Len returns the number of items in the cache.
+func (c *Cache) Len() int {
+	c.mu.RLock()
+	l := c.ll.Len()
+	c.mu.RUnlock()
+	return l
 }
 
 // Remove removes the provided key from the cache.
 func (c *Cache) Remove(key string) bool {
-	if c.cache == nil {
-		return false
-	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
 
 	if ele, hit := c.cache[key]; hit {
-		c.removeElement(ele)
+		c.removeElementLocked(ele)
 		return true
 	}
 
@@ -100,28 +170,23 @@ func (c *Cache) Remove(key string) bool {
 
 // RemoveOldest removes the oldest item from the cache.
 func (c *Cache) RemoveOldest() {
-	if c.cache == nil {
-		return
-	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.removeOldestLocked()
+}
+
+func (c *Cache) removeOldestLocked() {
 	ele := c.ll.Back()
 	if ele != nil {
-		c.removeElement(ele)
+		c.removeElementLocked(ele)
 	}
 }
 
-func (c *Cache) removeElement(e *list.Element) {
+func (c *Cache) removeElementLocked(e *list.Element) {
 	c.ll.Remove(e)
 	kv := e.Value.(*entry)
 	delete(c.cache, kv.key)
 	if c.OnEvicted != nil {
 		c.OnEvicted(kv.key, kv.value)
 	}
-}
-
-// Len returns the number of items in the cache.
-func (c *Cache) Len() int {
-	if c.cache == nil {
-		return 0
-	}
-	return c.ll.Len()
 }
--- a/prepared_cache.go
+++ b/prepared_cache.go
@@ -1,24 +1,24 @@
 package gocql
 
 import (
+	"time"
+
 	"github.com/gocql/gocql/internal/lru"
-	"sync"
 )
 
-const defaultMaxPreparedStmts = 1000
+const (
+	defaultMaxPreparedStmts    = 1000
+	defaultPreparedStmtsWindow = time.Second
+)
 
 // preparedLRU is the prepared statement cache
 type preparedLRU struct {
-	mu  sync.Mutex
 	lru *lru.Cache
 }
 
 // Max adjusts the maximum size of the cache and cleans up the oldest records if
 // the new max is lower than the previous value. Not concurrency safe.
 func (p *preparedLRU) max(max int) {
-	p.mu.Lock()
-	defer p.mu.Unlock()
-
 	for p.lru.Len() > max {
 		p.lru.RemoveOldest()
 	}
@@ -26,36 +26,27 @@ func (p *preparedLRU) max(max int) {
 }
 
 func (p *preparedLRU) clear() {
-	p.mu.Lock()
-	defer p.mu.Unlock()
-
 	for p.lru.Len() > 0 {
 		p.lru.RemoveOldest()
 	}
 }
 
 func (p *preparedLRU) add(key string, val *inflightPrepare) {
-	p.mu.Lock()
-	defer p.mu.Unlock()
 	p.lru.Add(key, val)
 }
 
 func (p *preparedLRU) remove(key string) bool {
-	p.mu.Lock()
-	defer p.mu.Unlock()
 	return p.lru.Remove(key)
 }
 
-func (p *preparedLRU) execIfMissing(key string, fn func(lru *lru.Cache) *inflightPrepare) (*inflightPrepare, bool) {
-	p.mu.Lock()
-	defer p.mu.Unlock()
-
+func (p *preparedLRU) execIfMissing(key string, fn func() *inflightPrepare) (*inflightPrepare, bool) {
 	val, ok := p.lru.Get(key)
 	if ok {
 		return val.(*inflightPrepare), true
 	}
 
-	return fn(p.lru), false
+	val, ok = p.lru.GetOrInsert(key, fn())
+	return val.(*inflightPrepare), ok
 }
 
 func (p *preparedLRU) keyFor(addr, keyspace, statement string) string {
--- a/session.go
+++ b/session.go
@@ -118,7 +118,7 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 		prefetch:        0.25,
 		cfg:             cfg,
 		pageSize:        cfg.PageSize,
-		stmtsLRU:        &preparedLRU{lru: lru.New(cfg.MaxPreparedStmts)},
+		stmtsLRU:        &preparedLRU{lru: lru.New(cfg.MaxPreparedStmts, cfg.PreparedStmtWindow)},
 		quit:            make(chan struct{}),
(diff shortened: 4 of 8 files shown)


dahankzter:
Ok so this isn't safe after all:
```go
func (c *Cache) Get(key string) (value interface{}, ok bool) {
	if c.cache == nil {
		return
	}
	if ele, hit := c.cache[key]; hit {
		c.ll.MoveToFront(ele)
		return ele.Value.(*entry).value, true
	}
	return
}```

This results in:
```
==================
WARNING: DATA RACE
Read at 0x00c420362300 by goroutine 9:
  container/list.(*List).MoveToFront()
      /home/travis/.gimme/versions/go1.10.linux.amd64/src/container/list/list.go:110 +0xb4
  github.com/gocql/gocql/internal/lru.(*Cache).Get()
      /home/travis/gopath/src/github.com/gocql/gocql/internal/lru/lru.go:81 +0xfe
  github.com/gocql/gocql.(*preparedLRU).execIfMissing()
      /home/travis/gopath/src/github.com/gocql/gocql/prepared_cache.go:52 +0x8c
  github.com/gocql/gocql.(*Conn).prepareStatement()
      /home/travis/gopath/src/github.com/gocql/gocql/conn.go:957 +0x22a
  github.com/gocql/gocql.(*Conn).executeQuery()
      /home/travis/gopath/src/github.com/gocql/gocql/conn.go:1071 +0x228
  github.com/gocql/gocql.(*Conn).query()
      /home/travis/gopath/src/github.com/gocql/gocql/conn.go:1354 +0x118
  github.com/gocql/gocql.(*Conn).awaitSchemaAgreement()
      /home/travis/gopath/src/github.com/gocql/gocql/conn.go:1367 +0x72a
  github.com/gocql/gocql.(*Conn).executeQuery()
      /home/travis/gopath/src/github.com/gocql/gocql/conn.go:1174 +0x12fe
  github.com/gocql/gocql.(*Query).execute()
      /home/travis/gopath/src/github.com/gocql/gocql/session.go:888 +0x56
```

**Comment by penberg** [comment]
Yeah, `Get` would need to take a write lock because it's updating the linked list that maintains the LRU ordering. But then we are no longer actually addressing the contention issue. Since with a lock-free hash map is not sufficient to maintain the LRU property, I am beginning to think your original approach of lock partition makes the most sense.

**Comment by dahankzter** [comment]
Updated PR:

Synchronization is moved into the LRU cache to allow for more granular locking.
This provides a good increase in the common fast path as well as safety in the slow path.

**Comment by dahankzter** [comment]
What do you think @Zariel @alourie @penberg ?

File: internal/lru/lru.go [hunk 1]
```diff
@@ -55,40 +61,69 @@ func New(maxEntries int) *Cache {
...
 
 // Get looks up a key's value from the cache.
 func (c *Cache) Get(key string) (value interface{}, ok bool) {
-	if c.cache == nil {
-		return
+	timestamp := time.Now()
+	c.lock.RLock()
+	if ele, ok := c.cache[key]; ok {
+		ent := ele.Value.(*entry)
+		if timestamp.After(ent.timestamp.Add(time.Second * 1)) {
```
penberg: We should probably make this magic number configurable via session or something.

internal/lru/lru.go
```diff
@@ -34,13 +38,15 @@ type Cache struct {
...
 	ll    *list.List
 	cache map[string]*list.Element
 }
 
 type entry struct {
-	key   string
-	value interface{}
+	key       string
+	value     interface{}
+	timestamp time.Time
```
penberg: Let's add a comment here that explains what `timestamp` is used for. That is, mention that we eliminate the need to take write-lock for every LRU lookup (needed to keep items ordered by LRU in `ll`) by delaying the update until some time period has passed (1 second in this series). We can do this because most recently accessed items are likely to be at the top of the list anyway.

Don't know if we need to mention this, but the technique is mentioned in the following blog post as "windowed updates":

https://www.openmymind.net/High-Concurrency-LRU-Caching/

internal/lru/lru.go
```diff
@@ -55,40 +61,69 @@ func New(maxEntries int) *Cache {
...
+		}
+		return ent.value, true
 	}
-	if ele, hit := c.cache[key]; hit {
-		c.ll.MoveToFront(ele)
+	c.lock.RUnlock()
+	return
+}
+
+func (c *Cache) GetOrInsert(key string, val interface{}) (interface{}, bool) {
```
penberg: Do we need to export this function? IOW, why is it not `getOrInsert`?

internal/lru/lru.go
```diff
@@ -55,40 +61,69 @@ func New(maxEntries int) *Cache {
...
+	value, ok := c.Get(key)
+	if ok {
+		return value, ok
+	}
+
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	// We need to check again
+	if ele, ok := c.cache[key]; ok {
```
penberg: We should have the same LRU list movement and timestamp update here.

prepared_cache.go
```diff
@@ -2,60 +2,49 @@ package gocql
...
 	if ok {
 		return val.(*inflightPrepare), true
 	}
 
-	return fn(p.lru), false
+	val, ok = p.lru.GetOrInsert(key, fn())
+	if ok {
+		return val.(*inflightPrepare), true
+	}
+	return val.(*inflightPrepare), false
```
penberg: Could be simplified to

```return val.(*inflightPrepare), ok```

No?

File: internal/lru/lru.go (hunk 4 above)
Comment (dahankzter) [reply internal/lru/lru.go]: Ah in the internal `add` we need or around it somewhere.

File: prepared_cache.go (hunk 5 above)
Comment (dahankzter) [reply prepared_cache.go]: Most likely.

File: internal/lru/lru.go (hunk 1 above)
Comment (dahankzter) [reply internal/lru/lru.go]: We can inject it via config vars.

File: internal/lru/lru.go (hunk 4 above)
Comment (dahankzter) [reply internal/lru/lru.go]: No actually the `add` already does this.

File: cassandra_test.go [hunk 6]
```diff
@@ -1484,16 +1484,14 @@ func TestPrepare_PreparedCacheEviction(t *testing.T) {
...
-	defer session.stmtsLRU.mu.Unlock()
-
 	//Make sure the cache size is maintained
 	if session.stmtsLRU.lru.Len() != session.stmtsLRU.lru.MaxEntries {
 		t.Fatalf("expected cache size of %v, got %v", session.stmtsLRU.lru.MaxEntries, session.stmtsLRU.lru.Len())
 	}
 
 	// Walk through all the configured hosts and test cache retention and eviction
 	for _, host := range session.cfg.Hosts {
+
```
mmatczuk: unnecessary whitespace

prepared_cache.go
```diff
@@ -2,60 +2,49 @@ package gocql
...
 	"github.com/gocql/gocql/internal/lru"
-	"sync"
 )
 
 const defaultMaxPreparedStmts = 1000
 
 // preparedLRU is the prepared statement cache
 type preparedLRU struct {
-	mu  sync.Mutex
 	lru *lru.Cache
```
mmatczuk: Why not use embedding and remove the proxy functions?

internal/lru/lru.go
```diff
@@ -18,7 +18,11 @@ limitations under the License.
...
 package lru
 
-import "container/list"
+import (
+	"container/list"
+	"sync"
+	"time"
+)
 
 // Cache is an LRU cache. It is not safe for concurrent access.
```
mmatczuk: Hmmmmmm @dahankzter?

internal/lru/lru.go
```diff
@@ -34,61 +38,96 @@ type Cache struct {
...
+	lock   sync.RWMutex
+	ll     *list.List
+	cache  map[string]*list.Element
+	window time.Duration
 }
 
 type entry struct {
-	key   string
-	value interface{}
+	key       string
```
mmatczuk: Why a key is needed?

internal/lru/lru.go
```diff
@@ -34,61 +38,96 @@ type Cache struct {
 	// executed when an entry is purged from the cache.
 	OnEvicted func(key string, value interface{})
 
-	ll    *list.List
-	cache map[string]*list.Element
+	lock   sync.RWMutex
```
mmatczuk: traditionally it's called mu and we put it below stuff it protects

internal/lru/lru.go
```diff
@@ -34,61 +38,96 @@ type Cache struct {
...
 func (c *Cache) Get(key string) (value interface{}, ok bool) {
-	if c.cache == nil {
-		return
+	timestamp := time.Now()
+	c.lock.RLock()
+	if ele, ok := c.cache[key]; ok {
+		ent := ele.Value.(*entry)
+		if timestamp.After(ent.timestamp.Add(c.window)) {
+			c.lock.RUnlock()
+			c.lock.Lock()
```
mmatczuk: Maybe refactor to a function?

internal/lru/lru.go
```diff
@@ -34,61 +38,96 @@ type Cache struct {
...
+		return ent.value, true
 	}
-	if ele, hit := c.cache[key]; hit {
-		c.ll.MoveToFront(ele)
+	c.lock.RUnlock()
+	return
+}
+
+func (c *Cache) GetOrInsert(key string, val interface{}) (interface{}, bool) {
+
```
mmatczuk: not needed whitespace

internal/lru/lru.go
```diff
@@ -34,61 +38,96 @@ type Cache struct {
...
+		}
+		return ent.value, true
 	}
-	if ele, hit := c.cache[key]; hit {
-		c.ll.MoveToFront(ele)
+	c.lock.RUnlock()
+	return
+}
+
+func (c *Cache) GetOrInsert(key string, val interface{}) (interface{}, bool) {
```
mmatczuk: Comment?

internal/lru/lru.go
```diff
@@ -100,9 +139,20 @@ func (c *Cache) Remove(key string) bool {
 
 // RemoveOldest removes the oldest item from the cache.
 func (c *Cache) RemoveOldest() {
-	if c.cache == nil {
-		return
-	}
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.removeOldest()
```
mmatczuk: Rename to removeOldestLocked, possily apply the pattern in other places too if needed

internal/lru/lru.go
```diff
@@ -118,10 +168,12 @@ func (c *Cache) removeElement(e *list.Element) {
 	}
 }
 
-// Len returns the number of items in the cache.
-func (c *Cache) Len() int {
-	if c.cache == nil {
-		return 0
+func (c *Cache) add(key string, val interface{}) {
```
mmatczuk: Move under Add and rename to addLocked?

prepared_cache.go
dahankzter: I guess it can be fine here but theoretically it makes the external API wider since the methods get promoted.

internal/lru/lru.go
dahankzter: We use the key to delete the entries from the map when they get evicted.

internal/lru/lru.go
dahankzter: The timestamp calculation?

dahankzter:
I will try to add more tests as @mmatczuk suggested. It would be really comfortable to have high coverage in this package.

alourie:
@dahankzter I would also love to see the profiling of this compared to the initial #1291. Really appreciate your work on this!

internal/lru/lru.go
```diff
@@ -34,64 +38,120 @@ type Cache struct {
...
 	}
-	return
+
+	c.addLocked(key, val)
+	return val, false
+}
+
+// Len returns the number of items in the cache.
+func (c *Cache) Len() int {
+	c.mu.RLock()
```
penberg: You could use `defer c.mu.RUnlock()` here.

internal/lru/lru.go
```diff
@@ -34,64 +38,120 @@ type Cache struct {
...
 	if c.MaxEntries != 0 && c.ll.Len() > c.MaxEntries {
-		c.RemoveOldest()
+		c.removeOldestLocked()
 	}
 }
 
 // Get looks up a key's value from the cache.
 func (c *Cache) Get(key string) (value interface{}, ok bool) {
-	if c.cache == nil {
-		return
```
penberg: You could extract the write-locked version of the lookup to a function and special case `c.window == 0` to call that directly.

internal/lru/lru.go
dahankzter: It felt so short and readable anyway. No need to push a defer on the stack.

dahankzter:
The profile looks like this for the test program:

```
(pprof) top20 -cum
Showing nodes accounting for 4267.27ms, 100% of 4267.28ms total
Dropped 24 nodes (cum <= 21.34ms)
Showing top 20 nodes out of 29
      flat  flat%   sum%        cum   cum%
         0     0%     0%  4267.28ms   100%  runtime.goexit
 4267.27ms   100%   100%  4267.27ms   100%  sync.(*Mutex).Unlock
         0     0%   100%  3955.18ms 92.69%  github.com/gocql/gocql.(*writeCoalescer).flush
         0     0%   100%  3955.18ms 92.69%  github.com/gocql/gocql.(*writeCoalescer).writeFlusher
         0     0%   100%   311.70ms  7.30%  main.Job
         0     0%   100%   241.54ms  5.66%  github.com/gocql/gocql.(*Conn).executeQuery
         0     0%   100%   241.54ms  5.66%  github.com/gocql/gocql.(*Query).Iter
         0     0%   100%   241.54ms  5.66%  github.com/gocql/gocql.(*Query).execute
         0     0%   100%   241.54ms  5.66%  github.com/gocql/gocql.(*Session).executeQuery
         0     0%   100%   241.54ms  5.66%  github.com/gocql/gocql.(*queryExecutor).attemptQuery
         0     0%   100%   241.54ms  5.66%  github.com/gocql/gocql.(*queryExecutor).do
         0     0%   100%   241.54ms  5.66%  github.com/gocql/gocql.(*queryExecutor).executeQuery
         0     0%   100%   241.54ms  5.66%  github.com/gocql/gocql.(*Conn).exec
         0     0%   100%   235.98ms  5.53%  github.com/gocql/gocql.(*Conn).Write
         0     0%   100%   235.98ms  5.53%  github.com/gocql/gocql.(*framer).finishWrite
         0     0%   100%   235.98ms  5.53%  github.com/gocql/gocql.(*framer).writeExecuteFrame
         0     0%   100%   235.98ms  5.53%  github.com/gocql/gocql.(*writeCoalescer).Write
         0     0%   100%   235.98ms  5.53%  github.com/gocql/gocql.(*writeExecuteFrame).writeFrame
         0     0%   100%   233.63ms  5.47%  sync.(*Cond).Wait
         0     0%   100%   209.82ms  4.92%  main.validationJob
```

```
(pprof) list execIfMissing
Total: 4.27s
ROUTINE ======================== github.com/gocql/gocql.(*preparedLRU).execIfMissing in /home/henrik/dev/gocql/prepared_cache.go
         0     2.44us (flat, cum) 5.7e-05% of Total
         .          .     43:	val, ok := p.lru.Get(key)
         .          .     44:	if ok {
         .          .     45:		return val.(*inflightPrepare), true
         .          .     46:	}
         .          .     47:
         .     2.44us     48:	val, ok = p.lru.GetOrInsert(key, fn())
         .          .     49:	return val.(*inflightPrepare), ok
         .          .     50:}
         .          .     51:
         .          .     52:func (p *preparedLRU) keyFor(addr, keyspace, statement string) string {
         .          .     53:	// TODO: maybe use []byte for keys?
```

So the contention in the `execIfMissing` seems virtually eliminated for this app. We generate a bounded number of statements so it is a fairly well behaved app. In the case where the statements are unbounded we will of course hit the write lock more.

session.go
```diff
@@ -128,7 +128,7 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 	s.nodeEvents = newEventDebouncer("NodeEvents", s.handleNodeEvent)
 	s.schemaEvents = newEventDebouncer("SchemaEvents", s.handleSchemaEvent)
 
-	s.routingKeyInfoCache.lru = lru.New(cfg.MaxRoutingKeyInfo)
+	s.routingKeyInfoCache.lru = lru.New(cfg.MaxRoutingKeyInfo, time.Second)
```
penberg: Perhaps it's safer to disable the LRU window update optimization for routing key info cache?

internal/lru/lru.go
penberg: Sure.

session.go
dahankzter: I don't know. What would the possible consequences be?

dahankzter:
With the window disabled:
```
(pprof) list Get
Total: 369.82ms
ROUTINE ======================== github.com/gocql/gocql/internal/lru.(*Cache).Get in /home/henrik/dev/gocql/internal/lru/lru.go
         0    49.81ms (flat, cum) 13.47% of Total
         .          .     89:// Get looks up a key's value from the cache.
         .          .     90:func (c *Cache) Get(key string) (value interface{}, ok bool) {
         .          .     91:	timestamp := time.Now()
         .          .     92:
         .          .     93:	if c.window == 0 {
         .    49.81ms     94:		return c.getWithLock(key, timestamp)
         .          .     95:	}
         .          .     96:
         .          .     97:	c.mu.RLock()
         .          .     98:	if ele, ok := c.cache[key]; ok {
         .          .     99:		ent := ele.Value.(*entry)
ROUTINE ======================== sync.(*Pool).Get in /usr/lib/go/src/sync/pool.go
         0    63.93ms (flat, cum) 17.29% of Total
         .          .    123:// the result of calling p.New.
         .          .    124:func (p *Pool) Get() interface{} {
         .          .    125:	if race.Enabled {
         .          .    126:		race.Disable()
         .          .    127:	}
         .    45.55ms    128:	l := p.pin()
         .          .    129:	x := l.private
         .          .    130:	l.private = nil
         .          .    131:	runtime_procUnpin()
         .          .    132:	if x == nil {
         .          .    133:		l.Lock()
         .          .    134:		last := len(l.shared) - 1
         .          .    135:		if last >= 0 {
         .          .    136:			x = l.shared[last]
         .          .    137:			l.shared = l.shared[:last]
         .          .    138:		}
         .    15.82ms    139:		l.Unlock()
         .          .    140:		if x == nil {
         .     2.56ms    141:			x = p.getSlow()
         .          .    142:		}
         .          .    143:	}
         .          .    144:	if race.Enabled {
         .          .    145:		race.Enable()
         .          .    146:		if x != nil {
(pprof) 
```
It is clear that the lock is much more contended now. The `web` rendered svg which isn't possible to upload is also very clear.

prepared_cache.go
dahankzter: I added a couple of tests to just hammer the cache as the one in the semaphore test.
A bug was actually found when run with `-race` so this was a very good idea! Thanks @mmatczuk !

session.go
penberg: Disabling the LRU optimization for routing key info just means we revert to the old behavior, no? It's safer because we don't have to think about it. :)

rkuska:
Isn't `Mutex` at `routingKeyInfoLRU` now uncecessary? Or you want to deal with `routingKeyInfoCache` locks in a different MR?

dahankzter:
I think you are right @rkuska ill remove it.

dahankzter:
I don't know why they are separate implementations of the wrapper class. Maybe ill unify them if possible, they work on different types.

session.go
```diff
@@ -471,16 +468,14 @@ func (s *Session) routingKeyInfo(ctx context.Context, stmt string) (*routingKeyI
...
 
 		key, _ := inflight.value.(*routingKeyInfo)
-
 		return key, nil
 	}
 
 	// create a new inflight entry while the data is created
-	inflight := new(inflightCachedEntry)
+	inflight = new(inflightCachedEntry)
 	inflight.wg.Add(1)
```
rkuska: I believe that these two lines (above) should be removed as inflight is already created by `execIfMissing`

dahankzter: You are right, thx!

internal/lru/lru.go
```diff
@@ -34,64 +38,130 @@ type Cache struct {
...
+	} else {
+		return nil, false
+	}
+}
+
+// GetOrInsert either retrieves the current entry for the key or inserts the
+// provided value for the key.
+// It returns true if the value was loaded and false if the provided value
+// was stored.
+func (c *Cache) GetOrInsert(key string, val interface{}) (interface{}, bool) {
```
rkuska: Wouldn't it be better to let it accept function? So the function is actually executed only if needed?

dahankzter: The wrapper code does this presently but I guess it's possible to push that into the cache as well.
Not sure what is best for a cache. Maybe the "loader function" is best but that one could be applied in a variety of pattern. Configurable on the cache instance and per each `Get` is also possible.

dahankzter:
What do you think @Zariel can we merge this? It accomplishes the goal to reduce contention and we can easily exchange to whatever other strategy the future demands later.

dahankzter:
Ping @Zariel what do you think? Merge or drop it, this PR has been inflight a very long time. Pun intended :)

Zariel:
I'd rather replace the cache with https://blog.dgraph.io/post/introducing-ristretto-high-perf-go-cache/

dahankzter:
That sounds good @Zariel do you have something working we could try?

PR #1296, opened 2019-04-06
partition prepared statementcache by host

Zariel:
Prepared statements are only valid for the host they are prepared on so
we can avoid contending when concurrently querying multiple hosts.

Store the map of hosts->LRU in a copy-on-write map.

updates #1291

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -1327,8 +1327,9 @@ func injectInvalidPreparedStatement(t *testing.T, session *Session, table string
 	conn := getRandomConn(t, session)
 
 	flight := new(inflightPrepare)
-	key := session.stmtsLRU.keyFor(conn.addr, "", stmt)
-	session.stmtsLRU.add(key, flight)
+	stmtLRU := session.stmtsLRU.forHost(conn.addr)
+	key := stmtLRU.keyFor("", stmt)
+	stmtLRU.add(key, flight)
 
 	flight.preparedStatment = &preparedStatment{
 		id: []byte{'f', 'o', 'o', 'b', 'a', 'r'},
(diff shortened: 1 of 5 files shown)


prepared_cache.go
```diff
@@ -1,37 +1,101 @@
...
+		}
+		p.mu.Unlock()
+	}
 
-	for p.lru.Len() > max {
-		p.lru.RemoveOldest()
+	return lru
+}
+
+func (p *preparedCache) addHost(current map[string]*preparedLRU, addr string) *preparedLRU {
```
dahankzter: This method is not synchronized. It is called from the event notification as well directly. If that is perfectly serialized then we are just fine but otherwise this might need protections.

prepared_cache.go
```diff
@@ -1,37 +1,101 @@
...
+		return
+	}
+
+	cow := make(map[string]*preparedLRU, len(v)+1)
+	for k, v := range v {
+		cow[k] = v
+	}
+
+	cow[addr] = &preparedLRU{lru: lru.New(p.maxStmnts)}
+	p.hosts.Store(cow)
```
dahankzter: Maybe this should also go via `addHost`?

prepared_cache.go
```diff
@@ -1,37 +1,101 @@
...
+	}
+
+	cow := make(map[string]*preparedLRU, len(v)-1)
+	for k, v := range v {
+		if k != addr {
+			cow[k] = v
+		}
 	}
+
+	p.hosts.Store(cow)
```
dahankzter: There are many places where the map is modified. Perhaps a unified small path to add/remove where they all happen?

dahankzter:
This looks good @Zariel perhaps we can join the windowed solution to this to allow for intra host scaling as well?

dahankzter:
Perhaps you can even take the sharded cache from #1293 as the upper layer to avoid any locks or copies at all in this layer. What do you think @Zariel ?

prepared_cache.go
```diff
@@ -1,37 +1,101 @@
...
+	"github.com/gocql/gocql/internal/lru"
 )
 
 const defaultMaxPreparedStmts = 1000
 
-// preparedLRU is the prepared statement cache
-type preparedLRU struct {
-	mu  sync.Mutex
-	lru *lru.Cache
+type preparedCache struct {
```
mmatczuk: @Zariel how about sing sync.Map for host to lru mapping?

cassandra_test.go
```diff
@@ -1484,40 +1485,35 @@ func TestPrepare_PreparedCacheEviction(t *testing.T) {
...
-	//Make sure the cache size is maintained
-	if session.stmtsLRU.lru.Len() != session.stmtsLRU.lru.MaxEntries {
-		t.Fatalf("expected cache size of %v, got %v", session.stmtsLRU.lru.MaxEntries, session.stmtsLRU.lru.Len())
-	}
-
 	// Walk through all the configured hosts and test cache retention and eviction
 	for _, host := range session.cfg.Hosts {
-		_, ok := session.stmtsLRU.lru.Get(session.stmtsLRU.keyFor(host+":9042", session.cfg.Keyspace, "SELECT id,mod FROM prepcachetest WHERE id = 0"))
+		stmtLRU := session.stmtsLRU.forHost(host + ":9042")
+		stmtLRU.mu.Lock()
```
mmatczuk: I do not think that it would resolve the lru lock contention problem. For instance with Scylla you may have clusters with only few nodes and lots of shards. Now the total amount of lock events is divided by few.

joao-r-reis:
Closing this because it's old (and #1291 is also being closed). If the issue is re-opened then some ideas can be taken from this PR.

PR #1299, opened 2019-04-11
conn: Non blocking Done

rkuska:
Instead of having blocking call to Done (it uses Mutex under the hood)
we move Done() to select and initiate a new context in case the context
passed is nil.
This should lower the contention caused by calls to ctx.Done().

There is an option to either use `TODO` or `Background` context I wasn't sure which one to choose. I created this PR because I noticed `*(cancelCtx).Done` few times in contention profiles when context is being passed to underlying Queries.

Related to: https://github.com/gocql/gocql/issues/1291

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -895,9 +895,9 @@ func (c *Conn) exec(ctx context.Context, req frameWriter, tracer Tracer) (*frame
 		timeoutCh = call.timer.C
 	}
 
-	var ctxDone <-chan struct{}
-	if ctx != nil {
-		ctxDone = ctx.Done()
+	if ctx == nil {
+		// Done channel is nil for top level context.
+		ctx = context.Background()
 	}
 
 	select {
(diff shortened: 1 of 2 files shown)


Zariel:
I don't understand what this achieves?

rkuska:
Context with cancel uses lock under the hood therefore moving `Done` call to select should be non blocking right?

https://github.com/golang/go/blob/master/src/context/context.go#L323

Zariel:
it still has to call the function to get the channel to use so it will still lock.

rkuska:
But it is now in select and there are 3 other branches to choose from. In the original form it blocks always.

nussjustin:
All the cases in the select are all evaluated before the select itself is run.

From the [spec section on "Select statements"](https://golang.org/ref/spec#Select_statements):

> For **all the cases** in the statement, the channel operands of receive operations and the channel and right-hand-side expressions of send statements **are evaluated exactly once, in source order, upon entering the "select" statement**. The result is a set of channels to receive from or send to, and the corresponding values to send. **Any side effects in that evaluation will occur irrespective of which (if any) communication operation is selected to proceed**.

So this

```go
select {
case <-fnA():
case <-B:
case <-fnC():
}
```

Is basically the same as writing

```go
tmpA := fnA()
tmpB := B
tmpC := fnC()
select {
case <-tmpA:
case <-tmpB:
case <-tmpC:
}
```

As you can see from this, your code does not change anything about the call to `ctx.Done()` as the call will still be executed even if the case is not selected.

rkuska:
@nussjustin Thank you! That actually makes sense. Closing the PR then.

PR #1300, opened 2019-04-12
Improve awaitSchemaAgreement

### LINKED ISSUE #1298: Invalid peers should be ignored when checking for schema agreement

3.11

0.0.0-20181117210152-33c0e89ca93a

### What did you do?
Ran migrations through gocqlx

Migrations finish successfully and quickly.

### What did you see instead?
Migrations take 60 seconds (The timeout to wait for schema agreement)

---

Cassandra is running as statefulset within kubernetes. Whenever a cassandra POD is recreated it will receive a new IP address. The old peers will remain in the peers table for some time though.

During initial connection gocql will handle this gracefully and ignore the peers that are not valid with an error message such as the below:
```
2019/04/10 14:20:16 Found invalid peer '[HostInfo connectAddress="10.42.1.7" peer="10.42.1.7" rpc_address="<nil>" broadcast_address="<nil>" preferred_ip="<nil>" connect_addr="10.42.1.7" connect_addr_source="connect_address" port=9042 data_centre="" rack="" host_id="a9bae1a8-379f-488e-a5f3-35e836b1d730" version="v0.0.0" state=UP num_tokens=0]' Likely due to a gossip or snitch issue, this host will be ignored
```

When executing queries that affect the schema version of the cluster no such check is done though. This means that `awaitSchemaAgreement` will wait until timeout for the schema versions to match, even though the cluster is fine.

`awaitSchemaAgreement` should make use of `isValidPeer` or similar and only consider valid peers when checking of schema agreements.

```
cqlsh> select peer, rpc_address from system.peers;

 peer       | rpc_address
------------+-------------
 10.42.6.23 |        null
  10.42.1.3 |   10.42.1.3
 10.42.6.10 |  10.42.6.10
 10.42.6.16 |  10.42.6.16
```

alourie:
Improve the way awaitSchemaAgreement queries the peers/local hosts and ignore invalid ones.

Fixes #1298

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1356,11 +1356,12 @@ func (c *Conn) query(ctx context.Context, statement string, values ...interface{
 
 func (c *Conn) awaitSchemaAgreement(ctx context.Context) (err error) {
 	const (
-		peerSchemas  = "SELECT schema_version, peer FROM system.peers"
+		peerSchemas  = "SELECT * FROM system.peers"
 		localSchemas = "SELECT schema_version FROM system.local WHERE key='local'"
 	)
 
 	var versions map[string]struct{}
+	var schemaVersion string
 
 	endDeadline := time.Now().Add(c.session.cfg.MaxWaitSchemaAgreement)
 	for time.Now().Before(endDeadline) {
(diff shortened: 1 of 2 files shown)


rusenask:
Hi, could you export this method please? I am adding some automated tests on my application and creating schemas for each table. This would make life a lot easier :)

conn.go
```diff
@@ -1356,28 +1356,35 @@ func (c *Conn) query(ctx context.Context, statement string, values ...interface{
 
 func (c *Conn) awaitSchemaAgreement(ctx context.Context) (err error) {
 	const (
-		peerSchemas  = "SELECT schema_version, peer FROM system.peers"
+		peerSchemas  = "SELECT * FROM system.peers"
 		localSchemas = "SELECT schema_version FROM system.local WHERE key='local'"
 	)
 
 	var versions map[string]struct{}
+	var schemaVersion string
```
Zariel: this isnt used?

alourie: a bit further in the loop, [L1395](https://github.com/gocql/gocql/pull/1300/files#diff-a2ddca1600f34abb91244424f8431c2cR1395) and on

conn.go
```diff
@@ -1356,28 +1356,35 @@ func (c *Conn) query(ctx context.Context, statement string, values ...interface{
...
-		for iter.Scan(&schemaVersion, &peer) {
-			if schemaVersion == "" {
-				Logger.Printf("skipping peer entry with empty schema_version: peer=%q", peer)
+		rows, err := iter.SliceMap()
+		if err != nil {
+			goto cont
+		}
+
+		for _, row := range rows {
+			host, err := c.session.hostInfoFromMap(row, c.session.cfg.Port)
```
amit-fourkites: We're getting error at this line. Please advice.
cannot use c.session.cfg.Port (type int) as type *HostInfo in argument to c.session.hostInfoFromMap

tigranbs: Yep! delayed our releases because of this ... seems CI failed on master!

What about making releases so that you won't push broken code to master?

alourie: Sorry about that. All tests passed fine, so we just missed this one. Anyway, a fix is on the way.

Issue #356, opened 2015-04-09
Token ring awareness API


dankinder:
Some CQL drivers are token aware and make the token ring available, for example so that we could ask it "what endpoints have this partition/token?" (like nodetool getendpoints).

Examples:
- Python: http://datastax.github.io/python-driver/api/cassandra/metadata.html#tokens-and-ring-topology
- Java: http://www.datastax.com/drivers/java/2.1/com/datastax/driver/core/Metadata.html

My particular use case: Go and Cassandra go together great with the help for gocql, and I would like to use gocql to do some batch-like processing over cassandra data. Spark et al are great for these but don't provide Go interfaces. Moreover my usecase actually doesn't take much advantage of the Spark/MapReduce paradigm, I really only want the data gravity aspect of it (moving compute close to data).

What would work well for me is to be able to run services (written in go) out on the cassandra nodes that have the capability of taking a given token and asking gocql "which hosts own this token"?, so that I can read locally for that token if this node has it and ignore that token if it doesn't. Then I can use my own orchestration mechanism to run these services on all Cassandra nodes and greatly cut down on on network congestion and latency.

A simple call like "which hosts/IPs own token X" or "does host Y own token X" would do the trick here and make is easy for gocql users to move computation close to their data.

0x6e6562:
It feels like this can follow on from #348 once it lands. We can start thinking about what the API would look like by sketching it out a bit.

Zariel:
I think it makes sense to have `Session` own the TokenRing privately and have methods like, GetHostForToken(), GetTokensForHost(). This way we have a single place which can update and handling incoming changes (im starting to think about how to nicely handle control connections with events which will tie into this).

justinretailnext:
As a TODO the TokenRing should be expanded to be replication strategy aware, something I was going to do as a follow on. This would answer the question of which hosts replicate a particular token.

I'm curious as to the use case for asking the tokens for a host though (GetTokensForHost()).

Zariel:
Was just brainstorming, I guess its not particularly useful to applications

nemothekid:
@Zariel @justinretailnext I'm not 100% sure how the current TokenAware routing is implemented, but knowing the host for a given primary key is useful for batching - that is you could group all your inserts with the same host into a single batch.

Zariel:
[...]

PR #1307, opened 2019-04-25
Implementing partitioned batches

snaffi:
Motivation
Current batch implementation can't be used with token aware host policy

`PartitionedBatch` allows you to group batch queries by hosts which selected by token aware policy. 
Thereby we reduce latency for heavy write applications.

### DIFF
--- a/session.go
+++ b/session.go
@@ -1987,20 +1987,13 @@ func NewErrProtocol(format string, args ...interface{}) error {
 // This limit is set by cassandra and could change in the future.
 const BatchSizeMaximum = 65535
 
-type partitionedBatch struct {
-	*Batch
-	routingKeyBuffer []byte
-	session          *Session
-	statement        string
-}
-
 // PartitionedBatch helps to partition batches by host using tokenAwareHostPolicy
 type PartitionedBatch struct {
 	session     *Session
 	statement   string
 	tokenRing   *tokenRing
 	batchTyp    BatchType
-	batches     map[string]*partitionedBatch
+	batches     map[string]*Batch
 	idempotent  bool
 	consistency Consistency
 }
@@ -2018,7 +2011,7 @@ func (s *Session) NewPartitionedBatch(statement string, typ BatchType, idempoten
 		statement:   statement,
 		tokenRing:   p.tokenRing.Load().(*tokenRing),
 		batchTyp:    typ,
-		batches:     make(map[string]*partitionedBatch),
+		batches:     make(map[string]*Batch),
 		idempotent:  idempotent,
 		consistency: consistency,
 	}, nil
@@ -2039,10 +2032,7 @@ func (b *PartitionedBatch) Query(args ...interface{}) error {
 
 	batch, ok := b.batches[hostID]
 	if !ok {
-		batch = &partitionedBatch{
-			Batch:   b.session.NewBatch(b.batchTyp),
-			session: b.session,
-		}
+		batch = b.session.NewBatch(b.batchTyp)
 		batch.routingKey = routingKey
 		batch.SetConsistency(b.consistency)
 		batch.IsIdempotent()
@@ -2059,7 +2049,7 @@ func (b *PartitionedBatch) Query(args ...interface{}) error {
 // Batches return batch object for future execution. You can execute it in parallel or consequentially
 func (b *PartitionedBatch) Batches() (batches []*Batch) {
 	for _, v := range b.batches {
-		batches = append(batches, v.Batch)
+		batches = append(batches, v)
 	}
 	return
 }
@@ -2107,8 +2097,4 @@ func (s *Session) getRoutingKey(ctx context.Context, stmt string, args ...interf
 	}
 	routingKey := buf.Bytes()
 	return routingKey, nil
-}
-
-func (b *partitionedBatch) GetRoutingKey() ([]byte, error) {
-	return b.routingKey, nil
 }
--- a/session.go
+++ b/session.go
@@ -1994,6 +1994,7 @@ type partitionedBatch struct {
 	statement        string
 }
 
+// PartitionedBatch helps to partition batches by host using tokenAwareHostPolicy
 type PartitionedBatch struct {
 	session     *Session
 	statement   string
--- a/session.go
+++ b/session.go
@@ -1497,6 +1497,7 @@ type Batch struct {
 	cancelBatch           func()
 	keyspace              string
 	metrics               *queryMetrics
+	routingKey            []byte
 }
 
 // NewBatch creates a new batch operation without defaults from the cluster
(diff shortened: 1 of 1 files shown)


snaffi:
Metrics
```
timer batch_lat
  count:            2000
  min:           1349488
  max:          32897844
  mean:          2655657.80
  stddev:        2777425.79
  median:        2084017.00
  75%:           2427595.50
  95%:           5578827.05
  99%:          20812701.96
  99.9%:        32835897.80
  1-min rate:        241.21
  5-min rate:        310.37
  15-min rate:       323.71
  mean rate:          80.00
timer single_lat
  count:            2000
  min:           6012577
  max:          55595891
  mean:         11259639.52
  stddev:        4346799.45
  median:       10212441.50
  75%:          12622448.00
  95%:          18267714.45
  99%:          29458340.58
  99.9%:        55315622.73
  1-min rate:         23.11
  5-min rate:          5.22
  15-min rate:         1.78
  mean rate:          68.39
```

Cluster consists of five machines

Keyspace

```
CREATE KEYSPACE test_updates WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}  AND durable_writes = true;

CREATE TABLE test_updates.updates (
    user_id uuid,
    update_id uuid,
    data text,
    PRIMARY KEY (user_id, update_id)
) WITH CLUSTERING ORDER BY (update_id ASC)
    AND bloom_filter_fp_chance = 0.01
    AND caching = {'keys': 'ALL', 'rows_per_partition': 'NONE'}
    AND comment = ''
    AND compaction = {'class': 'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy', 'max_threshold': '32', 'min_threshold': '4'}
    AND compression = {'chunk_length_in_kb': '64', 'class': 'org.apache.cassandra.io.compress.LZ4Compressor'}
    AND crc_check_chance = 1.0
    AND dclocal_read_repair_chance = 0.1
    AND default_time_to_live = 0
    AND gc_grace_seconds = 864000
    AND max_index_interval = 2048
    AND memtable_flush_period_in_ms = 0
    AND min_index_interval = 128
    AND read_repair_chance = 0.0
    AND speculative_retry = '99PERCENTILE';
```

Script

https://gist.github.com/snaffi/7565bc14981d039f0dccd02eb4745505

cristaloleg:
Kindly ping

s-kozlov:
@cristaloleg what I have to do?

cristaloleg:
Looks like nothing(?), ping for maintainers, your changes seems fine.

session.go
```diff
@@ -1990,3 +1990,82 @@ func NewErrProtocol(format string, args ...interface{}) error {
...
+	if host, _ := b.tokenRing.GetHostForPartitionKey(routingKey); host != nil {
+		hostID = host.HostID()
+	}
+
+	batch, ok := b.batches[hostID]
+	if !ok {
+		batch = b.session.NewBatch(b.batchTyp)
+		batch.routingKey = routingKey
+		batch.SetConsistency(b.consistency)
+		batch.IsIdempotent()
```
alourie: Are you trying to _set_ the idempotency or check it here? If check, the return value is not tested...if to set, use `Idempotent(bool)` call instead.

session.go
```diff
@@ -1990,3 +1990,82 @@ func NewErrProtocol(format string, args ...interface{}) error {
...
+	}
+	return nil
+}
+
+// Batches return batch object for future execution. You can execute it in parallel or consequentially
+func (b *PartitionedBatch) Batches() (batches []*Batch) {
+	for _, v := range b.batches {
+		batches = append(batches, v)
+	}
+	return
```
alourie: why not just use b.batches directly or even `b := b.batches; return b` ?

Zariel:
I would rather not add new API to do this, we should just do this for batches. The batch api is already too bloated

s-kozlov:
Could we make cluster API public? I need access to `tokenRing`, `getRoutingKey` etc, hence I could make an utils like package which will provide extended batch API

session.go
snaffi: `b.batches` is a map, since `Batches()` should return slice

vikmik:
Was this abandoned? I'm wondering if there are workarounds instead of this

Having to use the partition key as a proxy for splitting batches (and in `SELECT ... IN` queries) is limiting, especially in applications where the cardinality of partition keys is high. This causes splitting BATCH/SELECT...IN queries by partition key to produce 1 query per row, which does not always work well. Since gocql has all the information needed to split queries by host, it would be great if it was exposed (like this PR attempted for batches).

snaffi:
@vikmik unfortunately batch API bloated so this PR wouldn't be merged  ￼😢

vikmik:
> @vikmik unfortunately batch API bloated so this PR wouldn't be merged ￼

Thanks for trying though :)

Wondering if/how people attempt to solve this in practice without this. Seems like there's 3 options:
* split queries by partition keys, possibly requiring much higher query count / higher query parallelism (which can be good or bad depending on the data)
* consuming the resources of a coordinator node to distribute the work (increases cluster scaling requirements)
* change the schema to use lower-cardinality partition keys (seems like an antipattern as it encodes scaling requirements in the data - and probably impractical in a lot of cases)

is there another way? Doing this on the client side with driver support really seemed to be the right thing to do. I see there were other attempts to solve related problems ( #356 ).

Being new to this ecosystem, I'm a little confused by the fact that some of the performance talks / blog posts I've looked at say to split BATCH / SELECT IN queries to be local to a single node, but I don't see any practical way to do that without compromising on something else. 

Maybe this is usually a negligible performance factor in a lot of setups - as a newcomer I'm just puzzled that I can't implement what the performance best practices docs I've read/watched tell me to do :)

session.go
```diff
@@ -1990,3 +1990,82 @@ func NewErrProtocol(format string, args ...interface{}) error {
...
+
+func (b *PartitionedBatch) Query(args ...interface{}) error {
+	routingKey, err := b.getRoutingKey(context.Background(), b.statement, args...)
+	if err != nil {
+		return err
+	}
+
+	hostID := emptyHostID
+	if host, _ := b.tokenRing.GetHostForPartitionKey(routingKey); host != nil {
+		hostID = host.HostID()
```
tengu-alt, line 2034: We can't get hosts using `GetHostForPartitionKey()` since it was removed in #1404. I think it can be changed to `GetHostForToken`.

session.go
```diff
@@ -1990,3 +1990,82 @@ func NewErrProtocol(format string, args ...interface{}) error {
...
+	}
+
+	batch, ok := b.batches[hostID]
+	if !ok {
+		batch = b.session.NewBatch(b.batchTyp)
```
tengu-alt, line 2039: `session.NewBatch()` was deprecated

OleksiienkoMykyta:
> Could we make cluster API public? I need access to `tokenRing`, `getRoutingKey` etc, hence I could make an utils like package which will provide extended batch API

I hope to push the (https://github.com/apache/cassandra-gocql-driver/pull/1664) to merge, it will make tokenRing public, and looks like a query.GetRoutingKey already exposed to the user.
I'm not sure that this PR should be abandoned, it expands the API, but not so bad. It will help to reduce traffic cost, and latency.

Maybe I'm missing something, but looks pretty useful. @joao-r-reis, @jameshartig, what is your take on it?

OleksiienkoMykyta (commented):
It should be updated, but I think it's pretty good. Not sure that it should be closed, rather it is something to discuss.

PR #1313, opened 2019-05-20
Token aware batches

dahankzter:
This patch implements token awareness for batches. It does so by calculating the routing key based on the for batch entry.


session.go
```diff
@@ -1729,8 +1697,73 @@ func (b *Batch) attempt(keyspace string, end, start time.Time, iter *Iter, host
...
+	// try to determine the routing key
+	routingKeyInfo, err := b.session.routingKeyInfo(b.Context(), entry.Stmt)
+	if err != nil {
+		return nil, err
+	}
+
+	return b.routingKeyCreator.createRoutingKey(routingKeyInfo, entry.Args)
+}
+
+type routingKeyCreator struct {
```
Zariel: I think we can just get rid of this type, saving this byte allocation wont make a difference if people dont reuse the batch/query.

dahankzter:
Ok that makes sense.

session.go
dahankzter: There, the type removed and the method is a function. Should there perhaps be a pool of `[]byte` for the buffer?

PR #1318, opened 2019-06-11
SQL Scan implementation for UUID type

MrWaggel:
This type now satisfies the SQL Scanner.

To use UUIDs in SQL, use the binary(16) for the designated column.
`n,_ := gocql.RandomUUID()`
`Exec(n.Bytes())`
`Scan(&n)`

### DIFF
--- a/uuid.go
+++ b/uuid.go
@@ -270,3 +270,43 @@ func (u *UUID) UnmarshalText(text []byte) (err error) {
 	*u, err = ParseUUID(string(text))
 	return
 }
+
+// This satisfies the sql.Scanner interface.
+func (uuid *UUID) Scan(src interface{}) error {
+	switch src := src.(type) {
+	case nil:
+		return nil
+
+	case string:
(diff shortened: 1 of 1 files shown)


martin-sucha:
I don't think this belongs to the core gocql package. If you need a UUID type that works with both gocql and sql package, it is pretty simple to implement a type that satisfies both interfaces in a separate package:

```go
type MyUUID struct {
	gocql.UUID
}

func (uuid *MyUUID) UnmarshalCQL(info gocql.TypeInfo, data []byte) error {
	return gocql.Unmarshal(info, data, &uuid.UUID)
}

func (uuid MyUUID) MarshalCQL(info gocql.TypeInfo) ([]byte, error) {
	 return gocql.Marshal(info, uuid.UUID)
}

// This satisfies the sql.Scanner interface.
func (uuid *MyUUID) Scan(src interface{}) error {
	// ...
}
```

Are there use cases that wouldn't work with a custom type like above?

Also somehow related to #1373 .

Issue #1320, opened 2019-06-11
Add the ecAudit plugin for username/password auth


CH-JosephBironas:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?
3.11 + the ecAudit plugin

### What version of Gocql are you using?
multiple

### What did you do?
Tried to implement two monitoring systems that use gocql as a connector to a backend cassandra cluster with compliance requirements for audit logging.

### What did you expect to see?
I expect username/password auth to not be blocked by a whitelist of allowed plugins.

### What did you see instead?

2019/06/11 17:41:06 gocql: unable to dial control conn <redacted>: unexpected authenticator "com.ericsson.bss.cassandra.ecaudit.auth.AuditPasswordAuthenticator"                                {"level":"fatal","ts":1560274866.5549803,"caller":"collector/main.go:87","msg":"Failed to init storage factory","error":"gocql: unable to create session: control: unable to connect to initial hosts: unexpected authenticator \"com.ericsson.bss.cassandra.ecaudit.auth.AuditPasswordAuthenticator\"","stacktrace":"main.main.func1\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/cmd/collector/main.go:87\ngithub.com/jaegertracing/jaeger/vendor/github.com/spf13/cobra.(*Command).execute\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/github.com/spf13/cobra/command.go:762\ngithub.com/jaegertracing/jaeger/vendor/github.com/spf13/cobra.(*Command).ExecuteC\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/github.com/spf13/cobra/command.go:852\ngithub.com/jaegertracing/jaeger/vendor/github.com/spf13/cobra.(*Command).Execute\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/vendor/github.com/spf13/cobra/command.go:800\nmain.main\n\t/home/travis/gopath/src/github.com/jaegertracing/jaeger/cmd/collector/main.go:177\nruntime.main\n\t/home/travis/.gimme/versions/go1.12.1.linux.amd64/src/runtime/proc.go:200"}

---

Please add "com.ericsson.bss.cassandra.ecaudit.auth.AuditPasswordAuthenticator" plugin to the whitelist of acceptable plugins maintained here:
https://github.com/gocql/gocql/blob/master/conn.go#L27

[...]

PR #1321, opened 2019-06-11
Adding ecaudit plugin to connection whitelist.

CH-JosephBironas:
This is a seemingly trivial change to a plugin whitelist. It uses a drop-in replacement to cassandra username/password auth to create an audit log for compliance purposes.

Edit: To address issue #1320

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -29,6 +29,7 @@ var (
 		"com.instaclustr.cassandra.auth.SharedSecretAuthenticator",
 		"com.datastax.bdp.cassandra.auth.DseAuthenticator",
 		"io.aiven.cassandra.auth.AivenAuthenticator",
+		"com.ericsson.bss.cassandra.ecaudit.auth.AuditPasswordAuthenticator",
 	}
 )
 


CH-JosephBironas:
One test failed because of what looks like a flakey environment setup.
```
+ccm create test -v 2.2.13 -n 3 -d --vnodes '--jvm_arg=-Xmx256m -XX:NewSize=100m'
integration.sh: line 34: ccm: command not found
```

alourie:
@CH-JosephBironas you can use a custom Authenticator using #1242. That being said, if the use of this authenticator is common, I wouldn't mind adding it to the whitelist.

CH-JosephBironas:
The problem with it not being added to the whitelist is, I have to maintain custom forks of two other largish projects that depend on this one in order to implement the same user/pass auth that exist already in this library. If there are no objections to augmenting the white list, that is the least effort for at least me, and I imagine others who depend on audit logging in cassandra pre the 4.0 release.

CH-JosephBironas:
Confirmed the travisci build was flakey.

PR #1322, opened 2019-06-19
Refactor query metrics code

martin-sucha:
This removes code duplication and ensures the metrics mutex is
locked only once in addAttempts and addLatency.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -418,7 +418,7 @@ func TestQueryMultinodeWithMetrics(t *testing.T) {
 
 	for i, ip := range addresses {
 		host := &HostInfo{connectAddress: net.ParseIP(ip)}
-		queryMetric := qry.getHostMetrics(host)
+		queryMetric := qry.metrics.hostMetrics(host)
 		observedMetrics := observer.GetMetrics(host)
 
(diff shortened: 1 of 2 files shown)


alourie:
Looks fine to me. The only concern I'm having is the speculative execution, where there can be a few goroutines that are using the same Query object and all might try to update it at the same time (and try to obtain a lock for doing so). Would you mind having a couple of test spec runs to see that there are no issues?

Thanks!

martin-sucha:
> Looks fine to me. The only concern I'm having is the speculative execution, where there can be a few goroutines that are using the same Query object and all might try to update it at the same time (and try to obtain a lock for doing so).

I did not add any new lock here, the only change regarding locking is that hostMetrics struct is created atomically with updating it in addMetrics and addLatency instead of a sequence lock, unlock, lock, unlock as was there before. So I believe the only performance impact this can have is positive, since we are locking just once instead of twice there.

> Would you mind having a couple of test spec runs to see that there are no issues?

Is this something I should run? If so, how?

alourie:
@martin-sucha  It's actually fine. Thanks!

PR #1324, opened 2019-06-25
Add attempt number to ObservedQuery

martin-sucha:
This will allow the observer to know whether the attempt was
original query or retry.


martin-sucha:
Fixed the issue in test.

PR #1327, opened 2019-07-01
Update supported Go versions to 1.11 and 1.12

martin-sucha:
As the README states, the current and previous version of Go is
supported, so tests should be run accordingly.

Also note that this commit removes Go 1.10, the last version
that does not support Go modules.

Thanks @dahankzter for the suggestion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,8 @@ env:
       AUTH=false
 
 go:
-  - "1.10"
   - "1.11"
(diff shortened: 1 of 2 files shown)


alourie:
I have no problem with this, pending @Zariel approval.

martin-sucha:
Ping. Is there any reason for requiring more than one reviewer for this trivial change?

PR #1332, opened 2019-07-15
Fix for #1274

### LINKED ISSUE #1274: panic: first replica is not the primary replica for the token

ScyllaDB 2018.1.10

github.com/martin-sucha/gocql@v0.0.0-20190227121127-86b0b61a5e6c

### What did you do?

After our database cluster went dead due to a server-side database bug, the gocql client panicked.

I'd expect the client to be resilient against cluster failures. If it can't reconnect to the nodes, it should not panic, but return errors instead, so that the service can fail gracefully.

### What did you see instead?

The gocql driver panicked with:

```
github.com/gocql/gocql.(*networkTopology).replicaMap(0xc00101e090, 0xc00101e078, 0x0, 0x0, 0xc007394000, 0x100, 0x100, 0x4634f5)
	/go/pkg/mod/github.com/martin-sucha/gocql@v0.0.0-20190227121127-86b0b61a5e6c/topology.go:203 +0x1291
github.com/gocql/gocql.(*tokenAwareHostPolicy).updateKeyspaceMetadata(0xc00017cf80, 0xc0000380f1, 0x7)
	/go/pkg/mod/github.com/martin-sucha/gocql@v0.0.0-20190227121127-86b0b61a5e6c/policies.go:447 +0x395
github.com/gocql/gocql.(*tokenAwareHostPolicy).AddHost(0xc00017cf80, 0xc0003e02c0)
	/go/pkg/mod/github.com/martin-sucha/gocql@v0.0.0-20190227121127-86b0b61a5e6c/policies.go:478 +0x6e
github.com/gocql/gocql.(*Session).addNewNode(0xc0000edc00, 0xc0003e02c0)
	/go/pkg/mod/github.com/martin-sucha/gocql@v0.0.0-20190227121127-86b0b61a5e6c/events.go:188 +0x9f
github.com/gocql/gocql.(*Session).handleNodeUp(0xc0000edc00, 0xc000b6eab0, 0x10, 0x10, 0x2352, 0xc00352ce01)
	/go/pkg/mod/github.com/martin-sucha/gocql@v0.0.0-20190227121127-86b0b61a5e6c/events.go:273 +0x10d
github.com/gocql/gocql.(*Session).reconnectDownedHosts(0xc0000edc00, 0x12a05f200)
	/go/pkg/mod/github.com/martin-sucha/gocql@v0.0.0-20190227121127-86b0b61a5e6c/session.go:284 +0x112
created by github.com/gocql/gocql.(*Session).init
	/go/pkg/mod/github.com/martin-sucha/gocql@v0.0.0-20190227121127-86b0b61a5e6c/session.go:233 +0x4a5
```

martin-sucha:
This pull request is based on PR https://github.com/gocql/gocql/pull/1328 as I reused some of that code to setup the test to reproduce the issue, so you might want to review that first separately. Only the last commit in this branch is the actual fix, while the one before that adds a test to reproduce the race condition.

The commit description for the last commit follows:
```
Fix updating cluster state in tokenAwareHostPolicy

tokenAwareHostPolicy keeps track of hosts in the cluster, keeps the
token ring and a map of token to replicas that keep the data (which is
computed using a placement strategy).

One issue that is present in old the code is that it updates the host
list and token ring whenever a node goes up or down. While the replica
map is only updated in AddHost/RemoveHost, it could be computed based
on partial host list in case AddHost/RemoveHost is called whenever some
host is down. The placement strategy would not see all nodes in the
cluster in that case. In fact, HostUp is never called anywhere in the
code now and AddHost is called even in case when a node goes up. So this
commit changes tokenAwareHostPolicy to only update token ring and
replica map whenever hosts change in AddHost/RemoveHost and not
HostUp/HostDown.

There is also a race condition in updateKeyspaceMetadata when computing
replicas, as it accesses fields that are synchronized separately
(tokenRing and hosts), but the placement strategy assumes the values it
receives are consistent. This leads to panics triggered in replicaMap
in case the race is hit when hosts are rapidly added/removed or getting
up/down (such as when we lose connection to multiple nodes at the same
time).

One sequence of events that might lead to the race condition and panic
in networkTopology.replicaMap could be like below:

1. A: calls tokenAwareHostPolicy.AddHost
2. A: calls tokenAwareHostPolicy.HostUp
3. A: calls tokenAwareHostPolicy.updateKeyspaceMetadata
4. A: calls tr, _ := t.tokenRing.Load().(*tokenRing) to read token Ring
5. B: calls tokenAwareHostPolicy.RemoveHost
6. B: calls tokenAwareHostPolicy.HostDown which calls
      t.hosts.remove(host.ConnectAddress())
7. A: calls strat.replicaMap(t.hosts.get(), tr.tokens)

To fix the inconsistencies, we need to perform the entire state change
atomically.

I think we can just lock (for writing) during the entire state update
as host add/remove should be rare. The remaining reads in Pick still
use lock-free access, but we now copy and replace all the metadata with
a single atomic.Value store instead of multiple atomic.Value fields.

I left the hosts field as cowHostList type, but it could be replaced by
a simple unsynchronized type as all accesses to hosts are now protected
by `tokenAwareHostPolicy.mu`.

Fixes #1274
```

### DIFF
--- a/policies.go
+++ b/policies.go
@@ -398,21 +398,27 @@ func TokenAwareHostPolicy(fallback HostSelectionPolicy, opts ...func(*tokenAware
 	return p
 }
 
-type keyspaceMeta struct {
+// clusterMeta holds metadata about cluster topology.
+// It is used inside atomic.Value and shallow copies are used when replacing it,
+// so fields should not be modified in-place. Instead, to modify a field a copy of the field should be made
+// and the pointer in clusterMeta updated to point to the new value.
+type clusterMeta struct {
+	// replicas is map[keyspace]map[token]hosts
 	replicas map[string]map[token][]*HostInfo
+	tokenRing *tokenRing
 }
 
 type tokenAwareHostPolicy struct {
-	hosts       cowHostList
-	mu          sync.RWMutex
-	partitioner string
 	fallback    HostSelectionPolicy
 	session     *Session
+	shuffleReplicas          bool
 
-	tokenRing atomic.Value // *tokenRing
-	keyspaces atomic.Value // *keyspaceMeta
-
-	shuffleReplicas bool
+	// mu protects writes to hosts, partitioner, metadata.
+	// reads can be unlocked as long as they are not used for updating state later.
+	mu          sync.Mutex
+	hosts       cowHostList
+	partitioner string
+	metadata    atomic.Value // *clusterMeta
 }
 
 func (t *tokenAwareHostPolicy) Init(s *Session) {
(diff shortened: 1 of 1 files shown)


martin-sucha:
Also please let me know if you are okay with the 100ms sleep in the test.

martin-sucha:
Also, if there is a better name for `clusterMeta` struct, please let me know. There is already `clusterMetadata` type in `ring.go`

martin-sucha:
I can also rebase just the latest commit onto master if you prefer.

martin-sucha:
Okay, rebased just the latest commit onto master.

PR #1334, opened 2019-07-17
Test with latest go1.1{1,2} versions

freeformz:
Go support policy is the latest minor release of the latest two major versions.

In travis, specifying "1.11" pins the test to the original go1.11 release.


Zariel:
Thanks!

PR #1336, opened 2019-07-30
Round robin fixes

NWilson:
There were some nasty issues with the round-robin host policies!

The round-robin host policy wasn't actually returning all the nodes to each iterator, you'd get queries where it would try some nodes twice, some not at all, because the iterators were using a shared index.

The DC-aware round-robin policy was similarly broken, and wouldn't even return any remote nodes: as long as there were any local nodes, it would never query a remote node, so if you were in a cluster where the local nodes went down the driver wouldn't fall back to anything else!

I've closely copied the strategy used by the Datastax Java driver, so I think these fixes are right.

Added tests (which fail without the code changes). See individual commit messages for further details.

### DIFF
--- a/policies.go
+++ b/policies.go
@@ -344,23 +344,28 @@ func (r *roundRobinHostPolicy) SetPartitioner(partitioner string)   {}
 func (r *roundRobinHostPolicy) Init(*Session)                       {}
 
 func (r *roundRobinHostPolicy) Pick(qry ExecutableQuery) NextHost {
-	// i is used to limit the number of attempts to find a host
-	// to the number of hosts known to this policy
-	var i int
-	return func() SelectedHost {
-		hosts := r.hosts.get()
-		if len(hosts) == 0 {
-			return nil
-		}
-
+	hosts := r.hosts.get()
+	var startIdx int
+	if len(hosts) != 0 {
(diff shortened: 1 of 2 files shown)


policies.go
```diff
@@ -344,23 +344,28 @@ func (r *roundRobinHostPolicy) SetPartitioner(partitioner string)   {}
...
-	// to the number of hosts known to this policy
-	var i int
-	return func() SelectedHost {
-		hosts := r.hosts.get()
-		if len(hosts) == 0 {
-			return nil
-		}
-
+	hosts := r.hosts.get()
+	var startIdx int
```
alourie: I'd think it would be cleaner here to return `nil` immediately if `len(hosts) == 0`

NWilson: That's a stylistic change I'd be happy to do.

But are you sure? None of the existing Pick() implementations returns nil - the function they return can return nil, but the NextHost function itself is always non-nil.

I tried making the change, but one of the tests currently asserts that Pick() returns a non-nil function even when there are no hosts.

alourie: Uhm, ok, that's a valid point. Ignore then ;-)

Zariel:
fixed via #1351

PR #1338, opened 2019-08-04
Add helpful functions for TimeUUID

Bychin:
While working on an app with ScyllaDB, I had to write code for queries that select data from a specific time range (TimeUUID columns).
Similar drivers for Java or C# include functions for generating query-friendly UUIDs, so I decided to add an implementation to the Golang version.


alourie:
@Bychin thanks a lot for this! Would you mind adding a test or two for testing the new functionality? Thanks!

Bychin:
@alourie I have pushed second commit with tests. Is it enough?

alourie:
@Bychin that's fantastic, thanks!

PR #1344, opened 2019-09-06
Add go1.13 to the testing matrix / formatting updates

freeformz:
This adds go1.13.x to the testing matrix.

This also adds small formatting fixes to the .travis.yml file (yay
linters).

I suggest dropping official go1.11 support once go1.13.1 is out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
 
 branches:
   only:
-  - master
+    - master
 
 env:
   global:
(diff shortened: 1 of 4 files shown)


martin-sucha:
Hi @freeformz, thanks for your pull request!

I believe the tests are failing because omitting the `flag.Parse` in Go 1.13 means that `clusterHosts` variable value is no longer parsed from command-line, so we either need to call `testing.Init` just before `flag.Parse` or remove `flag.Parse` in a way that keeps the value of the flag being used.

I have prepared an alternative pull request at https://github.com/gocql/gocql/pull/1347 that does the latter.

alourie:
@martin-sucha @freeformz after #1347 is merged now, should we close this one?

Zariel:
fixed via #1347

PR #1352, opened 2019-09-23
Fix context leaking between goroutines

martin-sucha:
When multiple goroutines wait for the same inflight request and the first goroutine's context is canceled, all waiting goroutines receive context canceled error. These commits fix the issue.

```
Refactor prepareStatement

The handling of flights was interspersed with code that executes the
statement preparation, let's split it so it's more readable.
```
```
Fix context leak in prepareStatement

prepareStatement deduplicates in-flight requests for preparing the same
statement, however it failed the preparation in case the context passed
from the first goroutine requesting was canceled. prepareStatment
returned context.Canceled error to all goroutines, even if their context
was not canceled, so the context from first goroutine leaked to all
others..

We need to move the prepare statement call to a separate goroutine so
that it can continue even if the first goroutine that spawned the task
does not want to wait anymore, since there might be other goroutines
waiting for the prepareStatement to complete.
```
```
Refactor routingKeyInfo
Split the parts handling in-flight requests and caching from the actual
business logic to make the code more readable.
```
```
Fix context leak in routingKeyInfo

Context used in routingKeyInfo for preparing statements was used to
compute result shared across multiple goroutines. This means that when
the context was canceled, all the currently inflight goroutines
asking for the same routing key info got context.Canceled error even
though their context was not canceled.
```


martin-sucha:
Ping. @Zariel could you please check this MR?

nightlyone:
@martin-sucha Do these changes mean that I can never abort a preparable statement that runs the first time? 

AFAICS these changes mean that the slow path of the prepare statement is still always run in the background leading to hangs on session.Close in tests.

martin-sucha:
Hi @nightlyone,

> @martin-sucha Do these changes mean that I can never abort a preparable statement that runs the first time?
> 
> AFAICS these changes mean that the slow path of the prepare statement is still always run in the background leading to hangs on session.Close in tests.

Yes. I thought my changes wouldn't change behavior, but I missed that `call.timeout` is closed in `Conn.exec` when the context is canceled, which results in closing the stream.

So we'll need to keep track of the contexts and close the one used in `prepareStatement` when the last context from call is closed.

martin-sucha:
Okay, I've edited the commits so that the code tracks how many goroutines are waiting for the result and the background context for work is canceled when nobody waits anymore for the result.

Zariel:
I think a simpler and more targeted fix is a better idea here, #1368

nightlyone:
@martin-sucha is your issue now solved with Zariels fix? Then we could close the issue.

nightlyone:
@martin-sucha can this be closed now?

martin-sucha:
Sorry for the late response.

I believe the fix in #1368 only fixed half of the issue (in `prepareStatement`), the same issue is present in `routingKeyInfo` as that uses its own inflight mechanism, see https://github.com/gocql/gocql/blob/68f928edb90a8a33cc6a91eb5f8c4361edba73d4/session.go#L482-L500

However, I don't have a reproducer at hand at the moment since I'm no longer using the code that triggered the issue for me in the first place.

PR #1355, opened 2019-09-26
networkTopology: optimise replicaMap slightly

Zariel:
```
name                   old time/op    new time/op    delta
Topology_ReplicaMap-8    83.5ms ± 1%    54.0ms ± 1%  -35.29%  (p=0.000 n=8+9)

name                   old alloc/op   new alloc/op   delta
Topology_ReplicaMap-8    5.73MB ± 0%    0.83MB ± 0%  -85.44%  (p=0.000 n=9+8)

name                   old allocs/op  new allocs/op  delta
Topology_ReplicaMap-8     89.2k ± 0%      6.9k ± 0%  -92.22%  (p=0.000 n=10+10)
```

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -228,8 +228,9 @@ func (h *HostInfo) PreferredIP() net.IP {
 
 func (h *HostInfo) DataCenter() string {
 	h.mu.RLock()
-	defer h.mu.RUnlock()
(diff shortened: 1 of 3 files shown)


alourie:
:+1:

PR #1358, opened 2019-10-09
testing: increase local run reliability further

dahankzter:
Local stability increased by clearing the database before running
integration tests.

### DIFF
--- a/integration.sh
+++ b/integration.sh
@@ -75,10 +75,15 @@ function run_tests() {
 	else
 		sleep 1s
 		go test -tags "cassandra gocql_debug" -timeout=5m -race $args
+
+		ccm clear
+		ccm start --wait-for-binary-proto
+		sleep 1s
+
 		go test -tags "integration gocql_debug" -timeout=5m -race $args
 
 		ccm clear
-		ccm start
+		ccm start --wait-for-binary-proto
 		sleep 1s
 
 		go test -tags "ccm gocql_debug" -timeout=5m -race $args


dahankzter:
I am sorry @alourie but the previous commit was not sufficient :( This patch however seems to really do it now.

integration.sh
```diff
@@ -75,6 +75,11 @@ function run_tests() {
 	else
 		sleep 1s
 		go test -tags "cassandra gocql_debug" -timeout=5m -race $args
+
+		ccm clear
+		ccm start
+		sleep 1s
```
alourie: it seels 1s is not enough for some Travis tests to pass. Maybe try to up it to 10s?

dahankzter: Sure we can try that.

martin-sucha: Could we add a loop (with inner sleep maybe 1s or 100ms) that checks whether the database is ready instead? Not relying on a constant sleep should make it more reliablie and responsive (if we need to wait less than those 10 seconds)

dahankzter: We can absolutely do that @martin-sucha but it's more work and needs more testing so perhaps in another PR. What do you guys think @alourie ?

dahankzter: This seems to work as well @alourie thx for the suggestion. What do you think @martin-sucha this should be good enough. I will run it a few more times and then push an update.

dahankzter: Pushed a new version using the `--wait-for-binary-proto` flag instead @alourie

martin-sucha: Looks good to me! :+1:

alourie:
Awesome sauce @dahankzter. Thanks!

PR #1363, opened 2019-10-16
query: fixes pagination with query binding

dahankzter:
Paging state was not reset in query::Bind which effectively broke
paging when Bind was used.

Fixes: #1364

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -313,6 +313,50 @@ func TestPaging(t *testing.T) {
 	}
 }
 
+func TestPagingWithBind(t *testing.T) {
+	session := createSession(t)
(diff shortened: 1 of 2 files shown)


dahankzter:
What do you think @alourie ? This must be pretty common no?

alourie:
@dahankzter yep, sounds about right. Do you think we might reset anything else on `Bind`, it seems as if we're basically running a completely new query at this point.

dahankzter:
We may need others but this one was so clear and isolated that I thought it best to send it first.

alourie:
Fine by me, I'll merge it as obviously fixing an issue. Thanks!

PR #1368, opened 2019-11-02
conn: dont propogate context cancelled in prepare

### LINKED ISSUE #1341: cancellation of a single query affects multiple queries in the same session
3.4.0
latest master at https://github.com/gocql/gocql/commit/df4b9cc330307b4a196f255fc0d1e8a53cabfbcd
### What did you do?
- start a new cassandra session
- run a statement in multiple queries multiple times concurrently in multiple Go routines each.
- statements created with `session.Query(selectStatement, x).WithContext(ctx)`
- each statement gets a different `ctx`
- one of thoese `ctx` variables is an already canceled context
Only the statement with the canceled context should receive the cancellation error.
### What did you see instead?
All statements running concurrently with the cancelled statement get cancelled.

Zariel:
If a callers context is cancelled and they won the race to prepare a
statement, dont stop the prepare and fail it with context failed error
as other callers may be waiting for us to finish.

Achieve this by replacing the waitgroup used to wait for prepared to
finish with a channel which is closed by the preparer. Do the preparing
in a goroutine and use context.Background in the exec call so that the
prepare will always finish. All callers now use the same flow to wait
for the result of the prepare waiting against their own context so they
can bail out individually.

fixes #1341

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -945,8 +945,8 @@ type preparedStatment struct {
 }
 
 type inflightPrepare struct {
-	wg  sync.WaitGroup
-	err error
+	done chan struct{}
+	err  error
 
 	preparedStatment *preparedStatment
 }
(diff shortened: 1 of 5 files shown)


martin-sucha:
I had similar change in #1352 initially, see https://github.com/gocql/gocql/commit/c6e0cd0ceba051df23991e3e113c0cce040ffd93 and https://github.com/gocql/gocql/commit/f5c7a5e6d5df5306dfcf60b38d7ee1f42e2ed5bf.

The change in just `prepareStatement` is insufficient to fix the issue in all cases as the result is cached also in `routingKeyInfo` (I hit that case in my program that heavily executes a lot of concurrent `errgroup.Group`s that get cancelled a lot when fixing the issue myself).

I believe the effect of this MR would be the same as that of https://github.com/gocql/gocql/commit/c6e0cd0ceba051df23991e3e113c0cce040ffd93. How does this MR address the comment about hangs on session.Close in tests https://github.com/gocql/gocql/pull/1352#issuecomment-544267397 ?

nightlyone:
Doesn't this approach mean that the prepare will not be retried quickly enough?

So we won the race, but decided not to wait. Next request is coming in, the same query will by run, finds it already being in the prepare in flight state and now waits.

Imagine that with a slow node of a replica set of 3+ and now the first N requests time out waiting for this slow node until finally the next node is being accessed.

What am I missing here?

Zariel:
This has always happened, if replicas are shuffled this wont be a problem and prepare only happens per node once. 

Thanks @martin-sucha

PR #1369, opened 2019-11-05
Connect to peers synchronously until policy is ready

### LINKED ISSUE #1365: Create session blocked until connection to every peer is established
Yugabyte 2.0.1

master (07ace3bab0f84bb88477bab5d79ba1f7e1da0169)

### What did you do?
When creating a session without `DisableInitialHostLookup` then the session will not be blocked from being created if ANY of the peers are down until the reconnect policy fails. Instead, there should be an option to do the initial host lookup but only connect to other hosts asynchronously.

I expected the session to be created and asynchronously connect to other peers.

### What did you see instead?
Instead CreateSession is blocked until a connection to all peers is established.

---
Logs with `gocql_debug`:
2019/10/22 16:03:25 gocql: Session.handleNodeUp: 10.142.0.175:9042
2019/10/22 16:03:31 connection failed "10.142.0.174": gocql: no response to connection startup within timeout, reconnecting with *gocql.ConstantReconnectionPolicy

The downside of setting `DisableInitialHostLookup` is that it won't discover any peers until a topology event. I'd like to try to connect to the peers discovered originally but not make them block. At the risk of making yet another flag, I'd like to propose adding `DisableInitialSyncPeerConnect`. If that is set to true, then after getting the peers they're connected to asynchronously and Session.init would return once one is connected. Alternatively we could call the option `RequireSinglePeer` or something.

jameshartig:
Fixes #1365

### DIFF
--- a/connectionpool.go
+++ b/connectionpool.go
@@ -437,6 +437,8 @@ func (pool *hostConnPool) fill() {
 			}
 			return
 		}
+		// notify the session that this node is connected
+		go pool.session.handleNodeUp(pool.host.ConnectAddress(), pool.port)
 
 		// filled one
 		fillCount--
@@ -448,6 +450,11 @@ func (pool *hostConnPool) fill() {
 
 		// mark the end of filling
 		pool.fillingStopped(err != nil)
+
+		if err == nil && startCount > 0 {
+			// notify the session that this node is connected again
+			go pool.session.handleNodeUp(pool.host.ConnectAddress(), pool.port)
+		}
 	}()
 }
 
--- a/control.go
+++ b/control.go
@@ -285,8 +285,6 @@ func (c *controlConn) setupConn(conn *Conn) error {
 	}
 
 	c.conn.Store(ch)
-	c.session.handleNodeUp(host.ConnectAddress(), host.Port(), false)
-
 	return nil
 }
 
@@ -452,6 +450,8 @@ func (c *controlConn) query(statement string, values ...interface{}) (iter *Iter
 
 	for {
 		iter = c.withConn(func(conn *Conn) *Iter {
+			// we want to keep the query on the control connection
+			q.conn = conn
 			return conn.executeQuery(context.TODO(), q)
 		})
 
--- a/install_test_deps.sh
+++ b/install_test_deps.sh
@@ -8,9 +8,3 @@ git clone https://github.com/pcmanus/ccm.git
 pushd ccm
 ./setup.py install --user
 popd
-
-if [ "$1" != "gocql/gocql" ]; then
-    USER=$(echo $1 | cut -f1 -d'/')
-    cd ../..
-    mv ${USER} gocql
-fi
--- a/policies.go
+++ b/policies.go
@@ -853,6 +853,51 @@ func (d *dcAwareRR) Pick(q ExecutableQuery) NextHost {
 	return roundRobbin(int(nextStartOffset), d.localHosts.get(), d.remoteHosts.get())
 }
 
+// ReadyPolicy defines a policy for when a HostSelectionPolicy can be used. After
+// each host connects during session initialization, the Ready method will be
+// called. If you only need a single Host to be up you can wrap a
+// HostSelectionPolicy policy with SingleHostReadyPolicy.
+type ReadyPolicy interface {
+	Ready() bool
+}
+
+// SingleHostReadyPolicy wraps a HostSelectionPolicy and returns Ready after a
+// single host has been added via HostUp
+func SingleHostReadyPolicy(p HostSelectionPolicy) *singleHostReadyPolicy {
+	return &singleHostReadyPolicy{
+		HostSelectionPolicy: p,
+	}
+}
+
+type singleHostReadyPolicy struct {
+	HostSelectionPolicy
+	ready    bool
+	readyMux sync.Mutex
+}
+
+func (s *singleHostReadyPolicy) HostUp(host *HostInfo) {
+	s.HostSelectionPolicy.HostUp(host)
+
+	s.readyMux.Lock()
+	s.ready = true
+	s.readyMux.Unlock()
+}
+
+func (s *singleHostReadyPolicy) Ready() bool {
+	s.readyMux.Lock()
+	ready := s.ready
+	s.readyMux.Unlock()
+	if !ready {
+		return false
+	}
+
+	// in case the wrapped policy is also a ReadyPolicy, defer to that
+	if rdy, ok := s.HostSelectionPolicy.(ReadyPolicy); ok {
+		return rdy.Ready()
+	}
+	return true
+}
+
 // ConvictionPolicy interface is used by gocql to determine if a host should be
 // marked as DOWN based on the error and host info
 type ConvictionPolicy interface {
--- a/session.go
+++ b/session.go
@@ -226,26 +226,44 @@ func (s *Session) init() error {
 	}
 
 	hosts = hosts[:0]
-
-	var wg sync.WaitGroup
+	// each host will increment left and decrement it after connecting and once
+	// there's none left, we'll close hostCh
+	var left int64
+	// we will receive up to len(hostMap) of messages so create a buffer so we
+	// don't end up stuck in a goroutine if we stopped listening
+	connectedCh := make(chan struct{}, len(hostMap))
+	// we add one here because we don't want to end up closing hostCh until we're
+	// done looping and the decerement code might be reached before we've looped
+	// again
+	atomic.AddInt64(&left, 1)
 	for _, host := range hostMap {
 		host := s.ring.addOrUpdate(host)
 		if s.cfg.filterHost(host) {
 			continue
 		}
 
-		host.setState(NodeUp)
-		hosts = append(hosts, host)
-
-		wg.Add(1)
+		atomic.AddInt64(&left, 1)
 		go func() {
-			defer wg.Done()
 			s.pool.addHost(host)
+			connectedCh <- struct{}{}
+
+			// if there are no hosts left, then close the hostCh to unblock the loop
+			// below if its still waiting
+			if atomic.AddInt64(&left, -1) == 0 {
+				close(connectedCh)
+			}
 		}()
+
+		hosts = append(hosts, host)
+	}
+	// once we're done looping we subtract the one we initially added and check
+	// to see if we should close
+	if atomic.AddInt64(&left, -1) == 0 {
+		close(connectedCh)
 	}
 
-	wg.Wait()
-
+	// before waiting for them to connect, add them all to the policy so we can
+	// utilize efficiencies by calling AddHosts if the policy supports it
 	type bulkAddHosts interface {
 		AddHosts([]*HostInfo)
 	}
@@ -254,6 +272,15 @@ func (s *Session) init() error {
 	} else {
 		for _, host := range hosts {
 			s.policy.AddHost(host)
+		}
+	}
+
+	readyPolicy, _ := s.policy.(ReadyPolicy)
+	// now loop over connectedCh until it's closed (meaning we've connected to all)
+	// or until the policy says we're ready
+	for range connectedCh {
+		if readyPolicy != nil && readyPolicy.Ready() {
+			break
 		}
 	}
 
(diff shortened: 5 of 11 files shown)


Zariel:
-1: I would rather not add more configuration and magic knobs, instead make the pool dial connections as they are needed instead of all at startup.

fastest963:
@Zariel I definitely agree but I'm not sure how to maintain the existing functionality (if people are relying on it) while also doing connections on demand.

fastest963:
@Zariel any suggestions on how to proceed? I think there are 3 other options that come to mind:
1) I could make what I added the default but then I assume it would break anyone using `TokenAwareHostPolicy` or anything like that because those assume you have all of the peers already.
2) I could check for `AddHosts(hosts []*HostInfo)` and assume if you support that method then you need to connect to all of the peers?
3) I could add a new optional method to `HostSelectionPolicy` called `Ready() bool` that it will call after each host and once it returns true, it'll stop blocking and let the rest of the peers connect asynchronously.

I think the 3rd option is the best since 1 could break custom `HostSelectionPolicy`'s and 2 has hidden effects with that method being present or not.

fastest963:
@Zariel Any input here?

fastest963:
@Zariel ping

fastest963:
@martin-sucha, thanks for reviewing! For some reason I never was notified about your comments. Let me go over them today.

session.go
```diff
@@ -138,7 +138,8 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 	s.hostSource = &ringDescriber{session: s}
 
 	if cfg.PoolConfig.HostSelectionPolicy == nil {
-		cfg.PoolConfig.HostSelectionPolicy = RoundRobinHostPolicy()
+		// keep the previous default of blocking until all connections are made
```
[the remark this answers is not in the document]
fastest963: Good point, I'll change those policies to never be ready so they keep the same behavior.

session.go
```diff
@@ -226,27 +227,70 @@ func (s *Session) init() error {
...
+			host := s.ring.addOrUpdate(host)
+			if s.cfg.filterHost(host) {
+				continue
+			}
+
+			atomic.AddInt64(&left, 1)
+			go func() {
+				host.setState(NodeUp)
+				s.pool.addHost(host)
+				s.policy.AddHost(host)
```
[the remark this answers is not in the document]
fastest963: Unfortunately, I'd need to know the keyspace in order to know when a particular set of hosts covers all tokens. I think that's better written as a wrapper around the `HostSelectionPolicy` with a defined keyspace.

fastest963:
@martin-sucha I updated the PR and removed `Ready` methods from all of the standard policies. I also implemented support for `AddHosts` by trying to group as many hosts together as possible as they connect.

Someone can opt-into being ready after a single host is connected by wrapping a `HostSelectionPolicy` with `SingleHostReadyPolicy`. Alternatively, they can implement their own `Ready` method and do any custom behavior they want (for instance, being ready once all tokens are accounted for).

session.go
```diff
@@ -226,27 +226,107 @@ func (s *Session) init() error {
...
+
+			hosts = append(hosts, host)
+		}
+
+		if bulkAddHostsPolicy != nil {
+			bulkAddHostsPolicy.AddHosts(hosts)
+		} else {
+			for _, host := range hosts {
+				s.policy.AddHost(host)
+			}
```
martin-sucha, line 329: This branch still connects to the hosts serially. I think even in the case when we want to wait for all hosts to be connected we want to be connecting to them concurrently, so we could probably remove this branch altogether.

policies.go
```diff
@@ -880,6 +880,45 @@ func (d *dcAwareRR) Pick(q ExecutableQuery) NextHost {
...
+// ReadyPolicy defines a policy for when a HostSelectionPolicy can be used. After
+// each host connects during session initialization, the Ready method will be
+// called. If you only need a single Host to be up you can wrap a
+// HostSelectionPolicy policy with SingleHostReadyPolicy.
+type ReadyPolicy interface {
+	Ready() bool
+}
+
+// SingleHostReadyPolicy wraps a HostSelectionPolicy and returns Ready after a
+// single host has been added via AddHost
```
martin-sucha, line 892: I think this policy should wait for `HostUp`, not `AddHost`. It's not described in `HostStateNotifier` interface, but I believe `AddHost` and `RemoveHost` are for adding hosts (even down ones) and `HostUp`/`HostDown` are for notifying of host status changes. Even though most policies don't distinguish those methods, I think for ReadyPolicy it makes sense to take advantage of this difference. When using `HostUp`, we should be able to add (or bulk add) the hosts before starting to connect anywhere and then update the policy with HostUp as the connections go up.

martin-sucha:
> I also implemented support for `AddHosts` by trying to group as many hosts together as possible as they connect.

I believe we should just add the hosts in the beginning and then start connecting to them as I described in other comment. This will allow us to just do bulk add or add individually without needing to coalesce the add calls based on time.

policies.go
fastest963: I couldn't find a single thing calling `HostUp` so I didn't bother putting it in there. Where is that being called?

session.go
fastest963: 👍 can do!

fastest963:
> I believe we should just add the hosts in the beginning and then start connecting to them as I described in other comment.

I did that at first but then I realized that there are policies that assume `AddHost` is called with a host that already has a connection pool. For instance `tokenAwareHostPolicy` seems to assume that. From what I can tell `roundRobinHostPolicy` does as well. I think adding them all before connecting makes sense but I'm afraid it'll cause a lot of issues downstream.

martin-sucha:
> For instance `tokenAwareHostPolicy` seems to assume that. From what I can tell `roundRobinHostPolicy` does as well. I think adding them all before connecting makes sense but I'm afraid it'll cause a lot of issues downstream.

Both those policies check `HostInfo.IsUp` in `Pick` before returning the host, so I believe they should work. Could you please elaborate on why you think things might break in them?

fastest963:
> Both those policies check `HostInfo.IsUp` in `Pick` before returning the host, so I believe they should work.

Ah, I missed that. Thanks, let me make those changes then to add beforehand.

session.go
```diff
@@ -227,18 +227,46 @@ func (s *Session) init() error {
...
 			continue
 		}
 
-		host.setState(NodeUp)
-		s.pool.addHost(host)
+		atomic.AddInt64(&left, 1)
+		go func() {
+			host.setState(NodeUp)
+			s.pool.addHost(host)
+			s.policy.HostUp(host)
```
fastest963, line 250: I added this here because during testing, nothing was calling `HostUp` so the ReadyPolicy wasn't working.

fastest963:
@martin-sucha thanks for the quick reviews! I updated and made those changes. I added to the policy after spawning the connect goroutines because I had to call filter anyways and re-build the `hosts` slice. After adding to the policy then I wait until ready.

I'm still confused about `HostUp`. During my testing, `HostUp` was never called so I added a `HostUp` call explicitly after it connects. Should there be a better place to call that?

session.go
fastest963: Maybe in `session.addNewNode` we should instead call `policy.HostUp` instead of `policy.AddHost`?

control.go
```diff
@@ -285,8 +285,6 @@ func (c *controlConn) setupConn(conn *Conn) error {
 	}
 
 	c.conn.Store(ch)
-	c.session.handleNodeUp(host.ConnectAddress(), host.Port(), false)
```
fastest963, line 288: I removed this because there was an issue during init where this called `policy.HostUp` which marked the policy as ready but then session init returned an error because `s.pool.Size() == 0`.

fastest963:
I updated the change to fix the issues with `HostUp`. In init we only add the host to the pool and then we rely on the connection succeeding to call `HostUp`. Also, instead of calling `HostUp` when reconnecting (which was just bouncing the status between Up and Down), it just tries to fill the pool to reconnect.

fastest963:
I had to make some changes to how queries are re-fetched (from commit message):
> When a query is run on a specific connection (rather than the session) it
will be pinned to that connection. This prevents inconsistencies or fetch
to be called on the session and get a different connection.

This is because `init` calls `GetHosts` which might end up calling `nextIter.fetch` but since the session doesn't have any connections yet, it fails unless you use the same connection. A similar thing happens with `tokenAwareHostPolicy.SetPartitioner`, which ends up triggering a connection on the control connection.

connectionpool.go
```diff
@@ -440,6 +442,11 @@ func (pool *hostConnPool) fill() {
 
 		// mark the end of filling
 		pool.fillingStopped(err != nil)
+
+		if err == nil && startCount == 0 {
+			// notify the session that this node is connected again
+			go pool.session.handleNodeUp(pool.host.ConnectAddress(), pool.port, true)
+		}
```
martin-sucha, line 449: Why is this branch necessary at all? If `startCount==0` then the `go pool.session.handleNodeUp` call was already made on line 433 when we reach this point, wasn't it?

martin-sucha:
> I'm still confused about `HostUp`. During my testing, `HostUp` was never called so I added a `HostUp` call explicitly after it connects. Should there be a better place to call that?

Interesting. Looking at `git log -p -S '.HostUp'` output, it seems the last call to `HostUp` was removed in a1745690de78b20dcd11eef937432a4dace28b93, but the commit message does not mention it explicitly. @Zariel could you please comment on this? `HostUp` is supposed to be called after the host changes state so that it's `IsUp()` returns true, right?

connectionpool.go
fastest963: Woops, this should've been `startCount > 0` it's meant to be called when we didn't do the synchronous call above. I'll add a comment as well.

fastest963:
Updated and fixed all of those comments. I also added a few tests. Adding `go_import_path: github.com/gocql/gocql` checks out the code into that path no matter what repo, which makes it easy to test from a fork and I could remove that block of code that was moving the code.

install_test_deps.sh
```diff
@@ -8,9 +8,3 @@ git clone https://github.com/pcmanus/ccm.git
 pushd ccm
 ./setup.py install --user
 popd
-
-if [ "$1" != "gocql/gocql" ]; then
```
fastest963, line 12: This is now handled by `go_import_path: github.com/gocql/gocql`

session.go
```diff
@@ -872,7 +909,7 @@ func (q *Query) Latency() int64 {
 }
 
 func (q *Query) AddLatency(l int64, host *HostInfo) {
-	q.metrics.attempt(0, time.Duration(l) * time.Nanosecond, host, false)
```
fastest963, line 875: This was just caused by `go fmt`

test_server_test.go
```diff
@@ -0,0 +1,316 @@
+package gocql
```
fastest963, line 1: Moved this out because it is useful to more than just `conn_test.go`. For instance, in `session_unit_test.go`.

session_connect_test.go
```diff
@@ -1,93 +0,0 @@
-package gocql
```
fastest963, line 1: Unused anywhere

fastest963:
@martin-sucha thanks so much for the review!

fastest963:
@Zariel @alourie Any updates? Appreciate it!

fastest963:
@Zariel @alourie @jorgebay Can someone review this?

fastest963:
@Zariel Can you review this please?

fastest963:
@Zariel @alourie @jorgebay Can someone glance at this?

jameshartig:
@Zariel any update on this?

jameshartig:
@Zariel @martin-sucha any update on this? I added back some of the dead code I removed so the diff is smaller if that helps get this merged. We've been using this in production for over a year now. We've already ran into others that would benefit from this but I don't want to keep maintaining a fork.

session.go
```diff
@@ -226,26 +226,44 @@ func (s *Session) init() error {
...
-	wg.Wait()
+		hosts = append(hosts, host)
+	}
+	// once we're done looping we subtract the one we initially added and check
+	// to see if we should close
+	if atomic.AddInt64(&left, -1) == 0 {
+		close(connectedCh)
+	}
 
+	// before witing for them to connect, add them all to the policy so we can
```
martin-sucha, line 265: **nitpick**: typo `witing`→`waiting`

martin-sucha:
Except that one typo looks good to me. This will also be useful for us to have and I'd prefer to merge it here than merge into our fork.

jameshartig:
@martin-sucha thanks! I've fixed that typo.

martin-sucha:
Merged. Thanks!

PR #1376, opened 2019-12-03
Inject Dialer from ClusterConfig

rinx:
I'd like to inject net.Dialer from outside of this package.
I added an option to receive net.Dialer in ClusterConfig.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -144,7 +144,8 @@ type ClusterConfig struct {
 	// (default: 200 microseconds)
 	WriteCoalesceWaitTime time.Duration
 
-	// Dialer
+	// Dialer will be used to establish all connections created for this Cluster.
+	// If not provided, a default dialer configured with ConnectTimeout will be used.
 	Dialer *net.Dialer
 
 	// internal config for testing
--- a/cluster.go
+++ b/cluster.go
@@ -144,6 +144,9 @@ type ClusterConfig struct {
 	// (default: 200 microseconds)
 	WriteCoalesceWaitTime time.Duration
 
+	// Dialer
+	Dialer *net.Dialer
+
 	// internal config for testing
 	disableControlConn bool
 }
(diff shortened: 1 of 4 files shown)


rinx:
@alourie Could you please review it?

rinx:
@martin-sucha @Zariel @alourie

rinx:
@titanous @alexzorin @dpanth3r @juliendsv @martin-sucha @jacobgreenleaf @zackslash @zhixinwen @furmmon @vrischmann @nathany

alourie:
@rinx is there a specific reason you need this?

rinx:
@alourie Thanks for your response.
We use our own implementation of net.Dialer interface for our products.
The major Golang libraries for RDB/NoSQL (for example go-sql-driver/mysql, go-redis/redis, etc...) provide dialer registration functions/interfaces. This feature makes gocql more useful.

martin-sucha:
Looks good to me.

@alourie There are several possible use-cases for a custom dialer. One is if you need to connect through a proxy (for example over a ssh connection). Another use case is if you need to limit the range of IP addresses the client should connect to, for example if you connect to an address provided by a user and want to exclude certain IP ranges to help mitigate possible server-side request forgery attacks.

cluster.go
```diff
@@ -144,6 +144,9 @@ type ClusterConfig struct {
 	// (default: 200 microseconds)
 	WriteCoalesceWaitTime time.Duration
 
+	// Dialer
```
martin-sucha: Maybe the doc comment could be expanded like
```
// Dialer will be used to establish all connections created for this Cluster.
// If not provided, a default dialer configured with ConfigTimeout will be used.
```

alourie:
@martin-sucha that's a fair point.

@rinx it does look good, but I'd ask you to indeed update the comment for the dialer, so it's clear that it's a new option and the default stays as it is. Then I can merge it.

cluster.go
rinx: Thanks! I've added your comment.

rinx:
@alourie @martin-sucha Thanks for your review. I've just updated the comment for the dialer.

alourie:
While I'm ok with the change, I'll give @Zariel to voice an opinion. If not, I'll merge it tomorrow.

alourie:
@rinx sorry it took a tad longer, but we got there eventually :-)

Issue #1252, opened 2019-01-07
No prepared statement with ID X found


eliaslevy:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?

Scylla 2.3.1

### What version of Gocql are you using?

9de8c0414fd7ff72ee819f25a66e4942cb223873

### What did you do?

Data load into Scylla using INSERTs and UPDATEs via worker using parallel goroutines sharing a single gocql session.  All queries are marked as idempotent and are using `.WithTimestamp`.

Basic call structure is
```go
q := session.Query(statement, ...).
	WithTimestamp(microsecs).
	Idempotent(true)

if err := q(); err != nil {
	panic(err)
}
q.Release()
```

### What did you expect to see?

No errors.

### What did you see instead?

After running for nearly two hours the worker died after receiving a `No prepared statement with ID 5d348ed0f8b4217b44b5bfca8b70fc2f found` error.  Reading #152 I gather that this is an error that gocql should have been able to recover from, but it did not for some reason.

Looking a [the code handling the error](https://github.com/gocql/gocql/blob/70385f88b28b43805bd83d212169ab2d38810b15/conn.go#L1119-L1120), it seems that gocql will only try to reexecute the query if it successfully finds and removes the statement in its local statement cache.  The local cache is a LRU cache, which means that if the statement has fallen off the cache while the query is in flight, gocql may fail to reexecute the query resulting in the observed error.

The default cache size is 1,000.  It seems unlikely that the local statement cache was overflowed.  At the time of error the worker was executing ~25K inserts/updates per second, but the worker is using only 19 prepared statements (all with placeholders).  

---

If you are having connectivity related issues please share the following additional information

### Describe your Cassandra cluster
please provide the following information

- output of `nodetool status`
- output of `SELECT peer, rpc_address FROM system.peers`
- rebuild your application with the `gocql_debug` tag and post the output

syhlion:
I have same problem too

lppsuixn:
same problem too

ehsannm:
we have the same issue, and this is still open

Somebi:
bump, same here

alourie:
I might be able to have another look in a few days. Sorry, folks, can't get to it sooner.

mediocregopher:
[...]

PR #1378, opened 2019-12-09
conn: unconditionally retry query execution on errUnprepared

dahankzter:
The error handling for retrying or erroring out upon failure to find a server side prepared statement now simply attempts to prepare and execute again.

Fixes: #1252

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1192,11 +1192,8 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
 		return iter
 	case *RequestErrUnprepared:
 		stmtCacheKey := c.session.stmtsLRU.keyFor(c.addr, c.currentKeyspace, qry.stmt)
-		if c.session.stmtsLRU.remove(stmtCacheKey) {
-			return c.executeQuery(ctx, qry)
-		}
-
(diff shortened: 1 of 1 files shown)


martin-sucha:
Looks good to me.

When reading the code I noticed a small issue: When `RequestErrUnprepared` is received by many goroutines in `executeBatch`/`executeQuery`, the first goroutine will start the prepare again, but others will evict the newly started prepare from cache, rendering the inflight deduplication for prepare not being used. They should only evict the old prepared statement with id we know does not exist. But it's a separate issue that was there before so does not belong to this PR..

dahankzter:
Yes this is a frustrating state but like you said it requires some thought
so we can perhaps defer it to another or.

dahankzter:
What do think @alourie can we merge this?

alourie:
@dahankzter yea, I'll merge it. But could you guys open a ticket for the second issue? So that deduplication is properly done for all goroutines?

Thanks!

martin-sucha:
> @dahankzter yea, I'll merge it. But could you guys open a ticket for the second issue? So that deduplication is properly done for all goroutines?
> 
> Thanks!

I have created the issue here: https://github.com/gocql/gocql/issues/1390

PR #1385, opened 2019-12-24
ensure framer is not nil before accessing customPayload

burkad01:
iter.GetCustomPayload() causes a nil panic if an error occurs on the cassandra query as iter.framer is nil. This PR adds a nil check of the framer before accessing customPayload and returns nil in the case where the framer is nil. This is similar to how Warnings() function works.


cardonator:
@martin-sucha Is there any way we can get this very small QOL change merged? This bug has existed for nearly 3 years now. If you need an updated PR or something, let me know and I'll throw one together in 2 seconds.

martin-sucha:
Thanks for the nudge. This has slipped through the cracks, sorry about that.

Looks good to me. Merged.

cardonator:
Thanks, @martin-sucha ! 🙇

PR #1391, opened 2020-01-15
conn: fix race evicting statements on ErrUnPrepared

### LINKED ISSUE #1390: RequestErrUnprepared should evict only the preparedStatement with old id

68f928edb90a8a33cc6a91eb5f8c4361edba73d4

### What did you do?

When `RequestErrUnprepared` is received by many goroutines in `executeBatch`/`executeQuery`, the first goroutine to receive it will start the prepare again and insert a new `inflightPrepare` to the LRU cache, but the other goroutinces will evict the newly started prepare from cache, rendering the inflight deduplication for prepare not being used. They should only evict the old prepared statement with id we know does not exist.

Receiving RequestErrUnprepared while a prepareStatement is in progress should not start a new prepare on retry.

### What did you see instead?

Receiving RequestErrUnprepared while a prepareStatement is in progress will start a new concurrent prepareStatement on retry.

Zariel:
Ensure we only remove prepared statements on ErrUnPrepared when our
cached statements ID matches the unprepared statement.

fixes #1390

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1193,7 +1193,7 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
 		return iter
 	case *RequestErrUnprepared:
 		stmtCacheKey := c.session.stmtsLRU.keyFor(c.addr, c.currentKeyspace, qry.stmt)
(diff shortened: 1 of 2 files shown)


alourie:
LGTM

PR #1392, opened 2020-01-21
Add a variant of WithTimestamp that accepts time.Time

martin-sucha:
With just timestamp int64 parameter, it's possible to mistake the unit
used for the timestamp, so we document that it's microseconds since UNIX
epoch.

Since Go users are likely to already have a time.Time value representing
the time instant used for the timestamp, I'm adding a method that
accepts a time.Time directly and takes care of the correct unit
conversion itself.

### DIFF
--- a/batch_test.go
+++ b/batch_test.go
@@ -41,10 +41,11 @@ func TestBatch_WithTimestamp(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	micros := time.Now().UnixNano()/1e3 - 1000
+	timestamp := time.Now().Add(-1000 * time.Microsecond)
+	micros := timestamp.UnixNano() / 1e3
 
 	b := session.NewBatch(LoggedBatch)
-	b.WithTimestamp(micros)
+	b.WithTimestamp(timestamp)
 	b.Query("INSERT INTO batch_ts (id, val) VALUES (?, ?)", 1, "val")
 	if err := session.ExecuteBatch(b); err != nil {
 		t.Fatal(err)
--- a/session.go
+++ b/session.go
@@ -796,27 +796,31 @@ func (qm *queryMetrics) attempt(addAttempts int, addLatency time.Duration,
 
 // Query represents a CQL statement that can be executed.
 type Query struct {
-	stmt                  string
-	values                []interface{}
-	cons                  Consistency
-	pageSize              int
-	routingKey            []byte
-	pageState             []byte
-	prefetch              float64
-	trace                 Tracer
-	observer              QueryObserver
-	session               *Session
-	rt                    RetryPolicy
-	spec                  SpeculativeExecutionPolicy
-	binding               func(q *QueryInfo) ([]interface{}, error)
-	serialCons            SerialConsistency
-	defaultTimestamp      bool
+	stmt             string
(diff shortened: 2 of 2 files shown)


deckarep:
Forgive my ignorance here...

Doesn't Cassandra store Timestamp as milliseconds since UNIX epoch?  What am I missing?

martin-sucha:
> Forgive my ignorance here...
> 
> Doesn't Cassandra store Timestamp as milliseconds since UNIX epoch? What am I missing?

That's the thing, Cassadra uses various timestamp units in different places. It's confusing and easy to miss what is used where. Actually happened to me that I mixed it up, that's why I want to add less ambiguous API.

So to recap:

* the timestamp of the query (this one is handled in this MR) is in microseconds since epoch - [protocol spec](https://github.com/datastax/native-protocol/blob/1acbf1f40d5dfeb2a50de765bc22f12d5241ea1c/src/main/resources/native_protocol_v4.spec#L338-L343)
* TIMESTAMP column values are in milliseconds since epoch - [protocol spec](https://github.com/datastax/native-protocol/blob/1acbf1f40d5dfeb2a50de765bc22f12d5241ea1c/src/main/resources/native_protocol_v4.spec#L911-L915)
* return value of WRITETIME is in microseconds since epoch, unless it's called on counter column, when it's in milliseconds since epoch - [docs](https://docs.datastax.com/en/dse/6.7/cql/cql/cql_using/useWritetime.html) (with [another docs wrongly claiming seconds since epoch](https://docs.datastax.com/en/dse/6.7/cql/cql/cql_reference/refCqlFunction.html#refCqlFunction__writetime))

deckarep:
@martin-sucha - wow, thanks for clarifying...that does seem quite confusing and inconsistent from Cassandra's perspective.  

Thanks for enlightening me!

martin-sucha:
@alourie may I request a review of this MR please?

alourie:
@martin-sucha I apologise, will review this week

Zariel:
Can we just change the existing api, yes its breaking but progress requires it and I agree the current API is terrible

alourie:
@martin-sucha Chris rarely allows breaking API, so I'd take the opportunity while it stands :-)

martin-sucha:
Okay, I have updated the commit to change the signature of WithTimestamp. Perhaps we could add to  README/CONTRIBUTING a statement about what backward compatibility policy we want to have (or not have)?

martin-sucha:
Actually, changing the signature will break `scylladb/gocqlx` https://github.com/scylladb/gocqlx/blob/2265c3fdbb459cd8bd014fbd074b54810955e301/queryx_wrap.go#L71 as well cc @mmatczuk

martin-sucha:
I think we shouldn't break the API unless we want to create v2 version of gocql. Since gocql is using go modules, it should be backwards compatible as other modules expect it to be, see https://github.com/golang/go/wiki/Modules#semantic-import-versioning

@Zariel what about preparing a list of breaking changes for v2 api and doing a bigger backwards-incompatible release instead of breaking the compatibility just for this small change?

martin-sucha:
Since this is a breaking change I've rebased the merge request onto `v2.dev` branch instead of `master`

PR #1393, opened 2020-01-22
Fix goroutine stall on write to broken socket

funny-falcon-at-joomcode:
imho, combination of
```go
	_, w.err = w.buffers.WriteTo(w.c)
	if w.err != nil {
		w.buffers = nil
	}
	w.cond.Broadcast()
```
and
```go
	for len(w.buffers) != 0 {
		w.cond.Wait()
	}
```
certainly have to lead to goroutine stall-ness on any write error.

Attempt to fix it by changing condition on "flush count".

If I'm mistaken, then excuse me for disturbing.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -113,3 +113,4 @@ Thomas Meson <zllak@hycik.org>
 Martin Sucha <martin.sucha@kiwi.com>; <git@mm.ms47.eu>
 Pavel Buchinchik <p.buchinchik@gmail.com>
 Rintaro Okamura <rintaro.okamura@gmail.com>
+Yura Sokolov <y.sokolov@joom.com>; <funny.falcon@gmail.com>
--- a/conn.go
+++ b/conn.go
@@ -745,6 +745,7 @@ type writeCoalescer struct {
 	cond    *sync.Cond
 	buffers net.Buffers
 	timeout time.Duration
+	flushes uint64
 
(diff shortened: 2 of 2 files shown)


funny-falcon-at-joomcode:
I forgot a bit about flow of my thoughts, but looks like you re-found it correctly. Excuse me I didn't describe it thoroughly on PR creation.

And I were care in first place of latency of writer: when other writers win race for lock, one unfortunate writer could loop in this condition for a long time even in case of no errors. Then I found it could stuck forever in certain cases.

Probably "certainly have to lead" is an hyperbole. I was tired at that moment, and probably started to see `!=` as `==`.

martin-sucha:
@funny-falcon-at-joomcode could you please `git commit --amend` the commit message and update the pull request description to include the above steps describing the race condition?

Also could you please add a commit adding yourself to the `AUTHORS` file as described in [CONTRIBUTING.md](https://github.com/gocql/gocql/blob/master/CONTRIBUTING.md#minimum-requirement-checklist)?

I think once that is done @Zariel or @alourie will merge this pull request.

funny-falcon-at-joomcode:
Well, looking more closely I think I were mistaken: there is no way for goroutine to stall.
But still current wait loop condition could lead to larger resources utilization then it should be.
I've changed commit message, and added myself to AUTHORS

PR #1395, opened 2020-01-27
Add AwaitSchemaAgreement to Session

annismckenzie:
This adds `AwaitSchemaAgreement` to the `*gocql.Session` object by using the control connection and the already existing `awaitSchemaAgreement` functionality on the Connection object.

See https://github.com/scylladb/gocqlx/issues/106 and https://github.com/gocql/gocql/issues/1301 for details.

### DIFF
--- a/session.go
+++ b/session.go
@@ -283,6 +283,21 @@ func (s *Session) init() error {
 	return nil
 }
 
+// AwaitSchemaAgreement will wait until schema versions across all nodes in the
+// cluster are the same (as seen from the point of view of the control connection).
+// The maximum amount of time this takes is governed
+// by the MaxWaitSchemaAgreement setting in the configuration (default: 60s).
+// AwaitSchemaAgreement returns an error in case schema versions are not the same
+// after the timeout specified in MaxWaitSchemaAgreement elapses.
+func (s *Session) AwaitSchemaAgreement(ctx context.Context) error {
+	if s.cfg.disableControlConn {
+		return errNoControl
+	}
+	return s.control.withConn(func(conn *Conn) *Iter {
+		return &Iter{err: conn.awaitSchemaAgreement(ctx)}
+	}).err
+}
+
 func (s *Session) reconnectDownedHosts(intv time.Duration) {
 	reconnectTicker := time.NewTicker(intv)
 	defer reconnectTicker.Stop()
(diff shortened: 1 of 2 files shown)


session.go
```diff
@@ -283,6 +283,18 @@ func (s *Session) init() error {
...
 }
 
+// AwaitSchemaAgreement will query the state of the schema versions across
+// all nodes of the cluster. The maximum amount of time this takes is governed
+// by the MaxWaitSchemaAgreement setting in the configuration (default: 60s).
+func (s *Session) AwaitSchemaAgreement(ctx context.Context) error {
+	if s.cfg.disableControlConn {
+		return errNoControl
+	}
+	return s.control.withConn(func(conn *Conn) *Iter {
```
annismckenzie: I didn't use `s.control.awaitSchemaAgreement()` here because I wanted the context to be honored and used. I hope that's okay. I can change it if you want.

annismckenzie:
Could someone take a look at this? 😅

session.go
```diff
@@ -283,6 +283,18 @@ func (s *Session) init() error {
 	return nil
 }
 
+// AwaitSchemaAgreement will query the state of the schema versions across
+// all nodes of the cluster. The maximum amount of time this takes is governed
+// by the MaxWaitSchemaAgreement setting in the configuration (default: 60s).
```
martin-sucha: The documentation comment could be clarified  a little bit. `will query the state of the schema versions` seems to me like it could query just once, but the function retries in the case schema versions do not match.

```
// AwaitSchemaAgreement will wait until schema versions across all nodes in the
// cluster are the same (as seen from the point of view of the control connection).
// The maximum amount of time this takes is governed
// by the MaxWaitSchemaAgreement setting in the configuration (default: 60s).
// AwaitSchemaAgreement returns an error in case schema versions are not the same
// after the timeout specified in MaxWaitSchemaAgreement elapses.
```

session.go
martin-sucha: Looks okay to me.

session.go
annismckenzie: That's a very good point! I used your suggestion. 💃

session.go
annismckenzie: Very good, thank you!

PR #1397, opened 2020-01-31
change CustomDialer type *net.Dialer to func(context.Context, string, string) (net.Conn, error)

kpango:
Hi, thanks for providing a great library

I want to make some changes to #1376 
@rinx did a great job at #1376 and likes his idea, but there's a problem with dialer customization.
In #1376 Dialer type is *net.Dialer, but net.Dialer is struct we can not customize dialer we can just configure only net.Dialer's field configuration,
I wish I could use `func(context.Context, string, string) (net.Conn, error)` type for Dialer instead of net.Dialer, but how about?

FYI:
in [go-redis](https://github.com/go-redis/redis) they also provide the [custom dialer functionality](https://github.com/go-redis/redis/blob/master/cluster.go#L56)
and the use `func(context.Context, string, string) (net.Conn, error)` type for configuration

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -113,3 +113,4 @@ Thomas Meson <zllak@hycik.org>
 Martin Sucha <martin.sucha@kiwi.com>; <git@mm.ms47.eu>
 Pavel Buchinchik <p.buchinchik@gmail.com>
 Rintaro Okamura <rintaro.okamura@gmail.com>
+Yusuke Kato <i.can.feel.gravity@gmail.com>
--- a/cluster.go
+++ b/cluster.go
@@ -5,6 +5,7 @@
 package gocql
 
 import (
+	"context"
 	"errors"
 	"net"
 	"time"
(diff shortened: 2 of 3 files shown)


kpango:
@Zariel @alourie @martin-sucha @rinx I'd like to get your feedback.

martin-sucha:
Hmm, good point. When reviewing the #1376  it didn't occur to me that net.Dialer is a struct and not an interface. We definitely want a function/interface to be able to support the use-cases mentioned in #1376 .

The change here looks good to me.

I'm not sure what is the current backward-compatibility policy of gocql (i.e. whether we can change the type from *net.Dialer to `func`). @Zariel @alourie maybe it would be good to state what the project promises regarding BC-compatibility somewhere, I couldn't find it mentioned in the README.

Changing to func shouldn't cause too much problems as the field was introduced only recently and can be easily fixed by using `net.Dialer.WithContext` as the value.

There is also an option to use

```go
interface DialContexter {
DialContext(ctx context.Context, network, address string) (Conn, error)
}
```

as a middle ground which would work even if someone uses the field already and writes `*net.Dialer` there (but would break if they read the value back).

And if we want to be really strict about the backward compatibility, then we'd need to introduce a separate field.

Zariel:
made this an interface, its a non breaking api change and more flexible

kpango:
why this PR closed? should I re-send PR implementing Dialer interface ?

alourie:
@kpango closed via #1398

PR #1405, opened 2020-02-17
Fix selection of host by token

martin-sucha:
We've seen increased per-query latency after upgrading to latest gocql.
This commit fixes the latency issue for us.

This partially reverts 7b17705d75148afe48e336a49743178a201cdda3. There
were multiple unrelated changes in that commit. The relevant part of
that commit's description is:

    Fix finding the correct token in the token ring for host selection.

Unfortunately that description does not contain the details of the issue
it was aimed to fix.

The change in 7b17705d75148afe48e336a49743178a201cdda3 also made the two
binary search implementations (tokenRingReplicas.replicasFor and
tokenRing.GetHostForToken) inconsistent, reverting the change fixed this
discrepancy as well.

The token values returned in system.peers and system.local tables are
ends of ranges that the nodes are responsible for[1], so we never need
to rollunder.

When searching for a node responsible for token X (computed from
partition key), we need to find hostTokens with end token Y with the
lowest value such that Y >= X. In case X is larger than the largest
end token value in tokenRingReplicas, we wrap the ring and the node with
the lowest token is responsible (i.e. the hostTokens at index 0).

[1] https://docs.datastax.com/en/dse/6.7/dse-arch/datastax_enterprise/dbArch/archAboutDataDistribute.html

### DIFF
--- a/policies_test.go
+++ b/policies_test.go
@@ -106,8 +106,8 @@ func TestHostPolicy_TokenAware_SimpleStrategy(t *testing.T) {
 	// now the token ring is configured
 	query.RoutingKey([]byte("20"))
 	iter = policy.Pick(query)
-	iterCheck(t, iter, "0")
 	iterCheck(t, iter, "1")
+	iterCheck(t, iter, "2")
 }
 
 // Tests of the host pool host selection policy implementation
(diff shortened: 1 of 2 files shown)


martin-sucha:
Hi @Zariel I'm not sure whether it's the correct fix since I couldn't find good documentation on how the token ranges in Cassandra work exactly (if you happen to have a link to a good documentation that describes the edge cases, please let me know), but this change fixed a latency regression here. What was the reason for the rollunder change in https://github.com/gocql/gocql/commit/7b17705d75148afe48e336a49743178a201cdda3 ?

martin-sucha:
Okay I finally found a piece of official documentation I was looking for that describes the edge cases: https://docs.datastax.com/en/dse/6.7/dse-arch/datastax_enterprise/dbArch/archAboutDataDistribute.html

(I found some other contradicting pieces of information on the internet)

The docs confirm that the tokens returned from `system.peers` and `system.local` are end of the range (the end is included in the range).

I'm going to update the commit message to include the explanation of why this fixes the issue including a link to the documentation.

PR #1413, opened 2020-03-02
Fix data race when accessing hostMetrics

martin-sucha:
ObservedQuery and ObservedBatch expose a pointer to hostMetrics struct
to the respective observers. However, the fields of the returned
*hostMetrics are not safe to read, since the observer that might access
hostMetrics' fields is called without the lock held.

In order to return a safe value to clients, we need to make a copy of
the hostMetrics struct while the lock is still locked.

This commit also optimizes the count of lock/unlock calls, previously we
did 3 pairs (for updating attempts, for updating latency and for getting
hostMetrics), now we only do one.

### DIFF
--- a/session.go
+++ b/session.go
@@ -696,7 +696,11 @@ func (s *Session) MapExecuteBatchCAS(batch *Batch, dest map[string]interface{}) 
 }
 
 type hostMetrics struct {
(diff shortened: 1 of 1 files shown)


Zariel:
all these things need pulling off the query type and onto iter

Issue #1401, opened 2020-01-31
roundRobinHostPolicy uses lots of resources for every query


muirdm:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?
3.11.5

### What version of Gocql are you using?
1982a06ad6b987c24beec15336019914d71b2e31

### What did you do?
Due to #1400 we were accidentally using the round robin host selection policy for every query.

### What did you expect to see?
I expected to see normal client side performance.

### What did you see instead?
I did CPU profile and noticed 10% CPU used by `rand.NewSource`. In addition, it allocates a lot, taking up ~30% CPU via GC.

AlexeyUzhva:
Just faced same issue.

Confirming that it happens due to excessive memory allocations.

![image_2020_02_02T09_41_25_771Z](https://user-images.githubusercontent.com/1519328/73606418-16169100-45bb-11ea-9ee7-95aa1fcb8443.png)

AlexeyUzhva:
Submitted merge request for that. Please review. It's not only memory efficient, and also faster :)

Zariel:
can we just pool rands?

AlexeyUzhva:
Already attached MR which eliminates random and array copying. It just uses single atomic counter which is interpreted of start offset for host selection.

So, no copying at all, and no expensive random calls. Saves 60% of allocations of entire library!...

AlexeyUzhva:
@Zariel why do we need to have random at all? Random is slow, and also requires array copying.

In our project we use high throughput submission something around 300k queries per second. It means 300k arrays allocations per second, 300k shuffles per second.

This is considerable GC overhead just to make a copy of it.

While it's not necessary at all. It's enough just to loop array from incremental offset without any copies of it.

nightlyone:
@Zariel / @AlexeyUzhva I am not sure doing this completely without randomness is a good idea. If you have N machines accessing the same set of M Cassandra hosts and all N start talking to machine node X, overloading it and iterating all the way through the M Cassandra hosts. 

Would you mind randomizing at least the start of that iteration on session creation and maybe also when changing the set of hosts?

AlexeyUzhva:
Hi, randomising initial list of hosts sounds like a good idea. May add so.

AlexeyUzhva:
@nightlyone I've added to the MR randomising of initial offset. Now every client start counting from different (random) position. And it's still allocation free.

PR #1415, opened 2020-03-06
Randomising initial offset as per comments in #1401

AlexeyUzhva:
As @nightlyone mentioned, zero-offset may be suboptimal in hypothetical scenario of multiple clients coming into server at a same time.

While this might be very unlikely scenario, the fix is easy and makes no harm. 

Let's pick initial offset randomly (only first time).

### DIFF
--- a/policies.go
+++ b/policies.go
@@ -329,7 +329,9 @@ type NextHost func() SelectedHost
 // RoundRobinHostPolicy is a round-robin load balancing policy, where each host
 // is tried sequentially for each query.
 func RoundRobinHostPolicy() HostSelectionPolicy {
-	return &roundRobinHostPolicy{}
+	return &roundRobinHostPolicy{
+		lastUsedHostIdx: rand.Uint64(),
+	}
 }
 
 type roundRobinHostPolicy struct {


AlexeyUzhva:
Additionally fixes uint64 / int overflow (safe for long-running pools)

AlexeyUzhva:
@nightlyone I believe simplicity is a bigger virtue in this case. Selection of a first server doesn't need at all to be cryptographically strong. And the simplest one is best for future maintenance.

nightlyone:
@AlexeyUzhva sorry for the late answer, but the explanation is a bit longer.

I suggsted to re-introduce the seeding code from https://github.com/gocql/gocql/commit/aca01c205cbe3f6ae880c4d39f6116a944e5dfde#diff-6c19a87ca4beb90f7fb10e2297c83dd5L853-L865 

Reasons:

The changes in the pull request would use an **unseeded random source**.  So it will still have the effect that every client will select the same node. Just another node this time. This is the effect of unseeded randomness. 

https://godoc.org/math/rand documents at the package level documentation 

> "Top-level functions, such as Float64 and Int, use a default shared Source that produces a deterministic sequence of values each time a program is run. **Use the Seed function to initialize the default Source if different behavior is required for each run**."

So again, the "simple" way is still buggy without a seeded random source. And since seeding the global random source by just calling rand.Seed is very rogue in a library, I suggested re-introducing the correct code for only seeding the random sources of the gocql library as linked above.

Please be aware that random source are not concurrency safe as per the same package level documentation.

So once you re-introduced `func randSource()` you need to replace `rand.Uint64()` with `randSource.Uint64()` and are done.

AlexeyUzhva:
@nightlyone I'm not sure we talk about same thing.

I suggest to define a little bit more clear a scope for a statement: "every client select same node".

Node is being rotated on every request. Every requests shifts offset. This random is used only to do an initial setup. Once settled up it begins rolling by itself, changing node on every request.

So I hardly can come up with a scenario when 2 clients can consistently pick same nodes. For that to match those clients:

1. Must start at exactly same time.
2. AND run same number of requests at exactly same times each. 

----

If you would look into official DataStax driver sources you find exactly same logic:

1. Here is a one-time initial index initialisation - https://github.com/datastax/java-driver/blob/3.5.x/driver-core/src/main/java/com/datastax/driver/core/policies/RoundRobinPolicy.java#L63
2. Here is an increment on every requests - https://github.com/datastax/java-driver/blob/3.5.x/driver-core/src/main/java/com/datastax/driver/core/policies/RoundRobinPolicy.java#L115

Lots of people have been running that happily onto heavy productions for years.

martin-sucha:
Go uses seed `1` every time https://golang.org/src/math/rand/rand.go#L293 while Java uses a time based seed when you create a new instance of `java.util.Random` http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/tip/src/share/classes/java/util/Random.java#l105

AlexeyUzhva:
Ok. That can be easily fixed, will do a bit later. It's only 1 time seeding during policy creation, not per request.

nightlyone:
Thank you @martin-sucha that was indeed was I was trying to explain. @AlexeyUzhva seeding was really the only missing piece in the changes you added for randomization.

Since gocql once had a great implementation of it. I just suggested to add back that battle hardened code.

Issue #1343, opened 2019-09-04
Null tuple cause panic


supar:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?
cqlsh 5.0.1 | Cassandra 3.11.4 | CQL spec 3.4.4 | Native protocol v4

### What version of Gocql are you using?
f6df8288f9b44e6

### What did you do?
```
CREATE TABLE products (
            id timeuuid PRIMARY KEY,
            name text,
            category frozen<tuple<timeuuid, varchar>>
 );

INSERT INTO products(id, name) VALUES (cb403e20-ce63-11e9-b49a-d99f577eb025, 'Without category');

SELECT * FROM products;

 id                                   | name       | category
--------------------------------------+------------+----------
 cb403e20-ce63-11e9-b49a-d99f577eb025 | Without category |     null
```

### What did you expect to see?
no panic

### What did you see instead?
```
http: panic serving 172.17.0.1:57798: runtime error: index out of range
goroutine 61 [running]:
net/http.(*conn).serve.func1(0xc000348140)
	/usr/local/go/src/net/http/server.go:1769 +0x139
panic(0xa2b360, 0x11f5160)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/gocql/gocql.readInt(...)
	/go/src/github.com/gocql/gocql/frame.go:322
github.com/gocql/gocql.readBytes(...)
	/go/src/github.com/gocql/gocql/marshal.go:1788
github.com/gocql/gocql.unmarshalTuple(0xbcfd20, 0xc000372540, 0x0, 0x0, 0x0, 0x9e5280, 0xc000380420, 0x0, 0x0)
	/go/src/github.com/gocql/gocql/marshal.go:1810 +0x11a7
github.com/gocql/gocql.Unmarshal(0xbcfd20, 0xc000372540, 0x0, 0x0, 0x0, 0x9e5280, 0xc000380420, 0xc0000c6d68, 0x69a3c5)
	/go/src/github.com/gocql/gocql/marshal.go:162 +0xdaa
github.com/gocql/gocql.scanColumn(0x0, 0x0, 0x0, 0xc000024ff0, 0x3, 0xc000024ff8, 0x8, 0xc000025028, 0x8, 0xbcfd20, ...)
	/go/src/github.com/gocql/gocql/session.go:1292 +0x1b3
[...]

PR #1424, opened 2020-04-02
tuples: correct unmarshalling of unset tuple values

dahankzter:
In case the tuple value has never been written the unmarshalling
used to panic. This is now handled by inspecting the data from the
database and if it is missing (nil) it will set the tuple value to
the corresponding default value.

Fixes: #1343

### DIFF
--- a/tuple_test.go
+++ b/tuple_test.go
@@ -107,7 +107,11 @@ func TestTuple_TupleNotSet(t *testing.T) {
 
 	const id = 1
 
-	err = session.Query("INSERT INTO tuple_not_set_test(id) VALUES(?)", id).Exec()
+	err = session.Query("INSERT INTO tuple_not_set_test(id,coord) VALUES(?, (?,?))", id, 1, 2).Exec()
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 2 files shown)


piodul:
@alourie @Zariel could you take a look and review?

tuple_test.go
```diff
@@ -89,6 +89,45 @@ func TestTuple_NullTuple(t *testing.T) {
...
+		coord frozen<tuple<int, int>>,
+
+		primary key(id))`)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	const id = 1
+
+	err = session.Query("INSERT INTO tuple_not_set_test(id) VALUES(?)", id).Exec()
```
mmatczuk, line 110: In the test we should have 2 rows, one with data the other without.
After reading row 1 x and y should be set accordingly.
After reading row 2 x and y should be set to 0.

dahankzter: Good point @mmatczuk we have to ensure they are reset. I think they should be but a test for it makes sense.

dahankzter: It didn't reset it.... I was hoping the delegated code would handle this case but apparently not @mmatczuk .

dahankzter: Nvm I messed up the evaluation order of the two rows. It seems to work fine.

PR #1425, opened 2020-04-09
Added unit test for testing cowHostList.get() data race.

GingerMoon:
Hi @Zariel , I made this PR based on your TODO comment.
Any insights would be appreciated.

p.s. From some point of view, cowHostList is not a "copy on write" because its get() also returns a  copy. So it's possible to be refactored in future. The unit test added is a reminder when we do the refactor in future.

### DIFF
--- a/policies.go
+++ b/policies.go
@@ -20,7 +20,7 @@ import (
 
 // cowHostList implements a copy on write host list, its equivalent type is []*HostInfo
 type cowHostList struct {
-	list atomic.Value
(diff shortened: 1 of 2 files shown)


Zariel:
You are miss understanding how slice copies work in Go, this is working as intended, its copy-on-write in the sense that callers get the same backing memory of the slice, they get a different copy of the slice header but that is irrelevant.

GingerMoon:
Thanks @Zariel . Now I get the point. The unit test is added to ensure "they get a different copy of the slice header" insteand of the slice header itself which is stored in the list?

PR #1429, opened 2020-04-10
fill the connection pool in another goroutine to speed up Session.init

GingerMoon:
Today I found my client started very late. It's strange because my timeout config is not that long.
Then I found IPs of some of the nodes didn't work any more and pool.fill of those nodes takes much time, it sleeps and retrys several times. Also those pool.fill is invoked serially.
hostConnPool.fill only used in three places:
```
connectionpool.go
    addHost
        pool.fill()
    HandleError
        go pool.fill()
    Pick
        go pool.fill()
```
And neither of the three cares about the return of pool.fill(). It seems that only session.init() cares about the result of pool.fill().

### DIFF
--- a/session.go
+++ b/session.go
@@ -226,6 +226,8 @@ func (s *Session) init() error {
 		hostMap[host.ConnectAddress().String()] = host
 	}
 
+	hostAdded := make(chan interface{}, len(hostMap))
+
 	hosts = hosts[:0]
 	for _, host := range hostMap {
 		host = s.ring.addOrUpdate(host)
@@ -234,9 +236,16 @@ func (s *Session) init() error {
 		}
 
 		host.setState(NodeUp)
-		s.pool.addHost(host)
+		go func(host *HostInfo) {
+			s.pool.addHost(host)
+			hostAdded <- struct{}{}
+		}(host)
 
 		hosts = append(hosts, host)
+	}
+
+	if err := s.waitForConnAvailable(hostAdded); err != nil {
+		return err
 	}
 
 	type bulkAddHosts interface {
@@ -270,10 +279,6 @@ func (s *Session) init() error {
 		s.hasAggregatesAndFunctions = version.AtLeast(2, 2, 0)
 	}
 
-	if s.pool.Size() == 0 {
-		return ErrNoConnectionsStarted
-	}
-
 	// Invoke KeyspaceChanged to let the policy cache the session keyspace
 	// parameters. This is used by tokenAwareHostPolicy to discover replicas.
 	if !s.cfg.disableControlConn && s.cfg.Keyspace != "" {
@@ -281,6 +286,36 @@ func (s *Session) init() error {
 	}
 
 	return nil
+}
+
+func (s *Session) waitForConnAvailable(hostAdded chan interface{}) (err error) {
+	totalHosts := cap(hostAdded)
+	ch := make(chan error)
+	go func() {
+		addedHosts := 0
+		for range hostAdded {
+			addedHosts++
+
+			poolSize := s.pool.Size()
+			if poolSize > 0 {
+				ch <-nil
+			}
+			if addedHosts == totalHosts {
+				if poolSize == 0 {
+					ch <-ErrNoConnectionsStarted
+				}
+				return
+			}
+		}
+	}()
+
+	select {
+	case err = <-ch:
+	case <-s.ctx.Done():
+		err = s.ctx.Err()
+	}
+
+	return err
 }
 
 // AwaitSchemaAgreement will wait until schema versions across all nodes in the
--- a/session_test.go
+++ b/session_test.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strconv"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


GingerMoon:
By the way, this issue happens when client accesses(from outside of AWS) C* cluster deployed in aws via their public IPs.
Whatever Hostfilter and HostSelectionPolicy we configured, session.init() can be finished when there is any workable connection. Am I right?

martin-sucha:
This is somewhat related to https://github.com/gocql/gocql/pull/1369, that pull request aims to make the waiting policy configurable.

GingerMoon:
> This is somewhat related to #1369, that pull request aims to make the waiting policy configurable.

@martin-sucha Thanks for the information. Not sure whether we need to add another policy just to speed up the session's initialization. It seems that the new added policy in [#1369](https://github.com/gocql/gocql/pull/1369) also has some other purposes/functionalities.

I'd like to make filling pool asynchronously and before the end of session.init(), if there is any workable connection, then session is initialized correctly.
Regarding checking workable connection, it's similar to the logic of building connection(wait&retry).

I would appreciate if you could tell me if there is anything wrong in my understanding.

martin-sucha:
Waiting for connection to all nodes (the current state) in the cluster is not ideal, since a single database node that is down will force you to wait the entire timeout period (and currently addHost are called serially, which adds the startup latency). We definitely want to connect to database nodes concurrently.

Some programs don't care about the connections as long as the queries can be executed (for example command-line utilities, long-running jobs, etc.).

In some cases like API servers, you might want to be able to wait until the driver established connection to all the peers that it usually sends queries to. Queries should work even with just a single connection to the cluster as the coordinator node will then forward them to the correct node. However, this incurs latency costs for the hop (which might add up when you issue a lot of queries to your service). When doing rolling release/restarting your service, you might want to wait for all connections to be established before starting routing requests to the newly spawned instance so that you don't impact your service's latency during normal operations (when the database cluster is healthy).

For the reasons described above, some users (like me) might want to wait for nodes in the local DC to be reachable or to wait until we are connected to at least one replica for each token range, depending on how they use their cluster.

The waiting does not need to be done in `session.init` though, if it's possible to implement the waiting easily via other means (e.g. in the client application). There could be another method on session to wait, for example. Also I'm not convinced that the waiting policy should be tied to the `HostSelectionPolicy` as in #1369 since there are multiple possible waiting policies per `HostSelectionPolicy`. For some `HostSelectionPolicies`, like the token aware policy, we probably need to provide some waiting mechanism though, since implementing "wait for all tokens to be covered by at least one reachable node" is not easy to implement without access to the token ring.

session.go
```diff
@@ -270,6 +270,13 @@ func (s *Session) init() error {
 		s.hasAggregatesAndFunctions = version.AtLeast(2, 2, 0)
 	}
 
+	reconnectionPolicy := s.cfg.ReconnectionPolicy
+	for i := 0; i < reconnectionPolicy.GetMaxRetries(); i++ {
+		if s.pool.Size() > 0 {
+			break
+		}
+		time.Sleep(reconnectionPolicy.GetInterval(i))
+	}
```
martin-sucha, line 279: I'd prefer if this wait wasn't implemented as polling, since this might wait longer (or just a bit shorter) than necessary depending on the reconnection policy (and the steps there might be exponential).
Instead, we could (for example) close a channel the first time a pool is connected or the initial pool filling fails.

Or, we could remove the wait altogether along with the `if s.pool.Size() == 0 {` condition below, but that will break clients that rely on and check `ErrNoConnectionsStarted`.

GingerMoon: Thanks @martin-sucha for your insights.
My original idea was similar as "Instead, we could (for example) close a channel the first time a pool is connected or the initial pool filling fails.". 
But the following code snip executes not only during session.init() and I don't want add the code which is only related to session.init().
```
// Fill the connection pool
func (pool *hostConnPool) fill() {
...
        // fill only the first connection synchronously
	if startCount == 0 {
		err := pool.connect()
...
```
"Or, we could remove the wait altogether along with the if s.pool.Size() == 0 { condition below, but that will break clients that rely on and check ErrNoConnectionsStarted." Sounds not bad. I am still thinking if there is any other ways.

How about think of the following definition of session.init [the connection pool part]:
Only all of the connections to those literally configured hosts are ready, session.init() suceeds.
session.init wait until the configured host pool.filling == false, returned the NotAllConfiguredHostsConnected error.

GingerMoon:
When the system is totally warm up (all the connectable connection pools are connectable),  we maybe don't care "this incurs latency costs for the hop", do we?

GingerMoon:
@martin-sucha 
I made another version according to your comment "we could (for example) close a channel the first time a pool is connected or the initial pool filling fails.". A unit test is also added.

session.go
```diff
@@ -283,6 +288,24 @@ func (s *Session) init() error {
 	return nil
 }
 
+func (s *Session) waitForConnAvailable() error {
+	select {
+	case <-s.connAva:
+	case <-time.After(s.cfg.Timeout):
```
martin-sucha, line 294: Waiting for just `s.cfg.Timeout` is not correct, the connection initialization simply takes different amount of time. There is `s.cfg.ConnectTimeout`, but that is used just for dialing. When establishing a connection, there is retry policy. We also send some data back and forth, like query to set keyspace, so the connection takes more time to setup. In case the connection fails, we need to wait until we stop retrying.

If you did not change the `policyHostPool.addHost` to non-blocking, you would be able to handle all this waiting within `session.init` and without need for selecting a specific timeout. I believe it's beter to have the waiting implementation encapsulated just within that method. (See also #1369, where the waiting is all implemented within `session.init`)

GingerMoon: Done.

fastest963:
@GingerMoon I'm obviously bias but I'm concerned that your PR is changing the behavior in the default case. If someone is depending on being able to connect to more than 1 node, this will break that, there should at least be an opt-in way to get the new behavior. The library has been around for years before I opened an issue last October about this being an issue.

One issue I ran into with my PR (which was similar to yours at first) is that `s.control.connect(hosts)` ends up adding that host to the ring (and the policy) even though it only has a control connection to it. This was causing a race condition where we'd think we have a connection (because the pool already knows about the host) but it's not ready yet and then we'd prematurely return an error that we have no connections.

I instead changed the behavior to actually use `HostUp` and convert things over to calling that instead after a comment recommending I add all of the hosts first and then later treat them as up once we connected.

GingerMoon:
@martin-sucha  @fastest963 
Currently, session.pool[i].fill() is syncchronous, which means session.init() waits for all the connection pools fill are done, then init() checks if session.pool.Size() <= 0, and if yes, init returns ErrNoConnectionsStarted.

Theoretically, session.pool.Size() is 0  when:
1. There is no workable connection.
2. InvalidKeySpace.

Actually, if there is no workable connection, which means connections to the configured hosts are not workable, session.init returns the error before session.pool[i].fill().
So checking session.pool.Size() when session.init returns, seems only for the purpose of checking the InvalidKeySpace case.

So I'd like to remove ErrNoConnectionsStarted and added InvalidKeySpace error and make keyspace validation before hostConnPool.fill.
If this is workable, then we can safely remove session.pool.Size() check from session.init.

Any more insights would be appreciated!

martin-sucha:
#1369 is merged, so I think we can close this.

PR #1436, opened 2020-04-29
fix metadata views, incorrect query

beltran:
Fixes https://github.com/gocql/gocql/issues/1435

### DIFF
--- a/metadata.go
+++ b/metadata.go
@@ -22,7 +22,9 @@ type KeyspaceMetadata struct {
 	Tables          map[string]*TableMetadata
 	Functions       map[string]*FunctionMetadata
 	Aggregates      map[string]*AggregateMetadata
-	Views           map[string]*ViewMetadata
+	// Deprecated: use the MaterializedViews field instead.
+	Views             map[string]*ViewMetadata
+	MaterializedViews map[string]*MaterializedViewMetadata
 }
 
 // schema metadata for a table (a.k.a. column family)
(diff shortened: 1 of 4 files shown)


jorgebay:
I didn't have time to check it earlier, I'll review this on Monday.

metadata.go
```diff
@@ -84,10 +84,27 @@ type AggregateMetadata struct {
 
 // ViewMetadata holds the metadata for views.
 type ViewMetadata struct {
-	Keyspace   string
-	Name       string
-	FieldNames []string
-	FieldTypes []TypeInfo
+	Keyspace                string
+	Name                    string
+	BaseTableId             UUID
```
mpenick, line 89: What do you think about having a pointer to the base table? possibly instead of ID and name?

```go
BaseTable *TableMetadata
```

jorgebay:
Should we be concerned about breaking existing usage for `ViewMetadata` (udts)?

We have the option of leaving `ViewMetadata` as is and introduce a `MaterializedViewMetadata` for `system_schema.views`, wdyt?

metadata.go
beltran: I think it's a good idea, but it would only replace BaseTableName right? Because TableMetadata doesn't have the BaseTableId.

beltran:
Hmm, I may not fully understand what you mean, the problem I think it's that a bug that caused a change in the API was introduced. So you are suggesting adding below [here](https://github.com/gocql/gocql/blob/master/metadata.go#L25) a new line `MaterializedViews MaterializedViewMetadata` and leaving the data in `Views` as it is? Couldn't users get confused when they see both fields?

jorgebay:
Yeah, I understand that it may cause confusion to new users (we can document the behaviour.  Introducing a new map+struct and leave the existing functionality as it is will avoid breaking usages of views as udts.

beltran:
@jorgebay thank you for the feedback! I think this is ready for another looks. There's a failing test in one of the runs, but seems unrelated:
```
--- FAIL: TestPolicyConnPoolSSL (0.12s)
    conn_test.go:1191: write tcp 127.0.0.1:34668->127.0.0.1:34356: use of closed network connection
```

PR #1437, opened 2020-05-05
Add eof checks to unmarshalList/Map

martin-sucha:
There were missing checks for length which could lead to panic when the
server sends malformed data. This commits adds the missing checks.


jorgebay:
I think I'll be able to review this on Monday.

PR #1441, opened 2020-05-15
Adding ecaudit.auth.AuditAuthenticator to connection whitelist.

rsbala:
To address issue https://github.com/gocql/gocql/issues/1440.
This is a trivial change to add a new authenticator of an existing plugin - ecaudit to the whitelist.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -31,6 +31,7 @@ var (
 		"io.aiven.cassandra.auth.AivenAuthenticator",
 		"com.ericsson.bss.cassandra.ecaudit.auth.AuditPasswordAuthenticator",
 		"com.amazon.helenus.auth.HelenusAuthenticator",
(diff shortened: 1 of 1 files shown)


alourie:
@rsbala Do you think we should remove AuditPasswordAuthenticator then? Or keep it for backwards compatibility?

I'm afraid that if we keep it, then when removed from Ericsson's project completely, it'll still linger here forever.

rsbala:
Thanks @alourie for the quick approval.

Ecaudit 2.5.0 has been released just two months back. If we remove AuditPasswordAuthenticator, it will be an issue for those who are using ecaudit<2.5.0 and they can never use the latest version of gocql. But I agree with you that will linger forever. For now we might have to leave it for backward compatibility. Please let me know if that is fine.

alourie:
@rsbala fair enough. I'll merge this one, but would you mind opening a separate ticket to remove it? This way at least we'll have a reminder.

PR #1443, opened 2020-05-21
add types metadata

beltran:


metadata.go
```diff
@@ -25,6 +25,7 @@ type KeyspaceMetadata struct {
 	// Deprecated: use the MaterializedViews field instead.
 	Views             map[string]*ViewMetadata
 	MaterializedViews map[string]*MaterializedViewMetadata
+	Types             map[string]*TypeMetadata
```
jorgebay, line 28: `system_schema.types` hold the info for user-defined types. Should we name this new field `UserTypes` and the struct `UserTypeMetadata` to make it obvious?

metadata.go
```diff
@@ -25,6 +25,7 @@ type KeyspaceMetadata struct {
 	// Deprecated: use the MaterializedViews field instead.
```
jorgebay, line 25:
```suggestion
	// Deprecated: use the MaterializedViews field for views and UserTypes field for udts instead.
```

metadata.go
beltran: Good suggestion @jorgebay , done this.

PR #1453, opened 2020-06-03
Add gokit logging and Prometheus metrics.

tomwilkie:
Hello!  This is more of an RFC, the gocql has certainly not take an opinionated approach on logging or monitoring yet.  I think we should - and in the go world gokit and prometheus seem to have become the standard.

So do you think this is useful?  If Logger and Registerer are not set, this PR should be a no-op.

Signed-off-by: Tom Wilkie <tom.wilkie@gmail.com>


martin-sucha:
Hi,

since there could be gocql v2, this discussion is definitely interesting to have, thanks for bringing it up.

I'm not convinced (yet?) that gocql should use an opinionated solution. Could you elaborate on why do you think that we should? What are the advantages of using gokit logger instead of a simple Logger interface like was here before? (Apart from the fact that the `Logger` variable is global currently, but we should be able to remove that in v2 since that's a backwards incompatible change.) Similar with the Prometheus integration - why do you think it's better to use Prometheus instead of the Observer interfaces we have currently?

For example, at my work we currently use a different library both for logging and for metrics. We have some custom tags on metrics from gocql based on the data available in the observer interfaces. With the observer interface we have complete control over the metrics, while using auto registered prometheus metrics does not seem to be that flexible. You might want to be able to add custom tags or you might want to even remove tags (if you have many hosts and the cardinality of the metrics would be too high with using app host x cassandra host tags, you might want to have app host x rack or app host x DC, for example).

On the other hand, having some common/starter-pack set of gocql metrics could be useful as well, since there could be some dashboard template which users could just import and use. This use-case does not require to use Prometheus internally in the main gocql package - it could be implemented using the Observer interfaces in a separate package or repo.

As for the code, I quickly checked the pull request and I see that some logging messages were changed from using Logger.Printf to the gokit logger, but I don't see any code that would ensure these log messages will still be routed to the global logger if no gokit logger was provided in the session. Maybe I just missed it, so in case it's there somewhere, could you please point me to it?

Zariel:
I don't want add this to gocql, this is something that should live outside of the main driver, we shouldn't be in a position to impose these things on users.

mrVanboy:
What is the reasoning behind introducing a dependency on the `gokit/logging`? If the [interface](https://pkg.go.dev/github.com/go-kit/kit@v0.10.0/log?tab=doc#Logger)
```go
type Logger interface {
	Log(keyvals ...interface{}) error
}
```
is the best option, `gocql` can have its own implementation to utilize "duck typing" interfaces feature.

Zariel:
Thanks but I dont want to add dependencies on prometheus or gokit to gocql, prometheus can already be implemented using observers, if that is not possible please raise an issue.

tomwilkie:
(Started writing this response, but as it looks like Zariel won't accept this, this response is only part finished)

I wrote this PR to help debug an intermittent but persistent "no hosts in pool" bug.

# Re: Opinions

> I'm not convinced (yet?) that gocql should use an opinionated solution. Could you elaborate on why do you think that we should?

In general I look for libraries to be "batteries included" out of the box, and not require extension.  This library is aimed at large-scale systems, and such systems require metrics and logs to be operated.  So I'd hope a library like this would expose a reasonable set of metrics and logs by default.

All that being said I think "batteries included but extensible" is the way forward, so I agreed users should be able to add their own logging and monitoring.  I don't think this PR prevents that.

# Re: Logging

> What are the advantages of using gokit logger instead of a simple Logger interface like was here before? (Apart from the fact that the Logger variable is global currently, but we should be able to remove that in v2 since that's a backwards incompatible change.)

One could argue that the gokit logger is a simpler interface than the existing Logger interface... it only has one method.  But this is not something I'm super opinionated about.  

Not having the globals is important, and in general I don't think every library should invent their own logging interface.  Gokit logging seems popular, so why not use that?

>  I see that some logging messages were changed from using Logger.Printf to the gokit logger, but I don't see any code that would ensure these log messages will still be routed to the global logger if no gokit logger was provided in the session. Maybe I just missed it, so in case it's there somewhere, could you please point me to it?

Thats a good point, I missed this.  ~I will add.~

# Re: Metrics

> Similar with the Prometheus integration - why do you think it's better to use Prometheus instead of the Observer interfaces we have currently?

Unfortunately it is very hard to expose detailed metrics using the observer interface pattern, and this pattern leads to spaghetti code.  This pattern also puts quite a burden on the library authors to expose everything through very wide interfaces, which might explain why there are very few points to expose useful metrics and no good defaults.  For instance, I couldn't find a way to expose the metrics in this PR without modifying gocql.

As it stands, I expect most users of this library have no metrics.

tomwilkie:
(I'll be maintaining and extending this branch over at grafana/gocql if anyone is interested)

Zariel:
I would be happy to have these things live out of tree I just don't want to have them in the main tree. I can see the benefit of a `gocql-contrib` style repo which has prometheus metrics + work to improve the interfaces to support them. See #1459

PR #1466, opened 2020-06-26
feat(opentracing): add support for an opentracing query observer

FrankSpitulski:
Closes https://github.com/gocql/gocql/issues/951

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/golang/snappy v0.0.0-20170215233205-553a64147049
 	github.com/hailocab/go-hostpool v0.0.0-20160125115350-e80d13ce29ed
 	github.com/kr/pretty v0.1.0 // indirect
+	github.com/opentracing/opentracing-go v1.1.0
 	github.com/stretchr/testify v1.3.0 // indirect
 	gopkg.in/inf.v0 v0.9.1
 )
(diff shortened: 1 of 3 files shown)


FrankSpitulski:
the tags match https://github.com/opentracing-contrib/java-cassandra-driver as best as I can

nightlyone:
This could happily live in a different package as the implementation uses only exposed publically exposed interfaces. In fact we use the same interface (and a bit more machinery) to implement the opencensus version of tracing for gocql.

Background: The people behind opentracing and opencensus now work together on a new, unified [open telemetry standard](https://opentelemetry.io/about/) and stated they will obsolete opentracing as well as opencensus when that work is done, I would suggest NOT adding either of opentracing and opencensus to gocql and instead keep the APIs open enough to implement it outside of this package to lower future maintenance efforts.

FrankSpitulski:
Sure, that's how I'm using it now. I just figured you might want it in here since there was an open issue to add it where a contributor asked for someone to help.

Zariel:
I do not want to have this in the main tree, I am happy to have a gocql-contrib repo with this and other monitoring helpers in it.

martin-sucha:
Closing as this won't go to the main gocql/gocql tree/package. @Zariel if you create the gocql-contrib repo, perhaps we could move `HostPoolHostPolicy` there as well (and deprecate in gocql/gocql in v1), as it too just wraps an external dependency.

PR #1467, opened 2020-06-30
Conn.close calling Session ObserveDisconnect

xunilrj:
Implements https://github.com/gocql/gocql/issues/756 .

Possible problems:

1 - Implementations of "ConnectObserver" will not be valid anymore. See https://github.com/gocql/gocql/pull/1467/commits/fd6472f2981091234a8c3afc785ee74b404d515e#diff-f2fccfb96064097b73fbe8eeae3703eeR2037. To avoid this problem, one solution is to follow the issue suggestion and start using channels for notifications.

2 - Maybe not a realistic problem, but one can call "session.Close()" before the connection pools are full. Exactly what happens in the test. I chose the simple solution of waiting for the filling to finish before closing the Session. This explains the promotion of the fill WaitGroup to inside the "hostConnPool" struct and the "Wait()" in the "pool.Close()".

Example of the implementation.
```
func (o PrintConnectObserver) ObserveDisconnect(e gocql.ObservedDisconnect) {
    fmt.Println("ObserveDisconnect.Host: ", e.Host)
    fmt.Println("ObserveDisconnect.Start: ", e.Start)
    fmt.Println("ObserveDisconnect.End: ", e.End)
    fmt.Println("ObserveDisconnect.Err: ", e.Err)
}

cluster := gocql.NewCluster("172.18.0.3")
cluster.ConnectObserver = PrintConnectObserver{}

cluster.CreateSession()
session.Close() // <- generates the log below
```

Result:

```
ObserveDisconnect.Host:  [HostInfo hostname="172.18.0.3" connectAddress="172.18.0.3" peer="<nil>" rpc_address="<nil>" broadcast_address="<nil>" preferred_ip="<nil>" connect_addr="172.18.0.3" connect_addr_source="connect_address" port=9042 data_centre="" rack="" host_id="" version="v0.0.0" state=UP num_tokens=0]
ObserveDisconnect.Start:  2020-06-30 13:48:57.1785857 +0100 BST m=+0.022178701
ObserveDisconnect.End:  2020-06-30 13:48:57.1786369 +0100 BST m=+0.022229901
ObserveDisconnect.Err:  <nil>
````

Test:

```
go test -tags cassandra -run TestSessionObserveConnectAndDisconnect -v 
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -498,7 +498,7 @@ func (c *Conn) closeWithError(err error) {
 
 	// if error was nil then unblock the quit channel
 	c.cancel()
-	cerr := c.close()
+	cerr := c.close(err)
 
 	if err != nil {
 		c.errorHandler.HandleError(c, err, true)
@@ -508,8 +508,29 @@ func (c *Conn) closeWithError(err error) {
 	}
 }
 
-func (c *Conn) close() error {
-	return c.conn.Close()
+func (c *Conn) close(err error) error {
+	e := ObservedDisconnect{}
+
+	if c.session.connectObserver != nil {
+		e.Host = c.host
+		e.Start = time.Now()
+	}
+
+	r := c.conn.Close()
+
+	if c.session.connectObserver != nil {
+		// return specified error, otherwise return network error
+		if err != nil {
+			e.Err = err
+		} else {
+			e.Err = r
+		}
+
+		e.End = time.Now()
+		c.session.connectObserver.ObserveDisconnect(e)
+	}
+
+	return r
 }
 
 func (c *Conn) Close() {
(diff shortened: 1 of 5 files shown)


dahankzter:
One spontaneous thing is that it's not backwards compatible if I get the extension of the ConnectObserver interface. 
Why not just have another interface DisconnectObserver and another field in the session for an instance of that as well?

dahankzter:
I think item 2 is for another PR since it isn't really connected to the notification forwarding as far as I can tell. I suggest a new issue and a separate PR for this.

xunilrj:
1 - Ok.
2 - Ok.

Issue #1348, opened 2019-09-18
Upgrade gocql version causes for loop iter.Scan() to mess up slice 


Tommy-42:
### What version of Cassandra are you using?
3.11.4

### What version of Gocql are you using?
3540fc649cd7fc57cef5612b7bacac7a4fc443d6
16cf9ea1b3e28090d416d36528f304e4553e6b56

### What did you do?
I upgraded gocql from 3540fc6 to 16cf9ea

It appears that the code below doesn't works well anymore
```
var keys []models.Key
var key models.Key

for iter.Scan(&key.X, &key.Y, &key.Z) {
    keys = append(keys, key)
}
```

where :
X is `string`
Y is `[]byte`
Z is `[]byte`

### What did you expect to see?
---->
X: private
Y: foo
Z: bar
---->
X: public
Y: fuzz
Z: buzz

### What did you see instead?
---->
X: private
Y: fuzz
Z: buzz
---->
X: public
Y: fuzz
Z: buzz

---

for instance it was run & tested on go 1.10.8 on OSX

Everything works well with 3540fc6 but when updated to 16cf9ea the data got messed up somehow :(

Do you have any Idea why it could happen ?

To reproduce that I just need to upgrade the glide and fetch one of the version listed above.

Tommy-42:
I've manage to make it work by reseting the `key` value to default value

```
var keys []models.Key
var key models.Key

for iter.Scan(&key.X, &key.Y, &key.Z) {
    keys = append(keys, key)
    key = models.Key{} // fix here
}
```

I still don't know why it work without it with the old version ...

omosu:
Hi @Tommy-42, I have found changes in `Unmarshal` func.

https://github.com/gocql/gocql/commit/f596bd36e19ecaa7a9be478cf137fedc75036b02#diff-984b621d683cf81bbf404d291c4802a8R235

But for me is unclear what behavior should be: clean after append or gocql lib should copy values to the new array.

Tommy-42:
Yep, 
I don't mind cleaning it after or the lib doing it but either way it should be documented somewhere :D 
especially in the example in the readme

Zariel:
As far as I am aware this is behaving as expected, the address of `key.X` is the same for each iteration, the value is copied into the slice which has its own memory but the pointer to the slice is the same. See https://play.golang.org/p/ayzN8-UzASS

The interface for reading from the iter doesn't help this, use `iter.Scanner()` instead then it is more obviously correct.

[...]

PR #1478, opened 2020-08-12
Demonstrate unmarshalling with []byte fields

jeffwidman:
In https://github.com/gocql/gocql/pull/1167 the unmarshalling behavior
for `[]byte` was changed to re-use the existing slice if it is non-nil.

However, this results in extremely confusing behavior when unmarshalling 
into a struct because it causes those fields (and only those fields) to be all
mutated to the value of the final row.

Reverting this change would result in less confusing behavior, but also a
significant performance regression for folks trying to leverage this. 
So instead add an example to the `README` demonstrating the issue 
and the fix of re-setting the struct field value between iterations.

Fix https://github.com/gocql/gocql/issues/1348

H/T to @Tommy42 for reporting the issue and 
@continuum-oleksandr-mosur for tracking down the root cause.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ func main() {
 
 	/* Search for a specific set of records whose 'timeline' column matches
 	 * the value 'me'. The secondary index that we created earlier will be
-	 * used for optimizing the search */
+	 * used for optimizing the search. */
 	if err := session.Query(`SELECT id, text FROM tweet WHERE timeline = ? LIMIT 1`,
 		"me").Consistency(gocql.One).Scan(&id, &text); err != nil {
 		log.Fatal(err)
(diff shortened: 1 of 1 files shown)


Zariel:
this isn't a bug and is due to reusing the same memory for reading each row, the iter.Scan api is poor and really should be removed which I hope to do in v2

jeffwidman:
@Zariel did you read the PR? 

I never called the code implementation itself a bug, nor did I change the behavior. I simply documented it since the behavior can be a bit confusing/surprising... as demonstrated by the fact that multiple other devs spent hours trying to figure it out as well over in #1348. 

Don't you want better docs?

jeffwidman:
Nudge @Zariel can you address above?

martin-sucha:
Since #1513 the examples have been moved from README to package docs and we now also have this section in documentation: https://pkg.go.dev/github.com/gocql/gocql#hdr-Reusing_slices

If you think those docs can be improved please open a new pull request and I'll review it.

Issue #1481, opened 2020-08-14
CPU hogging / unresponsive application 


DigitalMarc:
### What version of Cassandra are you using?
3.11

### What version of Gocql are you using?
github.com/gocql/gocql v0.0.0-20200814012250-f09c24d7b32a

### What did you do?
execute a bunch of read queries in parallel

### What did you expect to see?
application keeps running forever

### What did you see instead?
The application runs fine using around 20-25% of the CPU with a very low memory footprint.
Sporadically, after running for a bit of time (in the couple of minutes range) the application starts behing unresponsive even to pprof endpoint queries and uses 90-100% of the CPU (1 core it seems) (very low memory footprint still).

### Describe your Cassandra cluster
The sample code to reproduce the issue has been made with a single cassandra 3.11 node.
Similiar behavior has been seen on various cassandra and scylla multi nodes clusters.

The following code as been used to ease the reproduction of the issue:
``` go
func main() {
	runtime.SetMutexProfileFraction(1)
	runtime.SetBlockProfileRate(int(time.Second.Nanoseconds() * 1))
	go func() {
		http.ListenAndServe(":7778", nil)
	}()

	cluster := gocql.NewCluster("192.168.0.10")
	cluster.ProtoVersion = 4
	cluster.Authenticator = gocql.PasswordAuthenticator{
		Username: "cassandra",
		Password: "cassandra",
	}

	session, err := cluster.CreateSession()
	if err != nil {
		panic(err)
	}

	for {
		for i := 0; i < 2000; i++ {
			go func() {
				query := "SELECT time, value FROM ks.table where id = 261713490 LIMIT 720"
				cqlQuery := session.Query(query)
				iter := cqlQuery.Iter()
				if err := iter.Close(); err != nil {
					fmt.Printf("err : %v \n", err)
				}
			}()
		}
		time.Sleep(500 * time.Millisecond)
	}
}
```

Here is the output with gocql_debug tag:
```
go run -tags "gocql_debug" main.go
2020/08/14 16:29:33 gocql: Session.handleNodeUp: 192.168.0.10:9042
2020/08/14 16:30:33 Session.ring:[192.168.0.10:UP]
```

I've attached whatever I can get from pprof (block; mutex; goroutine and cpu profile) before the application hangs:
[cpu.zip](https://github.com/gocql/gocql/files/5075163/cpu.zip)
[cpu.pdf](https://github.com/gocql/gocql/files/5075165/cpu.pdf)
[...]

PR #1486, opened 2020-08-28
conn: prevent potential race during writeCoalese stop method

Zariel:


cdent:
Is this potentially related to #1481 ?

DigitalMarc:
Hey, I've been testing with this change but the problem https://github.com/gocql/gocql/issues/1481 is still there. It took much longer to happen if that can be of any help. @cdent

conn.go
```diff
@@ -781,10 +781,18 @@ func (w *writeCoalescer) stop() {
...
+	if w.writeCh == nil {
+		return
+	}
+
 	w.flushLocked()
 	// nil the channel out sends block forever on it
 	// instead of closing which causes a send on closed channel
 	// panic.
+
+	w.quit <- struct{}{}
```
cdent, line 793: I have this pull request patched into my systems and am still sometimes getting "freezes". When I `SIGABRT` one o the goroutines that I can't explain (of the many that are left hanging around) is blocked on this line:

```
goroutine 55 [chan send, 66 minutes]:                                           
github.com/gocql/gocql.(*writeCoalescer).stop(0xc000100300)                     
    external/com_github_gocql_gocql/conn.go:793 +0x9d                           
github.com/gocql/gocql.(*writeCoalescer).writeFlusher(0xc000100300, 0xc9c300, 0xc0000a0000, 0x30d40)
    external/com_github_gocql_gocql/conn.go:840 +0x395                          
created by github.com/gocql/gocql.newWriteCoalescer                             
    external/com_github_gocql_gocql/conn.go:734 +0x164                          
```

Not clear if that's a bad thing or not, but seemed worth noting.

conn.go
```diff
@@ -723,22 +723,22 @@ func (c *deadlineWriter) Write(p []byte) (int, error) {
 	return c.w.Write(p)
 }
 
-func newWriteCoalescer(conn net.Conn, timeout time.Duration, d time.Duration, quit <-chan struct{}) *writeCoalescer {
+func newWriteCoalescer(ctx context.Context, conn net.Conn, timeout time.Duration, d time.Duration) *writeCoalescer {
```
nightlyone, line 726: What about deriving a context here, storing  cancel function and call that cancel stop instead of closing w.quit? This would limit that knowledge of  how to stop things to newWriteCoalescer and stop and keep writeFlusher the same.

That would also get rid of the danger of messing up channel sends, since the context package does that well enough already.

conn.go
```diff
@@ -781,10 +781,18 @@ func (w *writeCoalescer) stop() {
...
+		return
+	}
+
 	w.flushLocked()
 	// nil the channel out sends block forever on it
 	// instead of closing which causes a send on closed channel
 	// panic.
+
+	w.quit <- struct{}{}
+	close(w.quit)
```
nightlyone, line 794: the cancel call would go here then.

nightlyone (commented):
idea for simplifying the stop protocol.

PR #1487, opened 2020-08-31
Add Support For Secure Connection Bundle 

giltohar:
There is no existing gocql driver support for connecting to Cassandra Databases via Secure Connection Bundle, which our team at IBM required. IBM Cassandra is switching to Secure Connection Bundles and this will be needed for anyone using this golang driver to connect to them. We've been using this gocql driver now for a long time to connect to our previous Cassandra instance, and made the updates required to add connectivity support for Cassandra databases which use the Secure Connection Bundle.

### DIFF
--- a/session.go
+++ b/session.go
@@ -144,6 +144,13 @@ func addrsToHosts(addrs []string, defaultPort int, sniConfig *SNIConfig) ([]*Hos
 
 // NewSession wraps an existing Node.
 func NewSession(cfg ClusterConfig) (*Session, error) {
+
+	// Not able to support DisableInitialHostLookup and secure bundle because the generated hosts are not resolvable
+	// They hosts can only be resolved by getting the host lookup data.
+	if cfg.DisableInitialHostLookup && cfg.SecureConnectBundleFilename != "" {
+		return nil, ErrDisableInitialHostLookupNotAllowed
+	}
+
 	// Check that hosts in the ClusterConfig is not empty. If SecureConnectBudndle hosts are not expected
 	if len(cfg.Hosts) < 1 && cfg.SecureConnectBundleFilename == "" {
 		return nil, ErrNoHosts
--- a/session.go
+++ b/session.go
@@ -170,12 +170,7 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 		//    b) 'key' file
 		//    c) 'ca.crt' file
 
-		// create tmp directory with secureBundle filename as root to put files into
-		tmp, err := os.Getwd()
-		dir, err := ioutil.TempDir(tmp, cfg.SecureConnectBundleFilename)
-		if err != nil {
-			return nil, err
-		}
+		dir, err := ioutil.TempDir("", "securezip")
 		if err != nil {
 			return nil, err
 		}
--- a/session.go
+++ b/session.go
@@ -68,6 +68,8 @@ type Session struct {
 
 	cfg ClusterConfig
 
+	sniConfig *SNIConfig
+
 	ctx    context.Context
 	cancel context.CancelFunc
 
@@ -81,20 +83,28 @@ var queryPool = &sync.Pool{
 	},
 }
 
-func addrsToHosts(addrs []string, defaultPort int) ([]*HostInfo, error) {
+// addresToHosts - only used in session initialization to set up initial contact hosts.
+func addrsToHosts(addrs []string, defaultPort int, sniConfig *SNIConfig) ([]*HostInfo, error) {
 	var hosts []*HostInfo
-	for _, hostport := range addrs {
-		resolvedHosts, err := hostInfo(hostport, defaultPort)
-		if err != nil {
-			// Try other hosts if unable to resolve DNS name
-			if _, ok := err.(*net.DNSError); ok {
-				Logger.Printf("gocql: dns error: %v\n", err)
-				continue
+	if sniConfig == nil {
+		for _, hostport := range addrs {
+			resolvedHosts, err := hostInfo(hostport, defaultPort)
+			if err != nil {
+				// Try other hosts if unable to resolve DNS name
+				if _, ok := err.(*net.DNSError); ok {
+					Logger.Printf("gocql: dns error: %v\n", err)
+					continue
+				}
+				return nil, err
 			}
-			return nil, err
-		}
 
-		hosts = append(hosts, resolvedHosts...)
+			hosts = append(hosts, resolvedHosts...)
+		}
+	} else {
+		// SNI - therefor the addrs are really Cassandrea node host_ids and not resolvable ip addresses.
+		for _, contact := range addrs {
+			hosts = append(hosts, &HostInfo{hostId: contact})
+		}
 	}
 	if len(hosts) == 0 {
 		return nil, errors.New("failed to resolve any of the provided hostnames")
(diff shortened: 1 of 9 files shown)


giltohar:
@Zariel @martin-sucha hi my team has been using this gocql driver for a few years now.  Recently, we switched our database to a Datastax enterprise, which required us to make some updates some functionality to allow for connections to be made using secure connect bundles. We're using our forked version now and are interested in contributing it.

I was wondering if someone might be able to share some insight as to the issues in the build, if that is a blocker in us getting this code merged in.

giltohar:
@Zariel @martin-sucha found the issue and fixed :)

Zariel:
Im not happy with adding more and more specialisation to the driver at this point, it is becoming bloated and difficult to maintain. If this can't be implemented outside of the tree then I would like to know why and fix the driver so that it can be.

giltohar:
@Zariel I'm not sure of any way how to do that. Basically we need to use SNI and not rely on each host having its own specific IPADDRESS, instead they are identified by their Cassdandra node DB host id (not related to connect ipaddress or dns host names). This requires changes in the client due to how it initializes connections

giltohar:
Hi. @Zariel Any thoughts? We think this is a good improvement to the driver, as it makes it compatible to work with more Cassandra instances, for example the Cassandra instance of the Datastax cassandra, which is a major provider of Cassandra.

phact:
(Full disclosure, I work at DataStax)

Hi @Zariel thanks for all your work with this driver. Been using a bunch more c* and golang lately and it's been good to work with.

I wanted to point out that the SNI proxy functionality in this patch (thank you @giltohar!) is not DataStax Astra specific. As more and more folks use c* with k8s a desire to get at the database from a single endpoint is desirable and SNI is a nice way to achieve that and still maintain token awareness.

Here's how you would set up Cassandra with listio and SNI: https://www.solo.io/blog/step-by-step-datastax-cassandra-with-istio-and-sni-routing/

In general I think this patch is a nice addition given the changing cloud native landscape and a good way to help expand the reach of gocql to more users.

giltohar:
@Zariel any thoughts? :)

martin-sucha:
I think that this would be useful to have this functionality in this repository (as opposed to an external one). But we should add it in a way so that the implementation is isolated in one place, for example in a sub-package and could be theoretically implemented in an external package.

We currently have `Dialer` config option that establishes raw network connection to given TCP address, but the `Dialer` interface is not sufficient to implement TLS bundle. We need a more high level interface for "get a connection to given host" to be able to do that.

So what about an interface that would look similar to this (I don't have a better name than Dialer2 at the moment):

```go
type Dialer2 interface {
	// Dial establishes new connection to host.
	// It is responsibility of Dial to setup TLS if necessary.
	Dial(ctx context.Context, host *HostInfo) (net.Conn, error)
}
```

Interface like Dialer2 could be used to replace current `Dialer` and implement TLS bundle.
If it was possible to construct custom HostInfo from outside of the package, this could also replace `AddressTranslator` interface, since `AddressTranslator` could be implemented as a wrapper for some other `Dialer2`:

```go
type TranslatorDialer struct {
    dialer Dialer2
}

func (td *TranslatorDialer) Dial(ctx context.Context, host *HostInfo) (net.Conn, error) {
    newHost := host.Clone()
    newHost.HostName = "192.0.2.1"
    return td.dialer.Dial(ctx, newHost)
}
```

session.go
```diff
@@ -114,6 +161,112 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
...
+
+		dir, err := ioutil.TempDir("", "securezip")
+		if err != nil {
+			return nil, err
+		}
+		defer os.RemoveAll(dir) // the files are only needed until we create the tlsConfig, at that point they have been read in and processed, so not needed any longer and can be deleted at end of method.
+
+		// add each file from bundle to the directory
+		for _, f := range r.File {
+			fpath := filepath.Join(dir, f.Name)
```
martin-sucha, line 188: Extracting a file like this is a security issue - if the zip was malicious, it could replace any file it wanted. The documentation for `Name` says:

```
    // It is the caller's responsibility to sanitize it as
    // appropriate, including canonicalizing slash directions,
    // validating that paths are relative, and preventing path
    // traversal through filenames ("../../../").
```

session.go
```diff
@@ -114,6 +161,112 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
...
+			return nil, err
+		}
+		defer r.Close()
+
+		// 2. Put in temp directory (the following MUST BE FILES for this to work.
+		//    a) 'cert' file
+		//    b) 'key' file
+		//    c) 'ca.crt' file
+
+		dir, err := ioutil.TempDir("", "securezip")
```
martin-sucha, line 180: We definitely shouldn't write to disk in gocql. It's a network client so nobody expects it will write to disk. We cannot even assume that it's possible to write to disk, that could be forbidden for example due to host app's security policy.

It is not necessary to write to disk to setup `tls.Config either`, we could just:

1. Find entries in zip for `cert`, `key` and `ca.crt` and read the contents to memory
2. Use [tls.X509KeyPair](https://golang.org/pkg/crypto/tls/#X509KeyPair) to parse the PEM key pair and put the result to `tlsConfig.Certificates`
3. Create new cert pool `x509.NewCertPool()` and add the CA certificate to it using it's `AppendCertsFromPEM`, put this to `tlsConfig.RootCAs`

session.go
```diff
@@ -283,6 +457,31 @@ func (s *Session) init() error {
...
+	}
+	var metadata secureBundleMetadata
+
+	resp, err := client.Get(endpoint)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode > 299 {
+		// err := utils.NewError(resp.StatusCode, msg)
```
martin-sucha, line 476: We should probably handle the case here.

martin-sucha (changes requested):

Zariel:
My main concern is this touches so much of the codebase and that to me tells me the abstractions are wrong. On top of that passing in path to parse the data from is just plain wrong and should be accepted in a way the user does the setup work. Also I would rather more functionality is not added to the main gocql package where possible.

I think in general these sort of PR's could do with some design work and planning through issues.

mpenick:
We've build a proxy, called [cql-proxy](https://github.com/datastax/cql-proxy/) that allows gocql (or any other CQL driver) to work without this patch. It allows gocql applications to connect without any or very little modification.

martin-sucha:
Interesting, thank you for the link to the proxy! Today I learned that https://github.com/datastax/go-cassandra-native-protocol exists :)

I still think it would be useful to be be able to use the connection bundle with gocql directly. However, we need to use the correct abstractions to keep gocql maintainable. Something like the dialer based on `*Host` that would also handle TLS (mentioned above) and a custom function to retrieve the initial host list.

@mpenick Do you happen to have some simple server setup in docker-compose (or something similar) for testing on a local machine? Or would you be willing to contribute a patch? I'd be happy to review it.

martin-sucha:
https://github.com/gocql/gocql/pull/1629 implements interface similar to the one suggested in https://github.com/gocql/gocql/pull/1487#issuecomment-715931710, so it should now be possible to implement the secure connection bundle in an external package.

martin-sucha:
@mpenick Nice! Feel free to add a link to the gocqlastra library to gocql's README.

PR #1490, opened 2020-09-25
Init host pools in parallel.

muirdm:
Creating a session can take a long time if your cluster is big and
there's latency between you and the cluster. We are seeing it take
~10s to create a session to a 45 node cluster with ~30ms of latency.

This is somewhat related to #1369 and #1429, but this change tries to
do the simplest thing to speed up initial connection without changing
any behavior.

### DIFF
--- a/session.go
+++ b/session.go
@@ -227,17 +227,25 @@ func (s *Session) init() error {
 	}
 
 	hosts = hosts[:0]
+
+	var wg sync.WaitGroup
 	for _, host := range hostMap {
-		host = s.ring.addOrUpdate(host)
+		host := s.ring.addOrUpdate(host)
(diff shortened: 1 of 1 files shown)


andy-retailnext (approved):

Zariel:
thanks, nice little change that solves the problem! How much does this improve the init time for your example?

muirdm:
> How much does this improve the init time for your example?

It went down from ~10s to ~1.5s.

jameshartig:
While this speeds up init it doesn't fix the underlying issue I reported (a year ago today, actually) https://github.com/gocql/gocql/issues/1365 which is that init will fail if any of your servers happen to be down at that moment.

PR #1492, opened 2020-09-27
Connection Dial should be to connect address not hostname

suhailpatel:
Throughout gocql, this connection is mapped to the ConnectAddress which translates to an IP. It is expected that a connection created this far down the stack should be specific to the IP that was requested rather than any hostname (if one is set).

In some cases (mainly [when creating control connections](https://github.com/gocql/gocql/blob/393f0c961220893ab75e4627f1487608014502a7/control.go#L145)), the hostname is populated in the HostInfo struct with the user supplied hosts. If these map to a DNS name with multiple entries, the connection will be established to any of those IPs rather than specifically to the IP we want to connect to.

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -415,13 +415,6 @@ func (h *HostInfo) IsUp() bool {
 	return h != nil && h.State() == NodeUp
 }
 
-func (h *HostInfo) HostnameAndPort() string {
-	if h.hostname == "" {
-		h.hostname = h.ConnectAddress().String()
-	}
-	return net.JoinHostPort(h.hostname, strconv.Itoa(h.port))
-}
-
 func (h *HostInfo) String() string {
 	h.mu.RLock()
 	defer h.mu.RUnlock()
--- a/conn.go
+++ b/conn.go
@@ -213,7 +213,8 @@ func (s *Session) dialWithoutObserver(ctx context.Context, host *HostInfo, cfg *
 		dialer = d
 	}
 
-	conn, err := dialer.DialContext(ctx, "tcp", host.HostnameAndPort())
+	addr := net.JoinHostPort(ip.String(), strconv.Itoa(port))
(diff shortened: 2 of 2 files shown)


host_source.go
```diff
@@ -415,13 +415,6 @@ func (h *HostInfo) IsUp() bool {
 	return h != nil && h.State() == NodeUp
 }
 
-func (h *HostInfo) HostnameAndPort() string {
```
suhailpatel, line 418: This function became unused so I removed it for cleanliness

suhailpatel:
I did some archeology and found when this was tweaked in #1310. I haven't looked too deeply but I think that patch too has some cases around when more peers are discovered where hostname verification would fail too.

More crucially, since some refactorings of how TLS Sessions are established, using `EnableHostVerification` in `gocql.SslOptions' does not work at all because the `tls.Config.ServerName` is not being set (which is a pre-requisite of having `tls.Config.InsecureSkipVerify` be false and having host verification). This regression was introduced in #1368 where we changed from `tls.DialWithDialer` (which [under the hood](https://github.com/golang/go/blob/4b09c8ad6fb9d30b9c3417b5364809ff0006749d/src/crypto/tls/tls.go#L157-L164) sets the `tls.Config.ServerName` if one is not set) to wrapping the existing connection using `tls.Client`.

I think the tls bug is a bigger patch so should be addressed separately. I've opened https://github.com/gocql/gocql/pull/1493 as a result

host_source.go
martin-sucha: Unfortunately, removing the function breaks backwards compatibility - both `HostInfo` and `HostnameAndPort` are exported, so `HostnameAndPort` could be used by some gocql user. So we need to keep it.

martin-sucha (changes requested):

PR #1493, opened 2020-09-27
TLS Host Verification failing with missing ServerName

suhailpatel:
Using `EnableHostVerification` in `gocql.SslOptions` does not work at all and fails with the following error
```
2020/09/27 22:06:05 gocql: unable to create session: unable to discover protocol version: tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config
```

This is because the `tls.Config.ServerName` is not being set (which is a pre-requisite of having `tls.Config.InsecureSkipVerify` be false and having host verification). 

This regression was introduced in #1368 where we changed from `tls.DialWithDialer` (which [under the hood](https://github.com/golang/go/blob/4b09c8ad6fb9d30b9c3417b5364809ff0006749d/src/crypto/tls/tls.go#L157-L164) sets the `tls.Config.ServerName` if one is not set) to wrapping the existing connection using `tls.Client`.

I've erred on the side of fixing for backwards compatibility by introducing the `HostServerName` field but falling back to `host.hostname` if one isn't present.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -89,10 +89,14 @@ type SslOptions struct {
 	CertPath string
 	KeyPath  string
 	CaPath   string //optional depending on server config
-	// If you want to verify the hostname and server cert (like a wildcard for cass cluster) then you should turn this on
-	// This option is basically the inverse of InSecureSkipVerify
+
+	// If you want to verify the hostname and server cert (like a wildcard
+	// for a Cassandra cluster) then you should turn this on. This option
+	// is basically the inverse of InSecureSkipVerify. If you enable this
+	// option, make sure you set the HostServerName property too
 	// See InSecureSkipVerify in http://golang.org/pkg/crypto/tls/ for more info
 	EnableHostVerification bool
+	HostServerName         string
 }
 
 type ConnConfig struct {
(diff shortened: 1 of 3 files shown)


conn.go
```diff
@@ -89,10 +89,14 @@ type SslOptions struct {
...
-	// If you want to verify the hostname and server cert (like a wildcard for cass cluster) then you should turn this on
-	// This option is basically the inverse of InSecureSkipVerify
+
+	// If you want to verify the hostname and server cert (like a wildcard
+	// for a Cassandra cluster) then you should turn this on. This option
+	// is basically the inverse of InSecureSkipVerify. If you enable this
+	// option, make sure you set the HostServerName property too
 	// See InSecureSkipVerify in http://golang.org/pkg/crypto/tls/ for more info
 	EnableHostVerification bool
+	HostServerName         string
```
martin-sucha, line 99: When would a user want to set the HostServerName? Setting it here would imply that all the Cassandra nodes in the cluster have the same SNI servername.

There is also https://github.com/gocql/gocql/pull/1487 - a simple HostServerName looks insufficient for that use case, so I'd rather not introduce a new configuration option for this fix. We should export some interface in configuration that would work for that use case as well. However, that would be for a separate pull request.

martin-sucha (commented):
Could you please also add a test so that we won't repeat the regression again in the future?

martin-sucha:
I have merged a slightly different fix, thanks for reporting.

PR #1495, opened 2020-09-28
[Code Optimization], huge parameters need to passed as pointer 

tushardas:
Structs which are huge/heavy should be passed as pointers, in-order to avoid copying a large value to local parameters in the function.

ObservedQuery struct in the method of QueryObserver interface, is a heavy struct.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -325,12 +325,14 @@ type testQueryObserver struct {
 	verbose bool
 }
 
-func (o *testQueryObserver) ObserveQuery(ctx context.Context, q ObservedQuery) {
-	host := q.Host.ConnectAddress().String()
-	o.metrics[host] = q.Metrics
-	if o.verbose {
-		Logger.Printf("Observed query %q. Returned %v rows, took %v on host %q with %v attempts and total latency %v. Error: %q\n",
-			q.Statement, q.Rows, q.End.Sub(q.Start), host, q.Metrics.Attempts, q.Metrics.TotalLatency, q.Err)
+func (o *testQueryObserver) ObserveQuery(ctx context.Context, q *ObservedQuery) {
+	if q !=  nil {
+		host := q.Host.ConnectAddress().String()
+		o.metrics[host] = q.Metrics
+		if o.verbose {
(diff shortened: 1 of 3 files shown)


cassandra_test.go
```diff
@@ -159,10 +159,12 @@ func TestObserve(t *testing.T) {
 		observedStmt = ""
 	}
 
-	observer := funcQueryObserver(func(ctx context.Context, o ObservedQuery) {
-		observedKeyspace = o.Keyspace
-		observedStmt = o.Statement
-		observedErr = o.Err
+	observer := funcQueryObserver(func(ctx context.Context, o *ObservedQuery) {
+		if o != nil {
```
martin-sucha, line 163: We don't need the nil check here. QueryObserver won't ever be called with nil pointer.

martin-sucha (commented):
Could we also do the same change with BatchObserver so that the API is consistent between those?

Changing to pointer breaks backwards compatibility. The QueryObserver interface is documented as experimental, but it is definitely used by users. There is `v2.dev` branch in gocql where some backwards-compatibility breaking changes are planned.

It might be possible to pass the pointer without breaking compatibility for example by introducing QueryObserverV2 interface alongside the existing one, but the question is whether the cost of the interface check is less than the cost of the copying.

How much does this change improve performance for your use case? Is the cost of heap allocation lower than the cost of copying? Could you show a benchmark?

cassandra_test.go
tushardas: Fixed and similar changes made for BatchObserver as well.

Benchmarks:
```
QueryObserver

Benchmark_GOCQL-16              613           2045540 ns/op           11650 B/op         88 allocs/op
Benchmark_ForkedGOCQL-16        858           1432294 ns/op            6051 B/op         87 allocs/op
```

```
BatchObserver

Benchmark_BatchGOCQL-16             862           1320205 ns/op            7902 B/op         31 allocs/op
Benchmark_ForkedBatchGOCQL-16       956           1297137 ns/op            2551 B/op         30 allocs/op
```

Zariel:
This is an API change and potentially unnecessary, can you benchmark with benchcmp with at least 20 runs on each

Issue #1497, opened 2020-10-01
Query isn't well formed in the log using ObserveQuery interface


tushardas:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?
3.11.8

### What did you do?
I'm implementing the QueryObserver interface as below, for logging the query
```go
type queryLogger struct {
	Logger log.Logger
}

func (q queryLogger) ObserveQuery(ctx context.Context, o gocql.ObservedQuery) {
	q.Logger.Debugf("[CQL] query: %v %vms", o.Statement, o.End.Sub(o.Start).Milliseconds())
}
```

And below is the query which I'm using
```go
query := "INSERT INTO students (id, name, marks) VALUES (?, ?, ?)"
cassandra.Session.Query(query, 1, "Tushar",95).Exec()
```
### What did you expect to see?
[CQL] INSERT INTO students (id, name, marks) VALUES (1, "Tushar", 95)

### What did you see instead?
[CQL] INSERT INTO students (id, name, marks) VALUES (?,?,?)

---

tushardas:
Found that the issue is at https://github.com/gocql/gocql/blob/master/session.go#L1004

The receiver `q` contains a field `values` which stores the args passed in Query().

martin-sucha:
gocql does not substitute bound values in the query, those are [passed separately in cql protocol](https://github.com/datastax/native-protocol/blob/82b12f3892aeb360b3acefe2f915449698764251/src/main/resources/native_protocol_v4.spec#L302), so rendering 
```
[CQL] INSERT INTO students (id, name, marks) VALUES (1, "Tushar", 95)
```
is out of scope.

I guess we could add something like `Values []interface{}` field to ObservedQuery and set it to `Query.values`, so that you can log the values separately.

tushardas:
@martin-sucha  
That would help. 

I would like to work on these changes and submit a PR.

For the `ObserveBatch`, would it be good to have a `map[string][]interface{}`, where the key would be the statement and value would be the corresponding bound values?

martin-sucha:
> I would like to work on these changes and submit a PR.

Nice :+1: 

> For the `ObserveBatch`, would it be good to have a `map[string][]interface{}`, where the key would be the statement and value would be the corresponding bound values? In order for backward compatible changes, we can create a new field `Entries`, maybe?

We can't use the statement as key as there could be the same statement executed multiple sets of bound values. For example adding multiple records to the same partition. Consider the following batch:

```
[...]

PR #1498, opened 2020-10-05
Bound values added in fields of ObservedQuery and ObservedBatch struct

tushardas:
The statements passed in the statement fields of ObservedQuery and ObservedBatch, do not account for the bound values passed while executing a query. 

Refer issue: #1497 for more details

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -1867,7 +1867,7 @@ func TestBatchObserve(t *testing.T) {
 		observedErr      error
 		observedKeyspace string
 		observedStmts    []string
-		observedEntries  map[string][]interface{}
+		observedValues   [][]interface{}
 	}
 
 	var observedBatch *observation
@@ -1882,7 +1882,7 @@ func TestBatchObserve(t *testing.T) {
 			observedKeyspace: o.Keyspace,
 			observedStmts:    o.Statements,
 			observedErr:      o.Err,
-			observedEntries:  o.Entries,
+			observedValues:   o.Values,
 		}
 	}))
 	for i := 0; i < 100; i++ {
@@ -1909,8 +1909,8 @@ func TestBatchObserve(t *testing.T) {
 		if stmt != fmt.Sprintf(`INSERT INTO batch_observe_table (id,other) VALUES (?,%d)`, i) {
 			t.Fatal("unexpected query", stmt)
 		}
-		if observedBatch.observedEntries[stmt] != i {
-			t.Fatal("expecting bound value '%v', got %q", i, observedBatch.observedEntries[stmt])
+		if observedBatch.observedValues[i] != []interface{}{i} {
+			t.Fatal("expecting bound value '%v', got %q", []interface{}{i}, observedBatch.observedEntries[i])
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


session.go
```diff
@@ -1995,6 +2001,7 @@ type QueryObserver interface {
 type ObservedBatch struct {
 	Keyspace   string
 	Statements []string
+	Values     [][]interface{}
```
martin-sucha, line 2004: Could you please add documentation for the field?

For example:

```go
// Values holds a slice of bound values for each statement.
// Values[i] are bound values passed to Statements[i].
// Do not modify the values here, they are shared with multiple goroutines.
Values     [][]interface{}
```

session.go
```diff
@@ -1957,6 +1962,7 @@ func (t *traceWriter) Trace(traceId []byte) {
 type ObservedQuery struct {
 	Keyspace  string
 	Statement string
+	Values    []interface{}
```
martin-sucha, line 1965: Could you please add documentation for the field?

For example:

```go
// Values holds a slice of bound values for the query.
// Do not modify the values here, they are shared with multiple goroutines.
Values     [][]interface{}
```

martin-sucha (changes requested):

session.go
tushardas: @martin-sucha Documentation is added. Thanks

session.go
tushardas: Documentation is added. Thanks

martin-sucha (approved):

tushardas:
@Zariel @jeffwidman Can you please look into this PR?

jeffwidman:
@tushardas I'm not a maintainer so @martin-sucha would be the right person to chat with.

Also FYI there's a merge conflict you might want to first resolve.

cassandra_test.go
```diff
@@ -1907,6 +1909,9 @@ func TestBatchObserve(t *testing.T) {
 		if stmt != fmt.Sprintf(`INSERT INTO batch_observe_table (id,other) VALUES (?,%d)`, i) {
 			t.Fatal("unexpected query", stmt)
 		}
+		if observedBatch.observedValues[i] != []interface{}{i} {
```
martin-sucha, line 1912:
```
./cassandra_test.go:1912:38: invalid operation: observedBatch.observedValues[i] != []interface {} literal (slice can only be compared to nil)
./cassandra_test.go:1913:81: observedBatch.observedEntries undefined (type *observation has no field or method observedEntries)
```

`observedBatch.observedEntries` should be `observedBatch.observedValues` instead. But both lines could be replaced with a call to `assertDeepEqual(t, "observed value", []interface{}{i}, observedBatch.observedValues[i]`

https://github.com/gocql/gocql/blob/87a5d7a5ff74ffa9eb716f05f6e26191836bf03d/common_test.go#L263-L268

martin-sucha (commented):
Thanks for resolving the merge conflict! There is a build failure in tests that we need to resolve before merging (see inline comment).

cassandra_test.go
tushardas: @martin-sucha Fixed the tests. Thanks for suggesting the fix

martin-sucha:
Merged. Thanks!

PR #1503, opened 2020-10-24
Document type conversions for Marshal and Unmarshal

martin-sucha:
It was not documented anywhere how cql data types map to Go
data types.

I haven't included data for exact cases when type aliases are
supported.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -44,6 +44,51 @@ type Unmarshaler interface {
 
 // Marshal returns the CQL encoding of the value for the Cassandra
 // internal type described by the info parameter.
+//
+// nil is serialized as CQL null.
+// If value implements Marshaler, its MarshalCQL method is called to marshal the data.
+// If value is a pointer, the pointed-to value is marshaled.
+//
+// Supported conversions are as follows, other type combinations may be added in the future:
+//
+//  CQL type                    | Go type (value)    | Note
+//  varchar, ascii, blob, text  | string, []byte     |
+//  boolean                     | bool               |
+//  tinyint, smallint, int      | integer types      |
+//  tinyint, smallint, int      | string             | formatted as base 10 number
+//  bigint, counter             | integer types      |
+//  bigint, counter             | big.Int            |
+//  bigint, counter             | string             | formatted as base 10 number
+//  float                       | float32            |
+//  double                      | float64            |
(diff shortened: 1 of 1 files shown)


marshal.go
```diff
@@ -44,6 +44,51 @@ type Unmarshaler interface {
...
+//  tuple                       | struct             | fields are marshaled in order of declaration
+//  user-defined type           | UDTMarshaler       | MarshalUDT is called
+//  user-defined type           | map[string]interface{} |
+//  user-defined type           | struct             | cql tags are used for field names
+//  date                        | int64              | milliseconds since Unix epoch to start of day (in UTC)
+//  date                        | time.Time          | start of day (in UTC)
+//  date                        | string             | parsed using "2006-01-02" format
+//  duration                    | int64              | duration in nanoseconds
+//  duration                    | time.Duration      |
+//  duration                    | Duration           |
```
matoous, line 90: **thought**: What's the difference between `Duration` and `time.Duration` on the previous line?

marshal.go
```diff
@@ -44,6 +44,51 @@ type Unmarshaler interface {
...
+//  varint                      | integer types      |
+//  varint                      | big.Int            |
+//  varint                      | string             | value of number in decimal notation
+//  inet                        | net.IP             |
+//  inet                        | string             | IPv4 or IPv6 address string
+//  tuple                       | slice, array       |
+//  tuple                       | struct             | fields are marshaled in order of declaration
+//  user-defined type           | UDTMarshaler       | MarshalUDT is called
+//  user-defined type           | map[string]interface{} |
+//  user-defined type           | struct             | cql tags are used for field names
```
matoous, line 84: **though**: this might be more accurate:
```suggestion
//  user-defined type           | struct             | struct fields cql tags are used for column names
```

matoous (changes requested):

marshal.go
martin-sucha: gocql has its own [Duration](https://pkg.go.dev/github.com/gocql/gocql#Duration) type that corresponds to duration type stored in database, see https://cassandra.apache.org/doc/latest/cql/types.html#durations

marshal.go
martin-sucha: Updated, thanks!

matoous (approved):

PR #1504, opened 2020-10-24
Remove dead code

martin-sucha:
This code is not used anywhere.

### DIFF
--- a/common_test.go
+++ b/common_test.go
@@ -153,26 +153,6 @@ func createSession(tb testing.TB, opts ...func(config *ClusterConfig)) *Session 
 	return createSessionFromCluster(cluster, tb)
 }
 
-// createTestSession is hopefully moderately useful in actual unit tests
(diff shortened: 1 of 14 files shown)


hummerd:
Talking about dead code.
@martin-sucha Can you please look at code
https://github.com/gocql/gocql/blob/5913df4d474e0b2492a129d17bbb3c04537a15cd/conn.go#L894-L910

Line 897 is the only place where timer initialized, and we do not have call pool any more, so else path will never be executed. Also code in if statement can be simplified with merging it with lines 908-909.

What do you think?

martin-sucha:
Thanks! Indeed, lines 900-905 are never executed. Also, access to `call.timer` is not synchronized, but should be. We modify it after adding the `call` to `c.calls`, so the receive goroutine can already access it from line 870 onwards, but `call.timer` is modified later.

PR #1505, opened 2020-10-24
Do not allocate extra bytes in unmarshalUUID

hummerd:
If unmarshalUUID called with value of type gocql.UUID, we already have allocated bytes for UUID, so do not allocate extra UUID, but just copy data to value. 

Also add ability to unmarshal directly to array of 16 bytes.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -1711,7 +1711,7 @@ func marshalUUID(info TypeInfo, value interface{}) ([]byte, error) {
 }
 
 func unmarshalUUID(info TypeInfo, data []byte, value interface{}) error {
-	if data == nil || len(data) == 0 {
+	if len(data) == 0 {
 		switch v := value.(type) {
 		case *string:
 			*v = ""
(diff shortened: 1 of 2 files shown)


marshal.go
```diff
@@ -1726,6 +1726,19 @@ func unmarshalUUID(info TypeInfo, data []byte, value interface{}) error {
 		return nil
 	}
 
+	if len(data) != 16 {
+		return unmarshalErrorf("Unable to parse UUID: UUIDs must be exactly 16 bytes long")
```
martin-sucha, line 1730: **nitpick**: Error messages in Go should start with lowercase letters, please change `Unable` to `unable`.

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

(I know there is error starting with uppercase letter below, I have opened a separate pull request to make the other errors lowercase: #1506)

martin-sucha (changes requested):
Thanks for the pull request! Just one nit pick, otherwise looks good to me.

marshal.go
hummerd: Done!

martin-sucha (approved):

marshal.go
martin-sucha: Thanks!

PR #1508, opened 2020-11-03
Copy query when fetching next page

martin-sucha:
Currently the attempt count in query metrics is incremented both on
retries and when fetching next page, making it impossible to track just
the retry count. By shallow-copying the query and changing the metrics
we ensure that attempts are counted per-age (so attempts>0 means always
a retry).

We were copying the Query previously
(before 271c061c7f16702ca550b0e9721320b1c92e00cb), but I don't know why
exactly the copying was removed there.

In any case we can't change the paging state in the original Query
because there might be a separate goroutine spawned by speculative
execution that accesses the field. Even if we added a lock around
Query.pagingState field, the speculative execution could try to fetch
different page.

We could possibly move pagingState and metrics to nextIter or other
struct to avoid copying whole Query, but that would require changing
signature of executeQuery to take that struct as parameter instead.

Batches don't implement fetching subsequent pages, so removing the note
about pages from ObservedBatch.Attempts as well.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1205,12 +1205,16 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
 		}
 
 		if x.meta.morePages() && !qry.disableAutoPage {
+			newQry := new(Query)
+			*newQry = *qry
+			newQry.pageState = copyBytes(x.meta.pagingState)
+			newQry.metrics = &queryMetrics{m: make(map[string]*hostMetrics)}
+
 			iter.next = &nextIter{
-				qry: qry,
+				qry: newQry,
 				pos: int((1 - qry.prefetch) * float64(x.numRows)),
 			}
 
-			iter.next.qry.pageState = copyBytes(x.meta.pagingState)
 			if iter.next.pos < 1 {
 				iter.next.pos = 1
 			}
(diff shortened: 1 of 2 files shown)


martin-sucha:
Another reason why a `Query` needs to be immutable is that someone might call `Exec` or `Iter` multiple times for the same `Query`.

matoous (approved):
LGTM

conn.go
```diff
@@ -1173,12 +1173,16 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
...
 
 		if x.meta.morePages() && !qry.disableAutoPage {
+			newQry := new(Query)
+			*newQry = *qry
+			newQry.pageState = copyBytes(x.meta.pagingState)
+			newQry.metrics = &queryMetrics{m: make(map[string]*hostMetrics)}
+
 			iter.next = &nextIter{
-				qry: qry,
+				qry: newQry,
```
funny-falcon-at-joomcode, line 1182:
```
newQry := *qry
...
qry: &newQry,
```
I think it will be more ideomatic.

funny-falcon-at-joomcode:
Why it takes so long to merge?

martin-sucha:
> Why it takes so long to merge?

I just forgot to merge it here since we already have the change merged in our fork. Rebased now, will merge once the CI pipeline finishes.

PR #1510, opened 2020-11-18
Add Go version to issue template

martin-sucha:
Some issues might be reproducible only using specific
Go version. If someone is not using latest Go version,
we might want to ask them to upgrade and try to reproduce.

This change is inspired by the following comment:
https://github.com/gocql/gocql/issues/1481#issuecomment-728758618


cdent:
👍

PR #1514, opened 2020-12-08
Add port number into error message

alexus1024:
Just adding port number into "unable to dial control conn" error message to simplify debugging

### DIFF
--- a/control.go
+++ b/control.go
@@ -177,7 +177,7 @@ func (c *controlConn) shuffleDial(endpoints []*HostInfo) (*Conn, error) {
 			return conn, nil
 		}
 
-		Logger.Printf("gocql: unable to dial control conn %v: %v\n", host.ConnectAddress(), err)
+		Logger.Printf("gocql: unable to dial control conn %v:%v: %v\n", host.ConnectAddress(), host.Port(), err)
 	}
 
 	return nil, err
--- a/AUTHORS
+++ b/AUTHORS
@@ -116,3 +116,4 @@ Rintaro Okamura <rintaro.okamura@gmail.com>
 Yura Sokolov <y.sokolov@joom.com>; <funny.falcon@gmail.com>
 Jorge Bay <jorgebg@apache.org>
 Dmitriy Kozlov <hummerd@mail.ru>
+Alexey Romanovsky <alexus1024+gocql@gmail.com>


control.go
```diff
@@ -177,7 +177,7 @@ func (c *controlConn) shuffleDial(endpoints []*HostInfo) (*Conn, error) {
 			return conn, nil
 		}
 
-		Logger.Printf("gocql: unable to dial control conn %v: %v\n", host.ConnectAddress(), err)
+		Logger.Printf("gocql: unable to dial control conn %v:%v: %v\n", host.ConnectAddress(), host.Port(), err)
```
martin-sucha, line 180: Perhaps we should include the full address (host+port) in the error returned from `session.dialWithoutObserver` and not add the address and port here? That way the code that creates the error message and the code that actually does the dialing is at the same place, so it's harder for it become out of sync (plus we get the address/port added to the error for regular dials as well, not just control conns). Could you please move adding the host/port to the error message to `session.dialWithoutObserver`?

martin-sucha (changes requested):
Hi! Thanks for the pull request! Could you please also add yourself to AUTHORS file as described in [CONTRIBUTING.md](https://github.com/gocql/gocql/blob/93eedddb64669e218f31cff48059066b03fabf83/CONTRIBUTING.md)?

control.go
alexus1024: I will need to wrap error at conn.go:218. And at maybe line 226 too, for uniformity. Something like 
`return nil, fmt.Errorf("dial. %w", err)`
This can broke deprecated error checks like "err.Error() == "some cassa err".
And if I wrap it with %w - it allow right-wrote checks to work but will not compile with older golang.

Should I still wrap? And if yes, should I use %w?

alexus1024: Ok. There is no support for golang prior to 1.13, so question about %w is not actual

martin-sucha: Ah, you are right, there are places that check the exact error now, like in 
https://github.com/gocql/gocql/blob/93eedddb64669e218f31cff48059066b03fabf83/connectionpool.go#L510-L520

so the changes would be larger than I thought. Let's use your original change then and create a separate issue for version 2 of the driver, I'll keep it in mind during refactoring.

martin-sucha: I've created the issue for the error handling here: https://github.com/gocql/gocql/issues/1515

martin-sucha:
Merged. Thanks!

PR #1529, opened 2021-01-27
Tuple unmarshaling improvements

piodul:
This PR improves some aspects of tuple unmarshaling:

- In case of nested tuples which use pointers to represent their elements, now element is set to nil if the element was NULL in the query result - previously it was set to zero value.
- When unmarshaling the last element of a tuple, and the element was set to an empty value (e.g. "" for text), it was incorrectly treated as if it was NULL - this is now fixed.
- Very minor improvement to one of the tests related to tuples.


martin-sucha (changes requested):
Thanks for the pull request! The code looks good to me, there is just one nitpick. Could you also please [add yourself to AUTHORS file](https://github.com/gocql/gocql/blob/master/CONTRIBUTING.md#minimum-requirement-checklist)?

piodul:
@martin-sucha I added myself to the AUTHORS.

I also saw that the second commit (setting pointers to nil in nested tuples) broke the TestMarshalTuple test. The test wrote a tuple with one field set to nil, then deserialized it and expected that field to have an empty value instead. For representing that field, it used `*string` type.

Now I'm not sure which behavior is correct. Personally I think that the new behavior is more intuitive, but I would like to avoid unnecessarily breaking users' code. I updated the test in the PR, but I can drop that commit if it's better that way.

What do you think?

martin-sucha:
>  The test wrote a tuple with one field set to nil, then deserialized it and expected that field to have an empty value instead.

I think this was a bug. We return `nil` for `null` values everywhere else if the receiver is a pointer, specifically so it's possible to distinguish `null` from empty string if necessary.

cc @vrischmann

vrischmann:
I can't remember why I wrote that test that way at the time, but looking at it now I agree that the test seems wrong.

martin-sucha (approved):

piodul:
One more fix: `TestUnmarshalTuple` had to be changed in the same way as `TestMarshalTuple`.

martin-sucha:
Merged. Thanks!

PR #1540, opened 2021-03-02
Don't do network IO with already canceled context

martin-sucha:
It does not make sense to send requests where the results will
be immediately discarded. In fact, it can be harmful in situations
where a lot of contexts are canceled (e.g. because of timeouts).


matoous (approved):
lgtm.

PR #1542, opened 2021-03-05
Make cluster defaults publicly accessible

jeffwidman:
Scenario:
We wrap gocql in some of our own libraries. We allow callers to override some of the ClusterConfig default values. 
However, we also want to have our help messages etc print what the default values are. 
We can copy/paste, but would prefer to just inherit the `gocql` current defaults, whatever they happen to be.

So exposed them as constants.

Note: This PR is a draft to stimulate discussion, I threw it up using the Github web UI so not even sure if it compiles... I'd be more than happy tweak as desired and ensure it compiles / runs if the maintainers are open to exposing these... I just have found in the past that opening a PR results in a better discussion than merely filing an issue.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -11,6 +11,24 @@ import (
 	"time"
 )
 
+const (
+	DefaultCQLVersion = "3.0.0"
+	DefaultClusterTimeout = 600 * time.Millisecond
+	DefaultClusterConnectTimeout = 600 * time.Millisecond
+	DefaultPort = 9042
+	DefaultNumConns = 2
+	// TODO DefaultConsistentcy - should we use a global var and document it should not be mutated?
+	// TODO defaultMaxPreparedStmts should be made public
+	DefaultMaxRoutingKeyInfo = 1000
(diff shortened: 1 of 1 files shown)


jeffwidman:
nudge @martin-sucha @Zariel Any thoughts on this?

martin-sucha:
We are not likely to change the defaults because of backwards-compatibility.

If we ever wanted to change some default, having defaults exposed as constants would mean that we wouldn't be able to change the default conditionally (for example based on CQL version or server version) because in addition to zero value, the user could have used a constant to set the value e.g. `NumConns: DefaultNumConns`. I think it's better to have just one way to specify that the user wants default behavior in the API - using the zero value.

You need to handle the fields that you want to override in your code anyway, so I don't think it's an issue if you copy the defaults to your code as well. You would just need to update the defaults once/if the gocql defaults change (which is unlikely) and we don't know now if and how they may change in the future, so you might need to handle the change in a completely unexpected way.

So I would say we shouldn't expose the defaults as constants in gocql

PR #1555, opened 2021-05-03
Fix reconnecting to cluster

martin-sucha:
After merging https://github.com/gocql/gocql/pull/1369 it could happen
that after losing connection to a host, it is not reconnected.

That could happen in the following case:

1. We lose connection to a host.
2. We receive UP event for the host, the host is marked as UP even
   though we don't have active connections to it.
3. We don't reconnect to the host in reconnectDownedHosts because it's
   marked as UP already.

In PR 1369 we intended to change the host states so that a host is
marked as UP only after we have some connection to it. However, we also
removed the call to fill the pool on UP event in case the host already
existed.

This commit adds the call to fill the pool back to handleNodeUp.
We introduce handleNodeConnected for marking the host as up,
so that handleNodeUp handles only the UP event and the responsibilities
are clear.

Control connection is tried to be reconnected every second after
it is lost.
Before PR 1369, after establishing a control connection to a host,
we triggered pool refill for the same host.
PR 1369 removed this behaviour so that it is possible to wait
for the connection to be added to the pool in Session.init (and not
mark the host UP if we have only control connection.
This means that it could take up to cfg.ReconnectInterval after
estabilishing control connection to have some usable host in the pool
to service user queries after PR 1369.

To restore the old behavior, we add back triggering of pool refill after
control connection is connected. We can only do so after the session is
initialized, so that we don't the interfere with initialization code
in Session.init.

Related https://github.com/gocql/gocql/issues/915

### DIFF
--- a/events.go
+++ b/events.go
@@ -199,9 +199,7 @@ func (s *Session) addNewNode(ip net.IP, port int) {
 	hostInfo = s.ring.addOrUpdate(hostInfo)
 
 	if !s.cfg.filterHost(hostInfo) {
-		// we let the pool call handleNodeUp to change the host state
-		s.pool.addHost(hostInfo)
-		s.policy.AddHost(hostInfo)
+		s.startPoolFill(hostInfo)
 	}
 
 	if s.control != nil && !s.cfg.IgnorePeerAddr {
@@ -262,6 +260,27 @@ func (s *Session) handleNodeUp(eventIp net.IP, eventPort int) {
 	if host == nil {
 		s.addNewNode(ip, port)
 		return
+	}
+
+	if s.cfg.filterHost(host) {
+		return
+	}
+
+	if d := host.Version().nodeUpDelay(); d > 0 {
+		time.Sleep(d)
+	}
+	s.startPoolFill(host)
+}
+
+func (s *Session) startPoolFill(host *HostInfo) {
+	// we let the pool call handleNodeConnected to change the host state
+	s.pool.addHost(host)
+	s.policy.AddHost(host)
+}
+
+func (s *Session) handleNodeConnected(host *HostInfo) {
+	if gocqlDebug {
+		s.logger.Printf("gocql: Session.handleNodeConnected: %s:%d\n", host.ConnectAddress(), host.Port())
 	}
 
 	host.setState(NodeUp)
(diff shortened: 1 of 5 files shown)


events.go
```diff
@@ -264,6 +262,25 @@ func (s *Session) handleNodeUp(eventIp net.IP, eventPort int) {
...
+
+func (s *Session) startPoolFill(host *HostInfo) {
+	// we let the pool call handleNodeConnected to change the host state
+	s.pool.addHost(host)
+	s.policy.AddHost(host)
+}
+
+func (s *Session) handleNodeConnected(host *HostInfo) {
+	if gocqlDebug {
+		Logger.Printf("gocql: Session.handleNodeConnected: %s:%d\n", host.ConnectAddress(), host.Port())
```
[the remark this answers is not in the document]
martin-sucha: Indeed. I missed this during rebase. Fixed. Thanks!

session.go
```diff
@@ -70,8 +70,12 @@ type Session struct {
 	ctx    context.Context
 	cancel context.CancelFunc
 
-	closeMu  sync.RWMutex
+	// sessionStateMu protects isClosed and isInitialized.
+	sessionStateMu sync.RWMutex
+	// isClosed is true once Session.Close is called.
 	isClosed bool
+	// isInitialized is true once Session.init succeeds.
```
mrVanboy, line 77: I believe it could be helpful for other maintainers to mention `initialized()` method here in docs.

events.go
```diff
@@ -264,6 +262,25 @@ func (s *Session) handleNodeUp(eventIp net.IP, eventPort int) {
 		return
 	}
 
+	if !s.cfg.filterHost(host) {
+		if t := host.Version().nodeUpDelay(); t > 0 {
+			time.Sleep(t)
+		}
+		s.startPoolFill(host)
+	}
```
mrVanboy, line 270:
```suggestion
	if s.cfg.filterHost(host) {
		return
	}
	if d := host.Version().nodeUpDelay(); d > 0 {
		time.Sleep(d)
	}
	s.startPoolFill(host)
```

mrVanboy (commented):
a couple of stylistic things, otherwise LGTM

mrVanboy (approved):

PR #1558, opened 2021-05-12
Fix flaky test TestStartupTimeout

panlong222:
In conn_test.go, "TestStartupTimeout" at the very last step requires the log to contain "no response to connection startup within timeout".

However, during my test, it fails sometimes, not always, with log containing "no response received from cassandra within timeout period" instead.

Fixing the flaky test by allowing log to contain either message.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -238,7 +238,8 @@ func TestStartupTimeout(t *testing.T) {
 		t.Fatalf("Expected to receive no connections error - got '%s'", err)
 	}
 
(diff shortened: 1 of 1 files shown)


martin-sucha:
Hello. Thank you for the pull request!

The change looks good to me.

Could you please [add yourself to AUTHORS file](https://github.com/gocql/gocql/blob/master/CONTRIBUTING.md#minimum-requirement-checklist)?

PR #1561, opened 2021-05-27
Add support for LZ4 compression

horkhe:
The subject says it all.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,14 +21,25 @@ env:
   matrix:
     - CASS=2.1.21
       AUTH=true
+      COMPRESSOR=snappy
     - CASS=2.2.14
       AUTH=true
+      COMPRESSOR=snappy
     - CASS=2.2.14
       AUTH=false
+      COMPRESSOR=snappy
+    - CASS=2.2.14
+      AUTH=false
+      COMPRESSOR=lz4
     - CASS=3.0.18
       AUTH=false
+      COMPRESSOR=snappy
     - CASS=3.11.4
       AUTH=false
+      COMPRESSOR=snappy
+    - CASS=3.11.4
+      AUTH=false
+      COMPRESSOR=lz4
 
 go:
   - 1.15.x
@@ -44,7 +55,7 @@ install:
 
 script:
   - set -e
-  - PATH=$PATH:$HOME/.local/bin bash integration.sh $CASS $AUTH
+  - PATH=$PATH:$HOME/.local/bin bash integration.sh $CASS $AUTH $COMPRESSOR
   - go vet .
 
 notifications:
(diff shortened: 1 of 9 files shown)


martin-sucha:
Thanks for the pull request! Looks good to me. Please [add yourself to AUTHORS file]9https://github.com/gocql/gocql/blob/master/CONTRIBUTING.md#minimum-requirement-checklist)).

Personally I don't mind the added dependency on lz4 package (especially since the lz4-compressed data is prefixed with the length). @Zariel what do you think?

horkhe:
I do not think the build failure is related to my changes. At least I am able to run the tests my laptop.

horkhe:
@martin-sucha @Zariel hey guys, do you have any more comments? What should I do to get this merged?

martin-sucha:
As I said, looks good to me. I wanted to get @Zariel an opportunity to voice his opinion as this introduces a new dependency.
I'll merge this on June 21st if he does not respond by then.

martin-sucha:
Merged, thanks!

virensinhads:
@martin-sucha @Zariel This new dependency is causing issue with package import stating no such package exists.
This is the error i am getting "as its subpackage github.com/pierrec/lz4/v4 is missing. (Package is required by github.com/gocql/gocql@"
This https://github.com/pierrec/lz4/v4 throws a 404 error meaning no such package exists.

martin-sucha:
@virensinhads how do you manage your dependencies? I can't reproduce. See https://gist.github.com/martin-sucha/5c923731dc686cb75870a593696c34ed

Does running `go mod tidy` solve your issue?

virensinhads:
@martin-sucha basically i am using go version 1.15 and to manage dependencies i use dep ensure

martin-sucha:
I see, so the problem is that `dep` does have some problem to support `v2` and later versions of modules. :disappointed: 

We can either revert this change to preserve compatibility with dep and the lz4 compression would need to be in a separate Go module. Or we could leave it in the tree, since dep is deprecated in favor of Go modules (https://github.com/golang/dep#dep) and reverting could potentially break users that started using the LZ4 compressor (there probably won't be many now).

@zariel @horkhe @virensinhads Opinions? Or do you know of another option than the two listed above?

virensinhads:
@martin-sucha not many users would be probably using this at moment . This can be reverted but best would be to create versions for this so that the no one gets affected

martin-sucha:
> This can be reverted but best would be to create versions for this so that the no one gets affected

Could you elaborate on this please? Do you mean tagging gocql version? How would that help with regards to adding the new dependency?

martin-sucha:
@horkhe @virensinhads What about moving the lz4 compressor to a separate module https://github.com/gocql/gocql/pull/1567, will it work for both of you?

@Zariel or do you want to create a separate gocql/gocql-extras repo instead?

martin-sucha:
Ok, moved the lz4 code to a separate module. @horkhe please update your usages to the new import path.

PR #1562, opened 2021-06-15
Add rack aware round-robin host policy

horkhe:
### Summary
When Cassandra is provisioned in a cloud environment, in such a way that nodes are spread across multiple availability zones, it is more const effective to query nodes in the same availability zone. Currently `gocql` provides [DC-aware round-robin host policy](https://github.com/gocql/gocql/blob/master/policies.go#L772-L777), but it does not respect availability zone. This PR introduces rack-aware round-robin policy that prefers nodes in the same rack, if there are no nodes in the same rack then nodes from the same datacenter are preferred.

### Usage Guidelines
 * When Cassandra is provisioned in AWS with [Ec2Snitch](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/architecture/archSnitchEC2.html), it converts availability zone to datacenter/rack pair breaking it in half on the last dash, so **availability zone** `eu-central-1c` becomes **datacenter** `eu-central` and **rack** `1c`;
 * When Cassandra is provisioned in GCP with [GoogleCloudSnitch](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/architecture/archSnitchGoogle.html), it converts availability zone to datacenter/rack pair breaking it in half on the last dash, so **availability zone** `us-central1-a` becomes **datacenter** `us-central1` and **rack** `a`;

For maximum efficiency we recommend using this policy as a fallback for token aware host policy as follows:
```Go
cluster := gocql.NewCluster(servers)
cluster.PoolConfig.HostSelectionPolicy = gocql.TokenAwareHostPolicy(gocql.RackAwareRRHostPolicy(datacenter, rack))
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1435,7 +1435,7 @@ func (c *Conn) awaitSchemaAgreement(ctx context.Context) (err error) {
 		}
 
 		for _, row := range rows {
-			host, err := c.session.hostInfoFromMap(row, &HostInfo{connectAddress: c.host.ConnectAddress(), port: c.session.cfg.Port})
+			host, err := c.session.hostInfoFromMap(row, c.host.ConnectAddress(), c.session.cfg.Port)
 			if err != nil {
 				goto cont
 			}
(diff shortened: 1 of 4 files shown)


horkhe:
It was a bit premature of me to open this PR before the solution was battle tested in production. It was revealed that `gocql.TokenAwareHostPolicy(gocql.RackAwareRRHostPolicy(datacenter, rack))` combination does not work as expected. So I need to make changes to the solution. Another PR will be open when the change is implemented.

PR #1564, opened 2021-06-21
Use Github actions for CI

martin-sucha:
travis-ci.org switched to travis-ci.com and we wanted to switch to
Github Actions anyway.


.github/workflows/main.yml
```diff
@@ -0,0 +1,181 @@
...
+          go-version: ${{ matrix.go }}
+      - uses: actions/cache@v2
+        id: gomod-cache
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('go.mod') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      - name: Install CCM
+        run: pip install git+https://github.com/pcmanus/ccm.git@2.8.2
```
martin-sucha, line 58: Should be probably github.com/riptano/ccm instead of github.com/pcmanus/ccm as the latter redirects to the former address.

.github/workflows/main.yml
```diff
@@ -0,0 +1,181 @@
...
+          )
+
+          ccm remove test || true
+
+          ccm create test -v $VERSION -n 3 -d --vnodes --jvm_arg="-Xmx256m -XX:NewSize=100m"
+          ccm updateconf "${conf[@]}"
+
+          export JVM_EXTRA_OPTS=" -Dcassandra.test.fail_writes_ks=test -Dcassandra.custom_query_handler_class=org.apache.cassandra.cql3.CustomPayloadMirroringQueryHandler"
+
+          sleep 1s
```
martin-sucha, line 87: Do we need these sleeps?

.github/workflows/main.yml
medasx: Changed. 

We should change also [install_test_deps.sh](https://github.com/gocql/gocql/blob/d83b80dfb4800e8761487e2a0c86cf6864824a8e/install_test_deps.sh)

.github/workflows/main.yml
medasx: 1s sleeps removed.

PR #1567, opened 2021-06-29
Move lz4 compressor to separate module

martin-sucha:
Some users have problems using gocql/gocql with dep as
dep can't work with v4 dependency.
Moving the lz4 compressor to a separate module.
This is a breaking change, but there probably aren't many
people using the lz4 compressor yet.

### DIFF
--- a/compressor.go
+++ b/compressor.go
@@ -1,11 +1,7 @@
 package gocql
 
 import (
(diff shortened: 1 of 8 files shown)


Zariel:
Lgtm

Chris

PR #1570, opened 2021-07-01
Prevent session Close() from hanging up

yutod:
**Summary**
Check closed flag on controlConn HandleError to avoid reconnecting when session closes

**Problem**
When trying to close connection pool via `session.Close()`, the method hangs up due to the following line.
https://github.com/gocql/gocql/blob/master/session.go#L464
https://github.com/gocql/gocql/blob/master/conn.go#L546

**Description**
This PR is adding a flag that is turned on when close method is called and HandleError method skip the main logic when the flag is on. The implementation is following `hostConnPool` on connectionpool.go ([ref](https://github.com/gocql/gocql/blob/master/connectionpool.go#L579-L587))

### DIFF
--- a/control.go
+++ b/control.go
@@ -395,6 +395,10 @@ func (c *controlConn) HandleError(conn *Conn, err error, closed bool) {
 		return
 	}
 
+	if atomic.LoadInt32(&c.started) == -1 {
+		return
+	}
+
 	oldConn := c.getConn()
 
 	// If connection has long gone, and not been attempted for awhile,
--- a/AUTHORS
+++ b/AUTHORS
@@ -123,3 +123,4 @@ Piotr Dulikowski <piodul@scylladb.com>
 Tushar Das <tushar.das5@gmail.com>
 Maxim Vladimirskiy <horkhe@gmail.com>
 Bogdan-Ciprian Rusu <bogdanciprian.rusu@crowdstrike.com>
+Yuto Doi <yutodoi.seattle@gmail.com>


control.go
```diff
@@ -488,6 +499,14 @@ func (c *controlConn) close() {
 		c.quit <- struct{}{}
 	}
 
+	c.mu.Lock()
+	if c.closed {
```
martin-sucha, line 503: The CompareAndSwapInt32 above already checks if `controlConn.close()` was called, why introduce a separate field?

You could achieve the same effect with:

```go
func (c *controlConn) close() {
	if !atomic.CompareAndSwapInt32(&c.started, 1, -1) {
		return
	}
	ch := c.getConn()
	if ch != nil {
		ch.conn.Close()
	}
}
```
(although clearly `c.started` could be named better)

control.go
```diff
@@ -395,6 +399,13 @@ func (c *controlConn) HandleError(conn *Conn, err error, closed bool) {
 		return
 	}
 
+	c.mu.Lock()
+	if c.closed {
+		c.mu.Unlock()
+		return
+	}
+	c.mu.Unlock()
+
```
martin-sucha, line 408: Similar here, this could be just

```go
if atomic.LoadInt32(&c.started) == -1 {
	// controlConn.close() was already called
	return
}
```

martin-sucha (changes requested):
Thank you for the pull request!

I've added inline comments.

However, it is not clear to me from the commit description why exactly Close() hangs (and whether this is the correct fix). Could you please elaborate on the conditions / sequence of events that lead to `Close()` hanging?

yutod:
Thank you for this quick review!
As you commented, reusing `started` flag makes more sense rather than introducing a new field. I'll update it once we make sure this fix is a decent approach.

Here is the sequence of events to trigger the situation.
1. Session.Close() - [try to close the session](https://github.com/gocql/gocql/blob/master/session.go#L450)
2. s.control.close() - [close the control connection](https://github.com/gocql/gocql/blob/master/session.go#L464)
3. ch.conn.Close() - [close the control connection](https://github.com/gocql/gocql/blob/master/control.go#L493)
4. c.closeWithError(nil) - [close the connection without error](https://github.com/gocql/gocql/blob/master/conn.go#L555)
5. c.close() - [close the connection](https://github.com/gocql/gocql/blob/master/conn.go#L540) returns an error e.g. `write tcp 172.XX.XXX.X:41228->XX.XXX.XX.XXX:9142: i/o timeout`
6. c.errorHandler.HandleError(c, cerr, true) - [call error handler with returned error](https://github.com/gocql/gocql/blob/master/conn.go#L546)
7. c.reconnect(false) - HandleError method doesn't check if the connection is supposed to be closed or not, so that it tries to **reconnect** again - [here](https://github.com/gocql/gocql/blob/master/control.go#L406)
8. c.setupConn(newConn) - [try to set up a new connection](https://github.com/gocql/gocql/blob/master/control.go#L382)
9. c.conn.Store(ch) - [try to store atomic value, but never ends](https://github.com/gocql/gocql/blob/master/control.go#L287)

martin-sucha:
I see now. Thanks for the sequence!

So this looks like a regression introduced in https://github.com/gocql/gocql/pull/1555. That change exposed that we are reconnecting during control connection teardown.

We lock the session state mutex in `Session.Close()` at https://github.com/gocql/gocql/blob/769848eae4625444c6abdabc4a67eacb117c9200/session.go#L452, but we try to acquire it again in `controlConn.setupConn()` by calling `Session.initialized()` and we have a deadlock.

martin-sucha:
I think we should check in `controlConn.reconnect` instead of the error handler because reconnect could be called in other places. For example, we don't want to reconnect while closing the session because of heartbeat.

yutod:
@martin-sucha 
Thanks for the clarifying it. The newly opened [PR](https://github.com/gocql/gocql/pull/1571) makes more sense:)

PR #1572, opened 2021-08-03
Merge upstream changes

mmatczuk:
This merges all upstream changes up to 9a3953d.


mmatczuk:
Sorry I fail to change target repo each time I want to catch up with upstream.
Any chances we could avoid this work and merge our patches to gocql/gocql?
Cheers.

PR #1573, opened 2021-08-13
Adding TokenAwareAuthenticator plugin to connection whitelist.

tejavadali:
This is a seemingly trivial change to a plugin whitelist. It uses a drop-in replacement to cassandra username/password authenticator which also supports Token based authentication. Based on https://github.com/gocql/gocql/pull/1441

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -32,6 +32,7 @@ var (
 		"com.ericsson.bss.cassandra.ecaudit.auth.AuditPasswordAuthenticator",
 		"com.amazon.helenus.auth.HelenusAuthenticator",
 		"com.ericsson.bss.cassandra.ecaudit.auth.AuditAuthenticator",
+		"com.bloomberg.cassandra.auth.TokenAwareAuthenticator",
 	}
 )
 


martin-sucha:
Is `com.bloomberg.cassandra.auth.TokenAwareAuthenticator` commonly used? I couldn't find it when searching for `com.bloomberg.cassandra.auth.TokenAwareAuthenticator` on Google.

tejavadali:
> Is `com.bloomberg.cassandra.auth.TokenAwareAuthenticator` commonly used? I couldn't find it when searching for `com.bloomberg.cassandra.auth.TokenAwareAuthenticator` on Google.

This is a new Authenticator, to allow both Username Password Authentication and Token-based Authentication.

tejavadali:
> > Is `com.bloomberg.cassandra.auth.TokenAwareAuthenticator` commonly used? I couldn't find it when searching for `com.bloomberg.cassandra.auth.TokenAwareAuthenticator` on Google.
> 
> This is a new Authenticator, to allow both Username Password Authentication and Token-based Authentication.

Also, If you think there is a better way to whitelist a custom Authenticator, please suggest.

martin-sucha:
> Also, If you think there is a better way to whitelist a custom Authenticator, please suggest.

You can always implement the [Authenticator](https://pkg.go.dev/github.com/gocql/gocql#Authenticator) interface in your own package, the `PasswordAuthenticator` implementation is just a couple of lines of code. If it is an implementation that will only be used somewhere internally, it does not make sense to have it in the driver. That's why I was asking whether it is commonly used.

I don't think we should maintain an allowlist of authenticators (except for the most commonly used ones) as we don't have feedback on when to remove them from the list. People generally just want to add to the list. Maybe we should just have something like `AllowedAuthenticators []string` in `PasswordAuthenticator` to be able to override the allowlist by gocql users?

Are there any cases when that wouldn't work?

> This is a new Authenticator, to allow both Username Password Authentication and Token-based Authentication.

In this case, shouldn't the authenticator implementation in Go support also passing the token instead of a username+password?

tejavadali:
> > Also, If you think there is a better way to whitelist a custom Authenticator, please suggest.
> 
> You can always implement the [Authenticator](https://pkg.go.dev/github.com/gocql/gocql#Authenticator) interface in your own package, the `PasswordAuthenticator` implementation is just a couple of lines of code. If it is an implementation that will only be used somewhere internally, it does not make sense to have it in the driver. That's why I was asking whether it is commonly used.
> 
> I don't think we should maintain an allowlist of authenticators (except for the most commonly used ones) as we don't have feedback on when to remove them from the list. People generally just want to add to the list. Maybe we should just have something like `AllowedAuthenticators []string` in `PasswordAuthenticator` to be able to override the allowlist by gocql users?
>  
> Are there any cases when that wouldn't work?
> >  Just to make sure we are on the same plate, did you mean, that PasswordAuthenticator should carry a dynamic List of Authenticators It can allow?  This sounds neat to me.

> > This is a new Authenticator, to allow both Username Password Authentication and Token-based Authentication.
> 
> In this case, shouldn't the authenticator implementation in Go support also passing the token instead of a username+password?

>> As I mentioned, It supports both Password-based Authentication and token-based authentication simultaneously. For token-based Authentication, the username is "token" and the password carries the actual token.

tejavadali:
> > > Also, If you think there is a better way to whitelist a custom Authenticator, please suggest.
> > 
> > 
> > You can always implement the [Authenticator](https://pkg.go.dev/github.com/gocql/gocql#Authenticator) interface in your own package, the `PasswordAuthenticator` implementation is just a couple of lines of code. If it is an implementation that will only be used somewhere internally, it does not make sense to have it in the driver. That's why I was asking whether it is commonly used.
> > I don't think we should maintain an allowlist of authenticators (except for the most commonly used ones) as we don't have feedback on when to remove them from the list. People generally just want to add to the list. Maybe we should just have something like `AllowedAuthenticators []string` in `PasswordAuthenticator` to be able to override the allowlist by gocql users?
> > Are there any cases when that wouldn't work?
> > > Just to make sure we are on the same plate, did you mean, that PasswordAuthenticator should carry a dynamic List of Authenticators It can allow?  This sounds neat to me.
> 
> > > This is a new Authenticator, to allow both Username Password Authentication and Token-based Authentication.
> > 
> > 
> > In this case, shouldn't the authenticator implementation in Go support also passing the token instead of a username+password?
> 
> > > As I mentioned, It supports both Password-based Authentication and token-based authentication simultaneously. For token-based Authentication, the username is "token" and the password carries the actual token.


@martin-sucha  please kindly take a look at this [PR](https://github.com/gocql/gocql/pull/1574) which I think is a neat approach to handling these issues.

martin-sucha:
Yep, #1574 is the approach I meant. Thanks for the pull request! I've added a comment there.

martin-sucha:
Closing in favor of https://github.com/gocql/gocql/pull/1574

PR #1576, opened 2021-08-17
Fix pool imbalance when using DNS for contact points

### LINKED ISSUE #1575: Resolving DNS when connecting pool connections can lead to connection imbalances

Reproducible with any version. Tested with 3.11.10.

bc256bbb90de7113a74ad4d777beeec75eb9c4e7

go1.16.6 linux/amd64

### What did you do?

Create a `Cluster` object using a DNS name with multiple A records:

```
	cluster := gocql.NewCluster("somednsname.org") // Uses multiple A-records

	cluster.NumConns = 100 // Using this number to show the issue, it can happen with the default of 2

	session, err := gocql.NewSession(*cluster)
	if err != nil {
		log.Fatalf("unable to connect session: %v", err)
	}
```

100 connections per host for the 3 hosts in the cluster and 1 extra for the control connection.

```
$ sudo lsof -n -i TCP:9042 | grep gocql | awk '{ print $9}' | cut -d">" -f2 | sort | uniq -c
100 170.34.196.104:9042
101 178.91.75.34:9042 <-- one extra for the control connection
100 94.203.139.34:9042
```

### What did you see instead?

Imbalanced number of connections. 

```
$ sudo lsof -n -i TCP:9042 | grep gocql | awk '{ print $9}' | cut -d">" -f2 | sort | uniq -c
100 170.34.196.104:9042
1 178.91.75.34:9042
200 94.203.139.34:9042
```

```
$ sudo lsof -n -i TCP:9042 | grep gocql | awk '{ print $9}' | cut -d">" -f2 | sort | uniq -c
2 170.34.196.104:9042
102 178.91.75.34:9042
197 94.203.139.34:9042
```

### The problem ###

The host is created with the original DNS entry as the struct member `hostname`:
https://github.com/gocql/gocql/blob/bc256bbb90de7113a74ad4d777beeec75eb9c4e7/control.go#L151

which causes it to be re-resolved when making the connection: 

https://github.com/gocql/gocql/blob/bc256bbb90de7113a74ad4d777beeec75eb9c4e7/conn.go#L247

using the `hostname` retrieved from `HostInfo.HostnameAndPort()`:
https://github.com/gocql/gocql/blob/bc256bbb90de7113a74ad4d777beeec75eb9c4e7/host_source.go#L369

The problem is that pools are mapped using `ConnectAddress()` from the original DNS resolved IP address (https://github.com/gocql/gocql/blob/bc256bbb90de7113a74ad4d777beeec75eb9c4e7/connectionpool.go#L224), but when re-resolved in `dialer.DialContext()` it can result in a different addresses because A-records don't always come back in the same order. This causes pools to contain connections to multiple addresses instead of one and results in an imbalance.

mpenick:
When a hostname is used for contact points it's resolved initially as
part of the initialization process then `HostInfo` objects are created
with the original `hostname` and the resolved `connectAddress`.
`hostname` is then used to dial the pool connections when causes another
DNS resolve which could result in a different IP then the original
`connectAddress` because A records can change order for each resolve.
This results in a connection pool for a given IP address containing
connections to multiple different IP addresses.

This patch removes the second resolve when dialing by setting the
`hostname` member to the resolved IP in the initialization step.

Resolves #1575

### DIFF
--- a/control.go
+++ b/control.go
@@ -148,7 +148,7 @@ func hostInfo(addr string, defaultPort int) ([]*HostInfo, error) {
 	}
 
 	for _, ip := range ips {
-		hosts = append(hosts, &HostInfo{hostname: host, connectAddress: ip, port: port})
+		hosts = append(hosts, &HostInfo{hostname: ip.String(), connectAddress: ip, port: port})
 	}
 
 	return hosts, nil


martin-sucha:
Looks like the fix for this won't be as simple as this one line change.
We currently don't have tests for TLS connections, but this change could potentially break TLS if we don't store the hostname. Hostname is used for SNI, see https://github.com/martin-sucha/gocql/commit/43a8f9bf6605c25fcaa921fb7c9f8b3d6024f05d (although I'm not sure if there are users that actually use this).

mpenick:
Thanks for the heads up. I can take a look today at a fix that doesn't break TLS.

mpenick:
Pushed up another fix that keeps the original hostname used in the contact points. It should be noted that hosts discovered through the peers table won't have valid hostnames because that table only contains IP addresses so they cannot be verified using a hostname in the CN/SANs. I think that would mean the certificates would need to contain a valid CN/SAN for the hostname and a host specific IP address in another SAN.

conn.go
```diff
@@ -254,14 +253,9 @@ func (s *Session) dialWithoutObserver(ctx context.Context, host *HostInfo, cfg *
 		// be modified after being used.
 		tlsConfig := cfg.tlsConfig
 		if !tlsConfig.InsecureSkipVerify && tlsConfig.ServerName == "" {
-			colonPos := strings.LastIndex(addr, ":")
-			if colonPos == -1 {
```
mpenick, line 258: This check might not have been necessary because `addr := host.HostnameAndPort()` would have always added the `:`.

PR #1590, opened 2021-12-22
Add version2 of function/aggregate metadata

martin-sucha:
There were the following issues with handling of aggregates:

- compileMetadata panicked if final_func was null.
  final_func is optional[1], so null is perfectly valid value.
- The AggregateMetadata FinalFunc field could not store nil,
  so it is not possible to represent the null.
- Functions and aggregates are not uniquely identified just by name.
  There could be two functions with the same name, but different
  argument types.

I'm adding a new version of function and aggregate metadata to fix
those shortcomings. For backward compatibility, I leave the V1
metadata filled-in where possible.

Fixes gocql/gocql#1587

[1] https://cassandra.apache.org/doc/latest/cassandra/cql/functions.html#user-defined-aggregates-functions


medasx (approved):

Zariel:
no we have semver can we just replace the broken api and bump the major version

PR #1593, opened 2021-12-24
Don't panic when there is no Go type available

martin-sucha:
When there is no corresponding Go type available,
unmarshaling panics. We want to return an error instead.

Unfortunately, TypeInfo.New() can't return an error.
We can't change the signature of New as it might be used
externally. For New(), the only option is to keep panicking.
We now at least provide a more useful panic message.

We are adding a new method to interface TypeInfo that returns
as error instead of panicking. Changing the interface is a
breaking change for users that implement custom TypeInfo.
I don't expect any custom implementations of this interface
though, so I think this change is safe.

An alternative to adding the method to the interface would
be to add a separate interface and a function to use the
new interface when available, with fallback to TypeInfo:

    type TypeInfo2 interface {
        TypeInfo
        NewWithError() (interface{}, error)
    }

    func NewFromType(t TypeInfo) (interface{}, error) {
        if t2, ok := t.(TypeInfo2); ok {
            return t2.NewWithError()
        }
        return t.New(), nil
    }

However, I think that is overkill in this case as it's unlikely there
are custom implementations of TypeInfo and NewFromType still wouldn't
guarantee no panics.

Related gocql/gocql#1148

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -20,52 +20,64 @@ type RowData struct {
 	Values  []interface{}
 }
 
-func goType(t TypeInfo) reflect.Type {
+func goType(t TypeInfo) (reflect.Type, error) {
 	switch t.Type() {
 	case TypeVarchar, TypeAscii, TypeInet, TypeText:
-		return reflect.TypeOf(*new(string))
+		return reflect.TypeOf(*new(string)), nil
 	case TypeBigInt, TypeCounter:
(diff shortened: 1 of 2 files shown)


medasx (approved):

medasx:
Why aren't there any checks ? Actions seems to be setup-up correctly.

PR #1597, opened 2022-02-10
Make DESC a ColumnOrder, not a bool

Drahflow:
Discovered while trying something along the lines of https://play.golang.com/p/kMijXA7kW0E, i.e.
```
a := gocql.DESC
if someCondition() {
	a = gocql.ASC
}
```

https://go.dev/ref/spec#Constant_expressions says only the expression part repeats and gives
```
const (
	u         = iota * 42  // u == 0     (untyped integer constant)
	v float64 = iota * 42  // v == 42.0  (float64 constant)
	w         = iota * 42  // w == 84    (untyped integer constant)
)
```


martin-sucha (approved):
Looks good to me. Thanks!

PR #1605, opened 2022-03-15
Add ability to observe streams

martin-sucha:
This allows to get metrics like current count of inflight streams.

A StreamObserver that does not want to store any context per stream
can always return the same value.


martin-sucha:
@Zariel I'd like to merge this upstream. What do you think?

PR #1609, opened 2022-03-18
Check HostFilter for control connection

### LINKED ISSUE #1608: Connecting to a load balancer puts wrong node IP in ring

Yugabyte 2.8

v1.0.0

go 1.17

### What did you do?

```
cluster := gocql.NewCluster("my-load-balancer:9042")
```

```
yugabyte host up -- addr="192.168.0.2" datacenter="us-east1"
yugabyte host up -- addr="192.168.0.4" datacenter="us-east1"
yugabyte host up -- addr="192.168.0.3" datacenter="us-east1"
```

### What did you see instead?

```
yugabyte host up -- addr="192.168.0.2" datacenter="us-east1"
yugabyte host up -- addr="192.168.0.4" datacenter="us-east1"
yugabyte host up -- addr="192.168.0.5" datacenter="us-east1"
...
gocql: events: unable to fetch host info for (192.168.0.3:9042): host not found in peers table
```

The issue is that the `controlConn.setupConn` method calls `conn.localHostInfo` which ends up sending along the `connectAddress` to `hostInfoFromMap` and that always prefers the `connectAddress` when determining how it should connect to the node.

Unfortunately the `AddressTranslator` interface doesn't give the function access to the `HostInfo` so there's no way for me to know that `192.168.0.5` should actually be `192.168.0.3`.

I had thought to do something like:
```
gocql.HostFilterFunc(func(host *gocql.HostInfo) bool {
	host.SetConnectAddress(host.RPCAddress())
	return true
}
```

but the host filter is not called for the control connection. Calling it for the control connection feels like it's the right thing to do but if it returns false we would need to set up a new control connection using another host.

Finally, we could change the control connection setup to not pass `connectAddress` the first time but there might be people relying on that.

---

`my-load-balancer` resolves to `192.168.0.5`
```
admin@ycqlsh> SELECT peer,rpc_address FROM system.peers;

 peer        | rpc_address
-------------+-------------
 192.168.0.3 | 192.168.0.3
 192.168.0.4 | 192.168.0.4

(2 rows)
admin@ycqlsh> SELECT key, rpc_address FROM system.local;

 key   | rpc_address
-------+-------------
 local | 192.168.0.2
```

jameshartig:
Check the HostFilter when making the control connection. This allows you to filter out connecting to the wrong DC or an undesirable host for the control connection. This also has the benefit of allowing the HostFilter to modify the ConnectAddress which fixes #1608.

If you filter out the chosen host for the control connection then we'll try another one until we've exhausted all hosts.

### DIFF
--- a/control.go
+++ b/control.go
@@ -167,28 +167,6 @@ func shuffleHosts(hosts []*HostInfo) []*HostInfo {
 	return shuffled
 }
 
-func (c *controlConn) shuffleDial(endpoints []*HostInfo) (*Conn, error) {
-	// shuffle endpoints so not all drivers will connect to the same initial
-	// node.
-	shuffled := shuffleHosts(endpoints)
-
-	cfg := *c.session.connCfg
-	cfg.disableCoalesce = true
-
-	var err error
-	for _, host := range shuffled {
-		var conn *Conn
-		conn, err = c.session.dial(c.session.ctx, host, &cfg, c)
-		if err == nil {
-			return conn, nil
-		}
-
-		c.session.logger.Printf("gocql: unable to dial control conn %v:%v: %v\n", host.ConnectAddress(), host.Port(), err)
-	}
-
-	return nil, err
-}
-
 // this is going to be version dependant and a nightmare to maintain :(
 var protocolSupportRe = regexp.MustCompile(`the lowest supported version is \d+ and the greatest is (\d+)$`)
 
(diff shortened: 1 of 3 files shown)


jameshartig:
@martin-sucha can you re-run that failed test. I think it's just flaky.

control.go
```diff
@@ -273,18 +279,21 @@ type connHost struct {
...
-	}
-
 	// TODO(zariel): do we need to fetch host info everytime
 	// the control conn connects? Surely we have it cached?
 	host, err := conn.localHostInfo(context.TODO())
 	if err != nil {
 		return err
 	}
 
+	if c.session.cfg.filterHost(host) {
```
martin-sucha, line 289: Shouldn't we filter-out the host before we dial the connection?

control.go
```diff
@@ -167,25 +167,14 @@ func shuffleHosts(hosts []*HostInfo) []*HostInfo {
 	return shuffled
 }
 
-func (c *controlConn) shuffleDial(endpoints []*HostInfo) (*Conn, error) {
-	// shuffle endpoints so not all drivers will connect to the same initial
-	// node.
-	shuffled := shuffleHosts(endpoints)
-
+func (c *controlConn) dial(host *HostInfo) (*Conn, error) {
```
martin-sucha, line 170: It seems `controlConn.dial` is not used anywhere.

control.go
martin-sucha: I see, thanks! It seems we need to update the comment

```go
	// TODO(zariel): do we need to fetch host info everytime
	// the control conn connects? Surely we have it cached?
```

above as we now need to have fresh info for this use case every time, we can't just look it up by connect address anymore.

jameshartig: I updated that comment to reflect that we need up-to-date info now for this call

martin-sucha: Thanks!

martin-sucha (approved):

jameshartig:
@martin-sucha This should be good to be merged now 🙂

jameshartig:
@martin-sucha I actually just pushed a small tweak to reuse the existing `*HostInfo` so we're not making new ones in the control connection

jameshartig:
Rebased on master and switched from comparing the connect addresses to just using `*HostInfo.Equal`

jameshartig:
Not sure why `Build / Integration Tests (1.18, 3.11.10, false, snappy, ccm) (pull_request)` failed but can it be retried?

martin-sucha:
Merged. Thanks for the contribution!

PR #1615, opened 2022-04-07
Add HostAddressTranslator

### LINKED ISSUE #1608: Connecting to a load balancer puts wrong node IP in ring

jameshartig:
This is related to https://github.com/gocql/gocql/pull/1609 but only so much as it solves the problem in a better way. That PR is still necessary to prevent making control connections to unwanted nodes but rather than doing any `connectAddress` mucking in `FilterHost` this new interface provides the `HostInfo` struct necessary to do it here instead.

This is intended on completely replacing the old `AddressTranslator`.

Since translation might change the connect address I added checking against the original `peer` address as well for events.

Fixes #1608

### DIFF
--- a/address_translators.go
+++ b/address_translators.go
@@ -5,6 +5,7 @@ import "net"
 // AddressTranslator provides a way to translate node addresses (and ports) that are
 // discovered or received as a node event. This can be useful in an ec2 environment,
 // for instance, to translate public IPs to private IPs.
+// Deprecated: Use HostAddressTranslator
 type AddressTranslator interface {
 	// Translate will translate the provided address and/or port to another
 	// address and/or port. If no translation is possible, Translate will return the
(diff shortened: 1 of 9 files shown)


martin-sucha:
I'd rather add some kind of dialer as that would work also for other proxy use-cases (like the secure connection bundle - https://github.com/gocql/gocql/pull/1487#issuecomment-715931710). We also probably need to change gocql pool to compare nodes based on node UUID, not connect address.

martin-sucha:
@jameshartig does #1629 solve your use-case?

jameshartig:
> @jameshartig does #1629 solve your use-case?

Unfortunately, I don't think so because the `connectAddress` is still used by the ring and other things and the HostDialer doesn't let you override the `connectAddress`. For example `init` calls `hosts, err := addrsToHosts(s.cfg.Hosts, s.cfg.Port, s.logger)` and passes in the list of configured hosts. This ends up doing `hosts = append(hosts, &HostInfo{hostname: host, connectAddress: ip, port: port})` therefore solidifying the `connectAddress` as the original ip that you connected with rather than the address that the dialer actually used. So let's say you connect to a load balancer at `192.168.0.5` and that actually connects to an instance that's `192.168.0.4`. What would put a peer in the ring that has `connectAddress: 192.168.0.5` but there's no peer with that address. In my PR inside of `hostInfoFromMap` I called `translateAddressPort` so it could end up using the `rpc_address` which gets set as the `connectAddress`.

I do like your solution better as it solves a lot more use-cases but I'm not sure how to solve correcting the `connectAddress` when using a dialer unless we delayed putting the `connectAddress` on the `HostInfo` until after we call `Dial` and then get the address from the net.Conn?

martin-sucha:
> the connectAddress is still used by the ring

Please see https://github.com/gocql/gocql/pull/1632.

jameshartig:
> > the connectAddress is still used by the ring
> 
> Please see #1632.

That should solve the issue along with your PR. I'll close this. Thanks!

PR #1616, opened 2022-04-07
Mark node as down if pool filling failed and pool is empty

### LINKED ISSUE #1614: Failed host is never reconnected if NumConns > 1

Yugabyte 2.8

v1.0.0

go 1.17

### What did you do?
`cluster.NumConns = 2`

Node was terminated and both connections closed which called `closeWithError` which ended up calling `hostConnPool.HandleError` and `hostConnPool.fill`. Unfortunately `hostConnPool.fill` was called twice and the first time failed to connect but since `startCount` was still `1` (since the other connection is blocked on `pool.mu`). The second time didn't do anything in `fill` because `filling = true` since the first goroutine is stuck in `fillingStopped`.

I expected to see the node eventually reconnect.

### What did you see instead?
The host is in a stuck state where the pool is empty and set the ring status is `Up`. This means it will never reconnect.

Unfortunately, I didn't have a chance to look at the `system.peers` table while it was down but I assume that the node was still present otherwise it would've removed from the ring when it was refreshed.
---

- output of `nodetool status`

jameshartig:
Previously if the `NumConns > 1` there was a race to fill the pool and we only ever marked the host as down if the `startCount == 0` but if 2 connections failed simultaneously there would be a race and not always would the startCount be 0. Therefore we now check whenever filling stops to see if the pool is empty and if so, mark as down.

Fixes #1614

### DIFF
--- a/connectionpool.go
+++ b/connectionpool.go
@@ -432,13 +432,7 @@ func (pool *hostConnPool) fill() {
 
 		if err != nil {
 			// probably unreachable host
-			pool.fillingStopped(true)
-
-			// this is call with the connection pool mutex held, this call will
-			// then recursively try to lock it again. FIXME
-			if pool.session.cfg.ConvictionPolicy.AddFailure(err, pool.host) {
-				go pool.session.handleNodeDown(pool.host.ConnectAddress(), pool.port)
-			}
+			pool.fillingStopped(err)
 			return
 		}
 		// notify the session that this node is connected
@@ -453,7 +447,7 @@ func (pool *hostConnPool) fill() {
 		err := pool.connectMany(fillCount)
 
 		// mark the end of filling
-		pool.fillingStopped(err != nil)
+		pool.fillingStopped(err)
 
 		if err == nil && startCount > 0 {
 			// notify the session that this node is connected again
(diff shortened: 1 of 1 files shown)


connectionpool.go
```diff
@@ -486,7 +483,21 @@ func (pool *hostConnPool) fillingStopped(hadError bool) {
...
 	pool.mu.Unlock()
+
+	// if we errored and the size is now zero, make sure the host is marked as down
+	// see https://github.com/gocql/gocql/issues/1614
+	if gocqlDebug {
+		pool.logger.Printf("gocql: conns of pool after stopped %q: %v\n", host.ConnectAddress(), count)
+	}
+	if err != nil && count == 0 {
+		if pool.session.cfg.ConvictionPolicy.AddFailure(err, host) {
+			pool.session.handleNodeDown(host.ConnectAddress(), port)
```
martin-sucha, line 498: Previously, `handleNodeDown` was called in a new goroutine. Is it safe to call it directly?

jameshartig: Yeah, there was a comment about that. Previously it was called in a goroutine to avoid a deadlock but now it's called within `fillingStopped` which is called within an anonymous goroutine at the end of `fill`. So it's still in a goroutine w/r/t `fill`.

martin-sucha:
Merged. Thanks for the contribution!

PR #1627, opened 2022-06-21
Use legacy URL parsing for ccm

martin-sucha:
nodetool launched by ccm fails with:

 nodetool: Failed to connect to 'localhost:7100' - URISyntaxException: 'Malformed IPv6 address at index 7: rmi://[localhost]:7100'

Switching back to legacy parsing mode should fix the error.


martin-sucha:
Closing in favor of #1628

PR #1629, opened 2022-06-21
Add HostDialer interface

martin-sucha:
There are multiple use cases where the Dialer interface is not
sufficient.

When using TLS, users need to control also TLS context per host.
For example, host certificates might be either UUID in common name,
some hostname, IP address per host, etc.

Hosted instances or instances deployed in Kubernetes cluster tend
to be behind proxies. A proxy might use TLS server name indication
to identify which database host to connect to.


dial.go
```diff
@@ -0,0 +1,79 @@
...
+	// Conn used to communicate with the server.
+	Conn net.Conn
+
+	// DisableCoalesce disables write coalescing for the Conn.
+	// If true, the effect is the same as if WriteCoalesceWaitTime was configured to 0.
+	DisableCoalesce bool
+}
+
+// defaultHostDialer dials host in a default way.
+type defaultHostDialer struct {
```
jameshartig, line 30: Is there a way to expose this for HostDialer wrappers? If you only want to connect to a host using the host's RPCAddress this is a lot of code to copy/paste just to change the IP address used for connecting. Maybe instead there can be a public `func DefaultHostDialer(dialer Dialer, tlsConfig *tls.Config) HostDialer` that people can call to get the default one and then they can do:

```go
myDialer{
  next: DefaultHostDialer(...),
}
...
func (md *myDialer) DialHost(ctx context.Context, host *HostInfo) (*DialedHost, error) {
  host.SetConnectAddress(host.RPCAddress)
  return md.next.DialHost(host)
}
```

martin-sucha: Updating the connect address in the host seems racy, so I would rather pass the address as an argument.

In https://github.com/scylladb/gocql/pull/98, we have `func wrapTLS(conn net.Conn, addr string, tlsConfig *tls.Config) (*DialedHost, error) {` that is shared between multiple HostDialers. Perhaps we can export this function, so the usage would be like:

```go
myDialer{
	tlsConfig: &tls.Config{}, // or nil for disabled TLS
}
```
```go
func (md *myDialer) DialHost(ctx context.Context, host *gocql.HostInfo) (*gocql.DialedHost, error) {
	var d net.Dialer
	addr := host.RPCAddress()
	conn, err := d.DialContext(ctx, "tcp", addr)
	if err != nil {
		return nil, err
	}
	return gocql.WrapTLS(conn, addr, md.tlsConfig)
}
```

Alternately, we could pass the function to setup TLS as argument to the DialHost function. That would remove the need for the user to configure TLS context in the dialer explicitly if they don't want to:

```go
// TLSSetupFunc wraps a the conn into a TLS session if necessary.
// addr is used to set the hostname in the TLS context if not provided, it can be in form host or host:port
// where host is hostname or IP address.
type TLSSetupFunc func(conn net.Conn, addr string) (*gocql.DialedHost, error)

func (md *myDialer) DialHost(ctx context.Context, host *gocql.HostInfo, defaultWrapTLS TLSSetupFunc (*gocql.DialedHost, error) {
	var d net.Dialer
	addr := host.RPCAddress()
	conn, err := d.DialContext(ctx, "tcp", addr)
	if err != nil {
		return nil, err
	}
	return defaultWrapTLS(conn, addr)
}
```

What do you think?

cc @mmatczuk @piodul

martin-sucha: Okay, I have exported `WrapTLS`. It makes more sense to me than passing the address in parameter, as the default host dialer in the scylla fork replaces the connect address in shard-aware-port mode.

martin-sucha: I've noticed that WrapTLS did not use HandshakeContext, so I added context there.

PR #1630, opened 2022-06-21
Handle nil values when (un)marshalling lists and maps

sergeikaretnikov:
Handle `nil` values for supported protocol versions when (un)marshalling lists and maps according to the specs:
```
6.11 list

  A [int] n indicating the number of elements in the list, followed by n
  elements.  Each element is [bytes] representing the serialized value.

6.12 map

  A [int] n indicating the number of key/value pairs in the map, followed by
  n entries.  Each entry is composed of two [bytes] representing the key
  and value.
```
```
    [bytes]        A [int] n, followed by n bytes if n >= 0. If n < 0,
                   no byte should follow and the value represented is `null`.
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -127,3 +127,4 @@ Yuto Doi <yutodoi.seattle@gmail.com>
 Krishna Vadali <tejavadali@gmail.com>
 Jens-W. Schicke-Uffmann <drahflow@gmx.de>
 Ondrej Polakovič <ondrej.polakovic@kiwi.com>
+Sergei Karetnikov <sergei.karetnikov@gmail.com>
--- a/marshal.go
+++ b/marshal.go
@@ -1651,11 +1651,15 @@ func unmarshalList(info TypeInfo, data []byte, value interface{}) error {
 			if len(data) < m {
 				return unmarshalErrorf("unmarshal list: unexpected eof")
 			}
+
+			// In case m < 0, the value is null, and unmarshalData should be nil.
+			var unmarshalData []byte
 			if m >= 0 {
-				if err := Unmarshal(listInfo.Elem, data[:m], rv.Index(i).Addr().Interface()); err != nil {
-					return err
-				}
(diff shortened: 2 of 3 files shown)


martin-sucha (commented):
Hello @sergeikaretnikov ! Thanks for the contribution!

Could you please also [update the AUTHORS file](https://github.com/gocql/gocql/blob/master/CONTRIBUTING.md#minimum-requirement-checklist)?

martin-sucha:
I have checked and the server does not support nulls inside collections:

```
[cqlsh 5.0.1 | Cassandra 4.0.4 | CQL spec 3.4.5 | Native protocol v4]
Use HELP for help.
cqlsh> CREATE KEYSPACE cycling   WITH REPLICATION = {     'class' : 'SimpleStrategy',     'replication_factor' : 1    };
cqlsh> use cycling;
cqlsh:cycling> CREATE TABLE list_example (id text, l list<text>, PRIMARY KEY ((id)));
cqlsh:cycling> CREATE TABLE list_example2 (id text, l frozen<list<text>>, PRIMARY KEY ((id)));
cqlsh:cycling> insert into list_example (id, l ) VALUES ( 'a', ['b', null, 'c']);
InvalidRequest: Error from server: code=2200 [Invalid query] message="null is not supported inside collections"
cqlsh:cycling> insert into list_example2 (id, l ) VALUES ( 'a', ['b', null, 'c']);
InvalidRequest: Error from server: code=2200 [Invalid query] message="null is not supported inside collections"
```

However, we should still merge this pull request as the current gocql behavior is not correct - the user should receive the error message from the server and with current master we store a list with zero values instead.

<hr>

It seems that a similar issue can happen when marshaling maps too. Would you like to create another pull request for maps?

martin-sucha (approved):

martin-sucha:
Good point about the use in the IN clause.

The changes look good to me. I have merged them to master. Thanks for the contribution!

PR #1632, opened 2022-06-22
Identify nodes by their host_id instead broadcasted_address

zimnx:
Currently driver identifies nodes based on their broadcasted IP
address. In cloud case, broadcasted IP addresses are private and are not
meant to be used as a contact point, and they may change overtime.
Hence driver internals were changed to identify nodes based on their
host_id which is unique per node and it's persistant throught entire
node lifecycle.
Because CQL Events are still using broadcasted IP addresses, driver
keep mapping between already known IP addresses and host_ids.
Prepared statement cache key was also changed to host_id, to not
invalidate it upon IP change.

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -174,6 +174,22 @@ func (h *HostInfo) connectAddressLocked() (net.IP, string) {
 	return net.IPv4zero, "invalid"
 }
 
+// nodeToNodeAddress returns address broadcasted between node to nodes.
+// It's either `broadcast_address` if host info is read from system.local or `peer` if read from system.peers.
+// This IP address is also part of CQL Event emitted on topology/status changes,
+// but does not uniquely identify the node in case multiple nodes use the same IP address.
+func (h *HostInfo) nodeToNodeAddress() net.IP {
+	h.mu.RLock()
+	defer h.mu.RUnlock()
+
+	if validIpAddr(h.broadcastAddress) {
+		return h.broadcastAddress
+	} else if validIpAddr(h.peer) {
+		return h.peer
+	}
+	return net.IPv4zero
+}
+
 // Returns the address that should be used to connect to the host.
 // If you wish to override this, use an AddressTranslator or
 // use a HostFilter to SetConnectAddress()
@@ -237,6 +253,12 @@ func (h *HostInfo) HostID() string {
 	h.mu.RLock()
 	defer h.mu.RUnlock()
 	return h.hostId
+}
+
+func (h *HostInfo) SetHostID(hostID string) {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.hostId = hostID
 }
 
 func (h *HostInfo) WorkLoad() string {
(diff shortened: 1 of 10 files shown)


ring.go
```diff
@@ -73,40 +80,47 @@ func (r *ring) addHostIfMissing(host *HostInfo) (*HostInfo, bool) {
...
+	if r.hostIPToUUID == nil {
+		r.hostIPToUUID = make(map[string]string)
+	}
 
-	existing, ok := r.hosts[ip]
+	existing, ok := r.hosts[hostID]
 	if !ok {
-		r.hosts[ip] = host
+		r.hosts[hostID] = host
+		r.hostIPToUUID[host.ConnectAddress().String()] = hostID
```
jameshartig, line 96: From what I can tell reading the cassandra source, it looks like the topology events are triggered with the host set to `getNativeaddress` which returns the rpc_address. I think this should be changed to the rpc_address (maybe falling back to the connect address if that's empty?) to prevent any weird issues with connecting through a LB or proxy.

zimnx: Good catch, it was actually a bug in current driver. Nodes are identified between each other and in CQL event by their broadcast_address, which is kept in `broadcast_address` column in `system.local` and by `peer` column in `system.peers`. 
I added new method to HostInfo to extract correct IP and used it in IP to UUID mapping.

Luckily driver was prepared for not finding given host in every event handler so it will just minimize number of ring refreshes.

zimnx:
> We should probably use the host ID also as the key for prepared statement cache.

Good idea, changed the key to HostID.

jameshartig:
LGTM 👍

host_source.go
```diff
@@ -174,6 +174,21 @@ func (h *HostInfo) connectAddressLocked() (net.IP, string) {
 	return net.IPv4zero, "invalid"
 }
 
+// NativeAddress returns address broadcasted between node to nodes.
+// It's either `broadcast_address` if host info is read from system.local or `peer` if read from system.peers.
+// This address is also used in CQL Event emitted on topology/status changes.
+func (h *HostInfo) NativeAddress() net.IP {
```
martin-sucha, line 180: The type emitted in topology/status change events is `[inet]`:

https://github.com/apache/cassandra/blob/39e89fd636ee4343eb2201820da87881cbc749e2/doc/native_protocol_v4.spec#L760-L770

`[inet]` is IP address and port:

https://github.com/apache/cassandra/blob/39e89fd636ee4343eb2201820da87881cbc749e2/doc/native_protocol_v4.spec#L241-L245

It seems we actually don't get the port from `system.local` and `system.peers` though as the `broadcast_address` and `peer` columns are `inet` CQL data type (IP address only), which is distinct from the `[inet]`. See also https://github.com/gocql/gocql/issues/1634#issuecomment-1167002493

I don't know if we can do something else than ignore the port though as the port is not in system.peers table. We can at least clarify the doc comment a little bit:

```
// This IP address is also part of CQL Event emitted on topology/status changes,
// but does not uniquely identify the node in case multiple nodes use the same IP address.
```

ring.go
```diff
@@ -14,8 +13,9 @@ type ring struct {
 	endpoints []*HostInfo
 
 	// hosts are the set of all hosts in the cassandra ring that we know of
-	mu    sync.RWMutex
-	hosts map[string]*HostInfo
+	mu           sync.RWMutex
+	hosts        map[string]*HostInfo
```
martin-sucha, line 17: Please add doc comment that the key is hostID.

zimnx: Added

host_source.go
zimnx: Both Cassandra and Scylla returns keep IP address in `system.local` and `system.peers`. Extended the documentation with your suggestion.

martin-sucha: It seems that we can get the port from `system.peers_v2` for Cassandra 4.0 and up (see https://github.com/gocql/gocql/issues/1634#issuecomment-1167091619), so `NativeAddress()` should probably return both IP address and port. If we had NativeAddress() and NativePort() as separate methods, we would have to duplicate the selection logic.

Also there are `native_address` and `native_port` fields in `system.peers_v2`. Should NativeAddress() use those for Cassandra 4.0 or are those unrelated and we should rename the NativeAddress() method not to clash?

ring.go
martin-sucha: Thanks!

host_source.go
jameshartig: That table doesn't exist for Yugabyte so it would need to check the error and fallback if we end up trying to use that table: https://github.com/yugabyte/yugabyte-db/commit/fc5e713354c388391b01170871d4e65c8623f69f

martin-sucha: Currently https://github.com/gocql/gocql/pull/1635 has check for server version >= 4.0.0 to determine whether to use system.peers_v2. Would that check work for Yugabyte or should we also add the fallback in case the table does not exist? For Scylla it should work with the version check as it returns server version lower than 4 currently.

zimnx: Changed name to NodeToNodeAddress() to avoid clash with how columns are called in `system.peers_v2`.

zimnx: sure, fixed

PR #1635, opened 2022-06-28
Propagate right cql ports to address translator implementations

smiklosovic:
Hi @martin-sucha ,

I prepared the PR for https://github.com/gocql/gocql/issues/1634

This patch also includes changes from this [PR ](https://github.com/gocql/gocql/pull/1632) at the moment.

Please let me know what you think about this approach.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1620,7 +1620,7 @@ func (c *Conn) query(ctx context.Context, statement string, values ...interface{
 
 func (c *Conn) awaitSchemaAgreement(ctx context.Context) (err error) {
 	const (
-		peerSchemas  = "SELECT * FROM system.peers"
+		peerSchemasTemplate = "SELECT * FROM %s"
 		localSchemas = "SELECT schema_version FROM system.local WHERE key='local'"
 	)
 
@@ -1628,8 +1628,11 @@ func (c *Conn) awaitSchemaAgreement(ctx context.Context) (err error) {
 	var schemaVersion string
 
 	endDeadline := time.Now().Add(c.session.cfg.MaxWaitSchemaAgreement)
+
+	queryString := fmt.Sprintf(peerSchemasTemplate, peersTableName(c.host.version))
+
 	for time.Now().Before(endDeadline) {
-		iter := c.query(ctx, peerSchemas)
+		iter := c.query(ctx, queryString)
 
 		versions = make(map[string]struct{})
 
--- a/host_source.go
+++ b/host_source.go
@@ -500,12 +500,24 @@ func (s *Session) hostInfoFromMap(row map[string]interface{}, host *HostInfo) (*
 				return nil, fmt.Errorf(assertErrorMsg, "rpc_address")
 			}
 			host.rpcAddress = net.ParseIP(ip)
+		case "native_address":
+			ip, ok := value.(string)
+			if !ok {
+				return nil, fmt.Errorf(assertErrorMsg, "native_address")
+			}
+			host.rpcAddress = net.ParseIP(ip)
 		case "listen_address":
 			ip, ok := value.(string)
 			if !ok {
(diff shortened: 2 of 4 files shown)


smiklosovic:
This patch also includes changes from this [PR ](https://github.com/gocql/gocql/pull/1632)

host_source.go
```diff
@@ -497,12 +497,24 @@ func (s *Session) hostInfoFromMap(row map[string]interface{}, host *HostInfo) (*
...
 			if !ok {
 				return nil, fmt.Errorf(assertErrorMsg, "listen_address")
 			}
 			host.listenAddress = net.ParseIP(ip)
+		case "native_port":
+			native_port, ok := value.(int)
+			if !ok {
+				return nil, fmt.Errorf(assertErrorMsg, "native_port")
+			}
+			host.port = native_port
```
martin-sucha, line 517: It seems we should store the native port into a new field, there are different ports fields in system.local: `broadcast_port`, `listen_port`, `rpc_port` and in system.peers_v2: `peer_port`, `native_port`, `preferred_port`.
So it should be clear from the API which port we return, especially since we already have methods like BroadcastAddress/ListenAddress/RPCAddress/PreferredIP. We should probably add corresponding methods that return both address and port and deprecate the methods that return only address, for example:

```go
func (h *HostInfo) BroadcastAddressPort() (net.IP, int) {
	h.mu.RLock()
	defer h.mu.RUnlock()
	return h.broadcastAddress, h.broadcastPort
}
```

Is there some documentation on how `broadcast_port`, `listen_port`, `rpc_port` correspond to `peer_port`, `native_port`, `preferred_port`? I couldn't find any Cassandra docs on this. Is rpc_port the same as native_port?

conn.go
```diff
@@ -1662,16 +1662,26 @@ func (c *Conn) query(ctx context.Context, statement string, values ...interface{
...
 
 	endDeadline := time.Now().Add(c.session.cfg.MaxWaitSchemaAgreement)
+
+	var queryString string
+
+	if c.host.version.AtLeast(4, 0, 0) {
+		queryString = peerSchemasV2
+	} else {
+		queryString = peerSchemas
+	}
```
martin-sucha, line 1681: This version check is in three places. Perhaps we should have a function that returns peers table name based on the version?

```go
func peersTableName(version cassVersion) string {
	if version.AtLeast(4, 0, 0) {
		return "system.peers_v2"
	}
	return "system.peers"
}
```

martin-sucha (commented):
Thanks for the pull request! Please find my comments inline.

Also, could you please [update the AUTHORS file](https://github.com/gocql/gocql/blob/master/CONTRIBUTING.md#minimum-requirement-checklist)?

conn.go
smiklosovic: added.

host_source.go
smiklosovic: `broadcast_port` is by default `7000` and it corresponds to `storage_port` in `cassandra.yaml`. That is port via Cassandra nodes talk to each other internaly. `listen_port` is same, 7000. If you are not using `broadcast_address` (it is commented out, then it defaults to `listen_address`.

In times of pre-4.0, `rpc_port` in cassandra.yaml configuration was meant to represent Thrift port. Whole Thrift was abandoned / removed from 4.0 and afaik it seems like it is same as transport  / cql port.

peer_port is by default 7000 and that is used for internal communication between nodes, basically same as `storage_port` in the configration.

I am not sure about preferred port, that is not important in this context.

You see it is quite wild, I would just keep this as simple as possible.

martin-sucha: Thanks for the explanation!

martin-sucha (commented):
Please rebase onto latest master.

smiklosovic:
@martin-sucha rebased

martin-sucha:
Looks good to me. Merged. Thanks!

PR #1645, opened 2022-08-16
Change default port if provided in new cluster

michaelraney:
Currently, when creating a cluster, the contact points can contain the port as well.  Currently this does not change the port when a new cluster is configured, causing timeout issues to peers.  This especially impacting when using TLS which is using 9142 instead of 9042.  If user does not set Cluster.Port connection to peers failed. Traditionally, cassandra drivers do not support connecting to nodes on different ports. All nodes must have the same cql port.  The result is users may end up connected only to provided seeds initial host without connecting to peers.  This pull request adds port assignment to NewCluster by extracting it from the host string if provided.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"net"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


martin-sucha:
Thanks for the pull request!

Cassandra 4.0 supports different ports per node (see https://github.com/gocql/gocql/pull/1635/commits/8f058c60165fc3be0e4e483966fbbe76186e39b8). However, with Cassandra 4.0 and later, the `ClusterConfig.Port` will be ignored (instead gocql will use the port from `system.peers_v2`) so the change should be safe with Cassandra 4.0.

Could this change break something? What should happen if the seeds have different ports? For example shouldn't `gocql.NewCluster("hostname1:9142", "hostname2")` use 9042 instead of 9142? Shouldn't we try to infer the port only if all the ports are the same?

In any case, please [add an entry to the AUTHORS file](https://github.com/gocql/gocql/blob/master/CONTRIBUTING.md#minimum-requirement-checklist).

PR #1646, opened 2022-08-24
Fallback to `system.peers` when `system.peers_v2` is not available

mpenick:
As currently implemented `system.peers_v2` support doesn't work for DSE
clusters because it returns 4.0.x for the `release_version`. This
attempts to use the `system.peers_v2` table when `release_version` is >=
4.0.x, but falls back gracefully to `system.peers` when that table isn't
found.


martin-sucha (approved):
Looks good to me. Thanks!

PR #1654, opened 2022-09-06
Use Go 1.19 in CI

martin-sucha:
We generally support the last two stable versions of Go,
updating the CI to test against them.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


medasx:
What about keeping go1.17 a little bit longer ? go1.18 was quite a big release, it's not even fully supported by golanci-lint yet.

martin-sucha:
Go 1.17 is no longer supported by Go project, which only supports the last two stable releases. Go 1.17 won't receive any security updates. I think we should stick to the last two stable Go versions. If anyone uses Go in production they should have already switched to Go 1.18.

medasx (approved):

PR #1664, opened 2022-11-14
Expose cluster metadata and token ranges

cs-sotescu:
This PR does a few things:

- It refactors the cluster metadata handling by extracting it from the TokenAwareHostPolicy into its own manager;
- It manages cluster metadata for any host policy, in the session itself;
- It exposes the metadata through a new Session.ClusterMetadata method;
- It gives access to the token ring thorough ClusterMetadata.GetTokenRing;
- It publishes TokenRing.GetHostForToken and adds a new method, TokenRing.GetTokenRanges.
- The goal is to allow people to build host-targeted token range queries like `SELECT id,foo FROM mytable WHERE TOKEN(id)>$startToken AND TOKEN(id)<=$endToken`.


token.go
```diff
@@ -5,218 +5,262 @@
...
-	"bytes"
-	"crypto/md5"
-	"fmt"
-	"math/big"
-	"sort"
-	"strconv"
-	"strings"
-
-	"github.com/gocql/gocql/internal/murmur"
+    "bytes"
```
martin-sucha, line 8: **nitpick**: It seems that in some files, you have changed tabs to spaces. Could you please keep tabs?

token.go
```diff
@@ -5,218 +5,262 @@
...
-	partitioner partitioner
-
-	// tokens map token range to primary replica.
-	// The elements in tokens are sorted by token ascending.
-	// The range for a given item in tokens starts after preceding range and ends with the token specified in
-	// token. The end token is part of the range.
-	// The lowest (i.e. index 0) range wraps around the ring (its preceding range is the one with largest index).
-	tokens []hostToken
+// TokenRing is a data structure for organizing the relationship between tokens and hosts
+type TokenRing interface {
```
martin-sucha, line 144: **suggestion**: Don't export interface. Interfaces should usually be defined at the user side of an API. Adding a method to interface is a breaking change unless there is a private method that guarantees that users can't provide their own implementation. Also, we won't be providing multiple TokenRing implementations. So it seems better to export a pointer to struct directly.

Same with ClusterMetadata.

token.go
```diff
@@ -5,218 +5,262 @@
...
+    //GetHostForToken returns the host to which a token belongs
+    GetHostForToken(token Token) (host *HostInfo, endToken Token)
```
martin-sucha, line 146: **suggestion**: Rename the method to `HostForToken`, in Go usually there are no `Get` prefixes for getters.

token.go
```diff
@@ -5,218 +5,262 @@
...
-	Less(token) bool
+// Token is a token
+type Token interface {
+    fmt.Stringer
+    Less(Token) bool
+}
+
+//TokenRange represents a token range.
+//The start token is exclusive and the end token is inclusive.
+type TokenRange struct {
```
martin-sucha, line 34: Do we need `TokenRange`? It seems to me that handling the sorted slice of tokens should be straightforward without needing to duplicate the start/end tokens in TokenRange. I'd rather avoid exporting another type if not necessary.

martin-sucha (commented):
Thanks for the pull request and sorry for the late reply! I think this could be useful for full table scans, for example.

martin-sucha:
How will the API behave in case the token ring is not available (e.g. because of DisableInitialHostLookup option)? The documentation should explicitly mention that.

token.go
cs-sotescu: No, we don't. Removed.

token.go
```diff
@@ -202,7 +217,99 @@ func (t *tokenRing) String() string {
...
+
+	ranges := make([]Token, len(t.tokens))
+	for i := range t.tokens {
+		ranges[i] = t.tokens[i].token
+	}
+
+	return ranges
+}
+
+func buildTokenQueries(s *Session, t *TokenRing) []*Query {
```
tengu-alt, line 276: Please remove it, it's unused and the same code is already commented on above as an example.

token.go
```diff
@@ -202,7 +217,99 @@ func (t *tokenRing) String() string {
...
-func (t *tokenRing) GetHostForToken(token token) (host *HostInfo, endToken token) {
+// Tokens returns the token range corresponding to the primary replica.
+// The elements are sorted by token ascending.
+// The range for a given item starts after preceding range and ends with the token at the current position.
+// The end token is part of the range.
+// The lowest (i.e. index 0) range wraps around the ring (its preceding range is the one with the largest index).
+// You can obtain the owner host/vnode of the range by calling HostForToken with the end token.
+//
+// The following example constructs one TOKEN-based query for each token range:
+//	func buildTokenQueries(s *Session, t *TokenRing) []*Query {
```
OleksiienkoMykyta, line 228: As I understand, you just added the buildTokenQueries function to demonstrate how to use token-based queries. I think that this commented example would be enough, but the unexportable function (line 276) ```func buildTokenQueries(s *Session, t *TokenRing) []*Query {```  should be deleted, or used for testing.

cluster_metadata.go
```diff
@@ -0,0 +1,180 @@
...
+		m.hosts.add(host)
+	}
+
+	meta := m.getMetadataForUpdate()
+	meta.resetTokenRing(m.partitioner, m.hosts.get(), m.logger)
+	m.updateReplicas(meta, m.getKeyspaceName())
+	m.metadata.Store(meta)
+}
+
+func (m *clusterMetadataManager) removeHost(host *HostInfo) {
```
OleksiienkoMykyta, line 124: To be more consistent and help future developers, it would be nice to have all methods/functions commented on.

session.go
```diff
@@ -553,6 +562,12 @@ func (s *Session) KeyspaceMetadata(keyspace string) (*KeyspaceMetadata, error) {
 	return s.schemaDescriber.getSchema(keyspace)
 }
 
+// ClusterMetadata returns the cluster metadata.
+// The returned value is a snapshot of the metadata at the time of the call. Do not store for later reuse.
```
OleksiienkoMykyta, line 566: We should provide more details to the user here, it's a potential weak spot. Something like:
``` 
// ClusterMetadata returns a snapshot of the current cluster metadata at the time of the call.
// The metadata may change over time, so storing the returned pointer for later reuse can lead to inconsistencies.
// Potential Data Race Risk:
// If metaMngr is modified by another goroutine while being read, a data race may occur.
```

OleksiienkoMykyta (commented):
It should be updated, there are a lot of changes provided since the PR was made, but generally it's pretty good, and it helps with (https://github.com/apache/cassandra-gocql-driver/issues/356).
I think the mentioned fixes will make it a little better.
@jameshartig, @joao-r-reis, could you take a look at it, to push mentioned issue forward, it's still relevant and needs fix.

Issue #915, opened 2017-05-23
"gocql: no hosts available in the pool" cannot recover without restart


zhixinwen:
We had a cluster degrades due to increased traffic load, and on the client side we see "gocql: no hosts available in the pool" error for ~80% of the requests.

The error rate is very consistent, and the C* cluster is healthy except the load. We try to reduce the load on the cluster, but the error is still consistent. It lasts for two hours.

We restart the client, and the errors go away immediately.

Also one thing interesting is that after we restart the gocql client, we also see an immediate drop on coordinator latency and traffic on the C* cluster (Yes, client restarts first, then the latency and traffic on C* drops, not the other way around). It may be due to upstream users stop retry after we restart the server and no longer see "gocql: no hosts available in the pool", but we are not sure about the root cause yet.

zhixinwen:
After we close a connection due to too many timeouts, when would the connection be made again? 

I do see gocql wait for event from control connection to reconnect. But if no node is perceived as down by the control node, and it never emits a 'node up' event, would gocql try to recreate the connections?

zhixinwen:
@Zariel gocql closes a connection when it takes too long to respond, when would the connection be reconnected?

I may miss some of the code, looks like it will only reconnect when it receives "Node Up" event?

Zariel:
It should be periodically try to dial nodes that are down, if you can reproduce and build with `gocql_debug` you should see more info about the pool state.

zhixinwen:
I think the root cause of the problem here is:

1. When all connections to a hosts are closed, the host is not marked as down, so ReconnectInterval would not help
2. HandleError in hostConnPool would try to refill the connection pool when a connection is closed, but in an outage it is very likely to fail. Moreover it would not retry afterward. As a result, when the cluster recovers after all the connections are closed, gocql would not attempt to rebuild the connection pool

To fix this, we need to mark the host as DOWN, when we close all the connections on that host.

balta2ar:
I simulated network connectivity issues using iptables:
```
sudo iptables -I OUTPUT -p tcp -d 127.0.0.1 --dport 9042 -j DROP
```
After that upon queries I start to receive this error message: `gocql: no response received from cassandra within timeout period`
And soon after it this message: `gocql: no hosts available in the pool`

After deleting the rule (`sudo iptables -D OUTPUT 1`) gocql indeed does not recover automatically.

Zariel:
[...]

PR #1669, opened 2023-01-11
refresh ring when node UP event received for missing host

sseidman:
The current issue is that there are no events dispatched for `NEW_NODE` changes. In a 3 node cassandra cluster deployed on cloud infrastructure, a cassandra instance was replaced using the `replace_address` flag so that it was moved to a new cloud instance (new IP, new host ID). The following debug logs are an example of how the client is currently handling this change in topology. 
```
2023/01/11 20:59:35 gocql: handling frame: [topology_change change=NEW_NODE host=10.128.191.122 port=9042]
2023/01/11 20:59:35 gocql: handling frame: [status_change change=UP host=10.128.191.122 port=9042]
2023/01/11 20:59:35 gocql: dispatching event: &{change:UP host:[10 128 191 122] port:9042}
2023/01/11 20:59:35 gocql: Session.handleNodeUp: 10.128.191.122:9042
```
There is no event dispatched for the `NEW_NODE` event and it jumps straight to processing the `UP` event, so the new node is never added to session ring. If all nodes in the cluster are replaced in this manner, the eventual outcome is that clients lose connection to the cluster and begin outputting `gocql: no hosts available in the pool`.

After bisecting recent commits, I found[ this PR](https://github.com/gocql/gocql/pull/1632/files#top) introduced the bug into the client. It looks like there was seemingly a [fail-safe](https://github.com/gocql/gocql/pull/1632/files#diff-6515f1a2f64d91949b3a8dbf406c260e24e551fb470fc9d520cf88d1b05c8090L271) for when the `NEW_NODE` event was missed, but this function was removed in the previously mentioned PR.

This change proposes to refresh the hostSource ring on `UP` events when the host cannot be found in the ring. This ensures that the hostMap stays up to date even if `NEW_NODE` events are not processed.

This should fix https://github.com/gocql/gocql/issues/1668, https://github.com/gocql/gocql/issues/1667, and https://github.com/gocql/gocql/issues/1582

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -132,3 +132,4 @@ Stefan Miklosovic <smiklosovic@apache.org>
 Adam Burk <amburk@gmail.com>
 Valerii Ponomarov <kiparis.kh@gmail.com>
 Neal Turett <neal.turett@datadoghq.com>
+Steven Seidman <steven.seidman@datadoghq.com>
--- a/events.go
+++ b/events.go
@@ -264,6 +264,9 @@ func (s *Session) handleNodeUp(eventIp net.IP, eventPort int) {
 
 	host, ok := s.ring.getHostByIP(eventIp.String())
 	if !ok {
+		if err := s.hostSource.refreshRing(); err != nil && gocqlDebug {
+			s.logger.Printf("gocql: Session.handleNodeUp: failed to refresh ring: %w\n", err.Error())
+		}
 		return
 	}
 


sseidman:
To be clear, this is occurs when multiple frames are passed to `handleNodeEvent` [here](https://github.com/gocql/gocql/blob/master/events.go#L132). The `NEW_NODE` event can get overwritten by the `UP` event depending on what order they are processed:
```
2023/01/20 19:55:50 debug: begin frame processing loop
2023/01/20 19:55:50 debug: processing frame
2023/01/20 19:55:50 debug: topology frame event. change: NEW_NODE host: 10.131.0.96
2023/01/20 19:55:50 debug: host not duplicate, adding event to hostEvent map
2023/01/20 19:55:50 debug: setting event change type
2023/01/20 19:55:50 debug: processing frame
2023/01/20 19:55:50 debug: status frame event. change: UP host: 10.131.0.96
2023/01/20 19:55:50 debug: setting event change type
2023/01/20 19:55:50 debug: end frame processing loop
```

sseidman:
any thoughts on this?

mikelococo:
We're now running these changes on our sizeable production infrastructure, having built them off a private fork.

This issue doesn't just affect folks who cycle ip's on Cassandra restart, it also prevents node-rotation. Our IPs are stable across restarts, but we regularly perform vertical scaling operations with `-Dcassandra.replace_address_first_boot=<dead_node_ip>` where every Cassandra node is replaced in sequence with bigger hardware running on a new IP address. This operation was regularly causing serving impacts where the gocql clients would fail to re-establish connections to the new nodes and eventually resulting first in hotspots and then in complete loss of connectivity to the cluster. It appears that a growing number of folks are impacted by this issue.

We (sseidman, myself, and the folks we work with) are happy to do any reasonable amount of work required to make this fix (or a better designed fix, if the maintainers think this is a non-optimal approach) upstreamable. Whether that's more tests, docs, testing alternate approaches... I think there's significant value in addressing this bug and we're motivated to try to get a fix merged if there's a world where that can happen.

lwimmer:
#915 seems to be related too. Do you think this will fix it too?

martin-sucha:
Hi! Looks good to me. Sorry for the delay.

What do you guys think of merging #1680 instead? Would that one work for you? Seems that it handles more cases.

PR #1675, opened 2023-02-13
JSONB and KeyspaceMetadata fixes for Yugabyte

jameshartig:
* Add JSONB column support
* Fallback in metadata functions to fix KeyspaceMetadata
* Added yugabyte tests

Ran the latest tests against yb 2.16.1.0-b50

I didn't include them in `all` and I'm not sure how we want to automate testing going forward.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,8 +4,9 @@ on:
   push:
     branches:
       - master
+      - ci-test
   pull_request:
-    types: [ opened, synchronize, reopened ]
+    types: [opened, synchronize, reopened]
 
 env:
   CCM_VERSION: "6e71061146f7ae67b84ccd2b1d90d7319b640e4c"
@@ -16,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [ '1.19', '1.20' ]
+        go: ["1.19", "1.20"]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v4
@@ -32,7 +33,6 @@ jobs:
     name: Integration Tests
     runs-on: ubuntu-latest
     strategy:
-      fail-fast: false
       matrix:
         go: [ '1.19', '1.20' ]
         cassandra_version: [ '4.0.8', '4.1.1' ]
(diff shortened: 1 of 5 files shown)


jameshartig:
@martin-sucha can you take a look at this? Thanks!

martin-sucha:
Hi! Previously, @Zariel was against including ScyllaDB-specific code to this repository (see the Scylla pull requests), including Yugabyte-specific code seems like the same category of issue, so I guess we should agree first whether to include such code or not. It seems we should not prefer one DB over the other.

Personally I don't mind including database-specific code to the gocql/gocql repository, but there needs to be a clear way how to maintain it and test it. Ideally we should run integration tests in CI automatically.

Alternately, is there a way to include an extension mechanism so the required functionality could be provided by an external package instead?

jameshartig:
@Zariel would appreciate your thoughts as well

martin-sucha:
> I definitely understand the concern. I think it makes sense to include changes per-database provided they aren't huge.

Right, it seems similar to handling different Cassandra versions. The question is where is the line :)

> I'm definitely fine maintaining the YB-specific code since we use it in production along with this repo.

Help with maintaining is always appreciated. Maybe we could even increase the bus factor here if you are willing to help with maintaining even the non-YB-specific code. What do you think?

> One issue I see is that I called the value TypeJsonb but maybe should use a more specific name like TypeYBJsonb.

Yes, it seems that the YB prefix might be useful as the JSONB type is an extension of the cql protocol. Or maybe we could have a `yugabyte` package with YB-specific definitions? Although having a separate package just for a single definition seems like an overkill.

I was also thinking what would happen in case a different database uses the same type number for a different type. Should we check that the database is yugabyte before checking for the 0x0080 value or it is unlikely to happen in practice? Or maybe we should update the upstream Cassandra CQL documentation and mention the 0x0080 value there?

> I added a specific test file that can use the YB docker image to run tests and I confirmed that they pass. What's the path to adding that to CI?

Currently the CI is broken because Ubuntu 18.04 machine is no longer available. I tried some upgrades in https://github.com/gocql/gocql/pull/1685 but it seems that the upgrading is not enough. I think it might be better to just write the Docker setup in Go from scratch than rely on ccm, see https://github.com/gocql/gocql/issues/1686.

> The type might be a place where we could expose a RegisterType function that adds a custom type but it would need to handle registering functions as well for unmarshaling and marshaling. Also anyone using YB would need to copy/paste some code to handle that specific type which seems not ideal.

How is Jsonb type usually used? Do you unmarshal it to strings, []byte etc?

Would a type like

```go
package yugabyte

import "errors"

const TypeJsonb = gocql.Type(0x0080)

var ErrNotJsonb = errors.New("type is not jsonb")

type Jsonb []byte

func (jb *Jsonb) UnmarshalCQL(info gocql.TypeInfo, data []byte) error {
	if info.Type() != TypeJsonb {
		return ErrNotJsonb
	}
	if data != nil {
		*jb = append((*v)[:0], data...)
	} else {
		*jb = nil
	}
}
```

work or is it too restrictive? The downside would be that the users would have to use type in all structs and it won't automatically work with maps.

In any case it seems registering types with current (un)marshal interfaces is not straightforward. Global registration would work but I'd rather not add more global state. Per-session registration would need a different MarshalCQL/UnmarshalCQL signature that gets a callback function to unmarshal sub-values, since the usual implementations currently call the global gocql.Marshal/gocql.Unmarhal. Or we'd need to hack it and pass the type map inside gocql.TypeInfo.

> The metadata issues are just a quirk from YB and I don't think an extension could work for that. I also expect it to be fixed on YB's end eventually.

Yep, makes sense.

jameshartig:
@martin-sucha So the biggest issue I ran into when trying to remove the yugabyte-specific type and instead have some opt-in "extension" is the package-level `Unmarshal` and `Marshal` methods. Any ideas what I could do there? The 2 things I was thinking were:
1. `*Session` would have `Unmarshal` and `Marshal` methods and we'd deprecate the top-level ones. Any unmarshalling or marshalling orginating from the session would use the session-level functions.
2. Make `YugabyteUnmarshal` and `YugabyteMarshal` methods that you'd have to use instead if you're using Yugabyte but I doubt many people will realize that until they ran into an error and searched why they got that error. This would also pollute the package-level functions.

jameshartig:
Yugabyte merged a commit to bring the system_schema tables in alignment which means we don't need to do a bunch of this work. https://github.com/yugabyte/yugabyte-db/commit/7fc3b761dddadfe5ab330478de1faba897ad0003

We still need to handle the custom jsonb column. I'm going to see if there's a way for a library user to add a custom type.

jameshartig:
Superseded by https://github.com/apache/cassandra-gocql-driver/pull/1855

PR #1680, opened 2023-03-13
Always refresh ring on topology changes and reconnections

joao-r-reis:
Currently, there are some cases where gocql does not refresh the ring during a control connection reconnection. If topology changes are happening on the cluster during the reconnection, some `NEW_NODE` and `REMOVED_NODE` protocol events might be missed. 

To fix this, every successful reconnection on the control connection should trigger a ring refresh operation. This PR adds that.

Also, this PR fixes an issue where `NEW_NODE` events can be overwritten by `UP` events and cause certain nodes to not be added to the ring. This is an issue already described in #1669 but the fix proposed by this PR is a bit different than one outlined on that PR. This PR causes any topology event to trigger a ring refresh regardless of how many nodes were removed or added. This eliminates the need to have duplicated logic around adding and removing nodes in both `ringDescriber` and `Session`.

**Note**: this is not an implementation of #1681.

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -670,50 +670,6 @@ func (r *ringDescriber) GetHosts() ([]*HostInfo, string, error) {
 	return hosts, partitioner, nil
 }
 
-// Given an ip/port return HostInfo for the specified ip/port
-func (r *ringDescriber) getHostInfo(hostID UUID) (*HostInfo, error) {
-	var host *HostInfo
-	for _, table := range []string{"system.peers", "system.local"} {
-		iter := r.session.control.withConnHost(func(ch *connHost) *Iter {
-			if ch.host.HostID() == hostID.String() {
-				host = ch.host
-				return nil
-			}
-
-			if table == "system.peers" {
-				return ch.conn.querySystemPeers(context.TODO(), ch.host.version)
-			} else {
-				return ch.conn.query(context.TODO(), fmt.Sprintf("SELECT * FROM %s", table))
-			}
-		})
-
-		if iter != nil {
-			rows, err := iter.SliceMap()
-			if err != nil {
-				return nil, err
-			}
-
-			for _, row := range rows {
-				h, err := r.session.hostInfoFromMap(row, &HostInfo{port: r.session.cfg.Port})
-				if err != nil {
-					return nil, err
-				}
-
-				if h.HostID() == hostID.String() {
-					host = h
-					break
-				}
-			}
-		}
-	}
-
-	if host == nil {
-		return nil, errors.New("unable to fetch host info: invalid control connection")
-	}
-
-	return host, nil
-}
-
 func (r *ringDescriber) refreshRing() error {
 	// if we have 0 hosts this will return the previous list of hosts to
 	// attempt to reconnect to the cluster otherwise we would never find
--- a/host_source.go
+++ b/host_source.go
@@ -646,7 +646,7 @@ func isValidPeer(host *HostInfo) bool {
 		len(host.tokens) == 0)
 }
 
-// Return a list of hosts the cluster knows about
+// GetHosts returns a list of hosts found via queries to system.local and system.peers
 func (r *ringDescriber) GetHosts() ([]*HostInfo, string, error) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
@@ -670,11 +670,22 @@ func (r *ringDescriber) GetHosts() ([]*HostInfo, string, error) {
 	return hosts, partitioner, nil
 }
 
-func (r *ringDescriber) refreshRing() error {
-	// if we have 0 hosts this will return the previous list of hosts to
-	// attempt to reconnect to the cluster otherwise we would never find
-	// downed hosts again, could possibly have an optimisation to only
-	// try to add new hosts if GetHosts didn't error and the hosts didn't change.
+// debounceRingRefresh submits a ring refresh request to the ring refresh debouncer.
+func (s *Session) debounceRingRefresh() {
+	s.ringRefresher.debounce()
+}
+
+// refreshRing executes a ring refresh immediately and cancels pending debounce ring refresh requests.
+func (s *Session) refreshRing() error {
+	err, ok := <-s.ringRefresher.refreshNow()
+	if !ok {
+		return errors.New("could not refresh ring because stop was requested")
+	}
+
+	return err
+}
+
+func refreshRing(r *ringDescriber) error {
 	hosts, partitioner, err := r.GetHosts()
 	if err != nil {
 		return err
@@ -682,7 +693,6 @@ func (r *ringDescriber) refreshRing() error {
 
 	prevHosts := r.session.ring.currentHosts()
 
-	// TODO: move this to session
 	for _, h := range hosts {
 		if r.session.cfg.filterHost(h) {
 			continue
@@ -714,9 +724,6 @@ func (r *ringDescriber) refreshRing() error {
 		delete(prevHosts, h.HostID())
 	}
 
-	// TODO(zariel): it may be worth having a mutex covering the overall ring state
-	// in a session so that everything sees a consistent state. Becuase as is today
-	// events can come in and due to ordering an UP host could be removed from the cluster
 	for _, host := range prevHosts {
 		r.session.removeHost(host)
 	}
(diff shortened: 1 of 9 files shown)


martin-sucha:
Hi! Thanks for the pull request.

Did you mean to put removal of `addNewNode` to the [remove unused code](https://github.com/gocql/gocql/pull/1680/commits/fff05239a6593369c4da5cd90ec6970067587e6a) commit instead of `handleNewNode` and `handleRemovedNode`?

Could you please update the description of commit https://github.com/gocql/gocql/pull/1680/commits/326072be085e740a48f545521301691f99261e2d to what the description of the merge request says? `refresh ring improvements` is a bit vague. The commit title could be something like  `Always refresh ring on topology change`. That should make it easier to find the commit in commit log and see the more details with `git blame`.

It seems gocql currently does not de-bounce the ring refresh, so always refreshing the ring could theoretically cause a lot of loads if events come in quick succession. I suspect this is not a problem in practice, but we can still add the de-bounce (maybe in a separate pull request?) to guard against that.

Another thing that I noticed while reviewing the code is that `refreshRing` loads data from `system.peers`, but seems to not use data from `system.local`. That happens only in `addNewNode` resp. `ringDescriber.getHostInfo`, which is not used in current master. We should read from system.local in ringDescriber.GetHosts as well.

Otherwise this looks good to me.

It would be great if you could add a test to make sure that we don't introduce the same issues in the future.

joao-r-reis:
Thanks for the quick review. I've changed the commits a bit so they should be clearer now.

> Did you mean to put removal of addNewNode to the [remove unused code](https://github.com/gocql/gocql/pull/1680/commits/fff05239a6593369c4da5cd90ec6970067587e6a) commit instead of handleNewNode and handleRemovedNode?

I removed `addNewNode` initially and then I found out that `handleNewNode` and `handleRemovedNode` were also unused so I removed these 2 later as well (now they are in the same commit).

> It seems gocql currently does not de-bounce the ring refresh, so always refreshing the ring could theoretically cause a lot of loads if events come in quick succession. I suspect this is not a problem in practice, but we can still add the de-bounce (maybe in a separate pull request?) to guard against that.

The topology change events are being debounced but I agree that it would probably be a good idea to debounce the ring refresh operations themselves. I'm happy to address this in a subsequent PR or I can just create a new issue so another contributor can pick it up, let me know what you prefer.

> Another thing that I noticed while reviewing the code is that refreshRing loads data from system.peers, but seems to not use data from system.local. That happens only in addNewNode resp. ringDescriber.getHostInfo, which is not used in current master. We should read from system.local in ringDescriber.GetHosts as well.

I'll start working on this and I'll also look into writing a test for this.

martin-sucha:
> > Did you mean to put removal of addNewNode to the [remove unused code](https://github.com/gocql/gocql/pull/1680/commits/fff05239a6593369c4da5cd90ec6970067587e6a) commit instead of handleNewNode and handleRemovedNode?
> 
> I removed `addNewNode` initially and then I found out that `handleNewNode` and `handleRemovedNode` were also unused so I removed these 2 later as well (now they are in the same commit).

Techically, `handleNewNode` and `handleRemovedNode` are used until https://github.com/gocql/gocql/pull/1680/commits/7b506bcb0421440ad23c02b53d830b35c7509b60 so the removal should be part of that commit.

> 
> > It seems gocql currently does not de-bounce the ring refresh, so always refreshing the ring could theoretically cause a lot of loads if events come in quick succession. I suspect this is not a problem in practice, but we can still add the de-bounce (maybe in a separate pull request?) to guard against that.
> 
> The topology change events are being debounced but I agree that it would probably be a good idea to debounce the ring refresh operations themselves. I'm happy to address this in a subsequent PR or I can just create a new issue so another contributor can pick it up, let me know what you prefer.

Feel free to send a pull request, that would be great.

> > Another thing that I noticed while reviewing the code is that `refreshRing` loads data from `system.peers`, but seems to not use data from `system.local`. That happens only in `addNewNode` resp. `ringDescriber.getHostInfo`, which is not used in current master. We should read from system.local in ringDescriber.GetHosts as well.
> 
> I'll start working on this and I'll also look into writing a test for this.

Thanks!

joao-r-reis:
I've rebased my branch and pushed commits to make `GetHosts()` retrieve the local host info from `system.local` and to remove `getHostInfo()` which was only used by `addNewNode()` and this function was removed in one of my earlier commits because it was also unused.

For the debouncer, I would think having an event debouncer would be enough and I see that `gocql` already has one, is there a need to have two debouncers? What would be the goal of the ring refresh debouncer?

StevenLacerda:
@martin-sucha Can you take a look at the merge request? This should help tremendously with K8's environments. Thanks!

joao-r-reis:
I've implemented a debouncer, can you take a look and see if this is aligned with what you were thinking? I'm going to look into tests now.

host_source.go
```diff
@@ -721,3 +732,151 @@ func (r *ringDescriber) refreshRing() error {
...
+		timer:         time.NewTimer(ringRefreshDebounceTime),
+		ringDescriber: r,
+	}
+	d.timer.Stop()
+	go d.flusher()
+	return d
+}
+
+func (d *ringRefreshDebouncer) debounce() {
+	d.timer.Reset(ringRefreshDebounceTime)
```
joao-r-reis, line 765: I might add a lock here as well, not sure if concurrent `Reset` and `Stop` calls can be problematic here so should probably lock just in case

martin-sucha:
> I've added unit tests but I don't see a way to simulate a topology change for an integration test, does the current test code support writing an integration test that could test whether the control connection updates the topology when it reconnects?

Yeah, the current integration tests use ccm and it would probably be hard to write an integration test that changes topology. We should probably use Go code directly instead of ccm to setup the cluster. That is a bigger chunk of work though. I've opened https://github.com/gocql/gocql/issues/1686 to track that.

> Oh I've also changed the host_source_test.go tests to run with the unit flag instead of cassandra flag since this file seems to contain unit tests only.

Nice catch. Thanks!

host_source.go
```diff
@@ -721,3 +732,160 @@ func (r *ringDescriber) refreshRing() error {
...
+	refreshNowCh chan struct{}
+	quit         chan struct{}
+	refreshFn    func() error
+}
+
+func newRefreshDebouncer(interval time.Duration, refreshFn func() error) *refreshDebouncer {
+	d := &refreshDebouncer{
+		stopped:      false,
+		broadcaster:  nil,
+		refreshNowCh: make(chan struct{}),
```
martin-sucha, line 756: It seems that refreshNowCh should have capacity one instead of zero. The sender does non-blocking send, so with capacity zero, there is a chance that the refresh signal is lost when there is no receiver (for example, while refreshFn is running or while errorBroadcaster.broadcast is running).

martin-sucha (changes requested):
Thank you for preparing the changes! Looks good to me in general, I've added some inline comments.

joao-r-reis:
Managed to add a ccm test that tests whether the control connection refreshes the ring on reconnect. I used `DisableTopologyEvents` config and a custom `HostFilter` to avoid using real topology changes.

I'll take a look at your review feedback now, thanks.

host_source.go
joao-r-reis: Oh nice catch, I missed this. I've changed this to have a capacity of 1.

host_source.go
```diff
@@ -721,3 +732,160 @@ func (r *ringDescriber) refreshRing() error {
...
+}
+
+func (d *refreshDebouncer) stop() {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	if d.stopped {
+		return
+	}
+	d.stopped = true
+	d.quit <- struct{}{}
```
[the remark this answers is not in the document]
joao-r-reis: I've tried to keep the `stop()` behavior as close as possible to the `eventDebouncer` but I didn't copy the comment that is on that one. I've copied that comment now so it is clear that we are syncing with the flusher goroutine.

host_source.go
```diff
@@ -721,3 +732,160 @@ func (r *ringDescriber) refreshRing() error {
...
+}
+
+func (d *refreshDebouncer) stop() {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	if d.stopped {
+		return
+	}
+	d.stopped = true
+	d.quit <- struct{}{} // sync with flusher
```
martin-sucha, line 840: It seems we need to unlock `d.mu` before waiting for flusher to exit, otherwise there could be a deadlock similar to #1687. If refresh is in progress, `stop()` waits until flusher completes, which waits for refresh to complete first. The refresh might take a long time and something could call `refreshNow` or `debounce` during that time, which would be blocked on acquiring `d.mu`.

martin-sucha (changes requested):
Thanks! I've noticed one more thing, added inline comment.

StevenLacerda:
@martin-sucha @joao-r-reis Is this ready to be reviewed and merged? Really looking forward to getting this out as it will help several of our customers.

joao-r-reis:
I'm waiting on a review and ready to implement any changes that are suggested. I've addressed all comments of the previous reviews.

martin-sucha (approved):

martin-sucha:
Looks good to me. Thank you!

martin-sucha:
It seems there is a data race in the ccm test. @joao-r-reis could you please check https://github.com/gocql/gocql/pull/1691 ?

StevenLacerda:
@joao-r-reis @martin-sucha Any updates? What's needed at this point?

joao-r-reis:
> @joao-r-reis @martin-sucha Any updates? What's needed at this point?

This has already been released with v1.4.0 AFAIK

Issue #1582, opened 2021-10-04
Rolling restart of k8s db cluster causes all nodes to remain down and not recover


justinfx:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?

N/A (Yugabyte 2.9.0)

### What version of Gocql are you using?

gocql@v0.0.0-20200602185649-ef3952a45ff4
(https://github.com/yugabyte/gocql)

### What version of Go are you using?

1.17.0

### What did you do?

Scale all database nodes in cluster down from 3 -> 0 and then back from 0  -> 3

### What did you expect to see?

I expected to see the gocql client log the connection failure for the 3 database node ip addresses, but then recover with 3 new ip addresses for the 3 new database pods. 

### What did you see instead?

The client continues to see the original 3 node ips as DOWN and never recovers with the new nodes.

---

refs #1575 , https://github.com/yugabyte/yugabyte-db/issues/10182

Because the forked client I am using is outdated, it is possible that this issue is resolved in a newer release of gocql, or that it is still outstanding as per #1575 

My connection host list look like the following:
```
yb-tserver-0.yb-tservers.db-namespace.svc.cluster.local
yb-tserver-1.yb-tservers.db-namespace.svc.cluster.local
yb-tserver-2.yb-tservers.db-namespace.svc.cluster.local
```

In k8s these are 3 unique ip addresses in the cluster. When I scale down and back up, the pods have new ip addresses.

I am attaching the client application logs, using the build tag `gocql_debug`, showing two different scenarios. 

The first test is where I only scale from 3 -> 2 and then back up from 2 -> 3.
This log shows that it does seem to handle the old IP being down but then the new IP becoming part of the ring again.
[cql_scale_1.log](https://github.com/gocql/gocql/files/7281409/cql_scale_1.log)

The second test is where I scale all 3 nodes ->0 and then back up to 3 again. I am using a policy that scales each node, one at a time. The log shows the nodes going down, but then never recover with any sign of the new nodes. 
[cql_scale_all.log](https://github.com/gocql/gocql/files/7281421/cql_scale_all.log)
 
If I manually kill the database pods, depending on the order I choose, I may or may not see this topology event:
```
{"level":"info","ts":"2021-10-04T21:14:26.465Z","caller":"gocql@v0.0.0-20200602185649-ef3952a45ff4/events.go:91","msg":"gocql: handling frame: [topology_change change=NEW_NODE host=10.42.9.48 port=9042]"}
```

[...]

PR #1682, opened 2023-03-20
Fix k8s reconnect

djschaap:
With existing behavior, when a Cassandra node changes its IP address (while keeping the same host ID), the gocql client never utilizes that new IP address.  In the event ALL nodes change their IP addresses (over the lifetime of the gocql client), the cluster connection will become unusable.  Attempts to query Cassandra will be met with `gocql: no hosts available in the pool` errors.

I am experiencing this behavior connecting to a Cassandra cluster running on Kubernetes.  When a rolling restart occurs (StatefulSet change or otherwise), many of my long-lived gocql clients will become unresponsive.  Restarting the affected processes is required to restore Cassandra connectivity.

With these changes in place, when a `NEW_NODE` message is received for an existing host (host ID) with a new IP address, 1) the existing host (`HostInfo` struct) is removed and 2) the new host (`HostInfo` struct) is added.

Personal testing has shown this fix to work (i.e., the client learns the each new node IP address and maintains connectivity to the Cassandra cluster); more thorough testing would be welcome.

This may address #915 and/or #1582.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -132,3 +132,4 @@ Stefan Miklosovic <smiklosovic@apache.org>
 Adam Burk <amburk@gmail.com>
 Valerii Ponomarov <kiparis.kh@gmail.com>
 Neal Turett <neal.turett@datadoghq.com>
+Doug Schaapveld <djschaap@gmail.com>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Learning a new IP address for an existing node (identified by host ID) now triggers replacement of that host.
+  This fixes some Kubernetes reconnection failures. (#1682)
+
 ## [1.3.1] - 2022-12-13
 
 ### Fixed
--- a/events.go
+++ b/events.go
@@ -129,6 +129,14 @@ func (s *Session) handleKeyspaceChange(keyspace, change string) {
 	s.policy.KeyspaceChanged(KeyspaceUpdateEvent{Keyspace: keyspace, Change: change})
 }
 
+// handleNodeEvent handles inbound status and topology change events.
+//
+// Within each category (topology vs status), events are debounced by
+// host IP; only the latest event is processed.
+//
+// Processing topology change events before status change events ensures
+// that a NEW_NODE event is not dropped in favor of a newer UP event (which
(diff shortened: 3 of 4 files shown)


martin-sucha:
Hello! Thank you for the pull request.

You have a very good point that `refreshRing` needs to handle re-adding the hosts, so I'll merge https://github.com/gocql/gocql/pull/1682/commits/8593ee86bc9f5fd9523c4d74637702d0e7621d17

As for the handling of events, please see #1680, what do you think about merging that pull request instead?

lwimmer:
Thank you!

I'be briefly tested this PR and it seems to fix the issue described in #915 for us.

I've also tested #1680 and #1669, which did not fix the issue.

djschaap:
@martin-sucha At first glance, I'm not thrilled with refreshing the ring on every topology change.  I share your fears re: thrashing.  It also feels bad/inefficient to disregard the actual information in a NEW_NODE message (i.e., knowledge of the host that is actually added) and trigger a full-scale refresh.

OTOH, doing a full ring refresh might help with consistency when the client misses a node add or remove event (if that ever occurs - I can't say).

Just to ask - would either approach interfere or fail when `DisableInitialHostLookup` or `DisableTopologyEvents` is true?  Are there any race/locking conditions that would be more likely to be triggered by doing a full ring refresh (vs a simple remove/add node)?

TL;DR I prefer my targeted approach to handling topology events, but #1680 looks like it would be sufficient for the scenarios I've considered.  I view both as functionally equivalent.

Bolodya1997:
Hi, in our project we are extremely interested in this fix.

Is there any plan for merging this and adding to the new release? If so, when approximately it may become available?

martin-sucha:
All right, I've merged this PR together with #1669 so that we can cut a bug fix release. It seems the change to update the existing host's IP is needed to fix #915. The arguments for refreshing the whole ring (with the debouncer) are compelling though, since it will reduce the amount of code that needs to be maintained. @joao-r-reis could you please rebase #1680 and add the debouncer? I think with 8593ee86bc9f5fd9523c4d74637702d0e7621d17 included #1680 should work as a fix to #915 too. We can release those changes in a subsequent release, that way even if someone encounters an issue with refreshing the ring always, they could fall back to a release that contains only #1682 and #1669.

martin-sucha:
Tagged [1.3.2](https://github.com/gocql/gocql/releases/tag/v1.3.2)

PR #1683, opened 2023-03-24
DialHost must connect to the requested host

laurovenancio:
When a hostname resolves to multiple hosts, multiple HostInfos are generated. DialHost must connect to the host received as parameter. If the hostname is used to establish the connection, the dns could resolve to another host.
The hostname should still be used to verify the TLS connection.


laurovenancio:
> Please also update AUTHORS.md as per [CONTRIBUTING.md](https://github.com/gocql/gocql/blob/master/CONTRIBUTING.md#minimum-requirement-checklist) and we can merge this.

@martin-sucha Rebased and authors updated.

Issue #1315, opened 2019-05-27
Data race in Query


rkuska:
### What version of Cassandra are you using?
Doesn't depend on Cassandra version

### What version of Gocql are you using?
scylladb/gocql@v1.0.4 but I believe the bug is present also in gocql/gocql

### What did you do?
Added speculative retry to the query object.

delay=80ms, attempts=2

### What did you expect to see?
Successful query execution.

### What did you see instead?
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa7a5dd]

goroutine 8600 [running]:
github.com/gocql/gocql.(*Query).getHostMetrics(0xc003c566c0, 0xc000818160, 0x40c829)
	/go/pkg/mod/github.com/scylladb/gocql@v1.0.4/session.go:715 +0x2d
github.com/gocql/gocql.(*Query).AddAttempts(0xc003c566c0, 0x1, 0xc000818160)
	/go/pkg/mod/github.com/scylladb/gocql@v1.0.4/session.go:749 +0x35
github.com/gocql/gocql.(*Query).attempt(0xc003c566c0, 0xc000038091, 0x7, 0xbf33095fb1e88efa, 0x1ce1983c6, 0x19d7da0, 0xbf33095facc564c8, 0x1c8f65974, 0x19d7da0, 0xc003ad4750, ...)
	/go/pkg/mod/github.com/scylladb/gocql@v1.0.4/session.go:890 +0x59
github.com/gocql/gocql.(*queryExecutor).attemptQuery(0xc0007167a0, 0x10cb120, 0xc000f6d100, 0x10d5000, 0xc003c566c0, 0xc000e6a100, 0xc0002f8e50)
	/go/pkg/mod/github.com/scylladb/gocql@v1.0.4/query_executor.go:32 +0x126
github.com/gocql/gocql.(*queryExecutor).do(0xc0007167a0, 0x10cb120, 0xc000f6d100, 0x10d5000, 0xc003c566c0, 0xe)
	/go/pkg/mod/github.com/scylladb/gocql@v1.0.4/query_executor.go:112 +0x1e4
github.com/gocql/gocql.(*queryExecutor).run(0xc0007167a0, 0x10cb120, 0xc000f6d100, 0x10d5000, 0xc003c566c0, 0xc00161ba40)
[...]

PR #1684, opened 2023-03-30
query_executor: Fixed race between Query.Release() and speculative executions

wprzytula:
As described in #1315, there is a race condition when there is more than one concurrent execution (by means of speculative execution) and `Query.Release()` is called after one of them completes. `Query.Release()` calls `Query.reset()`, which in turns zeroes the whole `Query`. Then, after another execution completes, it tries to access metrics, but they are already set to nil by the call to `Release()`, so a segfault is triggered.

The solution is quite simple: a ref counter is introduced into the `Query`. It is obviously initially set to 1. Every execution fiber (i.e. every execution goroutine) has the refcount atomically incremented before it is started, and decrements the refcount on completion. `Query.Release()` merely decrements the refcount. `Query.reset()` is called by whichever goroutine that decrements the refcount to 0.

Fixes: #1315

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -134,3 +134,4 @@ Valerii Ponomarov <kiparis.kh@gmail.com>
 Neal Turett <neal.turett@datadoghq.com>
 Doug Schaapveld <djschaap@gmail.com>
 Steven Seidman <steven.seidman@datadoghq.com>
+Wojciech Przytuła <wojciech.przytula@scylladb.com>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Deadlock in Session.Close(). (#1688)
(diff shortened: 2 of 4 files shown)


mykaul:
@martin-sucha - will you be able to review this change? TIA!

martin-sucha (changes requested):
Thanks for the pull request! Code looks good to me. Could you please also update AUTHORS and CHANGELOG.md as per https://github.com/gocql/gocql/blob/master/CONTRIBUTING.md#minimum-requirement-checklist ?

wprzytula:
> Thanks for the pull request! Code looks good to me. Could you please also update AUTHORS and CHANGELOG.md as per https://github.com/gocql/gocql/blob/master/CONTRIBUTING.md#minimum-requirement-checklist ?

Done.

wprzytula:
@martin-sucha ping

martin-sucha (approved):

martin-sucha:
Thanks! Merged :rocket:

PR #1685, opened 2023-04-04
Fix CI after ubuntu 18.04 stopped working

martin-sucha:
ubuntu-18.04 is not available anymore:

https://github.com/actions/runner-images/issues/6002

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   build:
     name: Unit tests
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         go: [ '1.19', '1.20' ]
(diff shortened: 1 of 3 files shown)


jameshartig:
@martin-sucha Made some more changes to enable materialized views in 4.x but one of the materialized views tests is still failing :/

martin-sucha:
> @martin-sucha Made some more changes to enable materialized views in 4.x but one of the materialized views tests is still failing :/

Nice! Thanks!

It seems that `chunk_length_in_kb` is different. The [default has changed from 64KiB to 16KiB in Cassandra 4.0](https://github.com/apache/cassandra/blob/976b8395f981c5956ef617ea90b483e8e15d576d/doc/modules/cassandra/pages/getting_started/production.adoc#compression). I've added a commit to change the expected value depending on version.

martin-sucha:
@jameshartig it works now, thanks for help! :tada:

Issue #1687, opened 2023-04-06
Deadlock when calling Session.Close() while control connection is reconnecting


joao-r-reis:
I ran into this while testing #1680 

Calling `Session.Close()` while the control connection is reconnecting can lead to a deadlock. 

This can be reproduced with the following integration test:

```
func TestControlConn_CloseDeadlock(t *testing.T) {
	sleep := atomic.Value{}
	sleep.Store(false)
	hostFilter := HostFilterFunc(func(host *HostInfo) bool {
		if sleep.Load().(bool) {
			// this function will be called for every host so to get a 10s sleep we can divide the 10s by number of hosts
			time.Sleep(time.Duration(10/len(getClusterHosts())) * time.Second)
		}
		return true
	})
	session := createSession(t, func(config *ClusterConfig) {
		config.HostFilter = hostFilter
	})
	sleep.Store(true)
	session.control.getConn().conn.Close()
	time.Sleep(7 * time.Second)
	session.Close()
}
```

After a few seconds if you dump goroutines you will see that one of them will be stuck here:

```
runtime.gopark(proc.go:364)
runtime.chansend(chan.go:259)
runtime.chansend1(chan.go:145)
github.com/gocql/gocql.(*controlConn).close(control.go:484)
github.com/gocql/gocql.(*Session).Close(session.go:477)
github.com/gocql/gocql.TestControlConn_CloseDeadlock(integration_test.go:269)
testing.tRunner(testing.go:1446)
testing.(*T).Run.func1(testing.go:1493)
runtime.goexit(asm_amd64.s:1594)
testing.(*T).Run(testing.go:1493)
```

And another one here:

```
runtime.gopark(proc.go:364)
runtime.goparkunlock(proc.go:369)
runtime.semacquire1(sema.go:150)
sync.runtime_SemacquireMutex(sema.go:77)
sync.(*RWMutex).RLock(rwmutex.go:71)
github.com/gocql/gocql.(*Session).initialized(session.go:501)
github.com/gocql/gocql.(*controlConn).setupConn(control.go:291)
github.com/gocql/gocql.(*controlConn).reconnect(control.go:371)
github.com/gocql/gocql.(*controlConn).heartBeat(control.go:101)
github.com/gocql/gocql.(*controlConn).connect.func1(control.go:260)
runtime.goexit(asm_amd64.s:1594)
github.com/gocql/gocql.(*controlConn).connect(control.go:260)
```

[...]

PR #1688, opened 2023-04-06
Fix a deadlock in Session.Close() during control connection reconnect

martin-sucha:
We switched from separate mutex for closing to sessionStateMu in 312a614907c8d2e19f6b14aff16015436cb78423.
This change introduced a deadlock.

We don't need to hold the mutex for the whole duration of Close(), we only need to update the status atomically.
Previously IsClosed() returned true only after all closing is done (because of the deferred unlock).
We can emulate that by setting isClosed at the end of Close(). We need a new variable to ensure that Close() is only executed once.

Fixes https://github.com/gocql/gocql/issues/1687

### DIFF
--- a/session.go
+++ b/session.go
@@ -73,8 +73,10 @@ type Session struct {
 
 	// sessionStateMu protects isClosed and isInitialized.
 	sessionStateMu sync.RWMutex
-	// isClosed is true once Session.Close is called.
+	// isClosed is true once Session.Close is finished.
 	isClosed bool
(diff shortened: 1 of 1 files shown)


joao-r-reis:
I added the test mentioned in #1687 to this branch (locally), ran it and it passed multiple times. It consistently gets stuck when I run it in `master` so this seems to fix the issue.

martin-sucha:
Thanks for checking! I've noticed missing unlock in the return branch, so I added it.

PR #1692, opened 2023-04-27
Support larger dates in TypeDate marshalling

olib963:
We are using nanosecond precision when reading a `TypeDate` into a `time.Time` value, this limits the dates we can accurately read and write from cassandra to `[1677-09-22, 2262-04-11]` since outside of this range we end up with integer overflow. Dates outside of this range encode correctly and are persisted in Cassandra, but are then read as an incorrect value.

The new supported date range is `[-5877641-06-23, 5881580-07-11]` given that we are decoding the epoch day as a 32 bit int (`-2147483648` = minimum int = `-5877641-06-23`). I think it's possible to expand the range if ever needed by decoding a 64 bit int and using `time.Unix(seconds, 0)`, but I didn't think it was worth looking into that change.

Please let me know if I missed anything from the testing.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -137,3 +137,4 @@ Steven Seidman <steven.seidman@datadoghq.com>
 Wojciech Przytuła <wojciech.przytula@scylladb.com>
 João Reis <joao.reis@datastax.com>
 Lauro Ramos Venancio <lauro.venancio@incognia.com>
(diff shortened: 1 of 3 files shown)


martin-sucha (approved):
Looks good to me. Thank you for the pull request!

PR #1702, opened 2023-06-12
conn: Advertise driver's name & version in STARTUP

wprzytula:
Advertising driver's name in the system.clients table can be helpful when debugging issues, e.g. when a connection imbalance occurs and allows to narrow down the culprit application/driver better.

The main module name is stored in a const in version.go.
The replaced module name and the version are fetched from the binary runtime information, where it comes from GitHub metadata (release tag).
Credits to @zimnx for this idea!

I've tested this manually using my fork. The version has shown correctly in system.clients.


wprzytula:
The failure in CI seems to be unrelated to this PR.

martin-sucha:
Looks good to me. Merged. Thanks!

PR #1708, opened 2023-06-22
Reresolve hostnames as fallback when all hosts are unreachable

wprzytula:
If all nodes in the cluster change their IPs at one time, driver used to no longer be able to ever contact the cluster; the only solution was to restart the driver. A fallback is added to the control connection `reconnect()` logic so that when no known host is reachable, all hostnames provided in ClusterConfig (initial contact points) are reresolved and control connection is attempted to be opened to any of them. If this succeeds, a metadata fetch is issued normally and the whole cluster is discovered with its new IPs.

For the cluster to correctly learn new IPs in case that nodes are accessible indirectly (e.g. through a proxy), that is, by translated address and not `rpc_address` or `broadcast_address`, the code introduced in #1682 is extended to remove and re-add a host also when its translated address changed (even when its internal address stays the same).

As a bonus, a misnamed variable `hostport` is renamed to a suitable
`hostaddr`.

### DIFF
--- a/control.go
+++ b/control.go
@@ -282,7 +282,7 @@ func (c *controlConn) setupConn(conn *Conn) error {
 	}
 
 	if err := c.registerEvents(conn); err != nil {
-		return err
+		return fmt.Errorf("register events: %v", err)
 	}
 
 	ch := &connHost{
@@ -347,6 +347,20 @@ func (c *controlConn) reconnect() {
 	}
 	defer atomic.StoreInt32(&c.reconnecting, 0)
 
+	conn, err := c.attemptReconnect()
+
+	if conn == nil {
+		c.session.logger.Printf("gocql: unable to reconnect control connection: %v\n", err)
+		return
+	}
+
+	err = c.session.refreshRing()
+	if err != nil {
+		c.session.logger.Printf("gocql: unable to refresh ring: %v\n", err)
+	}
+}
+
+func (c *controlConn) attemptReconnect() (*Conn, error) {
 	hosts := c.session.ring.allHosts()
 	hosts = shuffleHosts(hosts)
 
@@ -363,6 +377,25 @@ func (c *controlConn) reconnect() {
 		ch.conn.Close()
 	}
 
+	conn, err := c.attemptReconnectToAnyOfHosts(hosts)
+
+	if conn != nil {
+		return conn, err
+	}
+
+	c.session.logger.Printf("gocql: unable to connect to any ring node: %v\n", err)
+	c.session.logger.Printf("gocql: control falling back to initial contact points.\n")
+	// Fallback to initial contact points, as it may be the case that all known initialHosts
+	// changed their IPs while keeping the same hostname(s).
+	initialHosts, resolvErr := addrsToHosts(c.session.cfg.Hosts, c.session.cfg.Port, c.session.logger)
+	if resolvErr != nil {
+		return nil, fmt.Errorf("resolve contact points' hostnames: %v", resolvErr)
+	}
+
+	return c.attemptReconnectToAnyOfHosts(initialHosts)
+}
+
+func (c *controlConn) attemptReconnectToAnyOfHosts(hosts []*HostInfo) (*Conn, error) {
 	var conn *Conn
 	var err error
 	for _, host := range hosts {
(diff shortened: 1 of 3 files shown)


wprzytula:
@martin-sucha Could you please take a look?

control.go
```diff
@@ -363,6 +363,37 @@ func (c *controlConn) reconnect() {
...
+		initial_hosts, resolvErr := addrsToHosts(c.session.cfg.Hosts, c.session.pool.port, c.session.logger)
+		if resolvErr == nil {
+			conn, err = c.attemptReconnectToAnyOfHosts(initial_hosts)
+		} else {
+			c.session.logger.Printf("gocql: error on resolving contact points' hostnames: %v\n", err)
+		}
+	}
+
+	if conn == nil {
+		c.session.logger.Printf("gocql: control unable to register events: %v\n", err)
```
martin-sucha, line 381: It seems to me that currently we can log an earlier error (from `conn, err := c.attemptReconnectToAnyOfHosts(hosts)`) later than an error from the resolving (since err is not updated), which might be confusing.

It seems that inverting the error handling logic of the function might make the flow clearer:

```go
func (c *controlConn) reconnect() {
	if atomic.LoadInt32(&c.state) == controlConnClosing {
		return
	}
	if !atomic.CompareAndSwapInt32(&c.reconnecting, 0, 1) {
		return
	}
	defer atomic.StoreInt32(&c.reconnecting, 0)

	conn, err := c.attemptReconnect()

	if conn == nil {
		c.session.logger.Printf("gocql: unable to reconnect control connection: %v\n", err)
		return
	}

	err = c.session.refreshRing()
	if err != nil {
		c.session.logger.Printf("gocql: unable to refresh ring: %v\n", err)
	}
}

func (c *controlConn) attemptReconnect() (*Conn, error) {
	hosts := c.session.ring.allHosts()
	hosts = shuffleHosts(hosts)

	// keep the old behavior of connecting to the old host first by moving it to
	// the front of the slice
	ch := c.getConn()
	if ch != nil {
		for i := range hosts {
			if hosts[i].Equal(ch.host) {
				hosts[0], hosts[i] = hosts[i], hosts[0]
				break
			}
		}
		ch.conn.Close()
	}

	conn, err := c.attemptReconnectToAnyOfHosts(hosts)

	if conn != nil {
		return conn, err
	}

	c.session.logger.Printf("gocql: unable to connect to any ring node: %v\n", err)
	c.session.logger.Printf("gocql: control falling back to initial contact points.\n")
	// Fallback to initial contact points, as it may be the case that all known initialHosts
	// changed their IPs while keeping the same hostname(s).
	initialHosts, resolvErr := addrsToHosts(c.session.cfg.Hosts, c.session.cfg.Port, c.session.logger)
	if resolvErr != nil {
		return nil, fmt.Errorf("resolve contact points' hostnames: %v", resolvErr)
	}

	return c.attemptReconnectToAnyOfHosts(initialHosts)
}
```

What do you think?

martin-sucha (commented):
@wprzytula thank you for the pull request! This is indeed a useful fix. I've added a few comments inline.

control.go
wprzytula: Definitely clearer. I'll refactor as you suggest.

wprzytula:
I've addressed comments and retested.

control.go
```diff
@@ -282,6 +282,7 @@ func (c *controlConn) setupConn(conn *Conn) error {
 	}
 
 	if err := c.registerEvents(conn); err != nil {
+		c.session.logger.Printf("gocql: control unable to register events: %v\n", err)
```
martin-sucha, line 285: We should either handle the error (log or otherwise) or return it, not both. Otherwise we'd be handling the same error in multiple places. In this case it seems to me that we should wrap the error like `return fmt.Errorf("register events: %v", err)` instead of logging here and let the caller handle it, as the setup of the control connection failed and we don't continue setupConn with unregistered events.

wprzytula:
@martin-sucha Done.

martin-sucha:
Merged. Thank you!

Would you be willing to contribute a test that would ensure that the function does not break in the future?

wprzytula:
> 

I'm afraid that automated testing of DNS changes is too much effort. The manual tests that I ran were complex: they involved:
- stopping systemd DNS service,
- running custom local DNS service that maps hostnames to "old" IPs,
- using a proxy on connections to all nodes, listening on "old" IPs,
- running a crafted test that periodically sends queries,
- breaking connections by stopping proxies,
- changing DNS rules to resolve to new IPs,
- reestablishing proxies on new IPs,
- waiting until all pools get populated again,
- asserting that it happens in reasonable time.

PR #1709, opened 2023-06-26
Use type-hinted-values to allow bound-values to be used outside of prepared-stmts

janmejay:
Allow use of bound-values outside of prepared-stmts.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -41,6 +41,12 @@ type Unmarshaler interface {
 	UnmarshalCQL(info TypeInfo, data []byte) error
 }
 
+// TypeHintedValue is a hinted value to power binding non-prepared stmts
+type TypeHintedValue struct {
+	Value interface{}
+	TInfo TypeInfo
+}
+
 // Marshal returns the CQL encoding of the value for the Cassandra
 // internal type described by the info parameter.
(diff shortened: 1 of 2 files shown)


marshal.go
```diff
@@ -41,6 +41,12 @@ type Unmarshaler interface {
 	UnmarshalCQL(info TypeInfo, data []byte) error
 }
 
+// TypeHintedValue is a hinted value to power binding non-prepared stmts
+type TypeHintedValue struct {
+	Value interface{}
+	TInfo TypeInfo
```
martin-sucha, line 47: It seems to me that it would be more readable if the field name was not abbreviated. For example `TypeInfo` or `Type`.

martin-sucha:
Hello!

First of all, what is the use case for this? It is not clear to me why it is needed. Please also update the commit message to  contain this information.

Secondly, if the use case is compelling, this pull request is adding a new API, so we'd need:

* Tests.
* An example.
* Description in the docs.

Also please update AUTHORS file, as described in [CONTRIBUTING.md](https://github.com/gocql/gocql/blob/master/CONTRIBUTING.md)

janmejay:
Sorry this was an accidental pull-request. It was meant for a different fork. Sorry about the confusion, please ignore this.

Issue #1621, opened 2022-05-10
Token aware host policy chooses primary replicas only


havaker:
If no keyspace is specified in ClusterConfig, then the token aware host policy has a degraded functionality.

### What version of Gocql are you using?

master

### What did you do?

I did not specify `.Keyspace` in `ClusterConfig` and used `gocql.TokenAwareHostPolicy(gocql.RoundRobinHostPolicy(), gocql.ShuffleReplicas())` as an host selection policy. I then run a few SELECT queries against a 3 node cluster.

```go
package main

import (
	"time"

	"github.com/gocql/gocql"
)

func main() {
	/* The example assumes the following CQL was used to setup the keyspace:
	   create keyspace example with replication = { 'class' : 'SimpleStrategy', 'replication_factor' : 3 };
	   create table example.tweet(timeline text, id UUID, text text, PRIMARY KEY(id));
	*/

	cluster := gocql.NewCluster("localhost:9042")

	//cluster.Keyspace = "example"

	cluster.PoolConfig.HostSelectionPolicy = gocql.TokenAwareHostPolicy(gocql.RoundRobinHostPolicy(), gocql.ShuffleReplicas())
	session, err := cluster.CreateSession()
	if err != nil {
		panic(err)
	}
	defer session.Close()

	someUUID := gocql.TimeUUID()
	for i := 0; i < 997; i++ {
		if err := session.Query(`SELECT id FROM example.tweet WHERE id = ?`, someUUID).Exec(); err != nil {
			panic(err)
		}

		time.Sleep(time.Second)
	}
}
```

### What did you expect to see?

I expected each of 3 nodes to receive some queries.

### What did you see instead?

Only the primary replica received the select queries (even though I was operating on a 3-node cluster, on a keyspace with replication factor 3 and passed `gocql.ShuffleReplicas()` to `TokenAwareHostPolicy)`.

### More

This issue persists if the `.Keyspace` field is not equal to a keyspace directly specified in a query string. For example:
```go
cluster.Keyspace = "example"

// later

session.Query(`SELECT id FROM differentkeyspacename.tweet WHERE id = ?`, someUUID).Exec(
```

The root of this problem is the way that `tokenAwareHostPolicy` manages replica maps for keyspaces.

[...]

PR #1714, opened 2023-07-11
Update Keyspace/Table name in prepared Query statement

sylwiaszunejko:
Previously  `gocql.TokenAwareHostPolicy` always used keyspace explicitly specified in `.Keyspace` in `ClusterConfig` regardless of the keyspace in the Query. This would mean that the token awareness feature would not work properly if the user did not specify `.Keyspace` in `ClusterConfig` or did a query to a different keyspace.

This PR introduces transferring Keyspace and Table names to the Query from the PREPARED response  and updating information about that every time this information is received. After this change, the token awareness feature can work properly without user having to specify a keyspace in `ClusterConfig`.

I created automated integration test ( `keyspace_table_test.go`) on 3 nodes cluster. 

Fixes: #1621

### DIFF
--- a/keyspace_table_test.go
+++ b/keyspace_table_test.go
@@ -1,0 +1,81 @@
+//go:build all || integration
+// +build all integration
+
+package gocql
+
+import (
+	"context"
+	"fmt"
+	"testing"
+)
+
+// Keyspace_table checks if Query.Keyspace() is updated based on prepared statement
+func TestKeyspaceTable(t *testing.T) {
+	cluster := createCluster()
+
+	fallback := RoundRobinHostPolicy()
+	cluster.PoolConfig.HostSelectionPolicy = TokenAwareHostPolicy(fallback)
+
+	session, err := cluster.CreateSession()
+	if err != nil {
+		t.Fatal("createSession:", err)
+	}
+
+	cluster.Keyspace = "wrong_keyspace"
+
+	keyspace := "test1"
+	table := "table1"
+
+	err = createTable(session, `DROP KEYSPACE IF EXISTS `+keyspace)
+	if err != nil {
+		t.Fatal("unable to drop keyspace:", err)
+	}
+
+	err = createTable(session, fmt.Sprintf(`CREATE KEYSPACE %s
+	WITH replication = {
+		'class' : 'SimpleStrategy',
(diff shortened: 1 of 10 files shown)


example_keyspace_table_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// cluster.Keyspace was set to "test", but during prepering statement
+	// Keyspace in Query should be changed to "example"
+	if qry.Keyspace() != "example" {
+		log.Fatal("The keyspace does not match the contents of the query.")
+	}
+
+	fmt.Println("Keyspace:", qry.Keyspace(), qry.Table())
+
+	// Keyspace: example tweet
```
avelanarius, line 59: I don't get the purpose of this file. Is this an example to be ran by a user or a proper test? (if so, this should be a proper check rather than a `Println`)

sylwiaszunejko: I have changed it to automated integration test.

keyspace_table_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	if err := session.control.awaitSchemaAgreement(); err != nil {
+		t.Fatal(err)
+	}
+
+	ctx := context.Background()
+
+	fmt.Println()
+	fmt.Println("Insert row")
+	fmt.Println()
```
avelanarius, line 64: I don't think those (and other) `Println` are really neccessary for a automatic test.

avelanarius:
cc @martin-sucha

keyspace_table_test.go
sylwiaszunejko: Deleted

keyspace_table_test.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	cluster.Keyspace = "wrong_keyspace"
+
+	keyspace := "test1"
+	table := "table1"
+
+	err = createTable(session, `DROP KEYSPACE IF EXISTS `+keyspace)
+	if err != nil {
+		panic(fmt.Sprintf("unable to drop keyspace: %v", err))
```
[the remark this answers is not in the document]
sylwiaszunejko: Done

martin-sucha (approved):
Looks good to me. Thank you!

PR #1715, opened 2023-07-14
chore: use buf.String() instead of string(buf.Bytes())

testwill:


martin-sucha:
Hi!

Thank you for the pull request! The change looks good to me, but I have two questions:

- You have [Github vigilant mode](https://docs.github.com/en/authentication/managing-commit-signature-verification/displaying-verification-statuses-for-all-of-your-commits#about-vigilant-mode) enabled, but the commit is not signed. Is that intentional?
- It seems to me that we could use [strings.Builder](https://pkg.go.dev/strings#Builder) instead of `*bytes.Buffer`. That should save an allocation. What do you think?

PR #1719, opened 2023-08-28
Increase default timeouts

martin-sucha:
Client timeouts need to be higher than server timeouts, so that work does not accumulate on the server with retries.

This was not true by default, the gocql default timeout was lower than the Cassandra default timeout.

Closes https://github.com/gocql/gocql/issues/1671
Closes https://github.com/gocql/gocql/issues/1701


tarzanek (approved):
lgtm, vdaka Martin!

cluster.go
```diff
@@ -51,14 +51,28 @@ type ClusterConfig struct {
...
+	// Client Timeout should always be higher than the request timeouts configured on the server,
+	// so that retries don't overload the server.
+	// Timeout has a default value of 11 seconds, which is higher than default server timeout for most query types.
+	// Timeout is not applied to requests during initial connection setup, see ConnectTimeout.
 	Timeout time.Duration
 
-	// Initial connection timeout, used during initial dial to server (default: 600ms)
+	// ConnectTimeout limits the time spent during connection setup.
+	// During initial connection setup, internal queries, AUTH requests will return an error if the client
+	// does not receive a response within the Timeout period.
```
mykaul, line 66:
```suggestion
	// does not receive a response within the connectTimeout period.
```

martin-sucha: Fixed. Thanks!

PR #1722, opened 2023-09-08
If DisableInitialHostLookup, Update Hosts On First Ring Refresh

### LINKED ISSUE #1721: If DisableInitialHostLookup, Network Instability Can Cause Prolonged "no hosts available in pool"

Reproducible on 3.x & 4.x

v1.6.0

1.20

### What did you do?

If `DisableInitialHostLookup` is set to `true`, the ID for each
`gocql.HostInfo` is set to a random UUID.

https://github.com/gocql/gocql/blob/db6d5564dd6843cc08cc1d6c3642612adf94c618/session.go#L244

When `refreshRing(...)` is invoked, `system.peers` is queried,
returning the new list of Hosts. However, these `gocql.HostInfo` have
the ID set as the real `host_id` from `system.peers`.

Therefore, when `r.session.ring.addHostIfMissing(h)` is invoked, the
ID does not match any of the random UUIDs, resulting in `hostIPToUUID`
pointing to the new ID, but `hostList` appends.

https://github.com/gocql/gocql/blob/db6d5564dd6843cc08cc1d6c3642612adf94c618/ring.go#L95-L101

`addHostIfMissing(h)` returns `!ok`, resulting in
`r.session.startPoolFill(h)` being invoked for each Host, which in turn
calls `s.pool.addHost(host)` (which checks by Host ID), and
`s.policy.AddHost(host)`. The Policy's `AddHost` however uses a
`cowHostList`, which defines equality by the `ConnectAddress`, not the
`ID`!

https://github.com/gocql/gocql/blob/b9737ddcadbbe8092b27df4f6ab2e6e9f3cf4c72/host_source.go#L138-L145

This mismatch of checks (ID vs. IP/Addr) results in the Policy
de-duplicating the Host, but the Pool does not. This becomes a problem
if a network flap occurs again. One variant of this problem is
`queryExecutor.do(...)` has `hostIter()` only able to return `nil`,
because `q.policy`'s hosts are empty, whereas `q.pool`'s `hostConnPools`
contains the random UUID's entry, with `filling` repeatedly being set
to `true` on `reconnect()` attempts. This effectively starves the
ability to get an Up host, resulting in non-stop
`gocql: no hosts available in pool` until the session is re-created
or the application is restarted.

The workaround being used at the moment is to set `DisableInitialHostLookup`
to `false` and using `SingleHostReadyPolicy` (to reduce the connection
time that having `DisableInitialHostLookup=true` was originally achieving).

That when `DisableInitialHostLookup=true`, and a ring refresh occurs, the code correctly de-deduplicates.

### What did you see instead?

Remaining Pools associated with non-existent UUIDs, which causes side-effects on subsequent network flaps.

amcrn:
If DisableInitialHostLookup is enabled, the Host IDs are random UUIDs. Therefore, on the first ring refresh, remove the hosts and re-add them back and refill the pools.

Closes #1721

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -138,6 +139,10 @@ type HostInfo struct {
 func (h *HostInfo) Equal(host *HostInfo) bool {
 	if h == host {
 		// prevent rlock reentry
+		return true
+	}
+
+	if h.HostID() == host.HostID() {
 		return true
 	}
 
@@ -402,10 +407,10 @@ func (h *HostInfo) HostnameAndPort() string {
 }
 
 func (h *HostInfo) ConnectAddressAndPort() string {
-        h.mu.Lock()
-        defer h.mu.Unlock()
-        addr, _ := h.connectAddressLocked()
-        return net.JoinHostPort(addr.String(), strconv.Itoa(h.port))
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	addr, _ := h.connectAddressLocked()
+	return net.JoinHostPort(addr.String(), strconv.Itoa(h.port))
 }
 
 func (h *HostInfo) String() string {
@@ -423,10 +428,11 @@ func (h *HostInfo) String() string {
 
 // Polls system.peers at a specific interval to find new hosts
 type ringDescriber struct {
-	session         *Session
-	mu              sync.Mutex
-	prevHosts       []*HostInfo
-	prevPartitioner string
+	session          *Session
+	mu               sync.Mutex
+	prevHosts        []*HostInfo
+	prevPartitioner  string
+	firstRingRefresh int32
 }
 
 // Returns true if we are using system_schema.keyspaces instead of system.schema_keyspaces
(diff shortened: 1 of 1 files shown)


martin-sucha:
Hi! Thanks for the pull request!

In general the change looks good to me.

Could you please update AUTHORS file as per [CONTRIBUTING.md](https://github.com/gocql/gocql/blob/master/CONTRIBUTING.md#minimum-requirement-checklist)?

Also it seems that some tests are now failing. Could you please check them?

I wonder if we could also add a test case to ensure that #1721 does not re-occur later. Maybe something like `TestRingDiscovery`? :thinking:

amcrn:
@martin-sucha Alright, I believe the tests should be passing now, and I added an integration test

amcrn:
Hey there @martin-sucha, any chance you could take a look at this for merge? Cheers!

OleksiienkoMykyta (commented):
Looks good, but it should be updated, to fit in, resolve conflicts, and write changes to CHANGELOG.md. Also check the CONTRIBUTING.md to commit it properly.

OleksiienkoMykyta:
@joao-r-reis, could you please check it to?
I think it will help to solve the [issue 1721](https://github.com/apache/cassandra-gocql-driver/issues/1721) faster.

Issue #1723, opened 2023-09-14
`HostFilter` Not Used for Control Connection


amcrn:
### What version of Cassandra are you using?

Reproducible on 3.x & 4.x

### What version of Gocql are you using?

v1.6.0

### What version of Go are you using?

1.20

### What did you do?

The `HostFilter` is not being used for the Control Connection (aka
the `controlConn`). This can be problematic, especially if the host
is being filtered due to it being in a bad state, as it can hold up
the creation of the Control Connection (and therefore the Session
itself) for quite some time.

### What did you expect to see?

Hosts filtered by a `HostFilter` are not eligible to be used in the
Control Connection.

### What did you see instead?

The opposite.

OleksiienkoMykyta:
Thank you for raising the issue! I already reviewed the fix and reminded the maintainer to check it and push it forward.

jameshartig:
Two things:

1. Why are you passing an ineligible host to NewSession? Is it coming from a DNS lookup?
2. What is the HostFilter that you're applying?

I'd like to understand what you're trying to achieve before we make changes. I'm mostly curious why you aren't filtering the hosts ahead of time if your HostFilter is a simple IP-based one. Any complicated HostFilter will not work before we make a control connection.

PR #1724, opened 2023-09-14
HostFilter Not Used for Control Connection

amcrn:
The `HostFilter` is not being used for the Control Connection (aka the `controlConn`). This can be problematic, especially if the host is being filtered due to it being in a bad state, as it can hold up the creation of the Control Connection (and therefore the Session itself) for quite some time.

Closes: #1723

### DIFF
--- a/session.go
+++ b/session.go
@@ -198,6 +198,22 @@ func (s *Session) init() error {
 	}
 	s.ring.endpoints = hosts
 
+	// filter hosts at the onset to avoid using using filtered hosts in the control connection while
+	// discovering the protocol and/or connecting.
+	if s.cfg.HostFilter != nil {
+		newHosts := make([]*HostInfo, 0, len(hosts))
+		for _, host := range hosts {
+			if !s.cfg.filterHost(host) {
+				newHosts = append(newHosts, host)
+			}
+		}
+		if len(newHosts) > 0 {
+			hosts = newHosts
+		} else {
+			s.logger.Printf("gocql: hostfilter removed all hosts, defaulting to initial host list\n")
+		}
+	}
+
 	if !s.cfg.disableControlConn {
 		s.control = createControlConn(s)
 		if s.cfg.ProtoVersion == 0 {


session.go
```diff
@@ -198,6 +198,22 @@ func (s *Session) init() error {
 	}
 	s.ring.endpoints = hosts
 
+	// filter hosts at the onset to avoid using using filtered hosts in the control connection while
+	// discovering the protocol and/or connecting.
+	if s.cfg.HostFilter != nil {
+		newHosts := make([]*HostInfo, 0, len(hosts))
```
testisnullus, line 204: We can refactor to something like:
```
newHosts := hosts[:0]
```

to avoid allocating a new slice, but in general, it is a good suggestion to fix an issue.

testisnullus:
LGTM

OleksiienkoMykyta (commented):
LGTM. Could you please make it up to date, update CHANGELOG.md, and refactor commit log according to CONTRIBUTING.md, as an example you can check https://github.com/apache/cassandra-gocql-driver/commit/f35d4637473831d0f6fbd3182ad5958f62588e1f, or any other last merged PR.
@joao-r-reis, @jameshartig, please check it out.

jameshartig:
Unfortunately, I don't think this is the right way for this to be fixed. The problem is that most `HostFilter`'s expect the information about the host to be filled in (such as DataCenter, ConnectAddress, HostID), but in this case, so early in the process we don't have the correct, or really any information. I'm concerned that this will be pointless (because they'll all be filtered) and very confusing.

Right after we connect, we filter the hosts which seems about as good as we can do. https://github.com/apache/cassandra-gocql-driver/blob/bf16ec371974a1c2082d525b6d78fac108f8c49b/session.go#L247

> This also fixes an issue if disableControlConn=true and/or DisableInitialHostLookup=true, where the Ring is updated with the host, prior to filtering:

My understanding is that the ring should always contain all hosts in the cluster and that we should only be filtering before adding a host to the pool so I believe this is working as intended.

joao-r-reis:
> > This also fixes an issue if disableControlConn=true and/or DisableInitialHostLookup=true, where the Ring is updated with the host, prior to filtering:
> 
> My understanding is that the ring should always contain all hosts in the cluster and that we should only be filtering before adding a host to the pool so I believe this is working as intended.

There's a separate PR to address the issue with `DisableInitialHostLookup=true`, I posted this[ comment](https://github.com/apache/cassandra-gocql-driver/pull/1790#pullrequestreview-2429884740) there. Are your thoughts in line with what my suggestion there @jameshartig ?

PR #1726, opened 2023-09-26
Set the right value in custom property of Custom type

glutamatt:
i found https://github.com/gocql/gocql/issues/1053 

Custom type without any clue make the metadata unusable when dealing with udts

i get the complexity of udt types but at least, custom value for custom types could be set when custom is used as fallback for even for udts

after this patch , one can have information about the underlying custom type ( and build feature based on metadata )


glutamatt:
my new test without adding the custom value : 
image

PR #1747, opened 2024-05-02
MaxRequestsPerConn cluster param was added

testisnullus:


testisnullus:
@martin-sucha Hello, will be glad to receive your review on this PR!

conn.go
```diff
@@ -257,7 +257,7 @@ func (s *Session) dialWithoutObserver(ctx context.Context, host *HostInfo, cfg *
 		errorHandler:  errorHandler,
 		compressor:    cfg.Compressor,
 		session:       s,
-		streams:       streams.New(cfg.ProtoVersion),
+		streams:       s.streamIDGenerator(cfg.ProtoVersion),
```
taaraora, line 260: what if we call it `newStreamIDGenerator`?
Also is it obligatory to pass config there if we access cfg from attached type?

host_source.go
```diff
@@ -402,10 +402,10 @@ func (h *HostInfo) HostnameAndPort() string {
 }
 
 func (h *HostInfo) ConnectAddressAndPort() string {
-        h.mu.Lock()
-        defer h.mu.Unlock()
-        addr, _ := h.connectAddressLocked()
-        return net.JoinHostPort(addr.String(), strconv.Itoa(h.port))
+	h.mu.Lock()
```
taaraora, line 405: I don't think it makes sense to format files which were not touched by the fix.

conn.go
testisnullus: Fixed, renamed and moved the function to the `streams` package.

host_source.go
testisnullus: image

Reverted these changes, but it seems that they stayed inside this PR :\

taaraora (approved):
LGTM

PR #1756, opened 2024-05-29
Implemented token bucket rate limiter

### LINKED ISSUE #1731: too many (100k) concurrent queries produce errors

4.1.3 or 5.0

1.6.0

1.21.1

### What did you do?
- I start local cassandra in docker with `docker run --rm --name=cassandra -p 9042:9042 cassandra:4.1.3`
- I run this program - https://github.com/isopov/gocqltest/blob/main/main.go

I expect the program to finish successfully. 

### What did you see instead?
I see errors
```
gocql: no streams available on connection
gocql: no streams available on connection
```
or
```
```
Note: changing `workers` const to 50_000 allows the program to finish gracefully (increasing `queries` const is possible also - program will work minutes instead of seconds but still will finish gracefully with 50_000 workers)

RostislavPorohnya:
Implemented Queries Rate Limiter which uses Token Bucket algorithm to the Session struct. Added RateLimiterConfig to the Cluster struct. Closes #1731. 

Rate limiter is a feature similar to REQUEST_THROTTLER_CLASS feature implemented in the Java driver.

This PR adds functionality to configure rate limiter, which allows Cassandra cluster to sustain heavy loads from client code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added Queries Rate Limiter which uses Token Bucket algorithm to the Session struct. 
+  Added RateLimiterConfig to the ClusterConfig struct. (#1731)
 
 ### Changed
 
--- a/cluster.go
+++ b/cluster.go
@@ -259,6 +259,9 @@ type ClusterConfig struct {
 
 	// internal config for testing
 	disableControlConn bool
+
+	// If Session has RateLimiterConfig then queries will be limited using RateLimiter
+	RateLimiterConfig *RateLimiterConfig
 }
 
 type Dialer interface {
--- a/rate_limiter.go
+++ b/rate_limiter.go
@@ -1,0 +1,76 @@
+package gocql
+
+import (
+	"sync"
+	"time"
+)
+
+// RateLimiterConfig holds the configuration parameters for the rate limiter, which uses Token Bucket approach.
+//
+// Fields:
+//
+// - rate: Allowed requests per second
+// - Burst: Maximum number of burst requests
+//
+// Example:
+//	RateLimiterConfig{
+//		rate:  300000,
+//		burst: 150,
+//	}
+type RateLimiterConfig struct {
+	rate  float64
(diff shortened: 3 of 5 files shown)


martin-sucha:
It seems to me that the rate limiter should be an interface instead, so that a client that wants to use a different implementation could supply their own.

Specifically, we could use the [golang.org/x/time/rate implementation](https://pkg.go.dev/golang.org/x/time/rate).

The waiting should be done by the rate limiter implementation itself, i.e. Allow should be blocking, so that the wait does not need to be in increments of 50 milliseconds, but could wait e.g. for external events as well.

The rate limiter should take the context into account, so that when the context is canceled or a deadline is reached, the limiter will stop waiting.

The behavior in this pull request could be implemented outside of gocql by wrapping of Session.Query. That said a proper implementation should probably be rate limiting the query attempts (`queryExecutor.attemptQuery`) instead of the high-level Session.Query call. The `Session.Query` also does not handle batches and `attemptQuery` is used for batches too.

It seems that there could be an interceptor-like interface that would allow to wrap attemptQuery calls. This could be used to supply a rate limiter implementation as well as replace the Query/Batch observers in version 2 of the gocql package.

Something the following draft, but the interface would need some refinement probably:

```go
type QueryAttemptInterceptor interface {
    // AttemptQuery executes the query on the given connection.
    // The attempt function does the actual work, the AttempQuery implementation is free to execute code
    // before/after the call to the attempt function or skip the call altogether.
    // If error is nil, the returned *Iter must be non-nil.
    AttemptQuery(ctx context.Context, query ExecutableQuery, conn *Conn, attempt func(ctx context, query ExecutableQuery, conn *Conn) *Iter) (*Iter, error))
}
```

@testisnullus @joao-r-reis what do you think, should we open a separate issue for the interceptor-like interface?

joao-r-reis:
I agree that making this an interceptor-like interface would open up more use cases beyond rate limiting so it seems like a good idea to me 👍

testisnullus:
I think it is a great suggestion, we can create an issue for it

OleksiienkoMykyta:
@RostislavPorohnya, what is its status? It's a little unclear, should it be closed in favor of https://github.com/apache/cassandra-gocql-driver/pull/1820?

PR #1758, opened 2024-05-31
Fix cowHostList can't have hosts with same `ConnectAddress`

### LINKED ISSUE #1757: cowHostList uses outdated logic for host uniqueness

any

`v1.6.0`

`1.22`

### What did you do?
My cluster is hidden behind proxy, all nodes are mapped to one address, but different ports.
For this purpose i have use custom `AddressTranslator` that maps `broadcase_address` of nodes nodes to host+port of the proxy.

It works great, but unfortunately if one node goes down I begin experiencing queries returning `ErrNoConnections`.
Source of the problem is the following, when nodes are being added to the `roundRobinHostPolicy`:
https://github.com/gocql/gocql/blob/6d279ed01c1069d14420584f3f7d5bc1794cf804/policies.go#L331-L333

It uses `cowHostList.add` to do the job, which does deduplication via `HostInfo.Equal`:
https://github.com/gocql/gocql/blob/6d279ed01c1069d14420584f3f7d5bc1794cf804/policies.go#L50-L60
  
Which relies on `h.ConnectAddress().Equal(host.ConnectAddress())`:
https://github.com/gocql/gocql/blob/6d279ed01c1069d14420584f3f7d5bc1794cf804/host_source.go#L135-L142

As result I end-up with only one node in the `roundRobinHostPolicy.hosts` pool and if this node is one that does not work, then I get `ErrNoConnections` after first retry.

Everything works as supposed to

### What did you see instead?
`ErrNoConnections`

dkropachev:
`cowHostList` uses `HostInfo.Equal` to for host uniqueness, which relies on `ConnectAddress.Equal`, which does not allow to have different hosts with same `ConnectAddress`.

It breaks cases like accessing cluster via [tcp](https://github.com/gocql/gocql/issues/1757) or [ssl proxy](https://github.com/gocql/gocql/issues/1452) 

Closes #1757

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -141,7 +141,7 @@ func (h *HostInfo) Equal(host *HostInfo) bool {
 		return true
 	}
 
-	return h.ConnectAddress().Equal(host.ConnectAddress())
+	return h.HostID() == host.HostID() && h.ConnectAddressAndPort() == host.ConnectAddressAndPort()
 }
 
 func (h *HostInfo) Peer() net.IP {
@@ -402,10 +402,10 @@ func (h *HostInfo) HostnameAndPort() string {
 }
 
 func (h *HostInfo) ConnectAddressAndPort() string {
-        h.mu.Lock()
-        defer h.mu.Unlock()
-        addr, _ := h.connectAddressLocked()
-        return net.JoinHostPort(addr.String(), strconv.Itoa(h.port))
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	addr, _ := h.connectAddressLocked()
+	return net.JoinHostPort(addr.String(), strconv.Itoa(h.port))
 }
 
 func (h *HostInfo) String() string {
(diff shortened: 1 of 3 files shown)


dkropachev:
> LGTM. Please, make it up to date, make the notation to the CHANGELOG.md, and change the commit message according to the CONTRIBUTING.md. The commit message format should be:
> 
> ```
> <short description>
> 
> <reason why the change is needed>
> 
> Patch by <authors>; reviewed by <Reviewers> for CASSGO-#####
> ```
> 
> NIT: We shouldn't add code that we don't add/change, ConnectAddressAndPort() as an example in this case.
> 
> @jameshartig, @joao-r-reis, could we merge it after the update?

Done

joao-r-reis:
I don't want to increase the scope of the 2.0 release but I can add a 2.x fixversion to the JIRA if there is one

PR #1761, opened 2024-05-31
Fix broken public interfaces

dkropachev:
Currently private interface `token` directly or indirectly present in lot's of interfaces and therefore makes them unimplementable: `SelectedHost`, `ExecutableQuery`, `HostSelectionPolicy`, `NextHost`

To fix it we need to make all parts of interfaces be public


dkropachev:
Wrong repo

Issue #1187, opened 2018-09-13
CASSGO-7 Change Batch API to be consistent with Query()


Zariel:
Make Batch behave the same way as Query so it looks like

```go
session.Batch().Query("SELECT * FROM table").Query("SELECT * FROM table").Exec()
```

This means renaming session.NewBatch -> Batch and removing session.ExecuteBatch

joao-r-reis:
https://issues.apache.org/jira/browse/CASSGO-7

OleksiienkoMykyta:
Currently, it's marked as deprecated to keep backward compatibility in the merged PR(https://github.com/apache/cassandra-gocql-driver/pull/1764), but generally, we don't use it anymore.
Solved. The issue should be closed.

joao-r-reis:
Hmm so `session.ExecuteBatch()` still exists and it is not marked as deprecated. 

I've created CASSGO-57 to deprecate it and CASSGO-58 to remove it in the future.

This current issue (#1187 CASSGO-7) has been resolved already so closing this.

PR #1762, opened 2024-06-03
The gobal NewBatch function was removed

tengu-alt:
Issue: #1433

PR #1764, opened 2024-06-04
CASSGO-7 Change Batch API to be consistent with Query()

tengu-alt:
According to issue #1187, I have refactored the Query() method for the batch and implemented the Exec() method for the batch structure to save the general pattern of the query execution, and saved the older method (session.ExecuteBatch()) for backward compatibility. 
This means the batch for now behaves the same way as query.

### DIFF
--- a/session.go
+++ b/session.go
@@ -731,6 +731,13 @@ func (b *Batch) execute(ctx context.Context, conn *Conn) *Iter {
 	return conn.executeBatch(ctx, b)
 }
 
+// Exec executes a batch operation and returns nil if successful
+// otherwise an error is returned describing the failure.
+func (b *Batch) Exec() error {
+	iter := b.session.executeBatch(b)
+	return iter.Close()
+}
+
 func (s *Session) executeBatch(batch *Batch) *Iter {
 	// fail fast
 	if s.Closed() {
@@ -1748,7 +1755,14 @@ type Batch struct {
 }
 
 // NewBatch creates a new batch operation using defaults defined in the cluster
+//
+// Deprecated: use session.Batch instead
 func (s *Session) NewBatch(typ BatchType) *Batch {
+	return s.Batch(typ)
+}
+
+// Batch creates a new batch operation using defaults defined in the cluster
+func (s *Session) Batch(typ BatchType) *Batch {
 	s.mu.RLock()
 	batch := &Batch{
 		Type:             typ,
@@ -1848,8 +1862,9 @@ func (b *Batch) SpeculativeExecutionPolicy(sp SpeculativeExecutionPolicy) *Batch
 }
 
 // Query adds the query to the batch operation
-func (b *Batch) Query(stmt string, args ...interface{}) {
+func (b *Batch) Query(stmt string, args ...interface{}) *Batch {
 	b.Entries = append(b.Entries, BatchEntry{Stmt: stmt, Args: args})
+	return b
 }
 
 // Bind adds the query to the batch operation and correlates it with a binding callback
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Detailed description for NumConns (CASSGO-3)
 
+- Change Batch API to be consistent with Query() (CASSGO-7)
+
 ### Fixed
 
 - Retry policy now takes into account query idempotency (CASSGO-27)
--- a/batch_test.go
+++ b/batch_test.go
@@ -47,9 +47,9 @@ func TestBatch_Errors(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	b := session.NewBatch(LoggedBatch)
-	b.Query("SELECT * FROM batch_errors WHERE id=2 AND val=?", nil)
-	if err := session.ExecuteBatch(b); err == nil {
+	b := session.Batch(LoggedBatch)
+	b = b.Query("SELECT * FROM gocql_test.batch_errors WHERE id=2 AND val=?", nil)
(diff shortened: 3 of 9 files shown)


testisnullus (approved):

joao-r-reis:
@tengu-alt we should also add a new `Batch()` function and deprecate the existing `NewBatch` (basically a "soft" rename of `NewBatch()` -> `Batch()` without breaking users) and deprecate `Session.ExecuteBatch`. 

`NewBatch` and `ExecuteBatch` can then be removed in 3.0 so that users have more than enough time to transition to the new names.

tengu-alt:
> @tengu-alt we should also add a new `Batch()` function and deprecate the existing `NewBatch` (basically a "soft" rename of `NewBatch()` -> `Batch()` without breaking users) and deprecate `Session.ExecuteBatch`.
> 
> `NewBatch` and `ExecuteBatch` can then be removed in 3.0 so that users have more than enough time to transition to the new names.

@joao-r-reis I do not clearly understand. The `NewBatch()` function is already marked as `deprecated` and `session.NewBatch`  is recommended to use instead. Also I can't name the function same as struct (Batch) because the `Batch` is already declared.

joao-r-reis:
`Session.NewBatch()` should be deprecated now, global `NewBatch()` can be removed since it was deprecated already and `Session.Batch()` should be added to replace `Session.NewBatch()` (keep the same functionality). 

`Session.Batch()` does not have a conflict with the type `Batch`, you can have both (in fact this already exists with `Query()`)

The original issue mentions the rename `Session.NewBatch()` to `Session.Batch()` already, my only additional suggestion here is to keep and deprecate `Session.NewBatch()` for now AND add the new `Session.Batch()`.

joao-r-reis:
global `NewBatch()` removal is already handled by https://github.com/apache/cassandra-gocql-driver/pull/1762 so it doesn't have to happen here

tengu-alt:
> `Session.NewBatch()` should be deprecated now, global `NewBatch()` can be removed since it was deprecated already and `Session.Batch()` should be added to replace `Session.NewBatch()` (keep the same functionality).
> 
> `Session.Batch()` does not have a conflict with the type `Batch`, you can have both (in fact this already exists with `Query()`)
> 
> The original issue mentions the rename `Session.NewBatch()` to `Session.Batch()` already, my only additional suggestion here is to keep and deprecate `Session.NewBatch()` for now AND add the new `Session.Batch()`.

I see, thank you. Done, also I've changed the `Session.NewBatch()` usage in .doc and tests.

session.go
```diff
@@ -731,6 +731,11 @@ func (b *Batch) execute(ctx context.Context, conn *Conn) *Iter {
 	return conn.executeBatch(ctx, b)
 }
 
+func (b *Batch) Exec() error {
+	iter := b.session.executeBatch(b)
+	return iter.Close()
+}
```
worryg0d, line 739: I think you should add a doc here because it is meant to be used by users. Maybe use the same description as for `Session.ExecuteBatch()`.

Also, add a doc for `Session.Batch()` too.

tengu-alt: Agree, done.

session.go
```diff
@@ -1781,6 +1790,28 @@ func (s *Session) NewBatch(typ BatchType) *Batch {
...
+		observer:         s.batchObserver,
+		session:          s,
+		Cons:             s.cons,
+		defaultTimestamp: s.cfg.DefaultTimestamp,
+		keyspace:         s.cfg.Keyspace,
+		metrics:          &queryMetrics{m: make(map[string]*hostMetrics)},
+		spec:             &NonSpeculativeExecution{},
+		routingInfo:      &queryRoutingInfo{},
+	}
+
```
joao-r-reis: duplicate code (`NewBatch`)

joao-r-reis (commented):
LGTM, just 1 comment about the duplicate code in the new function

session.go
tengu-alt: I kept the `session.NewBatch` and marked it as deprecated according to your suggestion.

joao-r-reis: Yes thanks but please create a private function that both can call or make one of them call the other so we don't have the duplicate code.

tengu-alt: Got it! Fixed.

joao-r-reis (approved):

joao-r-reis:
Please format your commit message following the rules stated in [CONTRIBUTING.md](https://github.com/apache/cassandra-gocql-driver/blob/trunk/CONTRIBUTING.md#commit-message) and add this JIRA to the `CHANGELOG.md` in the `Changed` section of `Unreleased`, thanks!

joao-r-reis:
Sorry I merged another PR which is causing a git conflict on `CHANGELOG.md` can you fix it and I'll merge this right after? Otherwise I can merge manually and fix it in place

tengu-alt:
> Sorry I merged another PR which is causing a git conflict on `CHANGELOG.md` can you fix it and I'll merge this right after? Otherwise I can merge manually and fix it in place

Fixed! Thank you for the hint.

joao-r-reis:
@tengu-alt we forgot to deprecate session.ExecuteBatch(), do you have some free time to work on this? I created https://issues.apache.org/jira/browse/CASSGO-57 for it

joao-r-reis:
Oh and there's a few other session methods to execute batches that should also be moved to the Batch type

tengu-alt:
> @tengu-alt we forgot to deprecate session.ExecuteBatch()

Got It I'll take it to work.

> Oh and there's a few other session methods to execute batches that should also be moved to the Batch type

I'll mark them as deprecated and create similar methods for batch. (If I understand correctly - these methods: `ExecuteBatchCAS`, `MapExecuteBatchCAS` )

PR #1767, opened 2024-06-06
Fix deadlock in refresh debouncer stop

### LINKED ISSUE #1752: CASSGO-41 Deadlock in refreshDebouncer when reconnection fails
astra-classic

v1.6.0

1.21

### What did you do?
Connection errors, I think due to overload, lead to frequent reconnection attempts and failures

Should retry until connection succeeds

### What did you see instead?
Deadlock

```
498297 goroutine 1324045437 [chan send, 113 minutes]:
498298 github.com/gocql/gocql.(*refreshDebouncer).stop(0xc0b826a7c0)
498299         /go/pkg/mod/github.com/gocql/gocql@v1.6.0/host_source.go:848 +0x8c
498300 github.com/gocql/gocql.(*Session).Close(0xc03efb0c00)
498301         /go/pkg/mod/github.com/gocql/gocql@v1.6.0/session.go:494 +0x105
498302 github.com/gocql/gocql.NewSession({{0xc24be58930, 0x3, 0x3}, {0x2ef55cf, 0x5}, 0x4, 0x12a05f200, 0x12a05f200, 0x0, 0x755a, ...})
498303         /go/pkg/mod/github.com/gocql/gocql@v1.6.0/session.go:180 +0x98d
498304 github.com/gocql/gocql.(*ClusterConfig).CreateSession(...)
498305         /go/pkg/mod/github.com/gocql/gocql@v1.6.0/cluster.go:289
```

It looks like a race condition between `(*refreshDebouncer).stop()` and `(*refreshDebouncer).flusher()`
1. `stop()` acquires `d.mu` and sets `d.stopped` to true
2. `flusher()` exits the `select` at the top of the loop and blocks on acquiring `d.mu`
3. `stop()` releases `d.mu` and tries to write to `d.quit`
4. `flusher()` acquires `d.mu` and returns because `d.stopped` is true
5. `stop()` is deadlocked because `d.quit` is unbuffered and the reader has stopped

joao-r-reis:
fix #1752

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -747,25 +747,27 @@ const (
 // debounces requests to call a refresh function (currently used for ring refresh). It also supports triggering a refresh immediately.
 type refreshDebouncer struct {
 	mu           sync.Mutex
-	stopped      bool
 	broadcaster  *errorBroadcaster
 	interval     time.Duration
 	timer        *time.Timer
 	refreshNowCh chan struct{}
-	quit         chan struct{}
 	refreshFn    func() error
+	quitCtxFn    context.CancelFunc
+	quitCtx      context.Context
+	stoppedCtxFn context.CancelFunc
+	stoppedCtx   context.Context
 }
 
 func newRefreshDebouncer(interval time.Duration, refreshFn func() error) *refreshDebouncer {
 	d := &refreshDebouncer{
-		stopped:      false,
 		broadcaster:  nil,
 		refreshNowCh: make(chan struct{}, 1),
-		quit:         make(chan struct{}),
 		interval:     interval,
 		timer:        time.NewTimer(interval),
 		refreshFn:    refreshFn,
 	}
+	d.stoppedCtx, d.stoppedCtxFn = context.WithCancel(context.Background())
+	d.quitCtx, d.quitCtxFn = context.WithCancel(context.Background())
 	d.timer.Stop()
 	go d.flusher()
 	return d
@@ -775,7 +777,7 @@ func newRefreshDebouncer(interval time.Duration, refreshFn func() error) *refres
 func (d *refreshDebouncer) debounce() {
 	d.mu.Lock()
 	defer d.mu.Unlock()
-	if d.stopped {
+	if d.isStopped() {
 		return
 	}
 	d.timer.Reset(d.interval)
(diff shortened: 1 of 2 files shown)


host_source_test.go
```diff
@@ -274,6 +274,32 @@ func TestRefreshDebouncer_EventsAfterRefreshNow(t *testing.T) {
 	}
 }
 
+// https://github.com/gocql/gocql/issues/1752
+func TestRefreshDebouncer_DeadlockOnStop(t *testing.T) {
+	// there's no way to guarantee this bug manifests because it depends on which `case` is picked from the `select`
+	// with 4 iterations of this test the deadlock would be hit pretty consistently
```
kevinkyyro, line 313: Does this comment indicate that you were reliably reproducing it without the fix in place or is it just speculating that 4 should be enough to hit it?

joao-r-reis: I was reliably reproducing it with 4 without the fix in place

kevinkyyro (approved):
Not sure if my review helps get this merged, but I checked out the branch and verified that the test reliably reproduces the bug without the fix and passes with the fix. Thanks!

host_source.go
```diff
@@ -747,25 +747,27 @@ const (
...
 type refreshDebouncer struct {
 	mu           sync.Mutex
-	stopped      bool
 	broadcaster  *errorBroadcaster
 	interval     time.Duration
 	timer        *time.Timer
 	refreshNowCh chan struct{}
-	quit         chan struct{}
 	refreshFn    func() error
+	quitCtxFn    context.CancelFunc
```
jameshartig: I would say this is a bit of an abuse of contexts. Instead I would keep the old behavior but just change a small thing. Add another `chan struct{}` here called `done` that is `defer close(done)` in `flusher`. Then in `stop()` keep what was there but don't write to `quit` (I'm not really sure what the purpose of that was anyways) and just close it. Then do `<-done` to wait for `flusher` to stop.

jameshartig (changes requested):
I think what was there was pretty close just need to make some small changes.

host_source.go
joao-r-reis: Your solution is much simpler and more obvious, Idk why I overcomplicated things. Done

PR #1768, opened 2024-06-07
CASSGO-12 Remove deprecated features

tengu-alt:
This PR removes some features that are marked as "Deprecated".
Such as: 
* global var Timeout limit;
    `TimeoutLimit` was removed because it was always equal to 0 and the only condition where it was used was always false by 
    default. 

* `MustParseConsistency()` function;
    It was removed due to its deprecation.

* `New()` method for `typeInfo{}` interface;
    It was removed due to its deprecation. Instead of the removed method, the `typeInfo{}` interface has the 
    `NewWithError()` that returns an error instead of panicking.

* View field for the KeyspaceMetadata structure;
    Removed because of new and more informative `MaterializedViews` field.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -45,7 +45,7 @@ import (
 	"time"
 	"unicode"
 
-	inf "gopkg.in/inf.v0"
+	"gopkg.in/inf.v0"
 )
 
 func TestEmptyHosts(t *testing.T) {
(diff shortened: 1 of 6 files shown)


metadata.go
```diff
@@ -1001,7 +936,9 @@ func getMaterializedViewsMetadata(session *Session, keyspaceName string) ([]Mate
 			memtable_flush_period_in_ms,
 			min_index_interval,
 			read_repair_chance,
-			speculative_retry
+			speculative_retry,
+			field_names,
+			field_types
```
joao-r-reis: I don't think these columns exist

joao-r-reis (commented):
Did you run any tests on this? The tests can't even compile atm. Idk why CI wasn't triggered for this maybe push another commit to see if CI gets triggered.

metadata.go
tengu-alt: Good catch! fixed.

tengu-alt:
> Did you run any tests on this? The tests can't even compile atm. Idk why CI wasn't triggered for this maybe push another commit to see if CI gets triggered.

I've run the `metadata_test.go` and it passed and `cassandra_test.go` is ignored by the build.

joao-r-reis:
There's a couple of tests in `cassandra_test.go` that are failing.

tengu-alt:
> There's a couple of tests in `cassandra_test.go` that are failing.

Fixed.
Also I checked the CI job in my fork, and it's passed.

cassandra_test.go
```diff
@@ -2521,32 +2481,6 @@ func TestKeyspaceMetadata(t *testing.T) {
 		t.Fatalf("expected state function %s, but got %s", "avgstate", aggregate.StateFunc.Name)
 	}
 
-	_, found = keyspaceMetadata.Views["basicview"]
-	if !found {
-		t.Fatal("failed to find the view in metadata")
-	}
-	_, found = keyspaceMetadata.UserTypes["basicview"]
```
joao-r-reis, line 2602: UserTypes need to continue working, I don't think we want to remove this part of the test.

metadata.go
```diff
@@ -331,20 +315,12 @@ func compileMetadata(
 		aggregates[i].StateFunc = *keyspace.Functions[aggregates[i].stateFunc]
 		keyspace.Aggregates[aggregates[i].Name] = &aggregates[i]
 	}
-	keyspace.Views = make(map[string]*ViewMetadata, len(views))
-	for i := range views {
-		keyspace.Views[views[i].Name] = &views[i]
+	types := make([]UserTypeMetadata, len(materializedViews))
+	for i, mView := range materializedViews {
+		types[i].Keyspace = mView.Keyspace
+		types[i].Name = mView.Name
```
joao-r-reis: `Views` were being used to populate the `FieldNames` and `FieldTypes` fields of the `UserType` objects, we need to keep that functionality while removing the public `Views` property itself which is redundant.

As it stands with your changes a user is not able to get access to the field names and types of a UDT from `KeyspaceMetadata.UserTypes`.

tengu-alt: Good point! I've change the `getViewsMetadata` func, saving old functionality, but for `UserType`.

joao-r-reis (approved):

jameshartig (approved):

PR #1772, opened 2024-06-23
Donation to Apache Cassandra and ASF

michaelsembwever:
Reassignment of copyright, where approved, and relicensing.
Added file headers, reference to past license and copyright included.

Change phrases of "gocql team" to "the Cassandra project".

New repository will be github.com/apache/cassandra-gocql-driver

Most of this PR needs to be done in the source that is presented in the IP donation.
See https://incubator.apache.org/ip-clearance/cassandra-gocql-driver.html


cassandra_test.go
```diff
@@ -2913,7 +2936,7 @@ func TestDiscoverViaProxy(t *testing.T) {
 	// This (complicated) test tests that when the driver is given an initial host
 	// that is infact a proxy it discovers the rest of the ring behind the proxy
 	// and does not store the proxies address as a host in its connection pool.
-	// See https://github.com/gocql/gocql/issues/481
+	// See https://github.com/apache/cassandra-gocql-driver/issues/481
```
martin-sucha, line 2939: Do you plan to rename this github repository to `apache/cassandra-gocql-driver` (which should create a redirect to the new location) or create a new fork and archive this project? If fork, do you plan to transfer the list of issues?

I'd like to understand if new URL will work after the transfer.

go.mod
```diff
@@ -1,4 +1,27 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+//
+// Content before git sha 34fdeebefcbf183ed7f916f931aa0586fdaa1b40
+// Copyright (c) 2016, The Gocql authors,
+// provided under the BSD-3-Clause License.
+// See the NOTICE file distributed with this work for additional information.
+//
+module github.com/apache/cassandra-gocql-driver
```
martin-sucha, line 24: Changing the module path seems like a backwards-incompatible change to me. 

If we change the module path, users will be able to use `github.com/apache/cassandra-gocql/driver` import path and go.mod entry to include the driver in their code. However, existing code libraries (like scylladb/gocqlx) that depend on gocql would still depend on the original gocql and not the new one. It seems to me that it will not be possible to use a [replace](https://go.dev/ref/mod#go-mod-file-replace) directive to switch to the new version either, because

> Regardless of whether a replacement is specified with a local path or module path, if the replacement module has a go.mod file, its module directive must match the module path it replaces.

On the other hand, the new module path would allow us to remove deprecated stuff as we don't need to keep the backwards-compatibility.

If we don't change the module path, all users of the library would probably have to use the replace directive in go.mod, which is not ideal either. Not sure how it works if redirects are involved.

martin-sucha (commented):
Hi Mick, thank you for the pull request! I have added some inline comments.

cassandra_test.go
michaelsembwever: Once the donation is voted upon and accepted, we will then ask ASF Infra to move this github repository.  

The best approach, if possible, is to assign ASF Infra admin privileges to the repo and they will make the move (transfer organisation and rename project).  With this approach GitHub automatically handles redirects (both in browser and from git on the command line, and in tools).

go.mod
michaelsembwever: Thanks for catching this.

I didn't realise that changing these coordinates would be so impactful.  I think it wise to undo this change for now.   The project can discuss whether it wants to do it at a later point, e.g. in a new major version.

michaelsembwever:
Do we want the `v2.dev` branch donated ?

cassandra_test.go
martin-sucha: Thank you! That resolves my question.

go.mod
martin-sucha: In that case, all the import paths in `.go` files may need to stay as `github.com/gocql/gocql` as well.

I don't have experience moving a public Go repo, we should verify everything will work as expected on some testing repository before we setup the redirect on the main gocql repository.

michaelsembwever: Removing the imports 👍 

But what about the `Installation` in the README.md …?
```
    go get github.com/apache/cassandra-gocql-driver
```

martin-sucha: We might need to do some research on how best to approach this.

It depends on what the redirect would achieve. If using the current module path works with the redirect, installation would probably work with `go get github.com/gocql/gocql`. I don't know if it would work with `go get github.com/apache/cassandra-gocql-driver`.

As for the API documentation, it would probably stay at https://pkg.go.dev/github.com/gocql/gocql too. It seems that pkg.go.dev uses the import path, so for example the documentation from kiwicom/gocql is not available there.

martin-sucha:
> Do we want the `v2.dev` branch also donated ?

I am not sure if it would be useful, since it is 230 commits behind and probably would require extensive merge/rebase and it might be easier to implement similar changes from scratch. On the other hand, I guess there is no harm in donating it in case someone finds some of the changes useful for v2.

go.mod
michaelsembwever: For docs there's both https://godoc.org/github.com/apache/cassandra-gocql-driver and https://pkg.go.dev/github.com/apache/cassandra-gocql-driver

I'll restore them both back to `gocql/gocql` for now.

michaelsembwever:
> I am not sure if it would be useful, since it is 230 commits behind and probably would require extensive merge/rebase and it might be easier to implement similar changes from scratch. 

Given the work involved, and the existing CLAs and general agreement in place, i think it makes sense we leave the branch behind.  When someone does that work then they contribute it and it comes in, normally as future contributions will.

address_translators.go
```diff
@@ -1,3 +1,32 @@
...
+ * Copyright (c) 2016, The Gocql authors,
+ * provided under the BSD-3-Clause License.
+ * See the NOTICE file distributed with this work for additional information.
+ */
+/*
+ * Content before git sha 34fdeebefcbf183ed7f916f931aa0586fdaa1b40
+ * Copyright (c) 2016, The Gocql authors,
+ * provided under the BSD-3-Clause License.
+ * See the NOTICE file distributed with this work for additional information.
+ */
```
martin-sucha, line 29: This notice is twice in `address_translators.go`

cluster.go
```diff
@@ -2,6 +2,29 @@
 // Use of this source code is governed by a BSD-style
```
martin-sucha, line 2: Are these notices left in the files intentionally?

doc.go
```diff
@@ -357,4 +357,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*
+ * Content before git sha 34fdeebefcbf183ed7f916f931aa0586fdaa1b40
+ * Copyright (c) 2016, The Gocql authors,
+ * provided under the BSD-3-Clause License.
+ * See the NOTICE file distributed with this work for additional information.
+ */
```
martin-sucha, line 382: `go doc` includes any comments immediately before a package declaration (without any newlines in between) into the package documentation.

So the copyright notices should be above the package documentation and there should be at least one newline between the package documentation and the package declaration. (Or below the package declaration instead).

You can see the package documentation rendered as text by running the `go doc` command.

michaelsembwever: TIL.

fixed in all files.

cluster.go
michaelsembwever: nope.  removed. (and copyright back to 2012 on those files that had this.)

address_translators.go
michaelsembwever: fixed, thanks.

README.md
```diff
@@ -1,16 +1,16 @@
-gocql
+Apache Cassandra GoCQL Driver
 =====
 
-[![Join the chat at https://gitter.im/gocql/gocql](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/gocql/gocql?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-![go build](https://github.com/gocql/gocql/actions/workflows/main.yml/badge.svg)
-[![GoDoc](https://godoc.org/github.com/gocql/gocql?status.svg)](https://godoc.org/github.com/gocql/gocql)
+[!Join the chat at https://the-asf.slack.com/archives/C05LPRVNZV1](https://the-asf.slack.com/archives/C05LPRVNZV1)
+![go build](https://github.com/apache/cassandra-gocql-driver/actions/workflows/main.yml/badge.svg)
+[![GoDoc](https://godoc.org/github.com/apache/cassandra-gocql-driver?status.svg)](https://godoc.org/github.com/apache/cassandra-gocql-driver)
```
martin-sucha, line 6: This documentation link should point to the same URL as the one below (https://godoc.org/github.com/gocql/gocql)

Resp. we could replace all godoc link to pkg.go.dev, but that should perhaps be a separate commit/MR.

martin-sucha (approved):

PR #1773, opened 2024-06-24
Proto v5 support was added

testisnullus:
The pull request involves adding support for Protocol Version 5 (Proto v5) and addresses several key issues related to compatibility and the implementation of new features introduced in Cassandra's native protocol.

Key Issues and Changes in the PR:

- New Frame Format Support. The PR implements changes to support the new frame format introduced in Proto v5. This format includes modifications that are not backward-compatible with previous versions, requiring updates to the driver's handling of message frames to interpret and process the protocol's data streams correctly. Also, this PR added a logic to process the CRC calculations that ensures reliable communication between the driver and Cassandra nodes using Proto v5.

- Backward Compatibility. Ensuring backward compatibility with older protocol versions while implementing Proto v5 was a significant focus. These changes ensure that the driver works correctly with the latest Cassandra versions, particularly those utilizing the newer features and optimizations of Proto v5.


conn.go
```diff
@@ -660,6 +673,16 @@ func (c *Conn) heartBeat(ctx context.Context) {
 }
 
 func (c *Conn) recv(ctx context.Context) error {
+	// If native proto v5+ is used and conn is set up, then we should
+	// unwrap payload body from v5 compressed/uncompressed frame
+	if c.version > protoVersion4 && c.connReady {
```
lukasz-antoniak, line 678: Maybe revert the condition. If connection has not been initialised, we should not relay on protocol version.

conn.go
```diff
@@ -215,6 +216,14 @@ type Conn struct {
 	host            *HostInfo
 	isSchemaV2      bool
 
+	// Only for proto v5+.
+	// Indicates if Conn is ready to use Native Protocol V5.
```
lukasz-antoniak, line 220: This flag indicates that `STARTUP` message has been completed, not that we can use V5. Please fix the comment, because it can be misleading.

conn.go
```diff
@@ -1394,9 +1483,10 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
 		params.skipMeta = !(c.session.cfg.DisableSkipMetadata || qry.disableSkipMetadata)
 
 		frame = &writeExecuteFrame{
-			preparedID:    info.id,
-			params:        params,
-			customPayload: qry.customPayload,
+			preparedID:         info.id,
+			preparedMetadataID: info.metadataID,
```
lukasz-antoniak, line 1487: Rename the field to `result_metadata_id`. `preparedMetadataID` can be misleading.

conn.go
```diff
@@ -1394,9 +1483,10 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
...
+			params:             params,
```
lukasz-antoniak, line 1488: `resultRowsFrame` does not handle `new_metadata_id` and does not propagate it to next iterator calls.
```
<new_metadata_id> is [short bytes] representing the new, changed resultset
metadata. The new metadata ID must also be used in subsequent executions of
the corresponding prepared statement, if any.
```
This is needed to solve [CASSANDRA-10786](https://issues.apache.org/jira/browse/CASSANDRA-10786).

conn.go
```diff
@@ -1223,9 +1310,10 @@ type StreamObserverContext interface {
 }
 
 type preparedStatment struct {
-	id       []byte
-	request  preparedMetadata
-	response resultMetadata
+	id         []byte
+	metadataID []byte
```
lukasz-antoniak, line 1314: Please change the name to `result_metadata_id` as in the protocol not to create confusion.

crc.go
```diff
@@ -0,0 +1,57 @@
...
+
+var (
+	// Initial CRC32 bytes: 0xFA, 0x2D, 0x55, 0xCA
+	initialCRC32Bytes = []byte{0xfa, 0x2d, 0x55, 0xca}
+)
+
+// ChecksumIEEE calculates the CRC32 checksum of the given byte slice.
+func ChecksumIEEE(b []byte) uint32 {
+	crc := crc32.NewIEEE()
+	crc.Reset()
```
lukasz-antoniak, line 31: Call to reset is not needed.

crc.go
```diff
@@ -0,0 +1,57 @@
...
+import "hash/crc32"
+
+var (
+	// Initial CRC32 bytes: 0xFA, 0x2D, 0x55, 0xCA
+	initialCRC32Bytes = []byte{0xfa, 0x2d, 0x55, 0xca}
+)
+
+// ChecksumIEEE calculates the CRC32 checksum of the given byte slice.
+func ChecksumIEEE(b []byte) uint32 {
+	crc := crc32.NewIEEE()
```
lukasz-antoniak, line 30: Every object contains quite big array `type Table [256]uint32`. Cassandra on the server side creates only one CRC object per thread. Maybe we should also create a pool of those objects to minimise memory consumption if we process a lot of concurrent requests.

conn.go
```diff
@@ -777,6 +800,48 @@ func (c *Conn) handleTimeout() {
...
+		payload, isSelfContained, err = readCompressedFrame(c.r, c.compressor)
+	} else {
+		payload, isSelfContained, err = readUncompressedFrame(c.r)
+	}
+	if err != nil {
+		return err
+	}
+
+	if isSelfContained {
+		// TODO handle case when there are more than 1 envelop inside the frame
```
lukasz-antoniak, line 821: This TODO has to be solved, because other V5 capable drivers do it as well. Quote from protocol:
```
If self contained, then the payload includes one or more complete envelopes and can be fully processed immediately.
```
Also implement tests for it.

crc.go
```diff
@@ -0,0 +1,57 @@
...
+	return crc.Sum32()
+}
+
+const (
+	crc24Init = 0x875060  // Initial value for CRC24 calculation
+	crc24Poly = 0x1974F0B // Polynomial for CRC24 calculation
+)
+
+// KoopmanChecksum calculates the CRC24 checksum using the Koopman polynomial.
+func KoopmanChecksum(buf []byte) uint32 {
```
lukasz-antoniak, line 43: Tests are missing for both CRC calculations. Somebody could unexpectedly change initial value and he would not realise it.

frame.go
```diff
@@ -1599,8 +1608,9 @@ func (f frameWriterFunc) buildFrame(framer *framer, streamID int) error {
 }
 
 type writeExecuteFrame struct {
-	preparedID []byte
-	params     queryParams
+	preparedID         []byte
+	preparedMetadataID []byte
```
lukasz-antoniak, line 1612: For clarity change the property name to`<result_metadata_id>`.

frame.go
```diff
@@ -2070,3 +2085,251 @@ func (f *framer) writeBytesMap(m map[string][]byte) {
...
+		headerSize       = 6
+		selfContainedBit = 1 << 17
+	)
+
+	payloadLen := len(payload)
+	if payloadLen > maxPayloadSize {
+		return nil, fmt.Errorf("payload length (%d) exceeds maximum size of 128 KiB", payloadLen)
+	}
+
+	header := make([]byte, headerSize)
```
lukasz-antoniak, line 2189: Can we here not create `header` array and then copy it over to `frame`, but instead just write to `frame`?

frame.go
```diff
@@ -2070,3 +2085,251 @@ func (f *framer) writeBytesMap(m map[string][]byte) {
...
+
+	compressedLen := len(compressedPayload)
+
+	// Compression is not worth it
+	if uncompressedLen < compressedLen {
+		// native_protocol_v5.spec
+		// 2.2
+		//  An uncompressed length of 0 signals that the compressed payload
+		//  should be used as-is and not decompressed.
+		compressedPayload = uncompressedPayload
```
lukasz-antoniak, line 2246: `TestLargeSizeQuery` will generate random text, which LZ4 compressor most likely will not encode to smaller representation. Create a test to make sure that LZ4 compression is applied and we do not fall under this condition.

frame.go
```diff
@@ -2070,3 +2085,251 @@ func (f *framer) writeBytesMap(m map[string][]byte) {
...
+
+	// Compute and write the CRC32 checksum of the payload
+	payloadChecksum := ChecksumIEEE(compressedPayload)
+	binary.LittleEndian.PutUint32(headerBuf[:], payloadChecksum)
+	buf.Write(headerBuf[:4])
+
+	return buf.Bytes(), nil
+}
+
+func readCompressedFrame(r io.Reader, compressor Compressor) ([]byte, bool, error) {
```
lukasz-antoniak, line 2282: We are missing tests to simulate various frame errors here, e.g. invalid length encoded in the frame, invalid CRC. Tests could use our frame building logic for a simple EXECUTE or QUERY, change payload and then make sure decoding fails with certain error.

conn.go
```diff
@@ -1394,9 +1483,10 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
 		params.skipMeta = !(c.session.cfg.DisableSkipMetadata || qry.disableSkipMetadata)
 
 		frame = &writeExecuteFrame{
```
lukasz-antoniak, line 1485: For BATCH message we are missing keyspace field.
```
0x0080: With keyspace. If set, <keyspace> must be present. <keyspace> is a
[string] indicating the keyspace that the query should be executed in.
It supercedes the keyspace that the connection is bound to, if any.
```

lukasz-antoniak, line 1485: Field `<now_in_seconds>` has not been added as per below quote and flags bitmap not updated.
```
Added now_in_seconds field in QUERY, EXECUTE, and BATCH messages (Sections 4.1.4, 4.1.6, and 4.1.7).
```

lukasz-antoniak (changes requested):

crc.go
worryg0d: As far as I see each instance of `crc32.digest` created by `crc32.NewIEEE` has a pointer to the global `crc32.IEEETable` table, so it doesn't allocate a new table on each func call.

```go
// New creates a new [hash.Hash32] computing the CRC-32 checksum using the
// polynomial represented by the [Table]. Its Sum method will lay the
// value out in big-endian byte order. The returned Hash32 also
// implements [encoding.BinaryMarshaler] and [encoding.BinaryUnmarshaler] to
// marshal and unmarshal the internal state of the hash.
func New(tab *Table) hash.Hash32 {
	if tab == IEEETable {
		ieeeOnce.Do(ieeeInit)
	}
	return &digest{0, tab}
}

// NewIEEE creates a new [hash.Hash32] computing the CRC-32 checksum using
// the [IEEE] polynomial. Its Sum method will lay the value out in
// big-endian byte order. The returned Hash32 also implements
// [encoding.BinaryMarshaler] and [encoding.BinaryUnmarshaler] to marshal
// and unmarshal the internal state of the hash.
func NewIEEE() hash.Hash32 { return New(IEEETable) }
```

```go
// IEEETable is the table for the [IEEE] polynomial.
var IEEETable = simpleMakeTable(IEEE)
```

conn.go
worryg0d: I agree. I changed the comment and renamed the flag to `startupCompleted`.
```
	// Only for proto v5+.
	// Indicates if STARTUP has been completed.
	// github.com/apache/cassandra/blob/trunk/doc/native_protocol_v5.spec
	// 2.3.1 Initial Handshake
	// 	In order to support both v5 and earlier formats, the v5 framing format is not
	//  applied to message exchanges before an initial handshake is completed.
	startupCompleted bool
```

conn.go
worryg0d: If I understand correctly you meant to change the order. Could you please add some details here?
```
if c.startupCompleted && c.version > protoVersion4 {
    ...
}
```

crc.go
worryg0d: I agree here and added some tests to cover this important part.

frame.go
worryg0d: Good catch! I moved the frame buffer initializing to allow the header to be written directly into the frame.

frame.go
worryg0d: I agree, I added a test that covers this case as well.

frame.go
worryg0d: I agree and added tests for `readCompressedFrame` and `readUncompressedFrame`.

worryg0d:
Thanks a lot for the feedback!

Most of the issues mentioned in this PR have been covered in my other PRs related to proto 5 support.

What do you think about continuing the discussion in PR #1822 which contains changes from all others? I tried to segregate different features on different PRs for easy review. However, it doesn't seem to be as easy as I thought.

@lukasz-antoniak What are your thoughts on this?

joao-r-reis:

worryg0d:

Replied in #1778.

conn.go
lukasz-antoniak: This has been addressed in #1822.

frame.go
lukasz-antoniak: This has been addressed in #1822.

PR #1778, opened 2024-07-15
Support of `keyspace` field for BATCH message

worryg0d:
This PR adds support for the **keyspace** field for the BATCH message as part of Native Protocol v5 Support.
This implementation is based on the current implementation for QUERY and EXECUTE messages in gocql.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -3288,3 +3288,46 @@ func TestQuery_NamedValues(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestBatchKeyspaceField(t *testing.T) {
+	session := createSession(t)
+	defer session.Close()
+
+	if session.cfg.ProtoVersion < protoVersion5 {
+		t.Skip("keyspace for BATCH message is not supported in protocol < 5")
+	}
+
+	err := createTable(session, "CREATE TABLE batch_keyspace(id int, value text, PRIMARY KEY (id))")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ids := []int{1, 2}
+	texts := []string{"val1", "val2"}
+
+	b := session.NewBatch(LoggedBatch)
+	b.Query("INSERT INTO batch_keyspace(id, value) VALUES (?, ?)", ids[0], texts[0])
+	b.Query("INSERT INTO batch_keyspace(id, value) VALUES (?, ?)", ids[1], texts[1])
+	err = session.ExecuteBatch(b)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var (
+		id   int
+		text string
+	)
+
+	iter := session.Query("SELECT * FROM batch_keyspace").Iter()
+	defer iter.Close()
+
+	for i := 0; iter.Scan(&id, &text); i++ {
+		if id != ids[i] {
+			t.Fatalf("expected id %v, got %v", ids[i], id)
+		}
+
+		if text != texts[i] {
+			t.Fatalf("expected text %v, got %v", texts[i], text)
+		}
+	}
+}
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -32,6 +32,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/stretchr/testify/require"
 	"io"
 	"math"
 	"math/big"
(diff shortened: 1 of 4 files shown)


conn.go
```diff
@@ -1554,6 +1554,10 @@ func (c *Conn) executeBatch(ctx context.Context, batch *Batch) *Iter {
 		customPayload:         batch.CustomPayload,
 	}
 
+	if c.version > protoVersion4 {
+		req.keyspace = c.currentKeyspace
```
lukasz-antoniak: The purpose of introducing `keyspace` field in the protocol is not to copy default keyspace that we used when establishing connection, but to be able to override keyspace on per-statement basis. You need to specify keyspace from `Batch` type.

cassandra_test.go
```diff
@@ -3288,3 +3288,46 @@ func TestQuery_NamedValues(t *testing.T) {
...
+
+func TestBatchKeyspaceField(t *testing.T) {
+	session := createSession(t)
+	defer session.Close()
+
+	if session.cfg.ProtoVersion < protoVersion5 {
+		t.Skip("keyspace for BATCH message is not supported in protocol < 5")
+	}
+
+	err := createTable(session, "CREATE TABLE batch_keyspace(id int, value text, PRIMARY KEY (id))")
```
lukasz-antoniak: You should create table in different keyspace, and then do `b.keyspace = "foo"`. Query `session.Query("SELECT * FROM batch_keyspace")` should also override the default keyspace.

cassandra_test.go
```diff
@@ -3288,3 +3288,46 @@ func TestQuery_NamedValues(t *testing.T) {
...
+	var (
+		id   int
+		text string
+	)
+
+	iter := session.Query("SELECT * FROM batch_keyspace").Iter()
+	defer iter.Close()
+
+	for i := 0; iter.Scan(&id, &text); i++ {
+		if id != ids[i] {
```
lukasz-antoniak: Maybe use `require.Equal(t, ...)`.

cassandra_test.go
```diff
@@ -3288,3 +3288,46 @@ func TestQuery_NamedValues(t *testing.T) {
...
+	err := createTable(session, "CREATE TABLE batch_keyspace(id int, value text, PRIMARY KEY (id))")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ids := []int{1, 2}
+	texts := []string{"val1", "val2"}
+
+	b := session.NewBatch(LoggedBatch)
+	b.Query("INSERT INTO batch_keyspace(id, value) VALUES (?, ?)", ids[0], texts[0])
```
lukasz-antoniak, line 3323: I think we should also do the test when one of queries also overrides the keyspace.

lukasz-antoniak (changes requested):

conn.go
worryg0d: I totally agree with you. I added the `SetKeyspace` field for the `Batch` to allow overriding keyspace for the specific batch.

cassandra_test.go
worryg0d: As far as I see from the [proto 5 spec](https://github.com/apache/cassandra/blob/985fd2931a5c13ba5e3809ce327dac078212fc83/doc/native_protocol_v5.spec#L617) we can not override the keyspace for the specific query in the Batch, only for the Batch itself.

```
4.1.7. BATCH

  Allows executing a list of queries (prepared or not) as a batch (note that
  only DML statements are accepted in a batch). The body of the message must
  be:
    <type><n><query_1>...<query_n><consistency><flags>[<serial_consistency>][<timestamp>][<keyspace>][<now_in_seconds>]
```

worryg0d:
@lukasz-antoniak Thanks a lot for your feedback!

I added some enhancements according to your review.

Could you please take a look at this?

cassandra_test.go
lukasz-antoniak: I see that my comment might not be understandable. You can simulate hardcoding keyspace in CQL string itself. But, yeah it is not a very useful test indeed.

joao-r-reis:

worryg0d:
Hello👋. I don't see any problem of closing this one and others PRs as well 👍

cassandra_test.go
lukasz-antoniak: This has been addressed in #1822.

conn.go
lukasz-antoniak: This has been addressed in #1822.

PR #1781, opened 2024-07-16
Support for `now_in_seconds` option field for QUERY, BATCH and EXECUTE messages

worryg0d:
This PR provides support for [now_in_seconds](https://github.com/apache/cassandra/blob/42d18a83e4468827da234c68fad3e3058b368d7c/doc/native_protocol_v5.spec#L1455C1-L1456C1) optional field for QUERY, BATCH and EXECUTE messages as part of Native Protocol v5 support.
The implementation is mostly based on previous implementations for such optional fields as _timestamp_, _keyspace_, etc.

### DIFF
--- a/batch_test.go
+++ b/batch_test.go
@@ -84,3 +84,68 @@ func TestBatch_WithTimestamp(t *testing.T) {
 		t.Errorf("got ts %d, expected %d", storedTs, micros)
 	}
 }
+
+func TestQuery_WithNowInSeconds(t *testing.T) {
+	session := createSession(t)
+	defer session.Close()
+
+	if session.cfg.ProtoVersion < protoVersion5 {
+		t.Skip("Query now in seconds are only available on protocol >= 5")
+	}
+
+	if err := createTable(session, `CREATE TABLE query_now_in_seconds (id int primary key, val text)`); err != nil {
+		t.Fatal(err)
+	}
+
+	// Using 30 minutes ago timestamp as a starting point for TTL
+	seconds := time.Now().Add(-time.Minute*30).UnixMicro() / 1e6
+
+	err := session.
+		Query("INSERT INTO query_now_in_seconds (id, val) VALUES (?, ?) USING TTL 3600", 1, "val").
+		WithNowInSeconds(int(seconds)).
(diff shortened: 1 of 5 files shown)


batch_test.go
```diff
@@ -84,3 +84,68 @@ func TestBatch_WithTimestamp(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var ttl int64
+	if err := session.Query(`SELECT TTL(val) FROM query_now_in_seconds WHERE id = ?`, 1).Scan(&ttl); err != nil {
+		t.Fatal(err)
+	}
+
+	if ttl > 1800 {
```
lukasz-antoniak, line 116: I think this test is slightly better: [NowInSecondsIT.java](https://github.com/apache/cassandra-java-driver/blob/4.x/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/NowInSecondsIT.java). Use `NowInSeconds` also in SELECT query, so that you can indeed read expected TTL difference.

frame.go
```diff
@@ -1512,7 +1515,13 @@ func (f *framer) writeQueryParams(opts *queryParams) {
 	}
 
 	if f.proto > protoVersion4 {
-		f.writeUint(uint32(flags))
+		flags := uint32(flags)
```
lukasz-antoniak, line 1518: I think changing flags to `int` type and truncating it for protocols before V5 would be cleaner.

frame.go
```diff
@@ -1512,7 +1515,13 @@ func (f *framer) writeQueryParams(opts *queryParams) {
...
+
+		if opts.nowInSeconds {
```
lukasz-antoniak, line 1520: After we refactor flag to be an integer, we can move this to a place where all flags are computed.

lukasz-antoniak (changes requested):

joao-r-reis:

worryg0d:

Replied in #1778.

batch_test.go
lukasz-antoniak: This has been addressed in #1822.

PR #1782, opened 2024-07-19
Support for `result_metadata_id` in PREPARED and EXECUTE messages

worryg0d:
This PR provides support for the `result_metadata_id` field in **[PREPARED](https://github.com/apache/cassandra/blob/315cbac25503e61c6f0832f12177cebe88698445/doc/native_protocol_v5.spec#L603)** response and **[EXECUTE](https://github.com/apache/cassandra/blob/315cbac25503e61c6f0832f12177cebe88698445/doc/native_protocol_v5.spec#L899)** messages.
This PR is part of Native Protocol v5 support.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1394,9 +1394,10 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
 		params.skipMeta = !(c.session.cfg.DisableSkipMetadata || qry.disableSkipMetadata)
 
 		frame = &writeExecuteFrame{
-			preparedID:    info.id,
-			params:        params,
-			customPayload: qry.customPayload,
+			preparedID:         info.id,
+			params:             params,
+			customPayload:      qry.customPayload,
+			preparedMetadataID: info.request.id,
 		}
 
 		// Set "keyspace" and "table" property in the query if it is present in preparedMetadata
--- a/conn.go
+++ b/conn.go
@@ -1223,9 +1223,10 @@ type StreamObserverContext interface {
 }
 
 type preparedStatment struct {
-	id       []byte
-	request  preparedMetadata
-	response resultMetadata
+	id               []byte
+	resultMetadataID []byte
+	request          preparedMetadata
+	response         resultMetadata
 }
 
 type inflightPrepare struct {
@@ -1284,7 +1285,8 @@ func (c *Conn) prepareStatement(ctx context.Context, stmt string, tracer Tracer)
 				flight.preparedStatment = &preparedStatment{
 					// defensively copy as we will recycle the underlying buffer after we
 					// return.
-					id: copyBytes(x.preparedID),
+					id:               copyBytes(x.preparedID),
+					resultMetadataID: copyBytes(x.resultMetadataID),
 					// the type info's should _not_ have a reference to the framers read buffer,
 					// therefore we can just copy them directly.
 					request:  x.reqMeta,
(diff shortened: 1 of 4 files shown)


testisnullus (approved):

conn.go
```diff
@@ -1394,9 +1394,10 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
 		params.skipMeta = !(c.session.cfg.DisableSkipMetadata || qry.disableSkipMetadata)
 
 		frame = &writeExecuteFrame{
-			preparedID:    info.id,
-			params:        params,
-			customPayload: qry.customPayload,
+			preparedID:         info.id,
+			params:             params,
+			customPayload:      qry.customPayload,
+			preparedMetadataID: info.request.id,
```
lukasz-antoniak: This is definitely not correct. You should send here `<result_metadata_id>` received from PREPARE response, not the `<id>` of PREPARE response.

frame.go
```diff
@@ -1604,23 +1611,31 @@ type writeExecuteFrame struct {
 
 	// v4+
 	customPayload map[string][]byte
+
+	// v5+
+	preparedMetadataID []byte
```
lukasz-antoniak: I would consider following naming convention from protocol specification, so this would be `result_metadata_id`.

frame.go
```diff
@@ -952,6 +955,10 @@ func (f *framer) parsePreparedMetadata() preparedMetadata {
 	// TODO: deduplicate this from parseMetadata
 	meta := preparedMetadata{}
 
+	if f.proto > protoVersion4 {
```
lukasz-antoniak: `<id>` field of PREPARE response is already present in V4. Here we are missing reading of `<result_metadata_id>`.

conn.go
lukasz-antoniak: Following part of specification has not been implemented:
```
If a RESULT/Rows message reports
changed resultset metadata with the Metadata_changed flag, the reported new
resultset metadata must be used in subsequent executions.
```

frame_test.go
```diff
@@ -127,3 +127,27 @@ func TestFrameReadTooLong(t *testing.T) {
 		t.Fatalf("expected to get header %v got %v", opReady, head.op)
 	}
 }
+
+func Test_framer_writeExecuteFrame_preparedMetadataID(t *testing.T) {
```
lukasz-antoniak, line 131: We need an integration test to validate `<result_metadata_id>` logic. Details of why it was included can be found in ticket [CASSANDRA-10786](https://issues.apache.org/jira/browse/CASSANDRA-10786).

conn.go
worryg0d: Good catch! Thanks a lot. Also, I added the described mechanism. Now the driver receives the `RESULT/ROWS` response it updates the prepared stmt `meta` and `result_metadata_id` in the stmtsLRU cache and re-run query execution.

frame.go
worryg0d: Also good catch! Thanks! I renamed it no `resultMetadataID`

frame.go
worryg0d: I'm sorry for the misunderstanding here. The `id` field in `preparedMetadata` doesn't represent `<id>` in the specification, but `result_metadata_id`. 
Also, the naming is terrible here, which leads to a big misunderstanding.

I moved parsing of the `result_metadata_id` from this method to the `parseResultPrepared` method, because the `result_metadata_id` field is a part of the **PREPARED** message, not its metadata.

frame_test.go
worryg0d: I totally agree with you here. We should run tests over C* to ensure the logic works fine. However, I struggled with triggering C* to return a `<new_metadata_id>` by changing the table columns number, but C* responded with `UNPREPARED` error message, not the `RESULT/ROWS`.

Could you please give me some advices on how to trigger this mechanism?

lukasz-antoniak: Can you merge all those V5 pull requests into one branch and one PR? I have a test nearly ready, but wanted to double test it.

worryg0d: @lukasz-antoniak Yes, here it is [1882](https://github.com/apache/cassandra-gocql-driver/pull/1822)

joao-r-reis:

worryg0d:

Replied in #1778.

frame.go
lukasz-antoniak: This has been addressed in #1822.

PR #1783, opened 2024-07-22
Added uint32 flags for QUERY and BATCH operations for native protocol v5

testisnullus:
The pull request #1783 for the Cassandra gocql driver introduces uint32 flags for QUERY and BATCH operations to support native protocol v5. It refactors the code to handle protocol-specific flag settings more efficiently by retaining byte-based flags for protocol versions less than 5 and introducing integer-based flags for protocol version 5 and higher. This refactoring improves code readability and maintainability, preparing the driver for future protocol changes.

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -150,14 +150,14 @@ const (
 	flagNoMetaData      int = 0x04
 
 	// query flags
-	flagValues                byte = 0x01
-	flagSkipMetaData          byte = 0x02
-	flagPageSize              byte = 0x04
-	flagWithPagingState       byte = 0x08
-	flagWithSerialConsistency byte = 0x10
-	flagDefaultTimestamp      byte = 0x20
-	flagWithNameValues        byte = 0x40
-	flagWithKeyspace          byte = 0x80
+	flagValues                uint32 = 0x01
+	flagSkipMetaData                 = 0x02
+	flagPageSize                     = 0x04
+	flagWithPagingState              = 0x08
+	flagWithSerialConsistency        = 0x10
+	flagDefaultTimestamp             = 0x20
+	flagWithNameValues               = 0x40
+	flagWithKeyspace                 = 0x80
 
 	// prepare flags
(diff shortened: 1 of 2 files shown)


frame.go
```diff
@@ -159,6 +159,15 @@ const (
...
 	flagWithKeyspace          byte = 0x80
 
+	flagValuesV5                uint32 = 0x01
+	flagSkipMetaDataV5                 = 0x02
+	flagPageSizeV5                     = 0x04
+	flagWithPagingStateV5              = 0x08
+	flagWithSerialConsistencyV5        = 0x10
+	flagDefaultTimestampV5             = 0x20
+	flagWithNameValuesV5               = 0x40
+	flagWithKeyspaceV5                 = 0x80
```
worryg0d: What do you think about changing only the type of existing flags? 
It allows us to avoid such code duplications. We may implement the new way of constructing the flag field as **uint32**. Also, if the proto version is less than 5, we can just cast it to **byte** inside the corresponding methods like `framer.writeQueryParams`.

frame.go
```diff
@@ -1452,49 +1452,66 @@ func (f *framer) writeQueryParams(opts *queryParams) {
...
+				flagsByte |= byte(flagDefaultTimestamp)
+			}
+			if len(opts.values) > 0 && opts.values[0].name != "" {
+				flagsByte |= byte(flagWithNameValues)
+				names = true
+			}
+		}
+		f.writeByte(flagsByte)
+
+	case f.proto >= protoVersion5:
```
worryg0d: I think we also may avoid these code duplications by implementing only the new way of constructing the flag. As I see, we may construct the flag as **uint32**, and in the end, we may check the proto version. If the version is less than 5 we may cast it to **byte** and write it as it is.
What do you think about this approach? Also, if we follow this approach, the `flagsByte` var may be omitted.

testisnullus: Make sense, fixed

worryg0d (commented):
lgtm

PR #1784, opened 2024-07-24
Adding support system_traces.coordinator_port and source_port

OleksiienkoMykyta:
The are 2 new columns in Cassandra: 
system_traces.sessions.coordinator_port 
system_traces.events.source_port 
They should be exposed to the user to have the opportunity to check trace info.
https://datastax-oss.atlassian.net/browse/PYTHON-1229

Notice: This fields exists in Cassandra V3 but always equals null, only from Cassandra v4 we can get the ports.
Also, I found some columns from system_traces.sessions that didn`t exposed for user as well yet (client | command | parameters  | request | started_at)

### DIFF
--- a/session.go
+++ b/session.go
@@ -2111,14 +2111,15 @@ func NewTraceWriter(session *Session, w io.Writer) Tracer {
 
 func (t *traceWriter) Trace(traceId []byte) {
 	var (
-		coordinator string
-		duration    int
+		coordinator     string
+		duration        int
+		coordinatorPort int
 	)
-	iter := t.session.control.query(`SELECT coordinator, duration
-			FROM system_traces.sessions
-			WHERE session_id = ?`, traceId)
+	iter := t.session.control.query(`SELECT coordinator, duration, coordinator_port
+      FROM system_traces.sessions
+      WHERE session_id = ?`, traceId)
 
-	iter.Scan(&coordinator, &duration)
+	iter.Scan(&coordinator, &duration, &coordinatorPort)
 	if err := iter.Close(); err != nil {
 		t.mu.Lock()
 		fmt.Fprintln(t.w, "Error:", err)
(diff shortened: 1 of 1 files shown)


session.go
```diff
@@ -2111,14 +2111,15 @@ func NewTraceWriter(session *Session, w io.Writer) Tracer {
...
-		coordinator string
-		duration    int
+		coordinator     string
+		duration        int
+		coordinatorPort int
 	)
-	iter := t.session.control.query(`SELECT coordinator, duration
-			FROM system_traces.sessions
-			WHERE session_id = ?`, traceId)
+	iter := t.session.control.query(`SELECT coordinator, duration, coordinator_port
```
testisnullus, line 2118: We can switch places of duration and coordinator_port fields a little bit:
```
SELECT coordinator, coordinator_port, duration
```

It will be more logical to display info about coordinator node this way

session.go
```diff
@@ -2127,26 +2128,27 @@ func (t *traceWriter) Trace(traceId []byte) {
...
-		traceId, coordinator, time.Duration(duration)*time.Microsecond)
+	fmt.Fprintf(t.w, "Tracing session %016x (coordinator: %s, duration: %v, coordinator_port: %d):\n",
+		traceId, coordinator, time.Duration(duration)*time.Microsecond, coordinatorPort)
 
-	iter = t.session.control.query(`SELECT event_id, activity, source, source_elapsed, thread
-			FROM system_traces.events
-			WHERE session_id = ?`, traceId)
+	iter = t.session.control.query(`SELECT event_id, activity, source, source_elapsed, source_port, thread
+      FROM system_traces.events
+      WHERE session_id = ?`, traceId)
```
testisnullus, line 2147: The same is true here. We can change the place of source_port to:

```
SELECT event_id, activity, source, source_port, source_elapsed, thread
```

Issue #1699, opened 2023-05-22
Could MapScan() get nil instead of zero value for null value？


fhwangyinan:
Hi,
I am new to using Gocql, and I'm currently trying to write a reusable function for selecting data. I want the function to return the query result dynamically. However, when using MapScan, null values are being converted to zero, which is causing confusion for me in distinguishing between zero and null. I have reviewed previous issues and documentation but haven't found a similar solution. Could you please provide some suggestions?

Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?
4.1.0

### What version of Gocql are you using?
v1.3.2

### What version of Go are you using?
v1.20

### What did you do?
```

func (c *CassandraClient) Select(keyspace string, table string, columns []string, where []Filter) ([]map[string]interface{}, error) {
	query := fmt.Sprintf("SELECT %s FROM %s.%s", strings.Join(columns, ","), keyspace, table)
	var whereClauses []string
	var args []interface{}
	for _, v := range where {
		whereClauses = append(whereClauses, fmt.Sprintf("%s %s ?", v.Key, v.Operator))
		args = append(args, v.Value)
	}
	if len(whereClauses) > 0 {
		query += " WHERE " + strings.Join(whereClauses, " AND ")
	}
	query += " ALLOW FILTERING"
	logrus.Infof("CQL: %s , Args: %s", query, args)
	iter := c.session.Query(query, args...).Iter()
	results := make([]map[string]interface{}, 0)

	for {
		m := make(map[string]interface{}, len(columns))

		if !iter.MapScan(m) {
			break
		}
		// Handle NaN values
		for key, value := range m {
			if f, ok := value.(float64); ok && math.IsNaN(f) {
				m[key] = "NaN"
			}
		}
		results = append(results, m)
	}

	if err := iter.Close(); err != nil {
		return nil, err
	}
	return results, nil
}
```

### What did you expect to see?
The null value in the m variable is distinguished from the zero value.

### What did you see instead?
null is converted to zero values.
---

If you are having connectivity related issues please share the following additional information

### Describe your Cassandra cluster
please provide the following information

- output of `nodetool status`
- output of `SELECT peer, rpc_address FROM system.peers`
- rebuild your application with the `gocql_debug` tag and post the output

martin-sucha:
Hi!

`gocql.Unmarshal` returns zero values, that is probably because a `gocql.Unmarshaler` can handle null values itself.

[...]

PR #1785, opened 2024-07-24
CASSGO-6 'Unable to discover cluster nodes with an empty rack name' issue fix

### LINKED ISSUE #1706: Unable to discover cluster nodes with an empty rack name

3.11.8, 4.0.3

1.5.2

1.20.3

We have a client who is using gocql to connect to a Cassandra cluster which has been configured with an empty rack name on all nodes. What we are seeing is that gocql is unable to discover the cluster topology and it only sends requests to the initial contact point(s).
When using the Java driver against this cluster it works just fine and discovers all of the nodes correctly. The difference appears to be that they filter out nodes that have a NULL rack name in system.peers[_v2] and gocql filters out nodes with an empty string in the rack name.

I have been able to reproduce this in a lab environment on Cassandra 3.11 and 4.0 by using GossipingPropertyFileSnitch and setting the rack to an empty string in cassandra-rackdc.properties.

tengu-alt:
This PR addresses issue #1706.

The driver was unable to discover nodes with an empty rack name because the `isValidPeer()` function incorrectly checked for an empty string in the rack value (which is a valid case). The correct validation should ensure the rack name is not null (as a null value indicates an invalid peer).

This PR fixes the method of retrieving the rack value from the `system.peers` and `system.peers_v2` tables. Now, the rack value is set as a pointer to a string, allowing the value to be checked for null.

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -332,23 +332,29 @@ func (iter *Iter) RowData() (RowData, error) {
 	values := make([]interface{}, 0, len(iter.Columns()))
 
 	for _, column := range iter.Columns() {
-		if c, ok := column.TypeInfo.(TupleTypeInfo); !ok {
-			val, err := column.TypeInfo.NewWithError()
-			if err != nil {
-				iter.err = err
-				return RowData{}, err
-			}
+		if column.Name == "rack" && column.Keyspace == "system" && (column.Table == "peers_v2" || column.Table == "peers") {
+			var strPtr = new(string)
 			columns = append(columns, column.Name)
-			values = append(values, val)
+			values = append(values, &strPtr)
 		} else {
-			for i, elem := range c.Elems {
-				columns = append(columns, TupleColumnName(column.Name, i))
-				val, err := elem.NewWithError()
+			if c, ok := column.TypeInfo.(TupleTypeInfo); !ok {
+				val, err := column.TypeInfo.NewWithError()
 				if err != nil {
 					iter.err = err
 					return RowData{}, err
 				}
+				columns = append(columns, column.Name)
 				values = append(values, val)
+			} else {
+				for i, elem := range c.Elems {
+					columns = append(columns, TupleColumnName(column.Name, i))
+					val, err := elem.NewWithError()
+					if err != nil {
+						iter.err = err
+						return RowData{}, err
+					}
+					values = append(values, val)
+				}
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


helpers.go
```diff
@@ -331,21 +331,27 @@ func (iter *Iter) RowData() (RowData, error) {
 	values := make([]interface{}, 0, len(iter.Columns()))
 
 	for _, column := range iter.Columns() {
-		if c, ok := column.TypeInfo.(TupleTypeInfo); !ok {
-			val, err := column.TypeInfo.NewWithError()
-			if err != nil {
-				return RowData{}, err
-			}
+		if column.Name == "rack" && column.Keyspace == "system" && (column.Table == "peers_v2" || column.Table == "peers") {
```
martin-sucha, line 335: This seems like a hack, plus it could break backwards compatibility for existing users of `RowData` (outside of gocql) that read from those tables.
Could the issue be fixed another way, for example improving the API/adding new API for all users?

tengu-alt: There is #1834, which allows to handle a nullable value. After #1834 is merged, I will refactor this PR using the nullable scan functional. This should help to avoid the hack.

host_source.go
```diff
@@ -157,6 +157,7 @@ type HostInfo struct {
 	state            nodeState
 	schemaVersion    string
 	tokens           []string
+	isRackNil        bool
```
joao-r-reis, line 182: We can change `rack` to be `*string`, it's a private field. A conversion has to happen in the `Rack()` function though.

helpers.go
joao-r-reis: I agree with addressing #1834 first, @martin-sucha can you provide feedback on #1834 ? I see [your original comment](https://github.com/apache/cassandra-gocql-driver/issues/1699#issuecomment-1556901238) on that issue, but I'm not sure if #1834 accurately implements what you were proposing.

host_source.go
tengu-alt: Sounds good to me!
Let's wait for the feedback that you have asked for.

joao-r-reis:
Just for future reference in case anyone checks the status of this: we're currently waiting on #1834 and #1834 is waiting on #1855

jameshartig:
I have an alternative solution that I can propose soon without depending on any other PRs.

tengu-alt:
> I have an alternative solution that I can propose soon without depending on any other PRs.

Sounds good, let's see it when it arrives.

jameshartig:
Alternative fix: https://github.com/apache/cassandra-gocql-driver/pull/1895

Issue #1665, opened 2022-11-17
`system.peers` queried event if `DisableInitialHostLookup` set to `true`


leesio:
# gocql github issue

### What version of Cassandra are you using?

We're actually using [AWS Keyspaces](https://aws.amazon.com/keyspaces/)

### What version of Gocql are you using?

The issue exists in `1.2.1`

### What version of Go are you using?

1.19

### What did you do?

Set the `DisableInitialHostLookup` to `true`

### What did you expect to see?

The driver would never query `system.peers` for host information

### What did you see instead?

On heartbeat failure, the driver tries to query `system.peers`.

---

Hi there,

Apologies in advance if this has been answered before - I searched through the issues and couldn't see anything related.

The [docs for the `DisableInitialHostLookup`](https://github.com/gocql/gocql/blob/6132f789ad22c1d19a66c58ac789402d5f1f9be5/cluster.go#L158-L163) config flag state:

> If DisableInitialHostLookup then the driver will not attempt to get host info from the `system.peers` table, this will mean that the driver will connect to hosts supplied and will not attempt to lookup the hosts information

It's true that the library avoids quering `system.peers` [on session initialization](https://github.com/gocql/gocql/blob/6132f789ad22c1d19a66c58ac789402d5f1f9be5/session.go#L216-L231) however, in the event of a heartbeat failure on the control connection, we [call `(*controlConn).reconnect` with the `refreshring` arg `true`](https://gkjithub.com/gocql/gocql/blob/6132f789ad22c1d19a66c58ac789402d5f1f9be5/control.go#L101). Which (as far as I can tell) will ultimately [query `system.peers`](https://github.com/gocql/gocql/blob/6132f789ad22c1d19a66c58ac789402d5f1f9be5/conn.go#L1623).

It seems to me that this is true to the _name_ of the config flag `DisableInitialHostLookup` but is not quite what the documentation suggests. I'm interested to know the maintainers' view on this and whether they'd consider a patch to change the reconnection logic to consider the `DisableInitialHostLookup` configuration. Perhaps something like this:

```
reconn:
	// try to connect a bit faster
	sleepTime = 1 * time.Second
	refreshRing := !c.session.cfg.DisableInitialHostLookup
	c.reconnect(refreshRing)
	continue
}
```

## Why do we care?

A little bit more context to explain why this matters to us.

[...]

Issue #1789, opened 2024-08-01
Fix disableHostLookup flag logic


RostislavPorohnya:
According to the Contributor comment on #1665 issue:
              It seems the behavior was like this since commit 83932d6de9af9257339038f8a3a017941c34921d that introduced the option. It seems reasonable to me to treat this as a bug since the documentation says that the driver won't try to discover the hosts. Although `DisableHostLookup` would indeed be better name for a config option like that. And it is not clear to me how users use the DisableInitialHostLookup option, resp. if disabling the lookup always can break someone's program.

Currently, gocql queries `system.peers` in these cases:

* During `Session.init` if `DisableInitialHostLookup` is `false`.
* When reconnecting control connection.
* When calling `Session.AwaitSchemaAgreement` explicitly.
* When a [schema change query successfully executes](https://github.com/gocql/gocql/blob/6132f789ad22c1d19a66c58ac789402d5f1f9be5/conn.go#L1438).
* When a [keyspace change event is received](https://github.com/gocql/gocql/blob/6132f789ad22c1d19a66c58ac789402d5f1f9be5/events.go#L128).
* When a node is added or removed from the cluster. (`Session.addNewNewNode`, `Session.handleNewNode`, `Session.handleRemovedNode`)

We would need to disable it in all these cases, not just during reconnection.

_Originally posted by @martin-sucha in https://github.com/apache/cassandra-gocql-driver/issues/1665#issuecomment-1319988023_
            
The flag should disable host lookup procedures in cases mentioned above.

OleksiienkoMykyta:
The issue is well described, but It looks a little bit stuck. Are there any blockers for it, or should it be added to v2.x?
As I understand from the review, we should keep backward compatibility to provide safe migration to the new version of the driver, so after some fixes, it can be merged into the current version.

PR #1790, opened 2024-08-01
CASSGO-5 Fix DisableInitialHostLookup flag ignored when querying system.peers

RostislavPorohnya:
This PR addresses issue #1665, which reports a bug where `system.peers` table is queried even if `DisableInitHostLookup` flag is set to `true`. According to the driver documentation, `system.peers` should not be queried when the flag is set to `true`

In this PR `DisableInitialHostLookup` flag renamed to `DisableHostLookup`, code logic is fixed to query `system.peers` only when the flag is set to `false`.


conn.go
```diff
@@ -1689,36 +1689,37 @@ func (c *Conn) querySystemLocal(ctx context.Context) *Iter {
...
+				}
 
-		if err = iter.Close(); err != nil {
-			goto cont
+				versions[host.schemaVersion] = struct{}{}
+			}
+
+			if err = iter.Close(); err != nil {
+				goto cont
+			}
```
joao-r-reis: I think it's fine if we query system.peers for the purpose of retrieving the schema version of each host as long as we don't update gocql's ring/hosts. If a user wants to use a static set of contact points and avoid using hosts from system tables there's no reason why they would have to give up the functionality of awaiting for schema agreement.

host_source.go
```diff
@@ -635,6 +635,11 @@ func (r *ringDescriber) getClusterPeerInfo(localHost *HostInfo) ([]*HostInfo, er
 		return nil, errNoControl
 	}
 
+	// Check if host lookup is disabled
+	if r.session.cfg.DisableHostLookup {
+		return []*HostInfo{}, nil
+	}
```
joao-r-reis: Instead of adding this, I think we should make the `ringRefresher` do a NOOP when this flag is set. `getClusterPeerInfo` should only be about retriving the peers info from the server so it should be up to the caller whether this should be done or not.

If we make the `func refreshRing(r *ringDescriber) error` function do nothing when the flag is set then it should solve all of the issues related to the driver not respecting the flag.

cluster.go
```diff
@@ -189,12 +189,12 @@ type ClusterConfig struct {
...
 	IgnorePeerAddr bool
 
-	// If DisableInitialHostLookup then the driver will not attempt to get host info
+	// If DisableHostLookup then the driver will not attempt to get host info
 	// from the system.peers table, this will mean that the driver will connect to
 	// hosts supplied and will not attempt to lookup the hosts information, this will
 	// mean that data_centre, rack and token information will not be available and as
 	// such host filtering and token aware query routing will not be available.
-	DisableInitialHostLookup bool
+	DisableHostLookup bool
```
joao-r-reis: I don't think this is worth breaking applications. I agree that removing the `Initial` word from the name does make sense but it's really not that big of a deal.

joao-r-reis (commented):
Added some comments.

In summary I think this PR should be limited to making the `refreshRing` function do nothing when the flag is set. That would ensure that the driver will keep using the contact points as the hosts forever and ignore any kind of topology changes or any host information from system tables. 

It's fine if the driver queries the system tables to check if there is schema agreement, I don't think it goes against the intention of the flag.

cluster.go
joao-r-reis: At the very least we would have to keep both flags (+ deprecate the old one) and map it to a single internal flag so that we can do a "safe" rename of the flag without breaking user applications.

OleksiienkoMykyta (commented):
It looks like your PR helps with (https://github.com/apache/cassandra-gocql-driver/issues/1789) and (https://github.com/apache/cassandra-gocql-driver/issues/1665), but what is the status of it? Should it be updated according to comments, or closed in favor of some other PR?

joao-r-reis:
If there's no update to this PR soon I'll descope it from the 2.0 release as this can be released in a minor released without an issue

RostislavPorohnya:
@joao-r-reis I'm going to update it by the end of next week, thanks for the comments!

RostislavPorohnya:
Hi. I’ve handed this PR over to @tengu-alt, who will finalize the remaining work. I’m switching to another task, so please feel free to tag him if you have any questions or need updates.

conn.go
tengu-alt: Agree, fixed.

host_source.go
tengu-alt: Agree, refactored

cluster.go
tengu-alt: Agree, I've returned the old name.

tengu-alt:
Made some updates: refactored tests, added new logic(`refreshRing()` from now does NOOP when `DisableInitialHostLookup` is true), and refactored code according to feedback. Also fixed the conflicts .

tengu-alt:
@joao-r-reis Could you please take a look?

CHANGELOG.md
```diff
@@ -50,6 +50,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Refactoring hostpool package test and Expose HostInfo creation (CASSGO-59)
 
+- Fix DisableInitialHostLookup flag ignored when querying system.peers (CASSGO-5)
+
+- Fix DisableInitialHostLookup flag ignored when querying system.peers (PR update) (CASSGO-5)
```
joao-r-reis: duplicate entry?

conn.go
```diff
@@ -1898,16 +1898,14 @@ func (c *Conn) querySystemLocal(ctx context.Context) *Iter {
...
-	var versions map[string]struct{}
+	versions := make(map[string]struct{})
 	var schemaVersion string
 
 	endDeadline := time.Now().Add(c.session.cfg.MaxWaitSchemaAgreement)
 
 	for time.Now().Before(endDeadline) {
 		iter := c.querySystemPeers(ctx, c.host.version)
 
-		versions = make(map[string]struct{})
```
joao-r-reis: why this change? resetting the map on each attempt made more sense to me

disable_host_lookup_test.go
```diff
@@ -0,0 +1,39 @@
+package gocql
+
+import (
+	"net"
+	"testing"
+)
+
+func TestDisableHostLookupRingRefresh(t *testing.T) {
```
joao-r-reis: can we not add this to an existing test file? Creating a new file just for this test seems excessive. Also this is running in "unit tests" step because it's missing the appropriate build flag at the top of the file. Also missing license header.

joao-r-reis:
Also, I'm considering pushing this to a 2.x minor release followup and descoping it from 2.0 because it doesn't really need a major release and I also want do some extensive testing with the `DisableInitialHostLookup` flag as I believe there's some other issues that still need to be fixed (e.g. system.local still being queried I think)

CHANGELOG.md
tengu-alt: Agree, deleted

conn.go
tengu-alt: Agree, fixed

disable_host_lookup_test.go
tengu-alt: Done, I've moved the test into the `cassandra_test.go`

PR #1791, opened 2024-08-01
CASSGO-13 Add support for cassandra 4.0 table options

OleksiienkoMykyta:
Cassandra 4.0 has some minor table metadata changes that need to be accounted for:
read_repair_chance is replaced with read_repair
dclocal_read_repair_chance has been removed
additional_write_policy has been added

In the PR implemented backward compatibility with previous versions, and added new types support. To make metadata table support easier for future Cassandra versions, hardcode scan from Cassandra were replaced with new "parseSystemSchemaViews" method which is much easier to expand, even if some fields were added in the middle of the table it wouldn`t be an issue anymore.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -2328,7 +2328,7 @@ func TestGetColumnMetadata(t *testing.T) {
 
 func TestMaterializedViewMetadata(t *testing.T) {
 	if flagCassVersion.Before(3, 0, 0) {
-		return
+		t.Skip("The Cassandra version is too old")
 	}
 	session := createSession(t)
 	defer session.Close()
@@ -2347,14 +2347,19 @@ func TestMaterializedViewMetadata(t *testing.T) {
 	expectedChunkLengthInKB := "16"
 	expectedDCLocalReadRepairChance := float64(0)
 	expectedSpeculativeRetry := "99p"
+	expectedAdditionalWritePolicy := "99p"
+	expectedReadRepair := "BLOCKING"
 	if flagCassVersion.Before(4, 0, 0) {
 		expectedChunkLengthInKB = "64"
 		expectedDCLocalReadRepairChance = 0.1
 		expectedSpeculativeRetry = "99PERCENTILE"
+		expectedReadRepair = ""
+		expectedAdditionalWritePolicy = ""
 	}
 	expectedView1 := MaterializedViewMetadata{
 		Keyspace:                "gocql_test",
 		Name:                    "view_view",
+		AdditionalWritePolicy:   expectedAdditionalWritePolicy,
 		baseTableName:           "view_table",
 		BloomFilterFpChance:     0.01,
 		Caching:                 map[string]string{"keys": "ALL", "rows_per_partition": "NONE"},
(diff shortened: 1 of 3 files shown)


cassandra_test.go
```diff
@@ -2278,12 +2283,17 @@ func TestMaterializedViewMetadata(t *testing.T) {
 		DefaultTimeToLive:       0,
 		Extensions:              map[string]string{},
 		GcGraceSeconds:          864000,
-		IncludeAllColumns:       false, MaxIndexInterval: 2048, MemtableFlushPeriodInMs: 0, MinIndexInterval: 128, ReadRepairChance: 0,
-		SpeculativeRetry: expectedSpeculativeRetry,
+		IncludeAllColumns:       false, MaxIndexInterval: 2048,
+		MemtableFlushPeriodInMs: 0,
+		MinIndexInterval:        128,
+		ReadRepair:              expectedReadRepair,
+		ReadRepairChance:        0,
```
testisnullus, line 2378: As far as I know, read_repair_chance was removed from C* 4.0. 

https://issues.apache.org/jira/browse/CASSANDRA-15604

OleksiienkoMykyta: Yes, but this test also runs for C* 3.0.0, so it covers data for all versions. We can make separate tests for each version, but as mentioned in the description for the current PR, the main aim is to implement a backward-compatible metadata getter, and this test case helps to test one.

metadata.go
```diff
@@ -975,69 +977,207 @@ func getViewsMetadata(session *Session, keyspaceName string) ([]ViewMetadata, er
...
+		return nil, fmt.Errorf(errorMessage, "extentions")
+	}
+
+	extensions := make(map[string]string, len(byteData))
+	for key, rowByte := range byteData {
+		extensions[key] = string(rowByte)
+	}
+
+	return extensions, nil
+}
```
worryg0d:
```suggestion
func bytesMapToStringMap(bytesMap map[string][]byte) map[string]string {
	stringMap := make(map[string]string, len(bytesMap))
	for key, bytes := range bytesMap {
		stringMap[key] = string(bytes)
	}

	return stringMap
}
```

Let's make this method take `map[string][]byte` as a param instead of an empty interface. This approach will allow you to return described in this func error more properly from `materializedViewMetadataFromMap`.
Also, what do you think about renaming this func to something like `bytesMapToStringMap`?

OleksiienkoMykyta: I agree, it will make bytesMapToStringMap more reusable and error handling much more readable.
Thank You very much!

joao-r-reis (approved):

metadata.go
```diff
@@ -139,6 +140,7 @@ type MaterializedViewMetadata struct {
 	MaxIndexInterval        int
 	MemtableFlushPeriodInMs int
 	MinIndexInterval        int
+	ReadRepair              string
```
joao-r-reis: Maybe a small comment here and `ReadRepairChance` to indicate that `ReadRepairChance` was removed in C* 4.0

OleksiienkoMykyta: Something like `ReadRepairChance        float64 // Deprecated since Cassandra v4, use ReadRepair instead.`?

joao-r-reis: Maybe something like:

```
ReadRepair              string    // Only present in Cassandra 4.0+
ReadRepairChance        float64   // Note: Cassandra 4.0 removed ReadRepairChance and added ReadRepair instead
```

jameshartig (approved):

joao-r-reis:
Can you resolve the conflict and add James to the commit message?

PR #1793, opened 2024-08-05
CASSGO-4 Support of sending queries to the specific node

worryg0d:
# Overview

This PR provides a mechanism that allows users to specify on which node the query will be executed. It is not a typical use case, but it makes sense with virtual tables. 
For example, when we want to retrieve metrics for a specific node, we have to send queries to the associated system view of this node.

# Implementation Overview

A new method ~`SetHost()`~ `SetHostID` for the `Query` allows users to specify the host (node) on which the driver must send the query. It is implemented by adding a `GetHost()` to the `ExecutableQuery` interface. When the host is specified for the query the driver ignores any retry, speculative, and host selection policies.

# Usage Example
Here is an example of retrieving metrics from `system_view.cql_metrics` for each node of the cluster:
```go
cluster := NewCluster("127.0.0.1")
session, err := cluster.CreateSession()
if err != nil {
	panic(err)
}
defer session.Close()

hosts, err := session.GetHosts()
if err != nil {
	panic(err)
}

type cqlMetric struct {
	name  string
	value float64
}

// hostId to []cqlMetric
cqlMetrics := map[string][]cqlMetric{}

for _, host := range hosts {
	iter := session.Query("SELECT * FROM system_views.cql_metrics").
		SetHostID(host.HostID()).
		Iter()

	metrics := make([]cqlMetric, 0, iter.NumRows())
	metric := cqlMetric{}
	for iter.Scan(&metric.name, &metric.value) {
		metrics = append(metrics, metric)
	}
	cqlMetrics[host.HostID()] = metrics
}
```

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -3312,7 +3312,6 @@ func TestUnsetColBatch(t *testing.T) {
 	}
 	var id, mInt, count int
 	var mText string
-
 	if err := session.Query("SELECT count(*) FROM gocql_test.batchUnsetInsert;").Scan(&count); err != nil {
 		t.Fatalf("Failed to select with err: %v", err)
 	} else if count != 2 {
@@ -3345,5 +3344,54 @@ func TestQuery_NamedValues(t *testing.T) {
 	var value string
 	if err := session.Query("SELECT VALUE from gocql_test.named_query WHERE id = :id", NamedValue("id", 1)).Scan(&value); err != nil {
 		t.Fatal(err)
+	}
+}
+
+// This test ensures that queries are sent to the specified host only
+func TestQuery_SetHostID(t *testing.T) {
+	session := createSession(t)
+	defer session.Close()
+
+	hosts := session.GetHosts()
+
+	const iterations = 5
+	for _, expectedHost := range hosts {
+		for i := 0; i < iterations; i++ {
+			var actualHostID string
+			err := session.Query("SELECT host_id FROM system.local").
+				SetHostID(expectedHost.HostID()).
+				Scan(&actualHostID)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			if expectedHost.HostID() != actualHostID {
+				t.Fatalf("Expected query to be executed on host %s, but it was executed on %s",
+					expectedHost.HostID(),
+					actualHostID,
+				)
+			}
+		}
+	}
+
+	// ensuring properly handled invalid host id
+	err := session.Query("SELECT host_id FROM system.local").
+		SetHostID("[invalid]").
+		Exec()
+	if !errors.Is(err, ErrNoConnections) {
+		t.Fatalf("Expected error to be: %v, but got %v", ErrNoConnections, err)
+	}
+
+	// ensuring that the driver properly handles the case
+	// when specified host for the query is down
+	host := hosts[0]
+	pool, _ := session.pool.getPoolByHostID(host.HostID())
+	// simulating specified host is down
+	pool.host.setState(NodeDown)
+	err = session.Query("SELECT host_id FROM system.local").
+		SetHostID(host.HostID()).
+		Exec()
+	if !errors.Is(err, ErrNoConnections) {
+		t.Fatalf("Expected error to be: %v, but got %v", ErrNoConnections, err)
 	}
 }
--- a/query_executor.go
+++ b/query_executor.go
@@ -41,6 +41,7 @@ type ExecutableQuery interface {
 	Keyspace() string
 	Table() string
 	IsIdempotent() bool
+	GetHostID() string
 
 	withContext(context.Context) ExecutableQuery
 
@@ -83,12 +84,32 @@ func (q *queryExecutor) speculate(ctx context.Context, qry ExecutableQuery, sp S
 }
 
 func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
-	hostIter := q.policy.Pick(qry)
+	var hostIter NextHost
+
+	// check if the host id is specified for the query,
+	// if it is, the query should be executed at the corresponding host.
+	if hostID := qry.GetHostID(); hostID != "" {
+		hostIter = func() SelectedHost {
+			pool, ok := q.pool.getPoolByHostID(hostID)
+			// if the specified host is down
+			// we return nil to avoid endless query execution in queryExecutor.do()
+			if !ok || !pool.host.IsUp() {
+				return nil
+			}
+			return (*selectedHost)(pool.host)
+		}
+	}
+
+	// if host is not specified for the query,
+	// then a host will be picked by HostSelectionPolicy
+	if hostIter == nil {
+		hostIter = q.policy.Pick(qry)
+	}
 
 	// check if the query is not marked as idempotent, if
 	// it is, we force the policy to NonSpeculative
 	sp := qry.speculativeExecutionPolicy()
-	if !qry.IsIdempotent() || sp.Attempts() == 0 {
+	if qry.GetHostID() != "" || !qry.IsIdempotent() || sp.Attempts() == 0 {
 		return q.do(qry.Context(), qry, hostIter), nil
 	}
 
--- a/session.go
+++ b/session.go
@@ -456,6 +456,7 @@ func (s *Session) Query(stmt string, values ...interface{}) *Query {
 	qry.session = s
 	qry.stmt = stmt
 	qry.values = values
+	qry.hostID = ""
 	qry.defaultsFromSession()
 	return qry
 }
@@ -949,6 +950,10 @@ type Query struct {
 
 	// routingInfo is a pointer because Query can be copied and copyable struct can't hold a mutex.
 	routingInfo *queryRoutingInfo
+
+	// hostID specifies the host on which the query should be executed.
+	// If it is empty, then the host is picked by HostSelectionPolicy
+	hostID string
 }
 
 type queryRoutingInfo struct {
@@ -1440,6 +1445,20 @@ func (q *Query) borrowForExecution() {
 
 func (q *Query) releaseAfterExecution() {
 	q.decRefCount()
+}
+
+// SetHostID allows to define the host the query should be executed against. If the
+// host was filtered or otherwise unavailable, then the query will error. If an empty
+// string is sent, the default behavior, using the configured HostSelectionPolicy will
+// be used. A hostID can be obtained from HostInfo.HostID() after calling GetHosts().
+func (q *Query) SetHostID(hostID string) *Query {
+	q.hostID = hostID
+	return q
+}
+
+// GetHostID returns id of the host on which query should be executed.
+func (q *Query) GetHostID() string {
+	return q.hostID
 }
 
 // Iter represents an iterator that can be used to iterate over all rows that
@@ -2057,6 +2076,11 @@ func (b *Batch) releaseAfterExecution() {
 	// that would race with speculative executions.
 }
 
+// GetHostID satisfies ExecutableQuery interface but does noop.
+func (b *Batch) GetHostID() string {
+	return ""
+}
+
 type BatchType byte
 
 const (
@@ -2187,6 +2211,11 @@ func (t *traceWriter) Trace(traceId []byte) {
 	if err := iter.Close(); err != nil {
 		fmt.Fprintln(t.w, "Error:", err)
 	}
+}
+
+// GetHosts return a list of hosts in the ring the driver knows of.
+func (s *Session) GetHosts() []*HostInfo {
+	return s.ring.allHosts()
 }
 
 type ObservedQuery struct {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Support of sending queries to the specific node with Query.SetHostID() (CASSGO-4)
+
 ### Changed
 
 - Don't restrict server authenticator unless PasswordAuthentictor.AllowedAuthenticators is provided (CASSGO-19)
--- a/connectionpool.go
+++ b/connectionpool.go
@@ -243,6 +243,13 @@ func (p *policyConnPool) getPool(host *HostInfo) (pool *hostConnPool, ok bool) {
 	return
 }
 
+func (p *policyConnPool) getPoolByHostID(hostID string) (pool *hostConnPool, ok bool) {
+	p.mu.RLock()
+	pool, ok = p.hostConnPools[hostID]
+	p.mu.RUnlock()
+	return
+}
+
 func (p *policyConnPool) Close() {
 	p.mu.Lock()
 	defer p.mu.Unlock()


martin-sucha:
Should the API allow for other routing preferences? For example only a single data center/rack? Or be extensible to allow targeting a specific shard in the scylladb/gocql fork?

Should the API allow to route to a different host during retry, i.e. allow passing a `HostSelectionPolicy` or a similar interface/function?

tengu-alt (approved):

OleksiienkoMykyta (approved):

worryg0d:
> Should the API allow for other routing preferences? For example only a single data center/rack? Or be extensible to allow targeting a specific shard in the scylladb/gocql fork?
> 
> Should the API allow to route to a different host during retry, i.e. allow passing a `HostSelectionPolicy` or a similar interface/function?

`HostInfo` provides API to retrieve information like dc or rack for the host, so users may filter retrieved hosts manually before binding them to the query.

In my opinion, re-routing queries bound to the specific hosts to others is not a good idea even during retries, at least as default behavior. Sending queries to a **specific** host is not a typical use case and if we try to do it, we assume that the specified host is reachable. Or, at least, we expect that the query was attempted to be sent to the specified host.

The idea of this API being extensible sounds good, but I am not sure about implementation. May you have some advice or ideas on how to properly implement this?

joao-r-reis:
I think keeping this focused on the virtual table use case is best. Other routing preferences like shard awareness for scylla should probably be handled at the `HostSelectionPolicy` level. For this, we could add a function on Query and Batch that allows the `HostSelectionPolicy` to be provided instead of using the session level one but this can be handled on a separate JIRA and PR, it's a different use case than this one which is focused on virtual table queries that should be strongly associated with a specific host.

If a user sets the host using `Query.SetHost()` then either we disable retries and speculative executions altogether for that query or we allow retries and speculative executions only on that host (basically the host selection policy should never be used in the context of this query).

In the future if we add a way to provide a `HostSelectionPolicy` in addition to `.SetHost()` then the driver should probably return an error saying that you can't simultaneously set a host and provide a policy on the same query.

joao-r-reis:
> When the queryExecutor.executeQuery is called then it checks if the provided ExecutableQuery implements the hostGetter interface as well. We need this interface to not change the API ExecutableQuery. Also, the ExecutableQuery is implemented by Query and Batch as well, so it is good to avoid adding redundant methods for Batch as well.
> 
> So, if provided ExecutableQuery implements hostGetter, then it type-asserts it and calls the underlying method to get the host. If the host is not nil, it wraps the host into hostIter func which just returns the specified host.

The discussion around the `ExecutableQuery` interface and the `Query` API in general probably has a decent overlap with the work being dicussed and implemented in https://issues.apache.org/jira/browse/CASSGO-22 , justfyi.

query_executor.go
```diff
@@ -83,7 +83,28 @@ func (q *queryExecutor) speculate(ctx context.Context, qry ExecutableQuery, sp S
 }
 
 func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
-	hostIter := q.policy.Pick(qry)
+	type hostGetter interface {
```
joao-r-reis: How about adding a private interface like:

```
type cqlQuery interface {
    ExecutableQuery
    getHost() *HostInfo
}
```

And make gocql internal code reference `cqlQuery` (name subject to change?) instead of `ExecutableQuery` so that we don't have to worry about breaking the public `ExecutableQuery` interface when we add new functions to `Query` or `Batch`?

joao-r-reis: Alternatively, we could just add new functions to `ExecutableQuery` because there's no place where a user can provide a custom implementation of `ExecutableQuery` to the driver is there? If the driver API doesn't expose any function that receives an object of type `ExecutableQuery` then we should be fine changing this interface since there's no real use case where a user might implement it on their application.

joao-r-reis: From a quick glance at the code, `ExecutableQuery` is provided to implementations of `HostSelectionPolicy` but it's the driver that calls `HostSelectionPolicy.Pick(ExecutableQuery)` with `Query` or `Batch` objects as parameters so I don't see an issue in adding `GetHost` to `ExecutableQuery`

worryg0d: I don't mind extending `ExecutableQuery` interface, but adding private methods is not the best thing to me. The driver exposes `HostSelectionPolicy` interface and allows user to provide their own implementations, but users cannot properly test them because `Pick()` accepts  `ExecutableQuery` that cannot be mocked for testing purposes due to private methods inside... Ofc they can just pass `Query` and `Batch` objects, but I'm unsure if it is enough for testing

joao-r-reis: These private methods would not be used in `HostSelectionPolicy` because `HostSelectionPolicy` declares the parameter as `ExecutableQuery` not `cqlQuery` or whatever the name would be for the private interface.

joao-r-reis: A user would be able to continue testing custom implementations of `HostSelectionPolicy`

joao-r-reis: I was suggesting changing `ExecutableQuery` references to the private interface only on internal code, `HostSelectionPolicy` is not internal so it would keep the `ExecutableQuery` declaration.

worryg0d: > Alternatively, we could just add new functions to ExecutableQuery

I was thinking about the second approach

joao-r-reis: But if we just add  `GetHost` to `ExecutableQuery` we don't need a private interface anyway so we can just go with that.

worryg0d: But it still remains untestable because `ExecutableQuery` already has some private methods. But this is out of the scope of this PR. Ok, I'm going to update implementation with this one.

One thing came in mind, this PR doesn't provide this API for `Batch`. Should I add this API to `Batch` to be consistent with the `Query`? This PR focuses on virtual tables, and main purpose of Batch is data modification, so it actually cannot be used with virtual tables...

worryg0d: I updated the implementation by adding `GetHost()` to `ExecutableQuery`.

Also, I looked over my implementation and found a bug leading to endless query execution... It appears when the host passed to `SetHost()` is not UP. I changed the behavior [here](https://github.com/worryg0d/gocql/blob/c1e46be251fc5779a2cb6a8a52cee03a8a161275/query_executor.go#L148) not to pick the next `selectedHost` when the selected one is not up and `GetHost()` returned no host.

query_executor.go
```diff
@@ -129,12 +145,13 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
...
 
 	var lastErr error
 	var iter *Iter
 	for selectedHost != nil {
 		host := selectedHost.Info()
-		if host == nil || !host.IsUp() {
+		if (host == nil || !host.IsUp()) && specifiedHost == nil {
 			selectedHost = hostIter()
 			continue
 		}
```
joao-r-reis: we should probably add another `if` here to exit right away if specified host is `!= nil` and the host is down

query_executor.go
joao-r-reis: > But it still remains untestable because ExecutableQuery already has some private methods.

I know it's out of scope of this PR but I want to understand this, what private methods are you referencing?

joao-r-reis: > One thing came in mind, this PR doesn't provide this API for Batch. Should I add this API to Batch to be consistent with the Query? This PR focuses on virtual tables, and main purpose of Batch is data modification, so it actually cannot be used with virtual tables...

I think it's fine if we add SetHost for query only, I don't see a use case for having it on batches atm

worryg0d: >  what private methods are you referencing?

Such methods as `execute()`, `attempt()`, etc.

joao-r-reis: Oh you're right, I didn't notice them. 👍 So it wouldn't actually make much of a difference if we added a private getHosts method there but we can keep it public

worryg0d: The only issue of making `GetHost()` public is a requirement to implement it for `Batch` only to satisfy the interface. So there will be a public API `Batch.GetHost()` which actually does noop. Maybe the first approach with internal query interface for internal use and the public one for exposing is better...

joao-r-reis: I think it's fine if `Batch.GetHost()` does a noop honestly, there's also a possibility that we might make it possible to set the host on batch queries in the future

worryg0d: Then let's leave this one as it is for now

query_executor.go
worryg0d: It makes sense not to send the query if the host is down. I added a check before this if that immediately returns an `ErrNoConnections` when the specified host is down.

joao-r-reis: 👍

joao-r-reis (approved):

worryg0d:
Force pushed due to conflicts with the trunk

jameshartig:
Once we introduce a new API it's very hard to change/remove it. Do we really think that accepting a `*HostInfo` is the right argument for `SetHost`? If you already had a host id there's no way for a user to construct a `*HostInfo`.

worryg0d:
Hello @jameshartig, thanks a lot for your feedback.

> If you already had a host id there's no way for a user to construct a *HostInfo.

You have a point, I really forgot that we can't easily create `HostInfo` obj outside the driver. Then I see two ways how to solve it:
1. Add `NewHostInfo(hostId string)` constructor. It is the easiest solution and `hostId` is enough to get conn pool from `queryExecutor.pool`.
2. Introduce a new type `Host` which will mostly duplicate `HostInfo` with public fields. We can't make the fields of `HostInfo` public because they're protected by the mutex. With a new type users will able to easily construct `Host` and users may easily get access to the host metadata with `session.GetHosts() ([]Host, err)`.

Honestly, I'd prefer the first one because with the second we'll have two `Host` and `HostInfo` types that both hold the same fields, but the one is used for internal and can't be created outside gocql and another that is mostly the same but is not used internally...

jameshartig:
> 1. Add `NewHostInfo(hostId string)` constructor. It is the easiest solution and `hostId` is enough to get conn pool from `queryExecutor.pool`

Is there a reason we wouldn't just make `SetHost` accept a `hostId` instead? That seems simpler? It would be easy to call with `*HostInfo` and you don't need a `*HostInfo` if you don't have one.

worryg0d:
> Is there a reason we wouldn't just make SetHost accept a hostId instead?

Honestly, I didn't think about it... I was refering to how such API is implemented in java driver, but we don't have to implement it the same way. I don't mind to reimplement it to use `string` instead.

@joao-r-reis What do you think about this? You've already approved this PR but we're going to change the implementation so we should agree on the proposal

joao-r-reis:
Honestly I'm fine with either, on one hand providing a host struct is a bit more consistent with other drivers I think but on the other hand making the parameter a host id seems simpler to use from the user's POV so if James prefers the host id approach then +1 for that one

joao-r-reis:
I will say that making it a string will make it more error prone because the user might use the wrong string format of the uuid. This will not be an issue if the user just calls `hostInfo.HostID()` but at that point might as well just make the parameter a `*hostInfo` instead. Wdyt about this @jameshartig

jameshartig:
> I will say that making it a string will make it more error prone because the user might use the wrong string format of the uuid. This will not be an issue if the user just calls `hostInfo.HostID()` but at that point might as well just make the parameter a `*hostInfo` instead. Wdyt about this @jameshartig

Hmm that is a good point, but by only accepting a `*HostInfo` the only way to use it is by calling `GetHosts` and then looping (and optionally filtering) yourself. That's not the end of the world but it does seem less than ideal.

jameshartig:
@worryg0d Actually thinking about it more, I dislike that `GetHosts` requires 2 round-trips, and that seems excessive to always require when you want to choose a host, especially if you already have the `HostId`. I don't think we should be encouraging people to hold onto `*HostInfo`'s which they might do if the only way to get one is calling `GetHosts`.

I'm doubling down on making it accept a `HostId string` and we can just look in the pool for that hostId and not need to do any round-trips.

We should mention in the method comment for `SetHost` that it's recommended to get the hostId from `*HostInfo.HostID()` to encourage people use it but it's not required.

worryg0d:
OK, I'll update this PR tomorrow. Once we're going to change it to take the host id then should it be something like `SetHostID(hostID string)`?. I am fine with both variants, but the second one has better semantics here

jameshartig:
> OK, I'll update this PR tomorrow. Once we're going to change it to take the host id then should it be something like `SetHostID(hostID string)`?. I am fine with both variants, but the second one has better semantics here

I like `SetHostID` because it leaves open the ability to add `SetHost` later if we needed to.

worryg0d:
I pushed [73590c2](https://github.com/apache/cassandra-gocql-driver/pull/1793/commits/73590c27f9d1e437983c1ff31b9ab52fed20c717) which changes of `SetHost()` -> `SetHostID(string)`.  

@jameshartig @joao-r-reis Could you please take a look at this? Thanks a lot!

session.go
```diff
@@ -1430,6 +1435,22 @@ func (q *Query) releaseAfterExecution() {
...
 }
 
+// SetHostID allows to define on which host the query should be executed.
+// If hostID is not set, then the HostSelectionPolicy will be used to pick a host.
+// It is recommended to get host id from HostInfo.HostID(), but it is not required.
+// It is strongly recommended to use WithContext() with this option because
+// if the specified host goes down during execution, the driver will try to send the query to this host until it succeeds
+// which may lead to an endless execution.
+func (q *Query) SetHostID(hostID string) *Query {
+	q.hostID = hostID
```
joao-r-reis, line 1455: I think we should check if it's a non empty string because if it's empty then the query will be executed as if `SetHostID()` was never called which is odd behavior. Either that or we change `q.hostID` to be `*string` so we can check if it was actually set or not

session.go
```diff
@@ -2177,6 +2203,15 @@ func (t *traceWriter) Trace(traceId []byte) {
 	}
 }
 
+// GetHosts returns a list of hosts found via queries to system.local and system.peers
+func (s *Session) GetHosts() ([]*HostInfo, error) {
```
joao-r-reis: Do we have no way of getting the hosts currently?

Also, we should document that this method performs 2 round trips, some users might be used to other drivers caching the hosts list.

query_executor.go
```diff
@@ -83,7 +84,27 @@ func (q *queryExecutor) speculate(ctx context.Context, qry ExecutableQuery, sp S
...
+	// check if the host id is specified for the query,
+	// if it is, the query should be executed at the corresponding host.
+	if hostID := qry.GetHostID(); hostID != "" {
+		pool, ok := q.pool.getPoolByHostID(hostID)
+		if !ok || !pool.host.IsUp() {
+			return &Iter{err: ErrNoConnections}, nil
+		}
+		hostIter = func() SelectedHost {
+			// forcing hostIter to always return the same host
+			// it makes any retries and speculative executions run on the specified host
```
joao-r-reis: Shouldn't we just prevent retries and speculative executions when the host is set? Having potentially endless executions on the same host doesn't look right to me.

query_executor.go
```diff
@@ -166,6 +187,8 @@ func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery, hostIter Ne
 
 		// Exit if the query was successful
 		// or query is not idempotent or no retry policy defined
+		// Also, if there is specified host for the query to be executed on
+		// and query execution is failed we should exit
```
joao-r-reis: Why was this comment added here? Is this the relevant place for it?

session.go
worryg0d: > Do we have no way of getting the hosts currently?

As I remember the driver doesn't expose API for this purpose. Users can get this via policies like `HostSelectionPolicy`, but directly no.

> some users might be used to other drivers caching the hosts list

I think more about how we can avoid 2 round trips and we can probably take known for the driver hosts from the pool `Session.executor.pool.hostConnPools`. Each pool contains a reference to `HostInfo` for which it is created.

query_executor.go
worryg0d: > Having potentially endless executions on the same host doesn't look right to me

Yea I don't like this one too. It makes sense to prevent speculative executions when we use this API, but disabling retry policies... Probably it makes sense once we use it we assume that the host is reachable.

However, providing context via `WithContext()` will prevent endless executions.

session.go
worryg0d: That's a good point. If we add a check then we should panic if it is empty which is probably not the best, but using `*string` seems to be fine

query_executor.go
worryg0d: This is left from the previous diff when it used `*HostInfo` and it should be removed. Thanks

cassandra_test.go
```diff
@@ -3338,3 +3338,56 @@ func TestQuery_NamedValues(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestQuery_SetHost(t *testing.T) {
```
jameshartig: `TestQuery_SetHostID`

query_executor.go
```diff
@@ -41,6 +41,7 @@ type ExecutableQuery interface {
 	Keyspace() string
 	Table() string
 	IsIdempotent() bool
+	GetHostID() string
```
jameshartig, line 44: Another option here rather than changing the `ExecutableQuery` is to do:

```go

type hostIDQuery interface {
  GetHostID() string
}
```

then later in `executeQuery` you'd do:

```go
var hostID string
if hostIDQry, ok := qry.(hostIDQuery); ok {
  hostID = hostIDQry.GetHostID()
}
```

Then we wouldn't need to break this interface and worry about modifying batch.

query_executor.go
jameshartig: Yeah I think retries makes sense. That can always be overwritten if they don't want that.

session.go
jameshartig: Passing a pointer seems a bit awkward to use and is very unusual given all of the other functions in this package. What happens if you call `RoutingKey(nil)` or `PageState(nil)`, etc. Setting it to empty could be a way to unset an existing HostID if one was set for some reason prior. When queries are immutable it could be useful to get a copy of a query without a specific HostID that previously had one set.

We should just document that sending an empty string will restore the default behavior.

session.go
jameshartig: We can't use the pool since it's filtered by the hostSelectionPolicy. We can use the session's `ring.allHosts()` method though.

jameshartig (changes requested):

cassandra_test.go
worryg0d: Good catch. Thanks

query_executor.go
worryg0d: It was the first approach I've proposed. We discussed this with @joao-r-reis. If you have an opinion on this, let's discuss it [here](https://github.com/apache/cassandra-gocql-driver/pull/1793#discussion_r1850752379).

TL;DR
We have two solutions how to avoid type casts and make the implementation more obvious:
1. Extend `ExecutableQuery` interface, because it can't be implemented outside gocql due to private methods like `execute`, so we cannot consider it a breaking change.
2. Extract private methods from `ExecutableQuery` to an internal interface which also should implement `ExecutableQuery` and use it internally.

session.go
worryg0d: I spent some time thinking about this more and I'm +1 on what @jameshartig says. We at least should have a way to restore `Query.hostID` to default. 

If `SetHostID` takes `*string` it is available, but this makes UX of this API way worse IMO. I don't like the idea of creating vars to hold some strings and pass a pointer to them. However, if API is `SetHostID(string)` there is no way to restore its behavior to default. Once `SetHostID()` is used, it can't be changed. Ofc we can expose API `Query.Default()` to restore this behavior, but this becomes something very odd to me.

All of this is based on the case when we want to copy the existing `Query` and somehow modify it, but I'm unsure how this matches query immutability. By this do we understand that once the `Query` obj is created it can't be modified? Either it means that the driver doesn't modify `Query` internally during its execution? As far as I know, gocql writes some metrics to the query obj.

session.go
worryg0d: Oh, I didn't realize `ring.allHosts()` exists. On any node events ring is updated so `ring.allHosts()` will return updated hosts. Thanks a lot 👍

cassandra_test.go
worryg0d: It's done

session.go
worryg0d: Changed this to use `ring.allHosts()`. However, I'm unsure if the comment I wrote is enough... `ring.allHosts()` returns only hosts known by the driver, which means if `ClusterConfig.HostFilter` is presented, `ring` won't have filtered hosts.

query_executor.go
worryg0d: Removed 🙌

query_executor.go
joao-r-reis: Yeah I don't think we need to treat `ExecutableQuery` as a "public" interface to be implemented by users, it's a bit messy right now and I actually want to spend some time investigating how we could clean up this part of the API as part of https://github.com/apache/cassandra-gocql-driver/pull/1848 ideally in time for the 2.0 release

session.go
joao-r-reis: When I suggested the pointer I wasn't actually saying we should make the parameter a pointer, I was saying we could make the private field a pointer so we know when the user actually set the value or not. The parameter would still be `string`.

joao-r-reis: but yeah I think we can just document it instead of checking and failing. To return an error we would have to change the method signature which would be awkward (or panic'ing but that's just a bad idea anyway)

query_executor.go
joao-r-reis: I'm more worried that someone might run into endless executions in production, face some kind of outage and only then realize that they need to provide context to ensure the goroutine doesn't get stuck on a potentially infinite loop. But I think in most cases the retry policy will have a max number of retries anyways so that's good enough for me, let's just make sure this is documented in the `SetHostID` method

worryg0d: I looked at `queryExecutor.do` more properly and discovered that context cancelation indeed won't protect users from endless executions here.

On line [158](https://github.com/apache/cassandra-gocql-driver/blob/4a3a53b65f0726bc30fe14bac9be9baa8497c8ea/query_executor.go#L158) will `continue` loop iteration if the specified host is down, so it never reaches [179](https://github.com/apache/cassandra-gocql-driver/blob/4a3a53b65f0726bc30fe14bac9be9baa8497c8ea/query_executor.go#L179).

This is not a big problem if we're going to disable retries when the `SetHostID` API is used. For this case `hostIter` should return the specified host only once.  Either `queryExecutor.do` should be changed

joao-r-reis: Yeah I think it makes a bit more sense for `hostIter` to just return the host once but retries might not work in this case right?

worryg0d: Right, if we do this way retries will be disabled

jameshartig: > On line [158](https://github.com/apache/cassandra-gocql-driver/blob/4a3a53b65f0726bc30fe14bac9be9baa8497c8ea/query_executor.go#L158) will continue loop iteration if the specified host is down, so it never reaches [179](https://github.com/apache/cassandra-gocql-driver/blob/4a3a53b65f0726bc30fe14bac9be9baa8497c8ea/query_executor.go#L179).

Can you file that as a bug? That could affect the normal flow as well separate from this. We should be tracking the last host and maybe erroring if we get the same host twice. I don't think we need to alter this just because of that specific case.

> Right, if we do this way retries will be disabled

I still think we should be aiming to make retries work, even if its in a follow-up issue. There are transient things that happen in the real world that might be annoying to handle if we just always disable retries.

query_executor.go
jameshartig: That makes sense. I was just suggesting so we didn't have to have a weird case for Batch. I think its fine to add and we can aim to improve it in a future MR.

session.go
jameshartig: I'm thinking we _could_ call it `GetAllHosts` to clarify that it will get all hosts but I'm not sure if that's strictly necessarily. We could call the host filter within the method when we're looping but tbh I'm not sure if users want it filtered or not. I think until we get feedback we just need to pick a path and be prepared to add another method (either add GetAllHosts later or GetHosts later that does filtering). @joao-r-reis do you have thoughts on the filtering vs not?

query_executor.go
jameshartig: > That could affect the normal flow as well separate from this.

I guess, correctly-behaving selection policies should already be returning not-down hosts but idk if that's guaranteed. There's also another `continue` on line 166 and 172 that seem like they could happen as well and cause an infinite loop. I think there are a few solutions to this but again, I think that can be explored in another MR.

worryg0d: > I still think we should be aiming to make retries work, even if its in a follow-up issue. There are transient things that happen in the real world that might be annoying to handle if we just always disable retries.

I agree, we can't just force disable them. If users really need to disable them, they can do it. There is a way to do it via `Idempotent()` and `RetryPolicy()`.

> We should be tracking the last host and maybe erroring if we get the same host twice

If we do so, it will break `DowngradingConsistencyRetryPolicy` which returns the retry type `Retry` which indicates that the driver should retry on the same host. It is described [here](https://github.com/apache/cassandra-gocql-driver/blob/4a3a53b65f0726bc30fe14bac9be9baa8497c8ea/query_executor.go#L201)

joao-r-reis: > We should be tracking the last host and maybe erroring if we get the same host twice.

There's nothing inherently wrong with a query plan that contains the same host more than once though, some apps could even rely on it potentially, the driver does support it currently if you provide your own policy

joao-r-reis: > correctly-behaving selection policies should already be returning not-down hosts but idk if that's guaranteed. 

We should definitely have that check internally in case a policy returns a downed host, a small optimization that doesn't cost anything basically.

I do think we can explore all of these things on a separate issue/PR like @jameshartig is suggesting

worryg0d: > I do think we can explore all of these things on a separate issue/PR

I have nothing against it

So, we're leaving this as potentially risky until this issue will be resolved in other PR

worryg0d:
Rebased on trunk and force pushed due to a weird cancellation of the workflow run.

query_executor.go
joao-r-reis: > So, we're leaving this as potentially risky until this issue will be resolved in other PR

Yeah I'm ok with that, can you document these issues/risks on a new bug ticket?

worryg0d: Yes, no problem

worryg0d: [CASSGO-50](https://issues.apache.org/jira/browse/CASSGO-50)

session.go
```diff
@@ -1442,6 +1447,23 @@ func (q *Query) releaseAfterExecution() {
...
+// if the specified host goes down during execution, the driver will try to send the query to this host until it succeeds
+// which may lead to an endless execution.
+// Setting hostID to empty restores behavior to default.
+func (q *Query) SetHostID(hostID string) *Query {
+	q.hostID = hostID
+	return q
+}
+
+// GetHostID returns id of the host on which query should be executed.
+func (q *Query) GetHostID() string {
```
ribaraka, line 1460: I'd name this method `GetTargetHostId` to make it more specific.

session.go
```diff
@@ -949,6 +950,10 @@ type Query struct {
 
 	// routingInfo is a pointer because Query can be copied and copyable struct can't hold a mutex.
 	routingInfo *queryRoutingInfo
+
+	// hostID specifies the host on which the query should be executed.
+	// If it is empty, then the host is picked by HostSelectionPolicy
+	hostID string
```
ribaraka, line 956: how about `queryTargetHostId` or `targetHostID`?

cassandra_test.go
```diff
@@ -3347,3 +3346,53 @@ func TestQuery_NamedValues(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestQuery_SetHostID(t *testing.T) {
+	// This test ensures that queries are sent to the specified host only
```
ribaraka: nitpick: It might be better to keep comments above the test function to clearly describe its overall purpose.

session.go
ribaraka: If you like it, then `SetTargetHostID` might be changed as well :smile:

session.go
ribaraka: If `HostFilter` is causing issues, consider mentioning it in the documentation. For example: '`SetHostID` will not work on filtered nodes (i.e., nodes excluded by `ClusterConfig.HostFilter`)

query_executor.go
worryg0d: I spent more time on this and found a way to avoid endless executions when `SetHostID()` is used with the downed host. I pushed in on my other branch [here](https://github.com/apache/cassandra-gocql-driver/commit/0e44de86776a74500aaae5f9cf18d9433a755e23). Could you please take a look and share your thoughts? 

Ofc this issue is still relevant when `HostSelectionPolicy` returns downed nodes. At least, this will resolve the described issue with `SetHostID()`. Thanks a lot to @ribaraka who pointed to this approach.

session.go
worryg0d: Hmm, they both have better semantics... If we want to be consistence with the same API in other drivers it should be `SetHost()`, but we don't really have to make the same...

jameshartig: I don't know if `Target` is specifically necessary since it seems kind of obvious what the point of the host is. I do agree that staying closer to the java API is probably ideal, which `SetHostID` tries to do.

session.go
jameshartig: it's an internal field so we can call it whatever. again, i think that it's obvious enough as-is without "target" but i'm fine internally if there's a need to clarify.

cassandra_test.go
jameshartig: Agreed

session.go
jameshartig: The comments for this method are getting lengthly. How about:


```go
// SetHostID allows to define the host the query should be executed against. If the
// host was filtered or otherwise unavailable, then the query will error. If an empty
// string is sent, the default behavior, using the configured HostSelectionPolicy will
// be used. A hostID can be obtained from HostInfo.HostID() after calling GetHosts().
```

I removed the WithContext part because we plan to fix that in a follow-up issue.

jameshartig (changes requested):

session.go
worryg0d: I prefer to leave `hostID` to be consistent with method names

session.go
worryg0d: Thanks a lot. It is much better!

> I removed the WithContext 

Oh I completely forgot about this, thanks lots!

query_executor.go
joao-r-reis: I think that approach is fine

worryg0d: If @jameshartig doesn't have any concerns about the proposed change I'll update this PR

jameshartig (approved):

jameshartig:
@worryg0d can you squash these commits and update the commit message to add my name to approvers

worryg0d:
@jameshartig I squashed commits and updated the message. Before merging it to the main could you please take a look at [this](https://github.com/apache/cassandra-gocql-driver/pull/1793#discussion_r1941037021)? There is an approach to avoid deadlocks when a downed host is selected by `SetHostID()`.

query_executor.go
jameshartig: @worryg0d I think those changes are fine. I don't like needing to check `IsUp` because that's already checked in the executor but that would trigger the infinite loop if you didn't so it's fine and we'll fix the loop separately.

jameshartig:
> @jameshartig I squashed commits and updated the message. Before merging it to the main could you please take a look at https://github.com/apache/cassandra-gocql-driver/pull/1793#discussion_r1941037021? There is an approach to avoid deadlocks when a downed host is selected by SetHostID().

Great! Yes, let's get those included in this and I'll approve.

worryg0d:
It's done! Could you please take a look at this?

jameshartig (approved):

query_executor.go
joao-r-reis: Do you guys think we need to address CASSGO-50 before 2.0 is out or can it be moved to a follow up 2.x minor release?

worryg0d: Well, the 2.0 release is already a bit overloaded, but if anybody has the extra capacity to work on this it would be nice to address this before the release.

Leaving this is not a good idea, because it may affect the end users and their applications.

@tengu-alt @OleksiienkoMykyta If it won't bother could you please take a look at CASSGO-50?

tengu-alt: > could you please take a look at [CASSGO-50](https://issues.apache.org/jira/browse/CASSGO-50)?

I can handle it.

PR #1794, opened 2024-08-08
Virtual keyspace metadata support immplementation

tengu-alt:
This PR provides access to the virtual metadata tables exposed through the `system_virtual_schema` keyspace.

Example query:
```CQL
cqlsh> SELECT * FROM system_virtual_schema.tables;

 keyspace_name         | table_name                       | comment
-----------------------+----------------------------------+----------------------------------------------------
          system_views |                    batch_metrics |               Metrics specific to batch statements
          system_views |                           caches |                                      system caches
          system_views |    cidr_filtering_metrics_counts |           Count metrics specific to CIDR filtering
          system_views | cidr_filtering_metrics_latencies |         Latency metrics specific to CIDR filtering
          system_views |                          clients |                        currently connected clients 
```

The `system_virtual_schema` keyspace contains `keyspaces`, `tables`, and `columns` that users can now parse by specifying a keyspace.

Users can retrieve parsed tables using the `VirtualKeyspaceMetadata(keyspace string)` method for the `Session`.

The implementation is based on how the driver retrieves regular `keyspace metadata`.

### DIFF
--- a/metadata.go
+++ b/metadata.go
@@ -50,6 +50,12 @@ type KeyspaceMetadata struct {
 	UserTypes         map[string]*UserTypeMetadata
 }
 
+// schema metadata for a virtual keyspace
+type VirtualKeyspaceMetadata struct {
+	Name   string
+	Tables map[string]*VirtualTableMetadata
+}
+
 // schema metadata for a table (a.k.a. column family)
 type TableMetadata struct {
 	Keyspace          string
@@ -66,6 +72,13 @@ type TableMetadata struct {
 	OrderedColumns    []string
 }
 
+type VirtualTableMetadata struct {
+	Keyspace string
+	Name     string
+	Comment  string
+	Columns  map[string]*VirtualColumnMetadata
+}
+
 // schema metadata for a column
 type ColumnMetadata struct {
 	Keyspace        string
@@ -78,6 +91,15 @@ type ColumnMetadata struct {
 	ClusteringOrder string
 	Order           ColumnOrder
 	Index           ColumnIndexMetadata
+}
+
+type VirtualColumnMetadata struct {
+	Keyspace        string
+	Table           string
+	Name            string
+	ClusteringOrder string
+	Kind            ColumnKind
+	Type            TypeInfo
 }
 
 // FunctionMetadata holds metadata for function constructs
@@ -231,12 +253,28 @@ type schemaDescriber struct {
 	cache map[string]*KeyspaceMetadata
 }
 
+// queries the cluster for schema information for a virtual keyspace
+type virtualSchemaDescriber struct {
+	session *Session
+	mu      sync.Mutex
+	cache   map[string]*VirtualKeyspaceMetadata
+}
+
 // creates a session bound schema describer which will query and cache
 // keyspace metadata
 func newSchemaDescriber(session *Session) *schemaDescriber {
 	return &schemaDescriber{
 		session: session,
 		cache:   map[string]*KeyspaceMetadata{},
+	}
+}
+
+// creates a session bound schema describer which will query and cache
+// virtual keyspace metadata
+func newVirtualSchemaDescriber(session *Session) *virtualSchemaDescriber {
+	return &virtualSchemaDescriber{
+		session: session,
+		cache:   map[string]*VirtualKeyspaceMetadata{},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


metadata.go
```diff
@@ -760,6 +879,37 @@ func getTableMetadata(session *Session, keyspaceName string) ([]TableMetadata, e
...
 }
 
+// query for only the table metadata in the specified keyspace from system_virtual_schema.tables
+func getVirtualTableMetadata(s *Session, keyspaceName string) ([]VirtualTableMetadata, error) {
+	const stmt = `
+		SELECT
+				table_name,
+				comment
+			FROM system_virtual_schema.tables
+			WHERE keyspace_name = ?`
```
dkropachev: Can you please fix indent:
```suggestion
	const stmt = `
		SELECT
			table_name,
			comment
		FROM system_virtual_schema.tables
		WHERE keyspace_name = ?`
```

metadata.go
```diff
@@ -947,6 +1097,52 @@ func getColumnMetadata(session *Session, keyspaceName string) ([]ColumnMetadata,
...
+func getVirtualColumnMetadata(s *Session, keyspaceName string) ([]VirtualColumnMetadata, error) {
+	const stmt = `
+		SELECT
+				table_name,
+				column_name,
+				clustering_order,
+				kind,
+				type
+			FROM system_virtual_schema.columns
+			WHERE keyspace_name = ?`
```
dkropachev: Same ident problem:
```suggestion
	const stmt = `
		SELECT
			table_name,
			column_name,
			clustering_order,
			kind,
			type
		FROM system_virtual_schema.columns
		WHERE keyspace_name = ?`
```

metadata.go
```diff
@@ -269,6 +307,23 @@ func (s *schemaDescriber) getSchema(keyspaceName string) (*KeyspaceMetadata, err
 	return metadata, nil
 }
 
+// returns the cached VirtualKeyspaceMetadata held by the describer for the named
+// keyspace.
+func (s *virtualSchemaDescriber) getSchema(keyspaceName string) (*VirtualKeyspaceMetadata, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	metadata, found := s.cache[keyspaceName]
```
dkropachev, line 306: Shouldn't you invalidate it at some point.
Say you upgrade cluster from one version to another, driver will still keep old schema.

Probably on `handleNodeUp` you could read `release_version` `select release_version from system.local where key = 'local'` and invalidate or update cache if it has changed.

Another problem if your cluster half on new version half on another.
Now you depends on where have your control connection landed, if it has landed on a node with new version, session will give you a new schema, while old nodes don't support it.
Way to solve it is to run queries not on control connection, but rather on the node with oldest version available, if you were do so, you can't pick all the nodes from the cluster, you have to adhere to `HostSelectionPolicy`, say if it is `dcAwareRR` you will need to filter out all nodes from other datacenters.

metadata.go
```diff
@@ -240,6 +262,13 @@ type schemaDescriber struct {
 	cache map[string]*KeyspaceMetadata
 }
 
+// queries the cluster for schema information for a virtual keyspace
+type virtualSchemaDescriber struct {
+	session *Session
+	mu      sync.Mutex
+	cache   map[string]*VirtualKeyspaceMetadata
```
dkropachev, line 260: I would recommend, in this particular case, use `sync.Map` instead of Mutex + map.

dkropachev (changes requested):

metadata.go
tengu-alt: agree, fixed

metadata.go
tengu-alt: fixed

metadata.go
tengu-alt: I agree with your point about invalidation, but virtual metadata is only requested on the user's purpose, not by some node event, also virtual tables can't be changed, if we talk about some `schemaAgreement` case. In my opinion, it is unnecessary to invalidate it on every `getSchema` request, and i think it is unnecessary to remove a control connection, it is good to keep the existing pattern of the metadata requests.

metadata.go
tengu-alt: I think it is preferable to use mutex, to keep the existing pattern, and avoid unnecessary code complexity.

metadata.go
dkropachev: If you clear cache a the betting of `getSchema` it means that there is going to be no cache, why just not drop it.
If you drop the cache then there is no point in having `virtualSchemaDescriber`, you can just move code to `Session`.

tengu-alt: I totally agree with you, but in my opinion, it is good to keep an existing pattern of processing the metadata (virtual or not)

Issue #1774, opened 2024-06-24
Trying to get cassandra active, idle, inUse connection


md-robiul-hassan-kr:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?
 Apache Cassandra 4.1 

### What version of Gocql are you using?
1.6.0

### What version of Go are you using?
go version go1.22.3 darwin/arm64

### What did you do?
Trying to get cassandra active, idle, inUse connection. 

### What did you expect to see?
get cassandra active, idle, inUse connection. 

### What did you see instead?
Nothing found. 
---

If you are having connectivity related issues please share the following additional information

### Describe your Cassandra cluster
please provide the following information

- Get active connections
- Get Idle connections
- Get in-use connection.

testisnullus:
Hello, @md-robiul-hassan-kr! Could you please add some context to the issue?

I assume that you want to be able to query the Cassandra active connections using the gocql driver, specifically the client connections, including IP addresses and connection times, am I right?

If we are going to think about idle connections, we can assume that these are established network connections between clients and Cassandra nodes that are not currently being used for read or write operations.

Also, in-use connections are established network connections between clients and a Cassandra node that are actively being used for executing read or write operations for the measured period of time.

Could you please explain what feature you want gocql to have to accomplish your request, please?

md-robiul-hassan-kr:
hi @testisnullus, your assumption is correct. I want to be able to know how many active, in-use, or idle connect I have to the database at a given time, so that I can use those metrics. 

A good example is [here](https://docs.datastax.com/en/developer/java-driver/4.9/manual/core/metrics/index.html) where those metrics are available through JVM, and I was wondering if there could be possibility to use from the gocql lib. 

```
datastax-java-driver.advanced.metrics {
  session.enabled = [ connected-nodes, cql-requests ]
  node.enabled = [ pool.open-connections, pool.in-flight ]
}
```

I would like to get the information for the connection pool mostly, so that can be used in grafana with prometheus.

OleksiienkoMykyta:
Reviewed related PR and reminded maintainers to check it. I hope it will speed up the issue resolution.

PR #1796, opened 2024-08-12
issue-1774: Method to retrieve info about client connections was added

testisnullus:
In this PR a new custom type `connectionType` was added to categorize different stages of client connections within a Cassandra cluster. Additionally, the `ClientConnection` struct was introduced to encapsulate detailed information about each connection, such as client address, connection stage, driver details, and SSL configuration. A new method, `RetrieveClientConnections`, was implemented to retrieve client connections based on their connection stage from the system_views.clients table.

### DIFF
--- a/session.go
+++ b/session.go
@@ -800,6 +800,95 @@ func (s *Session) MapExecuteBatchCAS(batch *Batch, dest map[string]interface{}) 
 	return applied, iter, iter.err
 }
 
+// connectionType is a custom type that represents the different stages
+// of a client connection in a Cassandra cluster. It is used to filter and categorize
+// connections based on their current state.
+type connectionType string
+
+const (
+	Ready      connectionType = "ready"
+	Connecting connectionType = "connecting"
+	Idle       connectionType = "idle"
+	Closed     connectionType = "closed"
+	Failed     connectionType = "failed"
+)
+
+// ClientConnection represents a client connection to a Cassandra node. It holds detailed
+// information about the connection, including the client address, connection stage, driver details,
+// and various configuration options.
+type ClientConnection struct {
+	Address         string
+	Port            int
+	ConnectionStage string
+	DriverName      string
+	DriverVersion   string
+	Hostname        string
+	KeyspaceName    *string
+	ProtocolVersion int
+	RequestCount    int
+	SSLCipherSuite  *string
+	SSLEnabled      bool
+	SSLProtocol     *string
+	Username        string
+}
+
+// RetrieveClientConnections retrieves a list of client connections from the
+// `system_views.clients` table based on the specified connection type. The function
+// queries the Cassandra database for connections with a given `connection_stage` and
+// scans the results into a slice of `ClientConnection` structs. It handles nullable
+// fields and returns the list of connections or an error if the operation fails.
+func (s *Session) RetrieveClientConnections(connectionType connectionType) ([]*ClientConnection, error) {
+	const stmt = `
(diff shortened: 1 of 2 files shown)


session.go
```diff
@@ -780,6 +780,87 @@ func (s *Session) MapExecuteBatchCAS(batch *Batch, dest map[string]interface{})
...
+	Username        string  `json:"username"`
+}
+
+// RetrieveClientConnections retrieves a list of client connections from the
+// `system_views.clients` table based on the specified connection type. The function
+// queries the Cassandra database for connections with a given `connection_stage` and
+// scans the results into a slice of `ClientConnection` structs. It handles nullable
+// fields and returns the list of connections or an error if the operation fails.
+func (s *Session) RetrieveClientConnections(connectionType connectionType) ([]*ClientConnection, error) {
+	query := "SELECT address, port, connection_stage, driver_name, driver_version, hostname, keyspace_name, protocol_version, request_count, ssl_cipher_suite, ssl_enabled, ssl_protocol, username FROM system_views.clients WHERE connection_stage = ?"
```
worryg0d: I think we should follow the general pattern of code styling for similar methods that are used for querying some system/system_views tables. So you should define this query as a string const and use idents as well. You may find some examples in the `Session.KeyspaceMetadata()` method

session.go
```diff
@@ -780,6 +780,87 @@ func (s *Session) MapExecuteBatchCAS(batch *Batch, dest map[string]interface{})
...
+	DriverVersion   string  `json:"driver_version"`
+	Hostname        string  `json:"hostname"`
+	KeyspaceName    *string `json:"keyspace_name"`
+	ProtocolVersion int     `json:"protocol_version"`
+	RequestCount    int     `json:"request_count"`
+	SSLCipherSuite  *string `json:"ssl_cipher_suite"`
+	SSLEnabled      bool    `json:"ssl_enabled"`
+	SSLProtocol     *string `json:"ssl_protocol"`
+	Username        string  `json:"username"`
+}
```
worryg0d: Are these json tags really necessary?

session.go
testisnullus: Ok, fixed

session.go
testisnullus: Not needed, removed

worryg0d (approved):

session.go
```diff
@@ -780,6 +780,95 @@ func (s *Session) MapExecuteBatchCAS(batch *Batch, dest map[string]interface{})
...
+	const stmt = `
+		SELECT address, port, connection_stage, driver_name, driver_version, 
+		       hostname, keyspace_name, protocol_version, request_count, 
+		       ssl_cipher_suite, ssl_enabled, ssl_protocol, username 
+		FROM system_views.clients 
+		WHERE connection_stage = ?`
+
+	iter := s.control.query(stmt, connectionType)
+	if iter.NumRows() == 0 {
+		return nil, ErrKeyspaceDoesNotExist
```
worryg0d: Oops, I missed this before. I think you should return an error related to connectionType, not keyspace

testisnullus: Oops, fixed, ty

RostislavPorohnya (approved):

OleksiienkoMykyta (commented):
As far as I can see, it helps with (https://github.com/apache/cassandra-gocql-driver/issues/1774) and looks good to me.
It should be updated, but I think it's a good feature.
@jameshartig, @joao-r-reis, could you please check it to help solve the issue?

PR #1800, opened 2024-08-23
CASSGO-18 Add MutualTls authenticators to defaultApprovedAuthenticators

tolbertam:
Allow MutualTlsWithPasswordFallbackAuthenticator and MutualTlsAuthenticator as possible authenticators.

MutualTlsWithPasswordFallbackAuthenticator should behave functionally the same as PasswordAuthenticator.

MutualTlsAuthenticator's current implementation doesn't send AUTHENTICATE messages to the client, but felt it was worth including here in case it is ever enhanced to possibly also require credentials.

patch by Andy Tolbert; reviewed by Martin Sucha for CASSANDRA-19858

PR #1801, opened 2024-08-23
CASSGO-19 Don't restrict server authenticator in PasswordAuthenticator

tolbertam:
Currently gocql will only allow authenticating with authenticators
defined in defaultApprovedAuthenticators in conn.go.

There have been multiple occurrences of implementers needing to update
this list, either when a vendor would like to add their authenticator,
or a new authenticator being added.

It would probably reduce friction to just accept any authenticator
provided by the server. From what I know, other drivers behave in this
way.

If a user wanted to restrict this, they could use the existing
configuration PasswordAuthenticator.AllowedAuthenticators.

patch by Andy Tolbert; reviewed by Joao Reis, Lukasz Antoniak for CASSGO-19

------

This is an alternative pr to #1800; since this is a behavioral change, it merits some discussion and may take longer to review.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -43,25 +43,11 @@ import (
 	"github.com/gocql/gocql/internal/streams"
 )
 
-var (
-	defaultApprovedAuthenticators = []string{
-		"org.apache.cassandra.auth.PasswordAuthenticator",
-		"com.instaclustr.cassandra.auth.SharedSecretAuthenticator",
-		"com.datastax.bdp.cassandra.auth.DseAuthenticator",
-		"io.aiven.cassandra.auth.AivenAuthenticator",
-		"com.ericsson.bss.cassandra.ecaudit.auth.AuditPasswordAuthenticator",
-		"com.amazon.helenus.auth.HelenusAuthenticator",
-		"com.ericsson.bss.cassandra.ecaudit.auth.AuditAuthenticator",
-		"com.scylladb.auth.SaslauthdAuthenticator",
-		"com.scylladb.auth.TransitionalAuthenticator",
-		"com.instaclustr.cassandra.auth.InstaclustrPasswordAuthenticator",
-	}
-)
-
-// approve the authenticator with the list of allowed authenticators or default list if approvedAuthenticators is empty.
+// approve the authenticator with the list of allowed authenticators. If the provided list is empty,
+// the given authenticator is allowed.
 func approve(authenticator string, approvedAuthenticators []string) bool {
 	if len(approvedAuthenticators) == 0 {
-		approvedAuthenticators = defaultApprovedAuthenticators
+		return true
 	}
 	for _, s := range approvedAuthenticators {
 		if authenticator == s {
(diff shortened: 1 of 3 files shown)


conn.go
```diff
@@ -43,25 +43,11 @@ import (
...
-		"com.scylladb.auth.TransitionalAuthenticator",
-		"com.instaclustr.cassandra.auth.InstaclustrPasswordAuthenticator",
-	}
-)
-
-// approve the authenticator with the list of allowed authenticators or default list if approvedAuthenticators is empty.
+// approve the authenticator with the list of allowed authenticators. If the provided list is empty,
+// the given authenticator is allowed.
 func approve(authenticator string, approvedAuthenticators []string) bool {
 	if len(approvedAuthenticators) == 0 {
```
tolbertam, line 49: alternatively, could test for nil, but felt would keep it this way for consistency.

lukasz-antoniak (approved):
Looks good to me. Indeed Java driver does not assert on authenticator returned by the server, but relies on client configuration.

doc.go
```diff
@@ -81,6 +81,16 @@
 //	 }
 //	 defer session.Close()
 //
+// By default, PasswordAuthenticator will attempt to authenticate regardless of what implementation the server returns
+// in its AUTHENTICATE message as its authenticator, (e.g. org.apache.cassandra.auth.PasswordAuthenticator).  If you
+// wish to restrict this you may use PasswordAuthenticator.AllowedAuthenticators:
```
joao-r-reis, line 86: nit: Maybe a comment on the interface function itself could also be useful

joao-r-reis (approved):
I agree with this change, other drivers already work like this.

joao-r-reis:
I'd rather merge this instead of #1800 but if there's an objection to this one we can merge #1800

tolbertam:
> I'd rather merge this instead of https://github.com/apache/cassandra-gocql-driver/pull/1800 but if there's an objection to this one we can merge https://github.com/apache/cassandra-gocql-driver/pull/1800

I agree with that sentiment, would prefer it be consistent with other drivers.

doc.go
tolbertam: Is the comment I changed at line 46 in conn.go ok or were you looking for documentation elsewhere (like on the `AllowedAuthenticators` field in `PasswordAuthenticator`?)

joao-r-reis: I was thinking about a comment on the type itself ( `conn.go` ) so that this information shows up on the docs section for the type (in pkg.go.dev) and when using an IDE:

```
type PasswordAuthenticator struct {
	Username                      string
	Password                       string

        // Setting this to nil or empty will allow any authenticator provided by the server
	AllowedAuthenticators []string
}
```

tolbertam: :+1: sounds great, agree that this is the most appropriate place.  Will make that change.

joao-r-reis: Maybe a comment on the type itself too.

```
// PasswordAuthenticator can be configured with an "allow list" of authenticators (can be set to nil or empty to allow all)
type PasswordAuthenticator struct {
	Username                      string
	Password                       string

        // Setting this to nil or empty will allow any authenticator provided by the server
	AllowedAuthenticators []string
}
```

tolbertam: made an attempt to document both.  Thought it would be good to make it clear that the default behavior of other drivers is to allow any authenticator,  generally I don't think people should configure this, and the presence of documentation may lead them to think they need to do so, so felt it was good to clarify that.   Hopefully that looks ok.

joao-r-reis: lgtm 👍

tolbertam: Thanks, i'll squash the commits and update the patch by/reviewed by :+1:

tolbertam: done, hopefully this is good to go!

PR #1802, opened 2024-08-23
CASSANDRA-19853 update issue templates and add pull request template

smiklosovic:


michaelsembwever:
we don't know if JIRA are to be used instead of GH issues for this repo yet.  i would rather see the group of active contributors have this decision and come to their own consensus.  in the meantime it's our job to inform them how and why the rest of the project does it a given way.  can we move this PR to draft in the meantime…?

(some things, like commit messages (which are fed into contribulyze), are more important to standardise on)

smiklosovic:
I was thinking that JIRA is a hard requirement for us. If it is for Java driver and Cassandra, I don't see the reason why it shouldn't be the case for this as well (can we even decide that on our own? Is not that something Apache wants?).

Anyway, I am open to the discussion here if JIRA tickets are evaluated as not necessary which would be quite surprising to see though. I think the note I added (JIRAs strongly recommended) is not making issue creations on GitHub impossible.

michaelsembwever:
There is no requirement from the foundation to which tracking system, if any, we use.  There is a requirement that all communication is recorded on the foundation's infrastructure.   This is solely a project decision.  For cassandra-website and cassandra-builds we have just PRs for some basic patches.   The java-driver project was previously using jira, so it made sense.  I would like to leave this to the subproject to figure out on its own, but it is our job to inform on project precedences – there is value in repeating them.

joao-r-reis:
Now that it is estabilished that we're going to use JIRA I think we can proceed with this PR @smiklosovic. Maybe we can be a bit more assertive and write in the issue template that all new issues should be opened in JIRA (GH issues will be closed when we are able to triage the existing ones and move the relevant ones to JIRA). 

I'll move the JIRA to the new CASSGO Project and I'll also be opening a PR soon to improve the commit message format example on `CONTRIBUTING.md` based on the Slack discussion we had with @martin-sucha a few days ago. We should probably use the same commit message example in `CONTRIBUTING.md` and the PR template.

joao-r-reis:
Oh actually this JIRA is part of a bigger parent JIRA that relates to the donation so Idk if we want to move all donation related JIRAs or decouple this one from the parent donation JIRA.

PR #1809, opened 2024-09-03
CASSGO-27 fix gocql-RetryPolicy-still-dont-use-query-idempotence

### LINKED ISSUE #1803: CASSGO-27 gocql RetryPolicies still don't use query idempotence
Issue https://github.com/apache/cassandra-gocql-driver/issues/1154 was meant to add idempotence check in gocql RetryPolicies

Issue was closed but Retry Policies still don't check if query IsIdempotent BUT documentations states that IsIdempotent affects retries 

Need to implement this logic or change documentation

OleksiienkoMykyta:
Closes #1803
RetryPolicy doesn't check the query's idempotency, but according to the [specification](https://pkg.go.dev/github.com/gocql/gocql#hdr-Retries_and_speculative_execution) queries with false idempotence shouldn't be retried.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Retry policy now takes into account query idempotency (CASSGO-27)
+
 ## [1.7.0] - 2024-09-23
 
 This release is the first after the donation of gocql to the Apache Software Foundation (ASF)
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -32,6 +32,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/stretchr/testify/require"
 	"io"
 	"math"
 	"math/big"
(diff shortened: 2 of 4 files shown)


SiyaoIsHiding:
Hi, I think this won't fix the issue because of this line https://github.com/apache/cassandra-gocql-driver/blob/92b4056c1a125e7094b397cfab8bd8ac8d48f26e/query_executor.go#L67-L69
the function `do` is only called when it's marked as NOT idempotent.

joao-r-reis:
@SiyaoIsHiding that line you linked is to avoid speculative executions if the query isn't idempotent which seems correct to me... It's not really related to retries. If it's idempotent then `run` will be called which will then call `do`

joao-r-reis:
@OleksiienkoMykyta can you add a test that reproduces the bug before this change to ensure we don't get a regression in the future? It should be possible to reproduce this by using a very small timeout

OleksiienkoMykyta:
> @OleksiienkoMykyta can you add a test that reproduces the bug before this change to ensure we don't get a regression in the future? It should be possible to reproduce this by using a very small timeout.

Yes, sure, I'll add some tests to cover this behavior.

OleksiienkoMykyta:
@joao-r-reis I have added some tests, it would be nice if you would check it.
 Also, I tried to test it with a timeout, but it doesn't reconcile query execution, just a single try, so I used a custom RetryPolicy.

joao-r-reis:
`TestQueryMultinodeWithMetrics` is panicking because it assumes the retry policy runs for all hosts but now it doesn't because of the idempotency change.

joao-r-reis (commented):
Changes look good 👍 , still need to fix failing tests though.

OleksiienkoMykyta:
> Changes look good 👍 , still need to fix failing tests though.

I have fixed it. The issue was in test which should make retries but doesn't set idempotence to query explicitly, now everything works properly.

joao-r-reis (approved):

joao-r-reis:
Please format your commit message following the rules stated in [CONTRIBUTING.md](https://github.com/apache/cassandra-gocql-driver/blob/trunk/CONTRIBUTING.md#commit-message) and add this JIRA to the `CHANGELOG.md` in the `Fixed` section of `Unreleased`, thanks!

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- CASSGO-27
```
jfleming-ic: Would be nice to have a short summary of what was fixed here, so users don't need to lookup what this ticket is, convention seems to be
"message" (ticket)

So for this, something like this would suffice

"Retry policy now takes into account query idempotency (CASSGO-27)"

jfleming-ic:
Left a small comment on the message in the release log, but happy to tick as is.

Jacko161:
Sorry was on the wrong github account on my last comment - happy to tick this, but would be good to update the changelog message if possible.

Jacko161 (approved):

joao-r-reis:
And also change the first line of the commit message and add Jackson as a reviewer.

The first line of the commit message could be as an example: `Change RetryPolicy so it checks query idempotence`

CHANGELOG.md
OleksiienkoMykyta: Sure, I have fixed it according to your suggestions

Issue #1786, opened 2024-07-29
CASSGO-39 Add query attempt interceptor


martin-sucha:
There are use-cases for wrapping the executions of individual query attempts.

We currently have QueryObserver and BatchObserver that allow observing the query attempts, but don't allow modification of behavior.

One use case that requires modification of behavior is a request rate limiter (#1756) or a semaphore to limit request concurrency.

It seems that instead of adding an individual interface for each use case, a single interceptor-like interface could serve all of them.

The following `QueryAttemptInterceptor` could be used to implement the rate limiter, however we will need to make changes the interface to be able to replace the `QueryObserver` and `BatchObserver`:

```
type QueryAttemptInterceptor interface {
    // AttemptQuery executes the query on the given connection.
    // The attempt function does the actual work, the AttempQuery implementation is free to execute code
    // before/after the call to the attempt function or skip the call altogether.
    // If error is nil, the returned *Iter must be non-nil.
    AttemptQuery(ctx context.Context, query ExecutableQuery, conn *Conn, attempt func(ctx context, query ExecutableQuery, conn *Conn) *Iter) (*Iter, error))
}
```

OleksiienkoMykyta:
@joao-r-reis, did I get it right, that this feature waits for resolution of the (https://github.com/apache/cassandra-gocql-driver/issues/1447), and can be solved as soon as the [PR](https://github.com/apache/cassandra-gocql-driver/pull/1820) will be merged?

PR #1810, opened 2024-09-03
Add query interceptors

BenEddy:
This PR introduces `QueryInterceptor` and `BatchInterceptor` interfaces for intercepting queries immediately before execution, which allows clients to inject logic that should apply to all queries. My personal use case is `context` manipulation and conditionally enabling cassandra tracing, but the interfaces could be used for rate limiting, query modification, logging, fault injection, metrics, etc. 

The approach here is similar to https://github.com/apache/cassandra-gocql-driver/issues/1786, but intercepts at the higher query execution level rather than the query attempt/request level. I like the idea of intercepting attempts, but I think the nuances of request handling (retry policies, speculative execution, client side host selection) could lead to unexpected behavior. For example, rewriting and replacing a query via attempt interceptor would lose query retry metrics and violate the retry policy. Or rate limiting a speculative execution request via attempt interceptor would defeat the purpose of speculative execution. However, if an attempt interceptor is preferred, it would not be difficult to move the interception point.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -214,6 +214,14 @@ type ClusterConfig struct {
 	// See https://issues.apache.org/jira/browse/CASSANDRA-10786
 	DisableSkipMetadata bool
 
+	// QueryInterceptor will set the provided query interceptor on all queries created from this session.
+	// Use it to intercept and modify queries by providing an implementation of QueryInterceptor.
+	QueryInterceptor QueryInterceptor
+
+	// QueryInterceptor will set the provided query interceptor on all queries created from this session.
+	// Use it to intercept and modify batches by providing an implementation of BatchInterceptor.
+	BatchInterceptor BatchInterceptor
+
 	// QueryObserver will set the provided query observer on all queries created from this session.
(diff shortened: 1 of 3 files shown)


hardikmadhu:
Hi @BenEddy ! Thanks for the work here! I have also been looking to do something similar to log some context. I have a couple of questions, but I'm a total golang noob, so ignore if some of these comments don't make sense.

1. As you elaborate in your description, if the idea is to log some context, configuring driver side rate limiter etc, wouldn't it be a better idea to add the intercept on Attempts? If you're in a real pickle, where you're actually trying to debug something, host timeouts, network issues, server load etc, having an intercept in the attempt would be invaluable. 
2. Since the query and batch intercept are almost identical, except the type of query, wouldn't it be possible to define an interface, that could take either query or batch as an argument and we'd be able to decide what intercept to execute after assessing the type?

BenEddy:
@hardikmadhu After more consideration, I think you're right that intercepting at a lower level is probably a better way to go. Opened https://github.com/apache/cassandra-gocql-driver/pull/1820 and will close this out for now.

PR #1812, opened 2024-09-04
CASSANDRA-19796 – cassandra-gocql-driver CI

smiklosovic:

PR #1813, opened 2024-09-05
Bump Go/C* versions

absurdfarce:
In prep for 1.7.0 release.

Ambition here is to match the ["Sunsetting Model"](https://github.com/apache/cassandra-gocql-driver?tab=readme-ov-file#sunsetting-model) that was in place with gocql.

[Test run](https://github.com/apache/cassandra-gocql-driver/actions/runs/10713198895) shows this passes all existing tests.


smiklosovic:
we are contemplating about doing this in https://github.com/apache/cassandra-gocql-driver/pull/1812

Would you mind to take it there?

absurdfarce:
@smiklosovic I actually think this should be handled independently.  We should upgrade the Go versions for 1.7.0 and I'd rather not bring any additional changes along with that if we can avoid it.  Ideally 1.7.0 would basically be what was in 1.6.0 along with a few changes related to the release/ASF conversion.

martin-sucha (approved):

.github/workflows/main.yml
```diff
@@ -34,8 +34,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        go: [ '1.19', '1.20' ]
-        cassandra_version: [ '4.0.8', '4.1.1' ]
+        go: [ '1.22', '1.23' ]
+        cassandra_version: [ '4.0.13', '4.1.6' ]
```
absurdfarce, line 38: 4.1.6 and 4.0.13 are the latest 4.1.x and 4.0.x as of this writing.  But now I'm wondering if we shouldn't be aiming for 4.1.x & 5.0.0 instead.

Trouble is stock ccm has issues with 5.0.0... and supporting a new major version of Cassandra probably needs a ticket (or perhaps even tickets) all to itself.

joao-r-reis (approved):

absurdfarce:
Closing in favor of #1825

Issue #1331, opened 2019-07-10
Scan of int value into `interface{}` fails


dahankzter:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?
3.0.7

### What version of Gocql are you using?
latest

### What did you do?
Same setup as in #1330 

Modified program to read as:

```go
var id gocql.UUID
data := make(map[string]interface{})
if !it.Scan(&id, &data) {
      break
}
```

### What did you expect to see?
A map of string to int values.

### What did you see instead?
The error message
`can not unmarshal int into *interface {}`

PR #1815, opened 2024-09-05
Unable to scan int into interface{} issue fix

tengu-alt:
Fix for the: #1331

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -44,7 +44,7 @@ import (
 	"time"
 	"unicode"
 
-	inf "gopkg.in/inf.v0"
+	"gopkg.in/inf.v0"
 )
 
 func TestEmptyHosts(t *testing.T) {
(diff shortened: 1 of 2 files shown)


tengu-alt:
Implementation for scanning into `interface{}` for: 

- Varchar as `string`
- Varint  as `bigInt`
- IntLike as `int`
- Bool as `bool`
- Float as `float32`
- Double as `float64`
- Decimal as `infDec`
- Time as `time.Duration`
- Timestamp as `time.Time`
- Date as `time.Time`
- Duration as `Duration`
- UUID as `string`
- Inet as `net.IP`

OleksiienkoMykyta (approved):

jameshartig:
I managed to work in `*interface{}` support into #1855 in order to simplify some logic. I also supported it in maps, lists, sets, tuples.

joao-r-reis:
Ok I think we should close this in favor of #1855 then since #1855 incorporates some cleanup of type encoding/decoding code, is that ok @tengu-alt ?

tengu-alt:
I see, let's close it then.

PR #1817, opened 2024-09-05
Update driver name after ASF donation

absurdfarce:
Issue originally identified by @tolbertam 

Update the driver name to allow servers to clearly distinguish between gocql and cassandra-gocql-driver.  Note that something [similar was done](https://github.com/apache/cassandra-java-driver/commit/16260261d3df50fcf24fac1fc2d37896c4a111bf#diff-c45a71e429526ff721761d6a9f3fc0cec3f82a3c44c3f81a1c2ed0157a9f5d70) for the Java driver when it was donated.


version.go
```diff
@@ -27,7 +27,7 @@ package gocql
 import "runtime/debug"
 
 const (
-	mainModule = "github.com/gocql/gocql"
+	mainModule = "github.com/apache/cassandra-gocql-driver"
```
joao-r-reis: Have you tested this? It's using this string to match the dependency path but I'm not sure where the dependency path comes from. If it comes from the actual module name then I don't think this will work.

martin-sucha: It comes from the module name as defined in `go.mod`, which won't change until v2.

The reason why it checks BuildInfo is so that when someone uses a fork with the replace directive, the replaced version/path is used.

We need to keep `mainModule` intact and do

```go
driverName = "github.com/apache/cassandra-gocql-driver"
```

instead of

```go
driverName = mainModule
```

joao-r-reis:
```
if d.Path == mainModule {
  driverName = "github.com/apache/cassandra-gocql-driver"
  driverVersion = d.Version
  if d.Replace != nil {
	  driverName = d.Replace.Path
	  driverVersion = d.Replace.Version
  }
  break
}
```

absurdfarce: 🤦 No, really... this is what I get for rushing.  Completely agree that "mainModule" is entirely the wrong name here although I still think that there's merit in encapsulating that value in a constant, albeit a better-named constant.

I'll tweak this a bit and see if I can come up with something that (a) is clearer and (b) is very much in the spirit of what's proposed above.

version.go
```diff
@@ -27,7 +27,7 @@ package gocql
 import "runtime/debug"
 
 const (
-	mainModule = "github.com/gocql/gocql"
+	defaultDriverName = "github.com/apache/cassandra-gocql-driver"
```
absurdfarce, line 30: Allows us to contain the default value in a distinct location (i.e. not embedded within the code below) with a _much_ more descriptive name.

version.go
```diff
@@ -28,6 +28,7 @@ import "runtime/debug"
 
 const (
 	mainModule = "github.com/gocql/gocql"
```
absurdfarce: Note: this string must remain as-is until we rename packages.  That won't happen in 1.7.0, however, since the goal of 1.7.0 is just to get an ASF-branded release containing the current contents of the repo out (with some minor modifications).

martin-sucha: Please add a note that the string must remain as-is until v2 release as a code comment.

version.go
martin-sucha: Perhaps we could rename `mainModule` to `upstreamModule` ?

absurdfarce: How about just "mainPacakge"?  That has the advantage of matching the terminology in the [runtime/debug docs](https://pkg.go.dev/runtime/debug#BuildInfo)... ?

version.go
absurdfarce: Done.

joao-r-reis (approved):

version.go
joao-r-reis: package in the Go land means something different than module, in this case it's referring to the module so I'm fine with keeping it `mainModule` or renaming it to `upstreamModule`, both work for me.

joao-r-reis: Oh wait we're looking at `d.Path` instead of `d.Main`, in that case it is referring to the package you're right.

joao-r-reis: `mainPackage` or `upstreamPackage` would be more suitable then

absurdfarce: Thanks @joao-r-reis .  I kinda like "mainPackage" in this case so I ran with it... if anybody has objections let me know!

martin-sucha (approved):
Looks good. Thank you!

martin-sucha:
Please squash the commits and update the commit message:

https://github.com/apache/cassandra-gocql-driver/blob/trunk/CONTRIBUTING.md#commit-message

version.go
```diff
@@ -38,8 +44,8 @@ func init() {
 	buildInfo, ok := debug.ReadBuildInfo()
```
jameshartig, line 44: Since there was some confusion in the PR it would be good to document the point of this code especially as it relates to the replace directive

jameshartig (approved):
This LGTM but I do think some comments would've gone a long way in clarifying what this is trying to do.

tolbertam (approved):
:+1: thank you for taking care of this!

version.go
absurdfarce: This should be addressed below... let me know if you were thinking of something else @jameshartig !

absurdfarce:
@tolbertam The test failures here seem to be coming from expired certs for TLS tests.  Is fixing that just a matter of getting #1805 merged?

tolbertam:
That's right @absurdfarce , #1805 aims to resolve the issues around the certs used by the tests expiring.

joao-r-reis (approved):

absurdfarce:
Thanks @tolbertam!  I'm gonna merge #1805 and confirm that that addresses the test failures before calling this one good.

absurdfarce:
Bah, I mucked this up with the merge commit 🤦 .  Closing this out and replacing it with #1824

Issue #1447, opened 2020-05-22
Dont mutate Query/Batch


Zariel:
Currently queries are mutated while being executed, the query metrics and the consistency are both changed. Instead move query metrics to the Iter and remove the need to change the queries consistency for speculation.

martin-sucha:
Also page state is mutated and the metrics are shared between multiple queries in a single iter (so attempt count is not accurate for queries) - #1508 and #1511 are related.

OleksiienkoMykyta:
According to the Jira ticket https://issues.apache.org/jira/browse/CASSGO-22 @joao-r-reis prototyping better API design, the developing work is in progress. 
As I understand following PRs paused until the API is finished
https://github.com/apache/cassandra-gocql-driver/pull/1848 
https://github.com/apache/cassandra-gocql-driver/pull/1820

PR #1820, opened 2024-09-12
CASSGO-39 Add query attempt interceptor

BenEddy:
As suggested in https://github.com/apache/cassandra-gocql-driver/issues/1786, this PR introduces a `QueryAttemptInterceptor` interface for intercepting query attempts, which allows clients to inject logic that should apply to all queries e.g. context manipulation, rate limiting, query modification, logging, fault injection, metrics, etc.

Alternative to https://github.com/apache/cassandra-gocql-driver/pull/1810.

### DIFF
--- a/example_interceptor_test.go
+++ b/example_interceptor_test.go
@@ -55,7 +55,7 @@ func (q MyQueryAttemptInterceptor) Intercept(
 
 	// Optionally bypass the handler and return an error to prevent query execution.
 	// For example, to simulate query timeouts.
-	if q.injectFault && attempt.Number == 1 {
+	if q.injectFault && attempt.Attempts == 0 {
 		<-time.After(1 * time.Second)
 		return gocql.NewIterWithErr(gocql.RequestErrWriteTimeout{})
 	}
--- a/example_interceptor_test.go
+++ b/example_interceptor_test.go
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*
+ * Content before git sha 34fdeebefcbf183ed7f916f931aa0586fdaa1b40
+ * Copyright (c) 2016, The Gocql authors,
+ * provided under the BSD-3-Clause License.
+ * See the NOTICE file distributed with this work for additional information.
+ */
+
+package gocql_test
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	gocql "github.com/gocql/gocql"
+)
+
+type MyQueryAttemptInterceptor struct {
+	injectFault bool
+}
+
+func (q MyQueryAttemptInterceptor) Intercept(
+	ctx context.Context,
+	query gocql.ExecutableQuery,
+	conn *gocql.Conn,
+	handler gocql.QueryAttemptHandler,
+) *gocql.Iter {
+	switch q := query.(type) {
+	case *gocql.Query:
+		// Inspect or modify query
+		query = q
+	case *gocql.Batch:
+		// Inspect or modify batch
+		query = q
+	}
+
+	// Inspect or modify context
+	ctx = context.WithValue(ctx, "trace-id", "123")
+
+	// Optionally bypass the handler and return an error to prevent query execution.
+	// For example, to simulate query timeouts.
+	if q.injectFault && query.Attempts() == 0 {
+		<-time.After(1 * time.Second)
+		return gocql.NewIterWithErr(gocql.RequestErrWriteTimeout{})
+	}
+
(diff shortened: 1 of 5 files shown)


example_interceptor_test.go
```diff
@@ -0,0 +1,92 @@
...
+		// Inspect or modify batch
+		query = q
+	}
+
+	// Inspect or modify context
+	ctx = context.WithValue(ctx, "trace-id", "123")
+
+	// Optionally bypass the handler and return an error to prevent query execution.
+	// For example, to simulate query timeouts.
+	if q.injectFault && query.Attempts() == 0 {
```
martin-sucha: Should the attempt number be a parameter to `Intercept`? It seems that calling `Attempts()` could be racy, especially when speculative execution is enabled:

1. Attempt 0 started
2. Attempt 1 started
3. Interceptor 0 called - gets attempt 0
4. Interceptor 1 called - gets attempt 0
5. attempts incremented
6. attempts incremented

Perhaps the signature should be `Intercept(ctx context.Context, attempt gocql.QueryAttempt, handler gocql.QueryAttemptHandler)`? That would enable adding more information about the attempt in the future.

```go
type QueryAttempt struct {
	// query to execute
	query gocql.ExecutableQuery
	// conn to use to execute the query
	conn *gocql.Conn
	// number of the attempt. 0 is the initial attempt, 1 is first retry, etc.
	number int
}
```

```go
type QueryAttemptHandler = func(context.Context, QueryAttempt) *Iter
```

BenEddy: @martin-sucha passing a `QueryAttempt` type seems like a great way to keep the interceptor API stable 👍

query_executor.go
```diff
@@ -48,16 +48,60 @@ type ExecutableQuery interface {
...
+// QueryAttemptInterceptor is the interface implemented by query interceptors / middleware.
+//
+// Interceptors are well-suited to logic that is not specific to a single query or batch.
+type QueryAttemptInterceptor interface {
+	// Intercept is invoked once immediately before a query execution attempt, including retry attempts and
+	// speculative execution attempts.
+
+	// The interceptor is responsible for calling the `handler` function and returning the handler result. Failure to
+	// call the handler will panic. If the interceptor wants to halt query execution and prevent retries, it should
+	// return an error.
```
martin-sucha: **nitpick**: The function does not have an `error` return value. Should we add an error return value or should we reference  `NewIterWithErr` instead?

**question**: How does returning an error prevent a retry?

BenEddy: > Should we add an error return value or should we reference `NewIterWithErr` instead?

Updated `Intercept` to optionally return an error and removed `NewIterWithErr`. 

> How does returning an error prevent a retry?

Returning a non-retriable error like `context.Canceled` would prevent retries, but admittedly that's not clear from the function docs. I removed the reference to preventing retries and will assume that users can probably figure out how to configure retry policies with whatever error handling and retry prevention makes sense for them.

query_executor.go
```diff
@@ -48,16 +48,64 @@ type ExecutableQuery interface {
...
+	interceptor QueryAttemptInterceptor
+}
+
+type QueryAttempt struct {
+	// The query to execute, either a *gocql.Query or *gocql.Batch.
+	Query ExecutableQuery
+	// The connection used to execute the query.
+	Conn *Conn
+	// The host that will receive the query.
+	Host *HostInfo
```
ribaraka: Do we need to add and expose these fields in `QueryAttempt`? 

If there’s no compelling need for direct connection manipulation, then maybe we could expose only a connection metadata that’s relevant for logging or monitoring purposes?

example_interceptor_test.go
```diff
@@ -0,0 +1,88 @@
...
+	defer session.Close()
+
+	ctx := context.Background()
+
+	var stringValue string
+	err = session.Query("select now() from system.local").
+		WithContext(ctx).
+		RetryPolicy(&gocql.SimpleRetryPolicy{NumRetries: 2}).
+		Scan(&stringValue)
+	if err != nil {
```
ribaraka, line 85: How about adding an option to apply the interceptor selectively to specific queries, like `.WithInterceptor()` method to the Query struct?

ribaraka:
https://issues.apache.org/jira/browse/CASSGO-39

joao-r-reis:
I don't know if I like the idea of having an `interceptor` that can modify the query/batch after it has already been submitted to the driver. Having something like the equivalent of a request tracker in the java driver would be nice but the goal there would be to inspect the queries, not modify them. Modifying query/batch also goes against #1447 which is being worked on atm.

joao-r-reis:
Rate limiting should probably also be handled in a different component with its own interface dedicated to rate limiting/throttling

joao-r-reis:
Hmm taking another look at #1786 it seems that the focus is indeed to be an interface generic enough to serve as a way to implement stuff like request limiting or query modification while the inspection stuff could be handled by the observers that we already have.

In that case we need to make sure the changes here are consistent with changes in #1447

query_executor.go
BenEddy: Removed the `Conn` field in favor of read-only `LocalAddr` and `RemoteAddr` connection metadata fields, but retained the `Host` field to match `ObservedQuery`.

example_interceptor_test.go
BenEddy: Not strongly opposed, but I'd favor using the query `context` to selectively enable an interceptor because it avoids introducing potential conflicts between a session-level interceptor config and query-level interceptor config.

joao-r-reis:
Since this PR doesn't require breaking changes to the API we can take a look at this in a minor release after 2.0.

example_interceptor_test.go
```diff
@@ -0,0 +1,139 @@
...
+		log.Fatalf("query failed %T", err)
+	}
+}
+
+type QueryAttemptInterceptorChain struct {
```
worryg0d: Just thinking about making this a part of gocql api.

As there are rate-limiting and logging/monitoring/tracing use cases, having an aggregator of multiple interceptors out of the box to free end-users from its implementation looks reasonable to me.

PR #1821, opened 2024-09-13
CASSGO-3 detailed description for NumConns

### LINKED ISSUE #1741: CASSGO-3 How to set limit for pool connection

4.x

v1.6.0

1.19

### What did you do?
I am creating a pool connection with cluster.NumConns = 20. so that it can create only 20 connection/pod.
container := dbType + "." + dbName
	username := String(container+".username", "")
	password := String(container+".password", "")
	host := String(container+".host", "")
	db := String(container+".db", "")
	dc := String(container+".dc", "")
	port1 := String(container+".port", "")

	// fmt.Println(container, " ", username, " ", password, " ", host, " ", db, " ", dc, " ", port1)
	port, err := strconv.Atoi(port1)

	HostList := strings.Split(host, ",")

	cluster := gocql.NewCluster(HostList...)
	pass := gocql.PasswordAuthenticator{username, password, nil}
	cluster.Keyspace = db
	cluster.Authenticator = pass
	cluster.Consistency = gocql.LocalQuorum
	//cluster.SerialConsistency = gocql.LocalSerial
	if err == nil {
		cluster.Port = port // Default port
	}

	if dbName == "cass_comp_gcp" || dbName == "cass_comp_gcp_newIN" {
		cluster.DisableInitialHostLookup = true
		cluster.SocketKeepalive = 10 * time.Second
	}
	cluster.NumConns = 20
	//
	cluster.PoolConfig.HostSelectionPolicy = gocql.DCAwareRoundRobinPolicy(dc)
	cluster.HostFilter = gocql.DataCentreHostFilter(dc)
	cluster.Timeout = 5 * time.Second
	if GetMode() != "prod" {
		cluster.ConnectTimeout = time.Second * 20
		//cluster.DisableInitialHostLookup = true
	}
	// cluster.SocketKeepalive = 10 * time.Second

	return cluster.CreateSession()

We are using the above function in our read API which is capable of serving around 5,000 requests per second. Our goal is to limit the number of connections created per pod to 20, even if there is an increase in request traffic. Although we have added a limit of 20 connections, we have observed that if the request count increases, it opens more than 20 connections, and some of them remain open even after the normal pace.

### What did you see instead?
It is creating around 200-600 connections per pod, which is not getting closed even after the normal pace. This is causing a load issue for Cassandra.
---

- output of `nodetool status`

OleksiienkoMykyta:
Closes #1741 .
There is some unclearness in the NumConns description because there are no details about what it affects, and does it sets any limits. I added a few more details to avoid confusion about it in the future.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -102,7 +102,13 @@ type ClusterConfig struct {
 	// Initial keyspace. Optional.
 	Keyspace string
 
-	// Number of connections per host.
+	// The size of the connection pool for each host.
+	// The pool filling runs in separate gourutine during the session initialization phase.
+	// gocql will always try to get 1 connection on each host pool
+	// during session initialization AND it will attempt
+	// to fill each pool afterward asynchronously if NumConns > 1.
+	// Notice: There is no guarantee that pool filling will be finished in the initialization phase.
+	// Also, it describes a maximum number of connections at the same time.
 	// Default: 2
 	NumConns int
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Remove global NewBatch function (CASSGO-15)
 
+- Detailed description for NumConns (CASSGO-3)
+
 ### Fixed
 
 - Retry policy now takes into account query idempotency (CASSGO-27)


cluster.go
```diff
@@ -102,7 +102,8 @@ type ClusterConfig struct {
 	// Initial keyspace. Optional.
 	Keyspace string
 
-	// Number of connections per host.
+	// The size of connection pool for each host. The Pool will be filled during the first request execution.
```
joao-r-reis: I'm not sure if this is 100% correct, if `InitialHostLookup`  is enabled (it's the default) then the pools for the discovered hosts will be filled during session initialization

OleksiienkoMykyta: You are right!
I double-checked it with the timeout, and it turned out that the pool filled during initialization. When I tested it last time the pool had not been filled before the query execution because [pool filling runs in separate gorutine](https://github.com/OleksiienkoMykyta/gocql/blob/974fa1211ccef10a418966fd7b00984a39ed9232/session.go#L292) and it took more time to finish than the query execution.  
Thank you for noticing the issue! 
I can replace it with something like 
```  
The size of the connection pool for each host. The pool filling runs in separate gourutine during the session initialization phase. 
Also, it describes a maximum number of connections at the same time.
Notice: There is no guarantee that pool filling will be finished in the initialization phase.
Default: 2 
```
What do you think?

joao-r-reis: Hmm interesting, did you use `SingleHostReadyPolicy`? If you don't use this policy then it should wait until all hosts have at least 1 connection in the pool before the session initialization is done.

Check [this](https://github.com/OleksiienkoMykyta/gocql/blob/974fa1211ccef10a418966fd7b00984a39ed9232/session.go#L326-L330) and [this](https://github.com/apache/cassandra-gocql-driver/blob/7b7e6affbfddf36e9cbc1068e96ecb9c7cc0af18/connectionpool.go#L454-L482).

Considering this, it would probably be good to mention that gocql will always try to get 1 connection on each host pool during session initialization AND it will attempt to fill each pool afterwards asynchronously if `NumConns > 1`

OleksiienkoMykyta: What about something like that? 
```
// The size of the connection pool for each host.
// The pool filling runs in separate gourutine during the session initialization phase.
// gocql will always try to get 1 connection on each host pool
// during session initialization AND it will attempt
// to fill each pool afterward asynchronously if NumConns > 1.
// Notice: There is no guarantee that pool filling will be finished in the initialization phase.
// Also, it describes a maximum number of connections at the same time.
```
I have added `gocql will always try to get 1 connection on each host pool during session initialization AND it will attempt to fill each pool afterward asynchronously if NumConns > 1` as you suggested, so there is no more room for unclearness.

joao-r-reis: 👍

joao-r-reis (approved):

Jacko161 (approved):

PR #1824, opened 2024-09-20
Update driver name after ASF donation, take two

absurdfarce:
Cleaned up version of #1817 without the merge commit.

Review was already completed on the previous PR.


absurdfarce:
I'm going to leave the reference pointing to the old PR since that's where all the review comments can be found.

PR #1827, opened 2024-09-24
Perform null check before dereference

shubhamjadhav:


taaraora:
Is there any description of the issue this PR is made for? @shubhamjadhav?

joao-r-reis:
We need information regarding the issue that is being fixed with this PR, closing PR for now.

PR #1822, opened 2024-09-16
CASSGO-1 CASSGO-30 Native Protocol 5 Support

worryg0d:
## Overview

C* 4.0 introduced Native Protocol 5, so we should make gocql to support it.
This PR consists of other my PRs, which also provides support for the v5 version of the protocol.

This PR provides support of Native Protocol 5 features as:
* New frames format [CASSANDRA-15299](https://issues.apache.org/jira/browse/CASSANDRA-15299)
* `result_metadata_id` field for **EXECUTE** and **PREPARED** messages [CASSANDRA-10786](https://issues.apache.org/jira/browse/CASSANDRA-10786)
* `now_in_seconds` field for `BATCH` and `QUERY` messages [CASSANDRA-14664](https://issues.apache.org/jira/browse/CASSANDRA-14664)
* `keyspace` field for `BATCH` and `QUERY` messages

closes #1750

PR #1828, opened 2024-10-10
CASSGO-11 Support vector type

### LINKED ISSUE #1734: CASSGO-11 Feature Request: Support Vector Type
Since Cassandra has introduced Vector type, and it's already supported in Python driver, I hope it can be added in gocql as well, and it will resolve this issue: https://github.com/datastax/gocql-astra/issues/17#issuecomment-1888111809

lukasz-antoniak:
Fix #1734.

Complete vector type support for GoCQL. Supports fixed and variable-length vector types (e.g. vector of floats, vector of list of text).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ on:
     types: [ opened, synchronize, reopened ]
 
 env:
-  CCM_VERSION: "6e71061146f7ae67b84ccd2b1d90d7319b640e4c"
+  CCM_VERSION: "4621dfee5ad73956b831091a8b863d100d25c610"
 
 jobs:
   build:
@@ -35,7 +35,7 @@ jobs:
       fail-fast: false
       matrix:
         go: [ '1.22', '1.23' ]
-        cassandra_version: [ '4.0.13', '4.1.6' ]
+        cassandra_version: [ '4.1.6', '5.0.3' ]
         auth: [ "false" ]
         compressor: [ "no-compression", "snappy", "lz4" ]
         tags: [ "cassandra", "integration", "ccm" ]
--- a/frame.go
+++ b/frame.go
@@ -34,6 +34,7 @@ import (
 	"io/ioutil"
 	"net"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 )
@@ -905,6 +906,22 @@ func (f *framer) readTypeInfo() TypeInfo {
 		collection.Elem = f.readTypeInfo()
 
 		return collection
+	case TypeCustom:
+		if strings.HasPrefix(simple.custom, VECTOR_TYPE) {
+			spec := strings.TrimPrefix(simple.custom, VECTOR_TYPE)
+			spec = spec[1 : len(spec)-1] // remove parenthesis
+			idx := strings.LastIndex(spec, ",")
+			typeStr := spec[:idx]
+			dimStr := spec[idx+1:]
+			subType := getCassandraLongType(strings.TrimSpace(typeStr), f.proto, nopLogger{})
+			dim, _ := strconv.Atoi(strings.TrimSpace(dimStr))
+			vector := VectorType{
+				NativeType: simple,
+				SubType:    subType,
+				Dimensions: dim,
+			}
+			return vector
+		}
 	}
 
 	return simple
--- a/helpers.go
+++ b/helpers.go
@@ -25,10 +25,12 @@
 package gocql
 
 import (
+	"encoding/hex"
 	"fmt"
 	"math/big"
 	"net"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
@@ -162,59 +164,165 @@ func getCassandraBaseType(name string) Type {
 	}
 }
 
-func getCassandraType(name string, logger StdLogger) TypeInfo {
-	if strings.HasPrefix(name, "frozen<") {
-		return getCassandraType(strings.TrimPrefix(name[:len(name)-1], "frozen<"), logger)
-	} else if strings.HasPrefix(name, "set<") {
+// TODO: Cover with unit tests.
+// Parses long Java-style type definition to internal data structures.
+func getCassandraLongType(name string, protoVer byte, logger StdLogger) TypeInfo {
+	if strings.HasPrefix(name, SET_TYPE) {
 		return CollectionType{
-			NativeType: NativeType{typ: TypeSet},
-			Elem:       getCassandraType(strings.TrimPrefix(name[:len(name)-1], "set<"), logger),
+			NativeType: NewNativeType(protoVer, TypeSet),
+			Elem:       getCassandraLongType(unwrapCompositeTypeDefinition(name, SET_TYPE, '('), protoVer, logger),
 		}
-	} else if strings.HasPrefix(name, "list<") {
+	} else if strings.HasPrefix(name, LIST_TYPE) {
 		return CollectionType{
-			NativeType: NativeType{typ: TypeList},
-			Elem:       getCassandraType(strings.TrimPrefix(name[:len(name)-1], "list<"), logger),
+			NativeType: NewNativeType(protoVer, TypeList),
+			Elem:       getCassandraLongType(unwrapCompositeTypeDefinition(name, LIST_TYPE, '('), protoVer, logger),
 		}
-	} else if strings.HasPrefix(name, "map<") {
-		names := splitCompositeTypes(strings.TrimPrefix(name[:len(name)-1], "map<"))
+	} else if strings.HasPrefix(name, MAP_TYPE) {
+		names := splitJavaCompositeTypes(name, MAP_TYPE)
 		if len(names) != 2 {
-			logger.Printf("Error parsing map type, it has %d subelements, expecting 2\n", len(names))
-			return NativeType{
-				typ: TypeCustom,
-			}
+			logger.Printf("gocql: error parsing map type, it has %d subelements, expecting 2\n", len(names))
+			return NewNativeType(protoVer, TypeCustom)
 		}
 		return CollectionType{
-			NativeType: NativeType{typ: TypeMap},
-			Key:        getCassandraType(names[0], logger),
-			Elem:       getCassandraType(names[1], logger),
+			NativeType: NewNativeType(protoVer, TypeMap),
+			Key:        getCassandraLongType(names[0], protoVer, logger),
+			Elem:       getCassandraLongType(names[1], protoVer, logger),
 		}
-	} else if strings.HasPrefix(name, "tuple<") {
-		names := splitCompositeTypes(strings.TrimPrefix(name[:len(name)-1], "tuple<"))
+	} else if strings.HasPrefix(name, TUPLE_TYPE) {
+		names := splitJavaCompositeTypes(name, TUPLE_TYPE)
 		types := make([]TypeInfo, len(names))
 
 		for i, name := range names {
-			types[i] = getCassandraType(name, logger)
+			types[i] = getCassandraLongType(name, protoVer, logger)
 		}
 
 		return TupleTypeInfo{
-			NativeType: NativeType{typ: TypeTuple},
+			NativeType: NewNativeType(protoVer, TypeTuple),
 			Elems:      types,
 		}
+	} else if strings.HasPrefix(name, UDT_TYPE) {
+		names := splitJavaCompositeTypes(name, UDT_TYPE)
+		fields := make([]UDTField, len(names)-2)
+
+		for i := 2; i < len(names); i++ {
+			spec := strings.Split(names[i], ":")
+			fieldName, _ := hex.DecodeString(spec[0])
+			fields[i-2] = UDTField{
+				Name: string(fieldName),
+				Type: getCassandraLongType(spec[1], protoVer, logger),
+			}
+		}
+
+		udtName, _ := hex.DecodeString(names[1])
+		return UDTTypeInfo{
+			NativeType: NewNativeType(protoVer, TypeUDT),
+			KeySpace:   names[0],
+			Name:       string(udtName),
+			Elements:   fields,
+		}
+	} else if strings.HasPrefix(name, VECTOR_TYPE) {
+		names := splitJavaCompositeTypes(name, VECTOR_TYPE)
+		subType := getCassandraLongType(strings.TrimSpace(names[0]), protoVer, logger)
+		dim, err := strconv.Atoi(strings.TrimSpace(names[1]))
+		if err != nil {
+			logger.Printf("gocql: error parsing vector dimensions: %v\n", err)
+			return NewNativeType(protoVer, TypeCustom)
+		}
+
+		return VectorType{
+			NativeType: NewCustomType(protoVer, TypeCustom, VECTOR_TYPE),
+			SubType:    subType,
+			Dimensions: dim,
+		}
 	} else {
+		// basic type
 		return NativeType{
-			typ: getCassandraBaseType(name),
+			proto: protoVer,
+			typ:   getApacheCassandraType(name),
 		}
 	}
 }
 
-func splitCompositeTypes(name string) []string {
-	if !strings.Contains(name, "<") {
-		return strings.Split(name, ", ")
+// Parses short CQL type representation (e.g. map<text, text>) to internal data structures.
+func getCassandraType(name string, protoVer byte, logger StdLogger) TypeInfo {
+	if strings.HasPrefix(name, "frozen<") {
+		return getCassandraType(unwrapCompositeTypeDefinition(name, "frozen", '<'), protoVer, logger)
+	} else if strings.HasPrefix(name, "set<") {
+		return CollectionType{
+			NativeType: NewNativeType(protoVer, TypeSet),
+			Elem:       getCassandraType(unwrapCompositeTypeDefinition(name, "set", '<'), protoVer, logger),
+		}
+	} else if strings.HasPrefix(name, "list<") {
+		return CollectionType{
+			NativeType: NewNativeType(protoVer, TypeList),
+			Elem:       getCassandraType(unwrapCompositeTypeDefinition(name, "list", '<'), protoVer, logger),
+		}
+	} else if strings.HasPrefix(name, "map<") {
+		names := splitCQLCompositeTypes(name, "map")
+		if len(names) != 2 {
+			logger.Printf("Error parsing map type, it has %d subelements, expecting 2\n", len(names))
+			return NewNativeType(protoVer, TypeCustom)
+		}
+		return CollectionType{
+			NativeType: NewNativeType(protoVer, TypeMap),
+			Key:        getCassandraType(names[0], protoVer, logger),
+			Elem:       getCassandraType(names[1], protoVer, logger),
+		}
+	} else if strings.HasPrefix(name, "tuple<") {
+		names := splitCQLCompositeTypes(name, "tuple")
+		types := make([]TypeInfo, len(names))
+
+		for i, name := range names {
+			types[i] = getCassandraType(name, protoVer, logger)
+		}
+
+		return TupleTypeInfo{
+			NativeType: NewNativeType(protoVer, TypeTuple),
+			Elems:      types,
+		}
+	} else if strings.HasPrefix(name, "vector<") {
+		names := splitCQLCompositeTypes(name, "vector")
+		subType := getCassandraType(strings.TrimSpace(names[0]), protoVer, logger)
+		dim, _ := strconv.Atoi(strings.TrimSpace(names[1]))
+
+		return VectorType{
+			NativeType: NewCustomType(protoVer, TypeCustom, VECTOR_TYPE),
+			SubType:    subType,
+			Dimensions: dim,
+		}
+	} else {
+		return NativeType{
+			proto: protoVer,
+			typ:   getCassandraBaseType(name),
+		}
+	}
+}
+
+func splitCQLCompositeTypes(name string, typeName string) []string {
+	return splitCompositeTypes(name, typeName, '<', '>')
+}
+
+func splitJavaCompositeTypes(name string, typeName string) []string {
+	return splitCompositeTypes(name, typeName, '(', ')')
+}
+
+func unwrapCompositeTypeDefinition(name string, typeName string, typeOpen int32) string {
+	return strings.TrimPrefix(name[:len(name)-1], typeName+string(typeOpen))
+}
+
+func splitCompositeTypes(name string, typeName string, typeOpen int32, typeClose int32) []string {
+	def := unwrapCompositeTypeDefinition(name, typeName, typeOpen)
+	if !strings.Contains(def, string(typeOpen)) {
+		parts := strings.Split(def, ",")
+		for i := range parts {
+			parts[i] = strings.TrimSpace(parts[i])
+		}
+		return parts
 	}
 	var parts []string
 	lessCount := 0
 	segment := ""
-	for _, char := range name {
+	for _, char := range def {
 		if char == ',' && lessCount == 0 {
 			if segment != "" {
 				parts = append(parts, strings.TrimSpace(segment))
@@ -223,9 +331,9 @@ func splitCompositeTypes(name string) []string {
 			continue
 		}
 		segment += string(char)
-		if char == '<' {
+		if char == typeOpen {
 			lessCount++
-		} else if char == '>' {
+		} else if char == typeClose {
 			lessCount--
 		}
 	}
@@ -233,20 +341,6 @@ func splitCompositeTypes(name string) []string {
 		parts = append(parts, strings.TrimSpace(segment))
 	}
 	return parts
-}
-
-func apacheToCassandraType(t string) string {
-	t = strings.Replace(t, apacheCassandraTypePrefix, "", -1)
-	t = strings.Replace(t, "(", "<", -1)
-	t = strings.Replace(t, ")", ">", -1)
-	types := strings.FieldsFunc(t, func(r rune) bool {
-		return r == '<' || r == '>' || r == ','
-	})
-	for _, typ := range types {
-		t = strings.Replace(t, typ, getApacheCassandraType(typ).String(), -1)
-	}
-	// This is done so it exactly matches what Cassandra returns
-	return strings.Replace(t, ",", ", ", -1)
 }
 
 func getApacheCassandraType(class string) Type {
@@ -297,6 +391,10 @@ func getApacheCassandraType(class string) Type {
 		return TypeTuple
 	case "DurationType":
 		return TypeDuration
+	case "SimpleDateType":
+		return TypeDate
+	case "UserType":
+		return TypeUDT
 	default:
 		return TypeCustom
 	}
--- a/helpers_test.go
+++ b/helpers_test.go
@@ -30,7 +30,7 @@ import (
 )
 
 func TestGetCassandraType_Set(t *testing.T) {
-	typ := getCassandraType("set<text>", &defaultLogger{})
+	typ := getCassandraType("set<text>", protoVersion4, &defaultLogger{})
 	set, ok := typ.(CollectionType)
 	if !ok {
 		t.Fatalf("expected CollectionType got %T", typ)
@@ -223,11 +223,68 @@ func TestGetCassandraType(t *testing.T) {
 				Elem:       NativeType{typ: TypeDuration},
 			},
 		},
+		{
+			"vector<float, 3>", VectorType{
+				NativeType: NativeType{
+					typ:    TypeCustom,
+					custom: VECTOR_TYPE,
+				},
+				SubType:    NativeType{typ: TypeFloat},
+				Dimensions: 3,
+			},
+		},
+		{
+			"vector<vector<float, 3>, 5>", VectorType{
+				NativeType: NativeType{
+					typ:    TypeCustom,
+					custom: VECTOR_TYPE,
+				},
+				SubType: VectorType{
+					NativeType: NativeType{
+						typ:    TypeCustom,
+						custom: VECTOR_TYPE,
+					},
+					SubType:    NativeType{typ: TypeFloat},
+					Dimensions: 3,
+				},
+				Dimensions: 5,
+			},
+		},
+		{
+			"vector<map<uuid,timestamp>, 5>", VectorType{
+				NativeType: NativeType{
+					typ:    TypeCustom,
+					custom: VECTOR_TYPE,
+				},
+				SubType: CollectionType{
+					NativeType: NativeType{typ: TypeMap},
+					Key:        NativeType{typ: TypeUUID},
+					Elem:       NativeType{typ: TypeTimestamp},
+				},
+				Dimensions: 5,
+			},
+		},
+		{
+			"vector<frozen<tuple<int, float>>, 100>", VectorType{
+				NativeType: NativeType{
+					typ:    TypeCustom,
+					custom: VECTOR_TYPE,
+				},
+				SubType: TupleTypeInfo{
+					NativeType: NativeType{typ: TypeTuple},
+					Elems: []TypeInfo{
+						NativeType{typ: TypeInt},
+						NativeType{typ: TypeFloat},
+					},
+				},
+				Dimensions: 100,
+			},
+		},
 	}
 
 	for _, test := range tests {
 		t.Run(test.input, func(t *testing.T) {
-			got := getCassandraType(test.input, &defaultLogger{})
+			got := getCassandraType(test.input, 0, &defaultLogger{})
 
 			// TODO(zariel): define an equal method on the types?
 			if !reflect.DeepEqual(got, test.exp) {
--- a/marshal.go
+++ b/marshal.go
@@ -172,6 +172,10 @@ func Marshal(info TypeInfo, value interface{}) ([]byte, error) {
 		return marshalDate(info, value)
 	case TypeDuration:
 		return marshalDuration(info, value)
+	case TypeCustom:
+		if vector, ok := info.(VectorType); ok {
+			return marshalVector(vector, value)
+		}
 	}
 
 	// detect protocol 2 UDT
@@ -276,6 +280,10 @@ func Unmarshal(info TypeInfo, data []byte, value interface{}) error {
 		return unmarshalDate(info, data, value)
 	case TypeDuration:
 		return unmarshalDuration(info, data, value)
+	case TypeCustom:
+		if vector, ok := info.(VectorType); ok {
+			return unmarshalVector(vector, data, value)
+		}
 	}
 
 	// detect protocol 2 UDT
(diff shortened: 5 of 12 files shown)


marshal.go
```diff
@@ -1709,6 +1719,163 @@ func unmarshalList(info TypeInfo, data []byte, value interface{}) error {
...
+}
+
+func writeUnsignedVInt(buf *bytes.Buffer, v uint64) {
+	numBytes := computeUnsignedVIntSize(v)
+	if numBytes <= 1 {
+		buf.WriteByte(byte(v))
+		return
+	}
+
+	numBytes = computeUnsignedVIntSize(v)
```
tengu-alt: Could you please explain why `numBytes` are computed two times?

tengu-alt:
Have you tested it with protoV5? If not I would be glad to handle it with #1822.

marshal.go
```diff
@@ -1709,6 +1719,163 @@ func unmarshalList(info TypeInfo, data []byte, value interface{}) error {
...
+		return true
+	case TypeCounter:
+		return true
+	case TypeDuration, TypeDate, TypeTime:
+		return true
+	case TypeDecimal, TypeSmallInt, TypeTinyInt, TypeVarint:
+		return true
+	case TypeInet:
+		return true
+	case TypeList, TypeSet, TypeMap, TypeUDT:
```
tengu-alt: It looks like `TypeTuple` is missing.

joao-r-reis:
> Have you tested it with protoV5?

Hmm protocol v5 shouldn't matter much here, the encoding of the type itself is the same regardless of whether v4 or v5 is being used.

marshal.go
lukasz-antoniak: Very good catch, thank you. I forgot about tuples. Fixed.

marshal.go
lukasz-antoniak: Copy-paste error, my bad. Fixed.

helpers.go
```diff
@@ -162,47 +164,83 @@ func getCassandraBaseType(name string) Type {
 	}
 }
 
-func getCassandraType(name string, logger StdLogger) TypeInfo {
+func getCassandraType(name string, protoVer byte, logger StdLogger) TypeInfo {
```
joao-r-reis: It's interesting that `protoVer` wasn't being set before in a lot of cases... Can we replace all instances of direct struct initialization by a call to the `NewNativeType()` function? It would make it clear in the future what fields need to be provided. Could also create a utility private function that always sets `custom` to `nil` to avoid having to provide `nil` on every call.

joao-r-reis: On a separate note, providing go docs for this function would be great. Something that clarifies that this function is meant to parse the "short" cql type names instead of the "long" java type names.

helpers.go
```diff
@@ -162,47 +164,83 @@ func getCassandraBaseType(name string) Type {
...
-			types[i] = getCassandraType(name, logger)
+			types[i] = getCassandraType(name, protoVer, logger)
 		}
 
 		return TupleTypeInfo{
-			NativeType: NativeType{typ: TypeTuple},
+			NativeType: NativeType{typ: TypeTuple, proto: protoVer},
 			Elems:      types,
 		}
+	} else if strings.HasPrefix(name, "udt<") {
```
joao-r-reis: Does this even exist? I think that the actual UDT name would appear here instead of `udt<>` so I think this entire `if` block is not used

helpers.go
```diff
@@ -236,19 +274,48 @@ func splitCompositeTypes(name string) []string {
...
+				skip -= 1
+				continue
+			}
+			idx := strings.Index(class, ":")
+			class = class[idx+1:]
+			act := getApacheCassandraType(class)
+			val := act.String()
+			switch act {
+			case TypeUDT:
+				val = "udt"
```
joao-r-reis: Shouldn't we just set this to the actual `udt` name? This would be what Cassandra returns in the short name format I believe.

helpers.go
```diff
@@ -236,19 +274,48 @@ func splitCompositeTypes(name string) []string {
 }
 
 func apacheToCassandraType(t string) string {
```
joao-r-reis: Similar to above, small comment to clarify that this function is meant to convert java style long cql type names into the standard short cql type names.

helpers.go
```diff
@@ -236,19 +274,48 @@ func splitCompositeTypes(name string) []string {
...
+			idx := strings.Index(class, ":")
+			class = class[idx+1:]
+			act := getApacheCassandraType(class)
+			val := act.String()
+			switch act {
+			case TypeUDT:
+				val = "udt"
+				skip = 2 // skip next two parameters (keyspace and type ID), do not attempt to resolve their type
+			case TypeCustom:
+				val = getApacheCassandraCustomSubType(class)
```
joao-r-reis: why not use the existing `getApacheCassandraType`?

helpers.go
```diff
@@ -236,19 +274,48 @@ func splitCompositeTypes(name string) []string {
...
 	t = strings.Replace(t, ")", ">", -1)
 	types := strings.FieldsFunc(t, func(r rune) bool {
 		return r == '<' || r == '>' || r == ','
 	})
-	for _, typ := range types {
-		t = strings.Replace(t, typ, getApacheCassandraType(typ).String(), -1)
+	skip := 0
+	for _, class := range types {
+		class = strings.TrimSpace(class)
+		if !isDigitsOnly(class) {
```
joao-r-reis: I think we should call `getApacheCassandraType(class)` and then if the returned value matches the vector type then we keep the "dimensions" parameter as is and if it matches UDT then we replace `UserType` with the actual UDT name and remove the other 2 parameters. No need to check `isDigitsOnly`. We need to use a `for` loop with an index instead of `for ... range` though.

metadata_test.go
```diff
@@ -636,12 +636,14 @@ func TestTypeParser(t *testing.T) {
...
-	// custom
+	// udt
 	assertParseNonCompositeType(
 		t,
 		"org.apache.cassandra.db.marshal.UserType(sandbox,61646472657373,737472656574:org.apache.cassandra.db.marshal.UTF8Type,63697479:org.apache.cassandra.db.marshal.UTF8Type,7a6970:org.apache.cassandra.db.marshal.Int32Type)",
-		assertTypeInfo{Type: TypeCustom, Custom: "org.apache.cassandra.db.marshal.UserType(sandbox,61646472657373,737472656574:org.apache.cassandra.db.marshal.UTF8Type,63697479:org.apache.cassandra.db.marshal.UTF8Type,7a6970:org.apache.cassandra.db.marshal.Int32Type)"},
+		assertTypeInfo{Type: TypeUDT, Custom: ""},
 	)
+
+	// custom
```
joao-r-reis, line 646: should add a vector metadata test here

metadata_test.go
```diff
@@ -700,7 +702,7 @@ func assertParseNonCompositeType(
 ) {
 
 	log := &defaultLogger{}
-	result := parseType(def, log)
+	result := parseType(def, 4, log)
```
joao-r-reis: Use the constant for the protocol version so it is easier to read.

metadata_test.go
```diff
@@ -731,7 +733,7 @@ func assertParseCompositeType(
...
```
joao-r-reis: same here

frame.go
```diff
@@ -928,6 +929,30 @@ func (f *framer) readTypeInfo() TypeInfo {
...
+			//   I think we could agree to use getTypeInfo() when parsing binary type definition
+			//   and parseType() would be responsible for parsing "custom" string definition.
+			//spec := strings.TrimPrefix(simple.custom, VECTOR_TYPE)
+			//spec = spec[1 : len(spec)-1] // remove parenthesis
+			//idx := strings.LastIndex(spec, ",")
+			//typeStr := spec[:idx]
+			//dimStr := spec[idx+1:]
+			//subType := getTypeInfo(strings.TrimSpace(typeStr), f.proto, nopLogger{})
+			//dim, _ := strconv.Atoi(strings.TrimSpace(dimStr))
+			result := parseType(simple.custom, simple.proto, nopLogger{})
```
joao-r-reis: I think parseType is not meant to be used this way, from what I've gathered it is only used to parse some table metadata fields IF the C* version is below 3.0.0 so it is very old code.

joao-r-reis: It looks like `getTypeInfo()` is meant to be used for both cases because it checks for the `org.apache.cassandra.db.marshal.` prefix. My opinion is that we should stay away from the `parseType` function and `typeParser` type since it looks like it hasn't been maintained for a very long time and it is only being used in C* versions below `3.0`. At some point we can delete all of that code when we decide to drop support for these old C* versions.

joao-r-reis (commented):
It's looking solid, I left a few comments. Also note that you left a few TODOs.

helpers.go
lukasz-antoniak: Vector type is implemented as special custom type and not top level type in the [protocol](https://github.com/datastax/go-cassandra-native-protocol/blob/main/specs/native_protocol_v5.spec#L822). `getApacheCassandraType` handles only top level types.

helpers.go
lukasz-antoniak: Changed it analogical to C# driver - UDTs parsed as the last resort.

helpers.go
lukasz-antoniak: Fixed.

metadata_test.go
lukasz-antoniak: Fixed.

metadata_test.go
lukasz-antoniak: Thank you, fixed.

metadata_test.go
lukasz-antoniak: Added. I kept basic vector test here and much more in `vector_test.go` file.

helpers.go
lukasz-antoniak: Updated the code. Hope this is what you had in mind. I had to leave `isDigitsOnly` function not to update vector dimension to `custom`.

helpers.go
lukasz-antoniak: Added. Good suggestion, this will help others understand the functionality.

metadata.go
```diff
@@ -1291,6 +1297,26 @@ func (t *typeParser) parse() typeParserResult {
...
+				class = class.params[0].class
+			}
+			types[i] = class.asTypeInfo()
+		}
+
+		return typeParserResult{
+			isComposite: true,
+			types:       types,
+			reversed:    reversed,
+		}
```
joao-r-reis: this code can be removed now right?

metadata.go
```diff
@@ -1332,15 +1360,60 @@ func (class *typeParserClassNode) asTypeInfo() TypeInfo {
...
+		dim, _ := strconv.Atoi(class.params[1].class.name)
+		return VectorType{
+			NativeType: NativeType{
+				typ:   TypeCustom,
+				proto: class.proto,
+			},
+			SubType:    class.params[0].class.asTypeInfo(),
+			Dimensions: dim,
+		}
+	}
```
joao-r-reis: this code can be removed now right?

helpers.go
```diff
@@ -162,47 +164,82 @@ func getCassandraBaseType(name string) Type {
...
+	} else if strings.Index(name, "<") == -1 {
+		// basic type
 		return NativeType{
-			typ: getCassandraBaseType(name),
+			proto: protoVer,
+			typ:   getCassandraBaseType(name),
+		}
+	} else {
+		// udt
+		idx := strings.Index(name, "<")
```
joao-r-reis: This doesn't exist. You're creating this artificial "udt" short type with the complete udt definition but this doesn't exist in CQL. This function is being used to parse CQL short types returned by C* AND short types that are the result of our "translation" code so it doesn't make sense to add custom logic here that applies to types that are the result of our "translation" code only. The "translation" code should never result in short type names that don't exist in CQL.

We have to handle UDT **long type** parsing **outside** of this function and this function should parse the udt name ONLY which is what C* returns in udt short types and then read the rest of the udt definition via system table.

For the UDT short type parsing you can simply add a TODO since this is not really the scope of the PR. 

For the purpose of this PR I think it's good enough to just handle UDT long type parsing outside of this function. You can still use this function for all types that are not UDT or custom though since UDTs and custom types are the only ones that can't be parsed in an equivalent way regardless of whether they are long or short (UDTs in short format don't have the complete definition and custom types in the short format can't be distinguished from a UDT). Whether it makes it easier to reuse this or just create a separate function to handle long types separately (and potentially getting rid of "translation" code) is up to you.

helpers.go
joao-r-reis: I don't see `isDigitsOnly` anywhere

helpers.go
joao-r-reis: Check my other comment that expands on this but my point is that you're creating a short type that does not exist in CQL and then adding handling logic for that "artificial" type on the short type handling function. Doesn't make a lot of sense.

helpers.go
lukasz-antoniak: Correct. check the logic of `apacheToCassandraType()` now.

metadata.go
lukasz-antoniak: If you remove vector block the following metadata_test.go will fail:
```
// vector
assertParseCompositeType(
	t,
	"org.apache.cassandra.db.marshal.VectorType(org.apache.cassandra.db.marshal.FloatType, 3)",
	[]assertTypeInfo{
		{Type: TypeFloat},
		{Type: TypeCustom, Custom: "3"},
	},
	nil,
)
```
Shall we leave only vector changes here?

metadata.go
lukasz-antoniak: If you remove vector block the following `metadata_test.go` will fail:
```
// vector
assertParseCompositeType(
	t,
	"org.apache.cassandra.db.marshal.VectorType(org.apache.cassandra.db.marshal.FloatType, 3)",
	[]assertTypeInfo{
		{Type: TypeFloat},
		{Type: TypeCustom, Custom: "3"},
	},
	nil,
)
```
Shall we leave only vector changes here?

joao-r-reis: That test should be removed as well yes, this code is not being used to parse vectors so the test is not relevant.

helpers.go
```diff
@@ -162,54 +164,152 @@ func getCassandraBaseType(name string) Type {
...
+	} else if strings.HasPrefix(name, UDT_TYPE) {
+		names := splitJavaCompositeTypes(strings.TrimPrefix(name[:len(name)-1], UDT_TYPE+"("))
+		fields := make([]UDTField, len(names)-2)
+
+		for i := 2; i < len(names); i++ {
+			spec := strings.Split(names[i], ":")
+			fieldName, _ := hex.DecodeString(spec[0])
+			fields[i-2] = UDTField{
+				Name: string(fieldName),
+				Type: getTypeInfo(spec[1], protoVer, logger),
```
joao-r-reis: should this be `getCassandraLongType`?

helpers.go
```diff
@@ -162,54 +164,152 @@ func getCassandraBaseType(name string) Type {
...
+		return UDTTypeInfo{
+			NativeType: NewNativeType(protoVer, TypeUDT),
+			KeySpace:   names[0],
+			Name:       string(udtName),
+			Elements:   fields,
+		}
+	} else if strings.HasPrefix(name, VECTOR_TYPE) {
+		names := splitJavaCompositeTypes(strings.TrimPrefix(name[:len(name)-1], VECTOR_TYPE+"("))
+		subType := getCassandraLongType(strings.TrimSpace(names[0]), protoVer, logger)
+		dim, _ := strconv.Atoi(strings.TrimSpace(names[1]))
```
joao-r-reis: Check error like in the `map` case

```
if err != nil {
    logger.Printf("gocql: error parsing vector dimensions: %v\n", err)
    return NewNativeType(protoVer, TypeCustom)
}
```

helpers.go
```diff
@@ -162,54 +164,152 @@ func getCassandraBaseType(name string) Type {
...
+		}
+	} else if strings.HasPrefix(name, LIST_TYPE) {
+		return CollectionType{
+			NativeType: NewNativeType(protoVer, TypeList),
+			Elem:       getCassandraLongType(strings.TrimPrefix(name[:len(name)-1], LIST_TYPE+"("), protoVer, logger),
+		}
+	} else if strings.HasPrefix(name, MAP_TYPE) {
+		names := splitJavaCompositeTypes(strings.TrimPrefix(name[:len(name)-1], MAP_TYPE+"("))
+		if len(names) != 2 {
+			logger.Printf("Error parsing map type, it has %d subelements, expecting 2\n", len(names))
```
joao-r-reis: I think gocql log statements are usually prefixed by `gocql: ` if I'm not mistaken

metadata.go
```diff
@@ -947,11 +947,11 @@ func getColumnMetadata(session *Session, keyspaceName string) ([]ColumnMetadata,
 	return columns, nil
 }
 
-func getTypeInfo(t string, logger StdLogger) TypeInfo {
+func getTypeInfo(t string, protoVer byte, logger StdLogger) TypeInfo {
 	if strings.HasPrefix(t, apacheCassandraTypePrefix) {
 		t = apacheToCassandraType(t)
```
joao-r-reis: we can get rid of `apacheToCassandraType` function now right? And just call `getCassandraLongType`?

joao-r-reis (commented):
I like it 👍 added a couple of small comments

helpers.go
lukasz-antoniak: Fixed.

helpers.go
lukasz-antoniak: Argh, good catch. Fixed.

helpers.go
lukasz-antoniak: Yes, updated.

metadata.go
lukasz-antoniak: Indeed, as always great review. Updated.

marshal.go
```diff
@@ -1709,6 +1719,169 @@ func unmarshalList(info TypeInfo, data []byte, value interface{}) error {
...
+			}
+			if rv.IsNil() {
+				return nil
+			}
+			rv.Set(reflect.Zero(t))
+			return nil
+		}
+		if k == reflect.Array {
+			if rv.Len() != info.Dimensions {
+				return unmarshalErrorf("unmarshal vector: array with wrong size")
```
joao-r-reis: NIT: maybe add the expected and provided sizes to the error message

vector_test.go
```diff
@@ -0,0 +1,403 @@
...
+					assertEqual(t, "vector", expected[i].String(), actual[i].String())
+				}
+			},
+		},
+		{name: "date", cqlType: TypeDate.String(), value: []time.Time{date1, date2, date3}},
+		{name: "time", cqlType: TypeTimestamp.String(), value: []time.Time{time1, time2, time3}},
+		{name: "smallint", cqlType: TypeSmallInt.String(), value: []int16{127, 256, -1234}},
+		{name: "tinyint", cqlType: TypeTinyInt.String(), value: []int8{127, 9, -123}},
+		{name: "duration", cqlType: TypeDuration.String(), value: []Duration{duration1, duration2, duration3}},
+		// TODO(lantonia): Test vector of custom types
```
joao-r-reis: NIT: a couple of TODOs in this test still

.github/workflows/main.yml
```diff
@@ -197,5 +242,6 @@ jobs:
           sleep 30s
       - name: Integration tests
         run: |
+          source ~/venv/bin/activate
```
joao-r-reis: just a reminder to remove the workflow changes from this PR before wrapping up (we'll have to potentially merge #1833 first) when we get a second +1

joao-r-reis (approved):
Good stuff 👍

marshal.go
lukasz-antoniak: 👍

vector_test.go
lukasz-antoniak: 👍

.github/workflows/main.yml
lukasz-antoniak: Yes, this PR will wait for #1833. I will remove workflow changes form this PR, rebase and squash commits.

jameshartig:
I'd be curious your thoughts on if implementing this after https://github.com/apache/cassandra-gocql-driver/pull/1855 would've made it easier.

joao-r-reis:
> I'd be curious your thoughts on if implementing this after #1855 would've made it easier.

@lukasz-antoniak can you explore this? I imagine that this implementation doesn't have much overlap with #1855 and I'd like to get it in before #1855 if this is the case but I'd like your thoughts on this @lukasz-antoniak 

If we do want to get this done before #1855 then we need a rebase on this PR and then I can review it again

helpers_test.go
```diff
@@ -223,11 +223,38 @@ func TestGetCassandraType(t *testing.T) {
...
+				NativeType: NativeType{
+					typ:    TypeCustom,
+					custom: VECTOR_TYPE,
+				},
+				SubType:    NativeType{typ: TypeFloat},
+				Dimensions: 3,
+			},
+		},
+		{
+			"vector<vector<float, 3>, 5>", VectorType{
```
joao-r-reis, line 237: It would be nice to have a couple of cases with more complex types like `vector<map<uuid,timestamp>, 5>` for example

joao-r-reis (commented):
Looks good, I think we need to move the C* versions to 4.1.x and 5.0.x in CI so that we get coverage for these tests before we approve the PR to be merged

worryg0d (approved):
Everything looks good to me, but we still need a couple of folks to review this one just to ensure we didn't miss something.

helpers_test.go
lukasz-antoniak: Added two more - one with map, and other with frozen tuple.

helpers.go
```diff
@@ -162,54 +164,154 @@ func getCassandraBaseType(name string) Type {
...
+
+		udtName, _ := hex.DecodeString(names[1])
+		return UDTTypeInfo{
+			NativeType: NewNativeType(protoVer, TypeUDT),
+			KeySpace:   names[0],
+			Name:       string(udtName),
+			Elements:   fields,
+		}
+	} else if strings.HasPrefix(name, VECTOR_TYPE) {
+		names := splitJavaCompositeTypes(strings.TrimPrefix(name[:len(name)-1], VECTOR_TYPE+"("))
```
ribaraka: In the `splitCQLCompositeTypes` and `splitJavaCompositeTypes`, the prefix trimming logic is repeated in multiple cases. It could be moved inside those funcs, which would make the code cleaner and reduce repetition.

helpers.go
```diff
@@ -162,54 +164,154 @@ func getCassandraBaseType(name string) Type {
 	}
 }
 
-func getCassandraType(name string, logger StdLogger) TypeInfo {
+// Parses long Java-style type definition to internal data structures.
+func getCassandraLongType(name string, protoVer byte, logger StdLogger) TypeInfo {
```
ribaraka, line 169: Should we add a TODO to cover `getCassandraLongType` with tests?

marshal.go
```diff
@@ -172,6 +172,11 @@ func Marshal(info TypeInfo, value interface{}) ([]byte, error) {
 		return marshalDate(info, value)
 	case TypeDuration:
 		return marshalDuration(info, value)
+	case TypeCustom:
+		switch info.(type) {
+		case VectorType:
+			return marshalVector(info.(VectorType), value)
+		}
```
ribaraka: you can simplify this block of code by assigning the type to var so you eliminate the need for the second type assertion:
```suggestion
	case TypeCustom:
		switch t := info.(type) {
		case VectorType:
			return marshalVector(t, value)
		}
	}
```

ribaraka: Also the switch statement appears to only handle the `VectorType` case without any `default` case.

marshal.go
```diff
@@ -276,6 +281,11 @@ func Unmarshal(info TypeInfo, data []byte, value interface{}) error {
 		return unmarshalDate(info, data, value)
 	case TypeDuration:
 		return unmarshalDuration(info, data, value)
+	case TypeCustom:
+		switch info.(type) {
+		case VectorType:
+			return unmarshalVector(info.(VectorType), data, value)
+		}
```
ribaraka: same as above

marshal.go
```diff
@@ -1716,6 +1726,169 @@ func unmarshalList(info TypeInfo, data []byte, value interface{}) error {
...
+				return nil, err
+			}
+			if isVectorVariableLengthType(info.SubType) {
+				writeUnsignedVInt(buf, uint64(len(item)))
+			}
+			buf.Write(item)
+		}
+		return buf.Bytes(), nil
+	}
+	return nil, marshalErrorf("can not marshal %T into %s", value, info)
```
ribaraka: I think it would be helpful to mention which types are actually marshalable (e.g., array, slice)

marshal.go
```diff
@@ -1716,6 +1726,169 @@ func unmarshalList(info TypeInfo, data []byte, value interface{}) error {
...
+				data = data[elemSize:]
+			}
+			err := Unmarshal(info.SubType, unmarshalData, rv.Index(i).Addr().Interface())
+			if err != nil {
+				return unmarshalErrorf("failed to unmarshal %s into %T: %s", info.SubType, unmarshalData, err.Error())
+			}
+		}
+		return nil
+	}
+	return unmarshalErrorf("can not unmarshal %s into %T", info, value)
```
ribaraka: same as above

marshal.go
```diff
@@ -1716,6 +1726,169 @@ func unmarshalList(info TypeInfo, data []byte, value interface{}) error {
...
+			return nil, marshalErrorf("expected vector with %d dimensions, received %d", info.Dimensions, n)
+		}
+
+		for i := 0; i < n; i++ {
+			item, err := Marshal(info.SubType, rv.Index(i).Interface())
+			if err != nil {
+				return nil, err
+			}
+			if isVectorVariableLengthType(info.SubType) {
+				writeUnsignedVInt(buf, uint64(len(item)))
```
ribaraka, line 1755: nitpick: These functions could use some documentation imo—particularly explaining how it works, the intended purpose, and what happens if `info.SubType` is not `isVectorVariableLengthType`.

marshal.go
```diff
@@ -1716,6 +1726,169 @@ func unmarshalList(info TypeInfo, data []byte, value interface{}) error {
...
+	case TypeCounter:
+		return true
+	case TypeDuration, TypeDate, TypeTime:
+		return true
+	case TypeDecimal, TypeSmallInt, TypeTinyInt, TypeVarint:
+		return true
+	case TypeInet:
+		return true
+	case TypeList, TypeSet, TypeMap, TypeUDT, TypeTuple:
+		return true
```
ribaraka: why don't combine it in one `return`?

marshal.go
OleksiienkoMykyta: The nested switch looks redundant.
Maybe something like this would be more suitable?
```
case TypeCustom:
	if vector, ok := info.(VectorType); ok {
		return marshalVector(vector, value)
	}
```

marshal.go
OleksiienkoMykyta: You mean that instead all of this cases could be single "if", or one case for all suitable types?
As I can see, it's structured by similar types for better readability, not sure that it should be merged into one.

marshal.go
```diff
@@ -1716,6 +1726,169 @@ func unmarshalList(info TypeInfo, data []byte, value interface{}) error {
...
+	var tmp = make([]byte, numBytes)
+	for i := extraBytes; i >= 0; i-- {
+		tmp[i] = byte(v)
+		v >>= 8
+	}
+	tmp[0] |= byte(^(0xff >> uint(extraBytes)))
+	buf.Write(tmp)
+}
+
+func readUnsignedVint(data []byte, start int) (uint64, int, error) {
```
OleksiienkoMykyta: `readUnsignedVint` -> `readUnsignedVInt`, it would be a bit more consistent.

OleksiienkoMykyta: Also, maybe it should be provided with unit tests, maybe even with fuzzing, to make sure that everything works properly.
Also, I think that it should be along with some documentation, to help developers dive into it faster.
Wdyt?

marshal.go
lukasz-antoniak: When dealing with fix-length types, serializing length of the element is not required. `isVectorVariableLengthType ` returns true for all types that require serialization of their length. Not sure what comment should be appropriate, which does not just read the code.

marshal.go
lukasz-antoniak: Readability is the reason.

marshal.go
lukasz-antoniak: There is already `encVint` so maybe let us leave it as is?

lukasz-antoniak: The description is in native protocol, so we do not need to duplicate it: https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v5.spec#L392C23-L399C45. Method signature already indicates what is the purpose. I can add a test.

OleksiienkoMykyta: It would be inconsistent with `writeUnsignedVInt`

helpers.go
lukasz-antoniak: Done.

helpers.go
lukasz-antoniak: For now added `TODO`. This is tested indirectly in e.g. `TestFunctionMetadata`.

marshal.go
lukasz-antoniak: Done.

marshal.go
lukasz-antoniak: Done.

marshal.go
lukasz-antoniak: The message is in-line with others. Do we want to make an exception?

tengu-alt: There is actually a messadges with the provided accepted types like in [here](https://github.com/apache/cassandra-gocql-driver/blob/trunk/marshal.go#L338)

tengu-alt:
Overall, looks good to me, except [this](https://github.com/apache/cassandra-gocql-driver/pull/1828?notification_referrer_id=NT_kwDOBPXNN7QxMjgwMDgzMDIyOTo4MzIxNzcxOQ#discussion_r2034742664) and [this](https://github.com/apache/cassandra-gocql-driver/pull/1828?notification_referrer_id=NT_kwDOBPXNN7QxMjgwMDgzMDIyOTo4MzIxNzcxOQ#discussion_r2034129465).

marshal.go
```diff
@@ -1716,6 +1724,169 @@ func unmarshalList(info TypeInfo, data []byte, value interface{}) error {
...
+				return unmarshalErrorf("unmarshal vector: array of size %d cannot store vector of %d dimensions", rv.Len(), info.Dimensions)
+			}
+		} else {
+			rv.Set(reflect.MakeSlice(t, info.Dimensions, info.Dimensions))
+		}
+		elemSize := len(data) / info.Dimensions
+		for i := 0; i < info.Dimensions; i++ {
+			offset := 0
+			if isVectorVariableLengthType(info.SubType) {
+				m, p, err := readUnsignedVint(data, 0)
```
ribaraka: The `start` parameter of the `readUnsignedVint` func seem to be redundunt. Could move inside the func. Or am I missing something?

marshal.go
ribaraka: Thanks for the explanation — that clarifies things. I was curios how data considered "Vector Variable Length" in this case. So if you like it, I suggest adding comment as following:
```
// isVectorVariableLengthType determines if a type requires explicit length serialization within a vector. 
Variable-length types need their length encoded before the actual data to allow proper deserialization.
Fixed-length types, on the other hand, don't require this kind of length prefix.
```

Also, if there are no serialization requirements for fixed-length types, should `TypeSmallInt` and `TypeTinyInt` be removed from `isVectorVariableLengthType`?

marshal.go
```diff
@@ -1716,6 +1724,169 @@ func unmarshalList(info TypeInfo, data []byte, value interface{}) error {
...
+		return true
+	case TypeList, TypeSet, TypeMap, TypeUDT, TypeTuple:
+		return true
+	case TypeCustom:
+		switch elemType.(type) {
+		case VectorType:
+			vecType := elemType.(VectorType)
+			return isVectorVariableLengthType(vecType.SubType)
+		}
+		return true
```
ribaraka: since we agreed on similar switch statement, let's then keep it consistent :pray: 

ctx: https://github.com/apache/cassandra-gocql-driver/pull/1828#discussion_r2035057901

marshal.go
ribaraka: Just showing what you said no to :rofl: 
```
switch elemType.Type() {
	case TypeVarchar, TypeAscii, TypeBlob, TypeText,
		TypeCounter,
		TypeDuration, TypeDate, TypeTime,
		TypeDecimal, TypeSmallInt, TypeTinyInt, TypeVarint,
		TypeInet,
		TypeList, TypeSet, TypeMap, TypeUDT, TypeTuple:
	return true
```

marshal.go
lukasz-antoniak: Done.

marshal.go
lukasz-antoniak: Fixed.

marshal.go
lukasz-antoniak: Done.

marshal.go
lukasz-antoniak: Added your comment. Unfortunately for those types server expects the length also.

marshal.go
lukasz-antoniak: This looks like a discussion about `{` put always on new line in C* code base. Changed as you wish.

marshal.go
lukasz-antoniak: Good catch. Changed to `if` statement like in other cases.

marshal.go
lukasz-antoniak: It is but I would leave it as is when data will not represent just unsigned Vint, but longer slice. But if that is a problem, I can change it.

marshal.go
ribaraka: I wasn't insisting, just sharing a visual.

marshal.go
ribaraka: Hardcoding is kind of distracting. We can always add the parameter back when there's an actual requirement for it. I'd probably remove it if there is no use of it. 
@joao-r-reis, I’d appreciate your thoughts on this when you have a moment.

lukasz-antoniak: Done.

joao-r-reis: > Hardcoding is kind of distracting. We can always add the parameter back when there's an actual requirement for it. I'd probably remove it if there is no use of it. @joao-r-reis, I’d appreciate your thoughts on this when you have a moment.

Agree

ribaraka (approved):
amazing job!

lukasz-antoniak:
@joao-r-reis: Changed C* versions to run integration tests from 4.0 and 4.1, to 4.1 and 5.0.

joao-r-reis (approved):
Good stuff

joao-r-reis:
I'll wait for approvals by @tengu-alt and @OleksiienkoMykyta before merging this

joao-r-reis:
After they approve please squash the commits and use a commit description according to our [guidelines](https://github.com/apache/cassandra-gocql-driver/blob/trunk/CONTRIBUTING.md#commit-message)

joao-r-reis:
@tengu-alt @OleksiienkoMykyta do you have time to review/approve this PR this week?

OleksiienkoMykyta:
> @tengu-alt @OleksiienkoMykyta do you have time to review/approve this PR this week?

Sure, I'll check it this week

tengu-alt (approved):
Nice work, looks good to me!

OleksiienkoMykyta (approved):
LGTM. Thanks for the great contribution

joao-r-reis:
@lukasz-antoniak can you squash and edit commit msg?

lukasz-antoniak:
Done.

joao-r-reis:
missing entry on `CHANGELOG` @lukasz-antoniak

Issue #1078, opened 2018-03-16
New Feature: Latency Aware Host Selection Policy


robusto:
This is another host selection ("load-balancing" policy) that exists in the Datastax drivers that would help with selecting an appropriate coordinator at query time. Having this available for operational and performance concerns when chained with DCAware and/or TokenAware in larger clusters.

Taken from the Datastax driver's [Java doc](https://docs.datastax.com/en/drivers/java/2.2/com/datastax/driver/core/policies/LatencyAwarePolicy.html):

> When used, this policy will collect the latencies of the queries to each Cassandra node and maintain a per-node latency score (an average). Based on these scores, the policy will penalize (technically, it will ignore them unless no other nodes are up) the nodes that are slower than the best performing node by more than some configurable amount (the exclusion threshold).
> 
> The latency score for a given node is a based on a form of exponential moving average. In other words, the latency score of a node is the average of its previously measured latencies, but where older measurements gets an exponentially decreasing weight. The exact weight applied to a newly received latency is based on the time elapsed since the previous measure (to account for the fact that latencies are not necessarily reported with equal regularity, neither over time nor between different nodes).
> 
> Once a node is excluded from query plans (because its averaged latency grew over the exclusion threshold), its latency score will not be updated anymore (since it is not queried). To give a chance to this node to recover, the policy has a configurable retry period. The policy will not penalize a host for which no measurement has been collected for more than this retry period.

[Java implementation: LatencyAwarePolicy.java](https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/policies/LatencyAwarePolicy.java)

It seems like we would start a single, dedicated goroutine in the policy's Init() method to update latency scores for all hosts in the pool on a configurable interval.

martin-sucha:
For the record, there is a related [host selection policy in Kiwi.com's fork](https://github.com/kiwicom/gocql/blob/354a9342c24f61f448e0eff351eda4438f8d7b55/tokenawarelatency.go), although that one is somewhat specialized, so cannot be used directly.

OleksiienkoMykyta:
According to the discussion under the following PR (https://github.com/apache/cassandra-gocql-driver/pull/1831) it is not relevant.
Candidate to be closed.

joao-r-reis:
[...]

PR #1831, opened 2024-10-15
CASSGO-54: Add a LatencyAwareHostPolicy that will deprioritize the host(s) with worse average latencies

Klose6:
Add a LatencyAwarePolicy that will deprioritize the host(s) with worse average latency to fix the [issue-1078](https://github.com/apache/cassandra-gocql-driver/issues/1078) and also as the counterpart of the LatencyAwarePolicy in the cassandra-java-driver client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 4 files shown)


Klose6:
May I ask anyone to take a look of this pull? thanks!

joao-r-reis:
I don't think there's consensus that we should have such a policy in the driver. It was originally decided on the java driver that there were a lot of negatives about this policy which prompted the java driver developers to not include it when they released version 4.0 of the driver.

Feel free to start a discussion thread on the C* mailing list though.

PR #1833, opened 2024-10-17
CASSGO-20 Fix integration tests

lukasz-antoniak:
- Java 17 is installed for future testing.
- C* 4.0.13 fails to start with CCM if only Java 17 is available.
- Creating environment variables expected by CCM.

### DIFF
--- a/.github/actions/setup-environment/action.yml
+++ b/.github/actions/setup-environment/action.yml
@@ -1,0 +1,50 @@
+name: Setup environment
+description: Setup environment for integration tests execution
+runs:
+  using: "composite"
+  steps:
+    - name: Set up cache for SDKMAN
+      uses: actions/cache@v3
+      with:
+        path: ~/.sdkman
+        key: ${{ runner.os }}-sdkman
+
+    - name: Set up cache for PIP
+      uses: actions/cache@v3
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+
+    - name: Install Java
+      shell: bash
+      run: |
+        echo "Installing SDKMAN..."
+        curl -s "https://get.sdkman.io" | bash
+        source "$HOME/.sdkman/bin/sdkman-init.sh"
+        echo "sdkman_auto_answer=true" >> ~/.sdkman/etc/config
+
+        echo "Installing Java versions..."
+        sdk install java 11.0.24-zulu
+        sdk install java 17.0.12-zulu
+
+        sdk default java 11.0.24-zulu
+        sdk use java 11.0.24-zulu
+
+        echo "JAVA11_HOME=$JAVA_HOME_11_X64" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


joao-r-reis (approved):
👍

.github/workflows/main.yml
```diff
@@ -135,10 +158,32 @@ jobs:
...
+          sdk default java 11.0.24-zulu
+          sdk use java 11.0.24-zulu
+          echo "JAVA_HOME=$JAVA_HOME_11_X64" >> $GITHUB_ENV
+          echo "PATH=$PATH" >> $GITHUB_ENV
       - name: Install CCM
-        run: pip install "git+https://github.com/riptano/ccm.git@${CCM_VERSION}"
+        run: |
+          python3 -m venv ~/venv
+          ~/venv/bin/pip install setuptools
+          ~/venv/bin/pip install "git+https://github.com/riptano/ccm.git@${CCM_VERSION}"
```
ribaraka: How about an idea to create a separate script file to combine the duplicated installation steps for Java and CCM? You can then reuse this script across all jobs for setup. It will reduce duplication and centralize maintenance. This way, if you need to update the installation logic for Java or CCM in the future, you can do it in one place instead of multiple jobs. The workflow becomes more nice and readable with a simple call to the script. 
Adding caching to the installation process can speed up builds by reusing dependencies (e.g., SDKMAN, Java versions, Python packages) across workflow runs. 

I’ve put together an example of this approach, which you can check out here: https://github.com/ribaraka/gocql/pull/26

joao-r-reis: I like that approach but just to clarify, the PR you linked is not a complete "replacement" to what we have because it's missing a few things like cassandra.yaml customization, the commands to run the tests, etc. It's meant more as an example of what could be incorporated to this current PR?

ribaraka: Yes, I quickly put together an example to clarify the idea.

joao-r-reis:
The GH action is failing justfyi

lukasz-antoniak:
Yes, I was actively developing it. @ribaraka, I have refactored the code to GitHub action instead of shell script. This way I do not need to repeat caching steps.

.github/actions/setup-environment/action.yml
```diff
@@ -0,0 +1,50 @@
...
+      shell: bash
+      run: |
+        echo "Creating Python virtual environment..."
+        VENV_DIR="$HOME/venv"
+        python3 -m venv $VENV_DIR
+        source $VENV_DIR/bin/activate
+        pip install --upgrade pip setuptools
+
+        echo "Installing CCM..."
+        pip install "git+https://github.com/riptano/ccm.git@${CCM_VERSION}"
```
ribaraka, line 50: No newline at end of file.

ribaraka: However, there's no newline in the .github/workflows/main.yml file, so I assume it's not required.

joao-r-reis (approved):

Jacko161 (approved):

joao-r-reis:
@lukasz-antoniak can you squash and format your commit message according to the guidelines? After that I'll merge this

lukasz-antoniak:
Squashed and applied commit message format. I did not include short description, because changes are simple.

Issue #910, opened 2017-05-15
Gocql doesn't parse Cassandra version correctly


szhou1234:
Currently Gocql only split version string with ".". However there are cases you want to use "3.0.10-xxx" as version number. In this case, Gocql will just panic. We should match the behavior with Java driver, which uses the below code for parsing version (from VersionNumber.java):

private static final String VERSION_REGEXP = "(\\d+)\\.(\\d+)(\\.\\d+)?(\\.\\d+)?([~\\-]\\w[.\\w]*(?:\\-\\w[.\\w]*)*)?(\\+[.\\w]+)?";

    /**
     * Parse a version from a string.
     * <p/>
     * The version string should have primarily the form X.Y.Z to which can be appended
     * one or more pre-release label after dashes (2.0.1-beta1, 2.1.4-rc1-SNAPSHOT)
     * and an optional build label (2.1.0-beta1+a20ba.sha). Out of convenience, the
     * "patch" version number, Z, can be omitted, in which case it is assumed to be 0.
     *
     * @param version the string to parse
     * @return the parsed version number.
     * @throws IllegalArgumentException if the provided string does not
     *                                  represent a valid version.
     */
    public static VersionNumber parse(String version) {
        if (version == null)
            return null;

        Matcher matcher = pattern.matcher(version);
        if (!matcher.matches())
            throw new IllegalArgumentException("Invalid version number: " + version);

        try {
            int major = Integer.parseInt(matcher.group(1));
            int minor = Integer.parseInt(matcher.group(2));

            String pa = matcher.group(3);
            int patch = pa == null || pa.isEmpty() ? 0 : Integer.parseInt(pa.substring(1)); // dropping the initial '.' since it's included this time

            String dse = matcher.group(4);
            int dsePatch = dse == null || dse.isEmpty() ? -1 : Integer.parseInt(dse.substring(1)); // dropping the initial '.' since it's included this time

            String pr = matcher.group(5);
            String[] preReleases = pr == null || pr.isEmpty() ? null : pr.substring(1).split("\\-"); // drop initial '-' or '~' then split on the remaining ones

            String bl = matcher.group(6);
            String build = bl == null || bl.isEmpty() ? null : bl.substring(1); // drop the initial '+'

            return new VersionNumber(major, minor, patch, dsePatch, preReleases, build);
        } catch (NumberFormatException e) {
            throw new IllegalArgumentException("Invalid version number: " + version);
        }
    }

martin-sucha:
[...]

PR #1836, opened 2024-10-25
CASSGO-49 fix for the cassandra version unmarshal

tengu-alt:
Fix for the #910. 
This PR fixes issue, when the driver is unable to unmarshal Cassandra version, which contains additional annotation.

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -56,6 +56,7 @@ const (
 
 type cassVersion struct {
 	Major, Minor, Patch int
+	Qualifier           string
 }
 
 func (c *cassVersion) Set(v string) error {
@@ -87,13 +88,30 @@ func (c *cassVersion) unmarshal(data []byte) error {
 
 	c.Minor, err = strconv.Atoi(v[1])
 	if err != nil {
-		return fmt.Errorf("invalid minor version %v: %v", v[1], err)
+		vMinor := strings.Split(v[1], "-")
+		if len(vMinor) < 2 {
+			return fmt.Errorf("invalid minor version %v: %v", v[1], err)
+		}
+		c.Minor, err = strconv.Atoi(vMinor[0])
+		if err != nil {
+			return fmt.Errorf("invalid minor version %v: %v", v[1], err)
+		}
+		c.Qualifier = v[1][strings.Index(v[1], "-")+1:]
+		return nil
 	}
 
 	if len(v) > 2 {
 		c.Patch, err = strconv.Atoi(v[2])
 		if err != nil {
-			return fmt.Errorf("invalid patch version %v: %v", v[2], err)
+			vPatch := strings.Split(v[2], "-")
+			if len(vPatch) < 2 {
+				return fmt.Errorf("invalid patch version %v: %v", v[2], err)
+			}
+			c.Patch, err = strconv.Atoi(vPatch[0])
+			if err != nil {
+				return fmt.Errorf("invalid patch version %v: %v", v[2], err)
+			}
+			c.Qualifier = v[2][strings.Index(v[2], "-")+1:]
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


host_source.go
```diff
@@ -56,6 +56,7 @@ const (
 
 type cassVersion struct {
 	Major, Minor, Patch int
+	AdditionalNotation  string
```
testisnullus: In Cassandra (and many Java-based projects), the bit after the dash is often called a qualifier or sometimes a suffix. In semantic versioning, it might be part of a pre-release segment or a build meta after a plus sign. However, Cassandra uses variations like snapshot, rc, etc., which are typically called qualifiers.

So, I suggest changing the name of the AdditionalNotation field to the Qualifier.

testisnullus (changes requested):

host_source.go
joao-r-reis: +1 to this suggestion

host_source.go
```diff
@@ -87,13 +88,29 @@ func (c *cassVersion) unmarshal(data []byte) error {
...
-		return fmt.Errorf("invalid minor version %v: %v", v[1], err)
+		vMinor := strings.Split(v[1], "-")
+		if len(vMinor) < 2 {
+			return fmt.Errorf("invalid minor version %v: %v", v[1], err)
+		}
+		c.Minor, err = strconv.Atoi(vMinor[0])
+		if err != nil {
+			return fmt.Errorf("invalid minor version %v: %v", v[1], err)
+		}
+		c.AdditionalNotation = vMinor[1]
```
joao-r-reis: Should return here as we're parsing the qualifier which should always be at the end. 

Also if the "suffix" of the version is something like `-qualifier1-qualifier2` we want this field to either contain the full suffix string `-qualifier1-qualifier2` or this field should be of type `[]string` and the slice should be `[qualifier1,qualifier2]`. [Here's the relevant code in the java driver](https://github.com/apache/cassandra-java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/api/core/Version.java#L122-L126). 

Personally I think it's fine if we just parse the full suffix string including the first `-`, if anyone is actually interested in parsing these qualifiers then they can do it on their app. In this case the name of the field should probably be `Suffix` instead of `Qualifier` though I think?

joao-r-reis (commented):
Added some comments. Let's get this in gocql 2.0, can you create a JIRA?

tengu-alt:
@joao-r-reis , @testisnullus I have renamed the `AdditionalNotation` field to the `Suffix`, and refactored the way we parse it in case it has more than one `qualifier`. I decided to retrieve it as one string. 

> Personally I think it's fine if we just parse the full suffix string including the first `-`

Could you please clarify why we need to contain the first `-`? I don't get the point of it.

host_source.go
jameshartig: How common will there be multiple qualifiers? I'm fine with `Suffix` but if there isn't realistically going to be more than one qualifier, then I think `Qualifier` is still more appropriate.

jameshartig:
I'm fine with including the `-` if its being called `Suffix` but I also don't know if its a big deal either way. It does look like the java driver removes the leading dashes but it also exposes an array.

jameshartig (approved):

host_source.go
joao-r-reis: Honestly I don't know, I just noticed there's logic to handle this in the java driver and thought we would do the same here but I don't remember the last time I saw a C* version with multiple qualifiers.

joao-r-reis:
James suggested we should keep `Qualifier` because we don't see versions with multiple qualifiers often (or at all recently). I think we can keep the first `-` out of it as it is currently on this PR and maybe just renaming it `Qualifier`, are you ok with this @jameshartig ?

jameshartig:
> James suggested we should keep `Qualifier` because we don't see versions with multiple qualifiers often (or at all recently). I think we can keep the first `-` out of it as it is currently on this PR and maybe just renaming it `Qualifier`, are you ok with this @jameshartig ?

Yep! That sounds good to me

joao-r-reis:
👍 @tengu-alt all we need is to rename `Suffix` to `Qualifier` and we're good to go then

tengu-alt:
Done, also updated the commit message and the changelog.

joao-r-reis (approved):

tengu-alt:
@joao-r-reis should it be merged?

PR #1837, opened 2024-10-28
CASSGO-26 consistency serial was added

### LINKED ISSUE #1832: CASSGO-26 Impossible to set CONSISTENCY SERIAL (select consistency)

ScyllaDB

1.7.0

1.22.5

The `.Consistency` options don't allow for setting `SERIAL` or `LOCAL_SERIAL`, despite them being supported by both Cassandra and Scylla for reads to use Paxos.

```go
type Consistency uint16

const (
	Any         Consistency = 0x00
	One         Consistency = 0x01
	Two         Consistency = 0x02
	Three       Consistency = 0x03
	Quorum      Consistency = 0x04
	All         Consistency = 0x05
	LocalQuorum Consistency = 0x06
	EachQuorum  Consistency = 0x07
	LocalOne    Consistency = 0x0A
)
```

Because the driver does not allow you to `session.Query("CONSISTENCY SERIAL").Exec()` (get's error `no viable alternative at input 'CONSISTENCY'`), it seems to be currently impossible to set SELECT consistency to SERIAL

OleksiienkoMykyta:
Closes #1832 
With the previous implementation, setting consistency SERIAL for Paxos reads was impossible. Fixed in this PR

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -192,6 +192,9 @@ const (
 
 type Consistency uint16
 
+// SerialConsistency is deprecated. Use Consistency instead.
+type SerialConsistency = Consistency
+
 const (
 	Any         Consistency = 0x00
 	One         Consistency = 0x01
@@ -201,6 +204,8 @@ const (
 	All         Consistency = 0x05
 	LocalQuorum Consistency = 0x06
 	EachQuorum  Consistency = 0x07
+	Serial      Consistency = 0x08
+	LocalSerial Consistency = 0x09
 	LocalOne    Consistency = 0x0A
 )
 
@@ -224,6 +229,10 @@ func (c Consistency) String() string {
 		return "EACH_QUORUM"
 	case LocalOne:
 		return "LOCAL_ONE"
+	case Serial:
+		return "SERIAL"
+	case LocalSerial:
+		return "LOCAL_SERIAL"
 	default:
 		return fmt.Sprintf("UNKNOWN_CONS_0x%x", uint16(c))
 	}
@@ -253,6 +262,10 @@ func (c *Consistency) UnmarshalText(text []byte) error {
 		*c = EachQuorum
 	case "LOCAL_ONE":
 		*c = LocalOne
+	case "SERIAL":
+		*c = Serial
+	case "LOCAL_SERIAL":
+		*c = LocalSerial
 	default:
 		return fmt.Errorf("invalid consistency %q", string(text))
 	}
@@ -260,6 +273,10 @@ func (c *Consistency) UnmarshalText(text []byte) error {
 	return nil
 }
 
+func (c Consistency) isSerial() bool {
+	return c == Serial || c == LocalSerial
+
+}
 func ParseConsistency(s string) Consistency {
 	var c Consistency
 	if err := c.UnmarshalText([]byte(strings.ToUpper(s))); err != nil {
@@ -273,41 +290,6 @@ func ParseConsistency(s string) Consistency {
 func ParseConsistencyWrapper(s string) (consistency Consistency, err error) {
 	err = consistency.UnmarshalText([]byte(strings.ToUpper(s)))
 	return
-}
-
-type SerialConsistency uint16
-
-const (
-	Serial      SerialConsistency = 0x08
-	LocalSerial SerialConsistency = 0x09
-)
-
-func (s SerialConsistency) String() string {
-	switch s {
-	case Serial:
-		return "SERIAL"
-	case LocalSerial:
-		return "LOCAL_SERIAL"
-	default:
-		return fmt.Sprintf("UNKNOWN_SERIAL_CONS_0x%x", uint16(s))
-	}
-}
-
-func (s SerialConsistency) MarshalText() (text []byte, err error) {
-	return []byte(s.String()), nil
-}
-
-func (s *SerialConsistency) UnmarshalText(text []byte) error {
-	switch string(text) {
-	case "SERIAL":
-		*s = Serial
-	case "LOCAL_SERIAL":
-		*s = LocalSerial
-	default:
-		return fmt.Errorf("invalid consistency %q", string(text))
-	}
-
-	return nil
 }
 
 const (
@@ -1441,7 +1423,7 @@ type queryParams struct {
 	values            []queryValues
 	pageSize          int
 	pagingState       []byte
-	serialConsistency SerialConsistency
+	serialConsistency Consistency
 	// v3+
 	defaultTimestamp      bool
 	defaultTimestampValue int64
@@ -1530,7 +1512,7 @@ func (f *framer) writeQueryParams(opts *queryParams) {
 	}
 
 	if opts.serialConsistency > 0 {
-		f.writeConsistency(Consistency(opts.serialConsistency))
+		f.writeConsistency(opts.serialConsistency)
 	}
 
 	if f.proto > protoVersion2 && opts.defaultTimestamp {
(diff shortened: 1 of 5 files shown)


joao-r-reis (commented):
Can't we just remove `SerialConsistency` altogether and use `Consistency` everywhere with these changes? I think it would make no difference and we'd get rid of a type that no longer has a reason to exist. Also we're missing a check on the `ClusterConfig` serial consistency to ensure it's a valid value.

On the other hand we could also keep `SerialConsistency` around to ensure that only `SERIAL` or `LOCAL_SERIAL` can be used when setting the serial consistency level. But to do this we couldn't make it `type SerialConsistency = Consistency` like it is on this PR. We'd have to keep the previous code for `SerialConsistency` AND add a `Serial`+`LocalSerial` for the `Consistency` type...

I believe that other drivers don't have a separate consistency level type for SERIAL CONSISTENCY so I'm +1 on the first option (i.e. remove `SerialConsistency` completely and just use `Consistency` everywhere with appropriate checks when setting the serial consistency).

OleksiienkoMykyta:
@joao-r-reis I had an idea to implement it as you said in the first option, but it's the breaking changes, so I decided to keep backward compatibility. In case this feature is added to v2, there is no reason to leave `SerialConsistency`.
I added the check to make sure that consistency is valid.

joao-r-reis:
> @joao-r-reis I had an idea to implement it as you said in the first option, but it's the breaking changes, so I decided to keep backward compatibility. In case this feature is added to v2, there is no reason to leave `SerialConsistency`. I added the check to make sure that consistency is valid.

Hmm what about if we have `type SerialConsistency = Consistency` like you did on this PR AND we replace all references of the `SerialConsistency` type (on parameters etc.) to `Consistency`? In this scenario I don't think there will be a breaking change. And we can mark `SerialConsistency` as `deprecated`

joao-r-reis:
So basically the only change we would add to this PR is make sure that `SerialConsistency` is not referenced anywhere (replacing with `Consistency`) and marking `SerialConsistency` as deprecated

session.go
```diff
@@ -144,6 +144,10 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 		return nil, errors.New("Can't use both Authenticator and AuthProvider in cluster config.")
 	}
 
+	if cfg.Consistency.IsSerial() {
+		return nil, fmt.Errorf("the default consistency level is not allowed to be SERIAL or LOCAL_SERIAL. Recived value: %v", cfg.Consistency)
```
joao-r-reis: No, it's fine if `Consistency` has `SERIAL` or `LOCAL_SERIAL` (it won't work with writes but sure, maybe the app only has reads). I meant that `cfg.SerialConsistency` has to be `SERIAL` or `LOCAL_SERIAL`

joao-r-reis (approved):

joao-r-reis:
Looks good, please format your commit message following the rules stated in [CONTRIBUTING.md](https://github.com/apache/cassandra-gocql-driver/blob/trunk/CONTRIBUTING.md#commit-message) and add this JIRA to the `CHANGELOG.md` in the `Added` section of `Unreleased`, thanks!

OleksiienkoMykyta:
> > @joao-r-reis I had an idea to implement it as you said in the first option, but it's the breaking changes, so I decided to keep backward compatibility. In case this feature is added to v2, there is no reason to leave `SerialConsistency`. I added the check to make sure that consistency is valid.
> 
> Hmm what about if we have `type SerialConsistency = Consistency` like you did on this PR AND we replace all references of the `SerialConsistency` type (on parameters etc.) to `Consistency`? In this scenario I don't think there will be a breaking change. And we can mark `SerialConsistency` as `deprecated`

I like this idea, it will solve the issue

session.go
OleksiienkoMykyta: I agree, it slipped my mind that consistency could be SERIAL for read operation, thank you for noticing it.

frame.go
```diff
@@ -202,6 +205,8 @@ const (
 	LocalQuorum Consistency = 0x06
 	EachQuorum  Consistency = 0x07
 	LocalOne    Consistency = 0x0A
+	Serial      Consistency = 0x08
+	LocalSerial Consistency = 0x09
```
ribaraka: NIT: move `Serial` and `LocalSerial` above `LocalOne` to maintain sequential order.

session.go
```diff
@@ -144,6 +144,10 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 		return nil, errors.New("Can't use both Authenticator and AuthProvider in cluster config.")
 	}
 
+	if cfg.SerialConsistency > 0 && !cfg.SerialConsistency.IsSerial() {
+		return nil, fmt.Errorf("the default SerialConsistency level is not allowed to be anything else but SERIAL or LOCAL_SERIAL. Recived value: %v", cfg.SerialConsistency)
+	}
```
ribaraka: How about adding a test for this?

ribaraka: And overall, how about adding another test for this new allowance usage with the two consistency levels?

ribaraka: And overall, how about adding another test to verify that these two consistency levels are allowed for use?

OleksiienkoMykyta: I have added some tests to cover each consistency level, please check it

ribaraka (approved):

frame.go
```diff
@@ -253,13 +262,21 @@ func (c *Consistency) UnmarshalText(text []byte) error {
...
+	case "LOCAL_SERIAL":
+		*c = LocalSerial
 	default:
 		return fmt.Errorf("invalid consistency %q", string(text))
 	}
 
 	return nil
 }
 
+func (c Consistency) IsSerial() bool {
```
jameshartig: If there isn't a reason for this to be public, let's keep it private.

jameshartig (changes requested):
Overall, I like it, let's just make that method be private if we don't need to expose it.

frame.go
OleksiienkoMykyta: I just wanted to provide the user with a method that helps to check whether the consistency is Serial, but if you think it's redundant, we can merge without it.

frame.go
```diff
@@ -275,39 +292,15 @@ func ParseConsistencyWrapper(s string) (consistency Consistency, err error) {
...
-
-func (s *SerialConsistency) UnmarshalText(text []byte) error {
-	switch string(text) {
-	case "SERIAL":
-		*s = Serial
-	case "LOCAL_SERIAL":
-		*s = LocalSerial
-	default:
-		return fmt.Errorf("invalid consistency %q", string(text))
+// MustParseConsistency is the same as ParseConsistency except it returns
```
jameshartig: We shouldn't be adding this back.

jameshartig (changes requested):
One small thing

frame.go
OleksiienkoMykyta: Do you mean to delete "MustParseConsistency" function, or just fix the description?

worryg0d: `MustParseConsistency` is removed in this [commit](https://github.com/apache/cassandra-gocql-driver/commit/8bb171c2edfc696f96dc99c8db3499742d69db13). 
I guess you accidentally added it back during conflict resolving

jameshartig: Yeah, don't re-add this function

OleksiienkoMykyta: @jameshartig, Sorry for the delay, I have been focused on reviewing issues. 
Can we merge it now?

jameshartig (approved):

jameshartig:
@OleksiienkoMykyta There's a merge conflict then this can be merged

OleksiienkoMykyta:
> @OleksiienkoMykyta There's a merge conflict then this can be merged

fixed

Issue #671, opened 2016-03-08
Marshalling errors should provide more context


jbdeboer:
Writing my program, I ran into the error:

```
2016/03/08 08:35:36 can not marshal int into timestamp
```

While accurate, I wasn't able to figure the accepted timestamp types until I dug through the source code.
A better error would be:

```
2016/03/08 08:35:36 can not marshal int into timestamp. Accepted types are Marshaler, int64, time.Time
```

Zariel:
Yep this could be improved, but the marshaller should allow you to marshal an int into a timestamp column

tengu-alt:
I'll handle it

OleksiienkoMykyta:
@tengu-alt, looks like you forgot to add closes to your PR, it should be closed as resolved.
@joao-r-reis, could you help with it, please?

joao-r-reis:
v2.0.0-rc1 with the fix for this has been released, it would be great if users could try it out and provide feedback on it before 2.0.0 GA

PR #1838, opened 2024-10-30
CASSGO-38 Marshal error messages enhancement

tengu-alt:
This PR provides marshal/unmarshal error messages enhancement.
According to #671 errors were updated by the supported types list.

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -110,6 +110,8 @@ type Unmarshaler interface {
 //	duration                    | time.Duration      |
 //	duration                    | gocql.Duration     |
 //	duration                    | string             | parsed with time.ParseDuration
+//
+// The marshal/unmarshal error provides a list of supported types when an unsupported type is attempted.
 func Marshal(info TypeInfo, value interface{}) ([]byte, error) {
 	if info.Version() < protoVersion1 {
 		panic("protocol version not set")
@@ -333,7 +335,7 @@ func marshalVarchar(info TypeInfo, value interface{}) ([]byte, error) {
 	case k == reflect.Slice && t.Elem().Kind() == reflect.Uint8:
 		return rv.Bytes(), nil
 	}
-	return nil, marshalErrorf("can not marshal %T into %s", value, info)
+	return nil, marshalErrorf("can not marshal %T into %s. Accepted types: Marshaler, string, []byte, UnsetValue.", value, info)
 }
 
 func unmarshalVarchar(info TypeInfo, data []byte, value interface{}) error {
@@ -372,7 +374,7 @@ func unmarshalVarchar(info TypeInfo, data []byte, value interface{}) error {
 		rv.SetBytes(dataCopy)
 		return nil
 	}
-	return unmarshalErrorf("can not unmarshal %s into %T", info, value)
+	return unmarshalErrorf("can not unmarshal %s into %T. Accepted types: Unmarshaler, *string, *[]byte", info, value)
 }
 
 func marshalSmallInt(info TypeInfo, value interface{}) ([]byte, error) {
@@ -450,7 +452,7 @@ func marshalSmallInt(info TypeInfo, value interface{}) ([]byte, error) {
 		}
 	}
 
-	return nil, marshalErrorf("can not marshal %T into %s", value, info)
+	return nil, marshalErrorf("can not marshal %T into %s. Accepted types: Marshaler, int16, uint16, int8, uint8, int, uint, int32, uint32, int64, uint64, string, UnsetValue.", value, info)
 }
 
 func marshalTinyInt(info TypeInfo, value interface{}) ([]byte, error) {
@@ -534,7 +536,7 @@ func marshalTinyInt(info TypeInfo, value interface{}) ([]byte, error) {
 		}
 	}
 
-	return nil, marshalErrorf("can not marshal %T into %s", value, info)
+	return nil, marshalErrorf("can not marshal %T into %s. Accepted types: Marshaler, int8, uint8, int16, uint16, int, uint, int32, uint32, int64, uint64, string, UnsetValue.", value, info)
 }
 
 func marshalInt(info TypeInfo, value interface{}) ([]byte, error) {
@@ -606,7 +608,7 @@ func marshalInt(info TypeInfo, value interface{}) ([]byte, error) {
 		}
 	}
 
-	return nil, marshalErrorf("can not marshal %T into %s", value, info)
+	return nil, marshalErrorf("can not marshal %T into %s. Accepted types: Marshaler, int8, uint8, int16, uint16, int, uint, int32, uint32, int64, uint64, string, UnsetValue.", value, info)
 }
 
 func encInt(x int32) []byte {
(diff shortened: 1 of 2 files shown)


marshal.go
```diff
@@ -333,7 +333,7 @@ func marshalVarchar(info TypeInfo, value interface{}) ([]byte, error) {
 	case k == reflect.Slice && t.Elem().Kind() == reflect.Uint8:
 		return rv.Bytes(), nil
 	}
-	return nil, marshalErrorf("can not marshal %T into %s", value, info)
+	return nil, marshalErrorf("can not marshal %T into %s. Accepted types: Marshaler, string, []byte, unsetColumn.", value, info)
```
worryg0d: `unsetColumn` is not an exported type, so the driver doesn't have to expose it to the user. There is a `UnsetValue` global var of type `unsetColumn` which is meant to be used to ignore writing. You can find it [here](https://github.com/apache/cassandra-gocql-driver/blob/2a5056d6b1d1d479ca5ee97ddd1d95a9d8e8c4e2/frame.go#L50).

marshal.go
```diff
@@ -2280,7 +2280,7 @@ func marshalUDT(info TypeInfo, value interface{}) ([]byte, error) {
 	}
 
 	if k.Kind() != reflect.Struct || !k.IsValid() {
-		return nil, marshalErrorf("cannot marshal %T into %s", value, info)
+		return nil, marshalErrorf("cannot marshal %T into %s. Accepted types: Marshaler, unsetColumn, UDTMarshaler, UDTMarshaler, map[string]interface{}, struct.", value, info)
```
worryg0d: Here **UDTMarshaler** is duplicated

marshal.go
tengu-alt: Thank you, refactored

marshal.go
tengu-alt: fixed.

marshal.go
worryg0d: I meant to change unsetColumn to UnsetValue, not to delete it at all

tengu-alt: I see, fixed.

ribaraka:
https://issues.apache.org/jira/browse/CASSGO-38

marshal.go
```diff
@@ -2208,7 +2210,7 @@ func unmarshalTuple(info TypeInfo, data []byte, value interface{}) error {
 		return nil
 	}
 
-	return unmarshalErrorf("cannot unmarshal %s into %T", info, value)
+	return unmarshalErrorf("cannot unmarshal %s into %T. Accepted types: struct, []interface{}, array, slice.", info, value)
```
joao-r-reis: Unmarshaler is also accepted I think. Also the accepted types other than `[]interface{}` have to be a pointer if I'm reading the code correctly (so `*struct, *array, *slice`)

marshal.go
```diff
@@ -2280,7 +2282,7 @@ func marshalUDT(info TypeInfo, value interface{}) ([]byte, error) {
 	}
 
 	if k.Kind() != reflect.Struct || !k.IsValid() {
-		return nil, marshalErrorf("cannot marshal %T into %s", value, info)
+		return nil, marshalErrorf("cannot marshal %T into %s. Accepted types: Marshaler, UDTMarshaler, map[string]interface{}, struct, UnsetValue.", value, info)
```
joao-r-reis, line 2285: It looks like `struct` is also accepted here

marshal.go
```diff
@@ -1706,7 +1708,7 @@ func unmarshalList(info TypeInfo, data []byte, value interface{}) error {
 		}
 		return nil
 	}
-	return unmarshalErrorf("can not unmarshal %s into %T", info, value)
+	return unmarshalErrorf("can not unmarshal %s into %T. Accepted types: slice, array.", info, value)
```
joao-r-reis: I think it has to be `*slice, *array`

marshal.go
```diff
@@ -2392,7 +2394,7 @@ func unmarshalUDT(info TypeInfo, data []byte, value interface{}) error {
 	}
 	k := rv.Elem()
 	if k.Kind() != reflect.Struct || !k.IsValid() {
-		return unmarshalErrorf("cannot unmarshal %s into %T", info, value)
+		return unmarshalErrorf("cannot unmarshal %s into %T. Accepted types: Unmarshaler, UDTUnmarshaler, *map[string]interface{}, struct.", info, value)
```
joao-r-reis: `*struct`

marshal.go
tengu-alt: Good catch! Thank you.

marshal.go
tengu-alt: Explain please, because `struct` is already here.

marshal.go
tengu-alt: Fixed.

marshal.go
tengu-alt: Fixed.

marshal.go
joao-r-reis: Oh I'm sorry, I was probably looking at the wrong place

joao-r-reis (approved):

jameshartig:
It's not very common in Go for the error to explicitly provide the supported types but I can see how that's useful. Ideally a developer would be able to know the supported types before encountering an error. In #671 what made you think that timestamp would support an int? Was it just common sense or did you read something indicating as much?

I'd like to see improvements to the docs and a general improvement in the supported types. That said, nothing is wrong with this PR.

tengu-alt:
> In https://github.com/apache/cassandra-gocql-driver/issues/671 what made you think that timestamp would support an int? Was it just common sense or did you read something indicating as much?

I don't clearly understand your question. The driver already supports `Timestamp` marshal into `int64`, as said in the [codebase](https://github.com/apache/cassandra-gocql-driver/blob/trunk/marshal.go#L1273) and [documentation](https://github.com/apache/cassandra-gocql-driver/blob/trunk/marshal.go#L87).

jameshartig:
> I don't clearly understand your question.

In #671 you ran into:

> 2016/03/08 08:35:36 can not marshal int into timestamp

Which meant you sent an int, that should be supported. I'm trying to understand how we can improve the documentation or API to make it more obvious which types are supported that doesn't involve someone having to parse an error after already writing code which assumed a certain type.

tengu-alt:
> In https://github.com/apache/cassandra-gocql-driver/issues/671 you ran into:

That was not me who ran into the error, the #671 was created by another person.

As Cassandra [doc](https://docs.datastax.com/en/cql-oss/3.x/cql/cql_reference/timestamp_type_r.html) says: `timestamp` is encoded as `int64`. The default `int` in GO is a 32bit integer, so it is not supported.

IMHO the documentation in the `marshal.go` is pretty obvious, and it appears when you call the `func Marshal()`.

jameshartig:
> That was not me who ran into the error, the https://github.com/apache/cassandra-gocql-driver/issues/671 was created by another person.

Oh apologies for the confusion.

> The default int in GO is a 32bit integer, so it is not supported.

Only on 32-bit systems, which are increasingly rare. We should be fine marshalling into Timestamp but maybe we just couldn't safely unmarshal

jameshartig (approved):

tengu-alt:
@joao-r-reis I have updated the `Changelog.md`, should it be merged?

PR #1847, opened 2024-12-06
WIP CASSGO-29

### LINKED ISSUE #465: refactor packages to separate frontend api to backend internal api
Currently everything exists within the gocql package, as the driver functionality expands the package gets more and more complicated.

We should maintain the current API in the gocql package, but move much of the internal things out into sub packages.

OleksiienkoMykyta:
Closes #465


OleksiienkoMykyta:
@joao-r-reis, could you please check it out? It is still a little bit muddy, but it is more about the architecture. Also, I use type aliases to provide users with functionality that should be able to use, but can't be easily separated from internal logic.

joao-r-reis (commented):
Hmm this will be harder than I thought. There's a lot of issues that I can see in this diff because previously exported stuff is going into an internal package.

Let's try a different approach, instead of trying to do the separation between "public" and "internal" + partioning code into different packages at the same time let's focus only on having a single internal package and let's move everything that we see is possible to move (and is unexported) to that new internal package. 

We can worry about creating multiple internal packages afterwards, the difficult and complex part is making this separation of public and internal so let's focus on that first.

In this first draft let's keep everything that is exported on the public `gocql` package for now.

OleksiienkoMykyta:
> Hmm this will be harder than I thought. There's a lot of issues that I can see in this diff because previously exported stuff is going into an internal package.
> 
> Let's try a different approach, instead of trying to do the separation between "public" and "internal" + partioning code into different packages at the same time let's focus only on having a single internal package and let's move everything that we see is possible to move (and is unexported) to that new internal package.
> 
> We can worry about creating multiple internal packages afterwards, the difficult and complex part is making this separation of public and internal so let's focus on that first.
> 
> In this first draft let's keep everything that is exported on the public `gocql` package for now.

Okay, I will make one more implementation, and then we can compare it.

OleksiienkoMykyta:
It should be closed in favor of https://github.com/apache/cassandra-gocql-driver/pull/1857

PR #1848, opened 2024-12-12
CASSGO-22: Refactor Query and Batch to be immutable

lukasz-antoniak:
Fixes [CASSGO-22](https://issues.apache.org/jira/browse/CASSGO-22).

### DIFF
--- a/query_executor.go
+++ b/query_executor.go
@@ -33,7 +33,7 @@ import (
 type ExecutableQuery interface {
 	borrowForExecution()    // Used to ensure that the query stays alive for lifetime of a particular execution goroutine.
 	releaseAfterExecution() // Used when a goroutine finishes its execution attempts, either with ok result or an error.
-	execute(ctx context.Context, conn *Conn) *Iter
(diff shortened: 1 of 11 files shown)


query_executor.go
```diff
@@ -122,11 +126,11 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
 	case iter := <-results:
 		return iter, nil
 	case <-ctx.Done():
-		return &Iter{err: ctx.Err()}, nil
+		return NewIterErr(qry, ctx.Err()), nil
 	}
 }
 
-func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery, hostIter NextHost) *Iter {
+func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery, it *Iter, hostIter NextHost) *Iter {
```
lukasz-antoniak, line 133: Since we move "mutable" query execution attributes into `Iter`, I think it does make sense to add it to the method chain that executes given query.

joao-r-reis:
Can I close this in favor of #1868 ?

PR #1852, opened 2024-12-26
CASSGO-40: skip metadata only if prepared result included metadata

jameshartig:
If the prepare result doesn't include metadata we shouldn't skip metadata when executing the query. Fixes CASSGO-40.

Patch by James Hartig for CASSGO-40

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Don't return error to caller with RetryType Ignore (CASSGO-28)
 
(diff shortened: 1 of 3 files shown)


joao-r-reis:
I believe this is already handled in #1822 can you double check? I believe this change was required to support the new result metadata id flag in v5

jameshartig:
@joao-r-reis It wasn't handled but we could do it in there if we wanted to. I left a comment [here](https://github.com/apache/cassandra-gocql-driver/pull/1822/files#r1934230576).

joao-r-reis (approved):

joao-r-reis:
Yeah let's merge this first, don't forget to add me as a reviewer to the commit msg 👍

PR #1855, opened 2025-01-02
CASSGO-43: externally-defined type registration

jameshartig:
CASSGO-43: externally-defined type registration
The new RegisteredTypes struct can be used to register externally-defined
types. You'll need to define your own marshalling and unmarshalling code
as well as a TypeInfo implementation. The name and id MUST not collide
with existing and future native CQL types.

A lot of the type handling was refactored to use the new format for
types.

inet columns are now unmarshaled as net.IP which is a breaking change.

```
goos: linux
goarch: amd64
pkg: github.com/gocql/gocql
cpu: AMD EPYC 7B13
                             │    old.txt    │                new3.txt                │
                             │    sec/op     │    sec/op     vs base                  │
SingleConn-16                   25.87µ ±  2%   26.49µ ±  1%   +2.40% (p=0.000 n=10)
ParseRowsFrame-16               870.5n ±  3%   683.1n ±  4%  -21.52% (p=0.000 n=10)
UnmarshalVarchar-16             47.29n ±  2%
UnmarshalUUID-16                3.161n ±  2%
Unmarshal_BigInt-16             17.61n ±  9%   13.47n ±  0%  -23.48% (p=0.000 n=10)
Unmarshal_Blob-16               19.19n ±  0%   14.79n ±  2%  -22.88% (p=0.000 n=10)
Unmarshal_Boolean-16            15.71n ±  0%   12.20n ±  1%  -22.32% (p=0.000 n=10)
Unmarshal_Date-16               18.53n ±  1%   15.98n ±  0%  -13.73% (p=0.000 n=10)
Unmarshal_Decimal-16           172.15n ±  2%   96.55n ±  1%  -43.92% (p=0.000 n=10)
Unmarshal_Double-16             16.33n ±  1%   12.21n ±  0%  -25.20% (p=0.000 n=10)
Unmarshal_Duration-16           26.03n ±  1%   22.30n ±  1%  -14.33% (p=0.000 n=10)
Unmarshal_Float-16              15.70n ±  0%   12.21n ±  0%  -22.17% (p=0.000 n=10)
Unmarshal_Int-16                17.80n ±  1%   14.43n ±  0%  -18.96% (p=0.000 n=10)
Unmarshal_Inet-16               28.68n ±  1%   25.25n ±  1%  -11.96% (p=0.000 n=10)
Unmarshal_SmallInt-16           17.92n ±  1%   14.45n ±  1%  -19.37% (p=0.000 n=10)
Unmarshal_Time-16               16.00n ±  1%   12.25n ±  3%  -23.47% (p=0.000 n=10)
Unmarshal_Timestamp-16          16.00n ±  2%   12.23n ±  0%  -23.62% (p=0.000 n=10)
Unmarshal_TinyInt-16            16.38n ±  3%   14.40n ±  1%  -12.09% (p=0.000 n=10)
Unmarshal_UUID-16               16.05n ±  1%   12.56n ±  1%  -21.77% (p=0.000 n=10)
Unmarshal_Varchar-16            18.86n ±  1%   14.81n ±  1%  -21.43% (p=0.000 n=10)
Unmarshal_List-16               176.2n ±  4%   172.9n ±  1%   -1.90% (p=0.002 n=10)
Unmarshal_Set-16                177.7n ±  1%   172.9n ±  0%   -2.67% (p=0.000 n=10)
Unmarshal_Map-16                374.0n ±  5%   348.0n ±  2%   -6.94% (p=0.000 n=10)
Unmarshal_TupleStrings-16       387.9n ±  1%   206.1n ±  1%  -46.85% (p=0.000 n=10)
Unmarshal_TupleInterfaces-16    485.4n ±  2%   289.6n ±  3%  -40.34% (p=0.000 n=10)
FramerReadTypeInfo-16           207.8n ±  3%   168.6n ±  5%  -18.89% (p=0.000 n=10)
ConnStress-16                   14.10µ ± 16%   16.05µ ± 19%        ~ (p=0.363 n=10)
ConnRoutingKey-16               275.3n ±  4%   276.9n ±  2%        ~ (p=0.869 n=10)
WikiCreateSchema-16             515.9m ±  3%   484.9m ±  3%   -6.02% (p=0.000 n=10)
WikiCreatePages-16              1.534m ±  1%   1.469m ±  1%   -4.21% (p=0.000 n=10)
WikiSelectAllPages-16           1.982m ±  3%   1.900m ±  1%   -4.13% (p=0.000 n=10)
WikiSelectSinglePage-16         1.529m ±  2%   1.472m ±  0%   -3.72% (p=0.000 n=10)
WikiSelectPageCount-16          1.691m ±  2%   1.629m ±  2%   -3.67% (p=0.000 n=10)
FramerReadCol-16                107.7n ±  3%   124.6n ±  1%  +15.79% (p=0.000 n=10)
geomean                         375.5n         391.1n        -15.90%                ¹
¹ benchmark set differs from baseline; geomeans may not be comparable

                             │    old.txt     │                 new3.txt                 │
                             │      B/op      │     B/op      vs base                    │
SingleConn-16                  3.155Ki ± 0%     3.157Ki ± 0%   +0.06% (p=0.000 n=10)
ParseRowsFrame-16               1128.0 ± 0%       872.0 ± 0%  -22.70% (p=0.000 n=10)
UnmarshalVarchar-16              32.00 ± 0%
UnmarshalUUID-16                 0.000 ± 0%
Unmarshal_BigInt-16              0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Blob-16                0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Boolean-16             0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Date-16                0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Decimal-16             96.00 ± 0%       48.00 ± 0%  -50.00% (p=0.000 n=10)
Unmarshal_Double-16              0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Duration-16            0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Float-16               0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Int-16                 0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Inet-16                4.000 ± 0%       4.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_SmallInt-16            0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Time-16                0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Timestamp-16           0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_TinyInt-16             0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_UUID-16                0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Varchar-16             0.000 ± 0%       0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_List-16                32.00 ± 0%       32.00 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Set-16                 32.00 ± 0%       32.00 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Map-16                 280.0 ± 0%       280.0 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_TupleStrings-16       158.00 ± 0%       62.00 ± 0%  -60.76% (p=0.000 n=10)
Unmarshal_TupleInterfaces-16    190.00 ± 0%       94.00 ± 0%  -50.53% (p=0.000 n=10)
FramerReadTypeInfo-16           160.00 ± 0%       96.00 ± 0%  -40.00% (p=0.000 n=10)
ConnStress-16                  2.956Ki ± 0%     2.958Ki ± 0%        ~ (p=0.238 n=10)
ConnRoutingKey-16                96.00 ± 0%       96.00 ± 0%        ~ (p=1.000 n=10) ¹
WikiCreateSchema-16            59.90Ki ± 3%     57.90Ki ± 1%   -3.34% (p=0.004 n=10)
WikiCreatePages-16             4.357Ki ± 0%     4.355Ki ± 0%        ~ (p=0.236 n=10)
WikiSelectAllPages-16          38.32Ki ± 0%     38.32Ki ± 0%        ~ (p=0.322 n=10)
WikiSelectSinglePage-16        3.783Ki ± 0%     3.781Ki ± 0%        ~ (p=0.239 n=10)
WikiSelectPageCount-16         3.215Ki ± 0%     3.212Ki ± 0%        ~ (p=0.232 n=10)
FramerReadCol-16                 99.00 ± 0%       67.00 ± 0%  -32.32% (p=0.000 n=10)
geomean                                     ²                 -10.43%                ³ ²
¹ all samples are equal
² summaries must be >0 to compute geomean
³ benchmark set differs from baseline; geomeans may not be comparable

                             │    old.txt    │                new3.txt                │
                             │   allocs/op   │ allocs/op   vs base                    │
SingleConn-16                   37.00 ± 0%     37.00 ± 0%        ~ (p=1.000 n=10) ¹
ParseRowsFrame-16               21.00 ± 0%     13.00 ± 0%  -38.10% (p=0.000 n=10)
UnmarshalVarchar-16             1.000 ± 0%
UnmarshalUUID-16                0.000 ± 0%
Unmarshal_BigInt-16             0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Blob-16               0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Boolean-16            0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Date-16               0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Decimal-16            4.000 ± 0%     3.000 ± 0%  -25.00% (p=0.000 n=10)
Unmarshal_Double-16             0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Duration-16           0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Float-16              0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Int-16                0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Inet-16               1.000 ± 0%     1.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_SmallInt-16           0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Time-16               0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Timestamp-16          0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_TinyInt-16            0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_UUID-16               0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Varchar-16            0.000 ± 0%     0.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_List-16               2.000 ± 0%     2.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Set-16                2.000 ± 0%     2.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_Map-16                5.000 ± 0%     5.000 ± 0%        ~ (p=1.000 n=10) ¹
Unmarshal_TupleStrings-16       8.000 ± 0%     4.000 ± 0%  -50.00% (p=0.000 n=10)
Unmarshal_TupleInterfaces-16   10.000 ± 0%     6.000 ± 0%  -40.00% (p=0.000 n=10)
FramerReadTypeInfo-16           4.000 ± 0%     4.000 ± 0%        ~ (p=1.000 n=10) ¹
ConnStress-16                   35.00 ± 0%     35.00 ± 0%        ~ (p=1.000 n=10)
ConnRoutingKey-16               4.000 ± 0%     4.000 ± 0%        ~ (p=1.000 n=10) ¹
WikiCreateSchema-16             858.0 ± 3%     746.0 ± 1%  -13.05% (p=0.000 n=10)
WikiCreatePages-16              56.00 ± 2%     56.00 ± 2%        ~ (p=1.000 n=10)
WikiSelectAllPages-16           343.0 ± 0%     343.0 ± 0%        ~ (p=1.000 n=10) ¹
WikiSelectSinglePage-16         50.00 ± 0%     50.00 ± 0%        ~ (p=1.000 n=10) ¹
WikiSelectPageCount-16          45.00 ± 0%     45.00 ± 0%        ~ (p=1.000 n=10) ¹
FramerReadCol-16                3.000 ± 0%     3.000 ± 0%        ~ (p=1.000 n=10) ¹
geomean                                    ²                -6.38%                ³ ²
¹ all samples are equal
² summaries must be >0 to compute geomean
³ benchmark set differs from baseline; geomeans may not be comparable
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -249,6 +249,14 @@ func (s *Session) dialWithoutObserver(ctx context.Context, host *HostInfo, cfg *
 		writeTimeout = cfg.WriteTimeout
 	}
 
+	logger := cfg.Logger
+	if logger == nil {
+		logger = s.logger
+		if logger == nil {
+			logger = &defaultLogger{}
+		}
+	}
+
 	ctx, cancel := context.WithCancel(ctx)
 	c := &Conn{
 		r: &connReader{
@@ -274,7 +282,7 @@ func (s *Session) dialWithoutObserver(ctx context.Context, host *HostInfo, cfg *
 		},
 		ctx:            ctx,
 		cancel:         cancel,
-		logger:         cfg.logger(),
+		logger:         logger,
 		streamObserver: s.streamObserver,
 		writeTimeout:   writeTimeout,
 	}
@@ -674,7 +682,7 @@ func (c *Conn) processFrame(ctx context.Context, r io.Reader) error {
 		return fmt.Errorf("gocql: frame header stream is beyond call expected bounds: %d", head.stream)
 	} else if head.stream == -1 {
 		// TODO: handle cassandra event frames, we shouldnt get any currently
-		framer := newFramer(c.compressor, c.version)
+		framer := newFramer(c.compressor, c.version, c.session.types)
 		if err := framer.readFrame(r, &head); err != nil {
 			return err
 		}
@@ -683,7 +691,7 @@ func (c *Conn) processFrame(ctx context.Context, r io.Reader) error {
 	} else if head.stream <= 0 {
 		// reserved stream that we dont use, probably due to a protocol error
 		// or a bug in Cassandra, this should be an error, parse it and return.
-		framer := newFramer(c.compressor, c.version)
+		framer := newFramer(c.compressor, c.version, c.session.types)
 		if err := framer.readFrame(r, &head); err != nil {
 			return err
 		}
@@ -713,7 +721,7 @@ func (c *Conn) processFrame(ctx context.Context, r io.Reader) error {
 		panic(fmt.Sprintf("call has incorrect streamID: got %d expected %d", call.streamID, head.stream))
 	}
 
-	framer := newFramer(c.compressor, c.version)
+	framer := newFramer(c.compressor, c.version, c.session.types)
 
 	err = framer.readFrame(r, &head)
 	if err != nil {
@@ -1188,7 +1196,7 @@ func (c *Conn) execInternal(ctx context.Context, req frameBuilder, tracer Tracer
 	}
 
 	// resp is basically a waiting semaphore protecting the framer
-	framer := newFramer(c.compressor, c.version)
+	framer := newFramer(c.compressor, c.version, c.session.types)
 
 	call := &callReq{
 		timeout:  make(chan struct{}),
--- a/frame.go
+++ b/frame.go
@@ -33,8 +33,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net"
-	"runtime"
-	"strconv"
+	"reflect"
 	"strings"
 	"time"
 )
@@ -379,13 +378,16 @@ type framer struct {
 	buf []byte
 
 	customPayload map[string][]byte
+
+	types *RegisteredTypes
 }
 
-func newFramer(compressor Compressor, version byte) *framer {
+func newFramer(compressor Compressor, version byte, r *RegisteredTypes) *framer {
 	buf := make([]byte, defaultBufSize)
 	f := &framer{
 		buf:        buf[:0],
 		readBuffer: buf,
+		types:      r,
 	}
 	var flags byte
 	if compressor != nil {
@@ -494,16 +496,7 @@ func (f *framer) readFrame(r io.Reader, head *frameHeader) error {
 	return nil
 }
 
-func (f *framer) parseFrame() (frame frame, err error) {
-	defer func() {
-		if r := recover(); r != nil {
-			if _, ok := r.(runtime.Error); ok {
-				panic(r)
-			}
-			err = r.(error)
-		}
-	}()
-
+func (f *framer) parseFrame() (frame, error) {
 	if f.header.version.request() {
 		return nil, NewErrProtocol("got a request frame from server: %v", f.header.version)
 	}
@@ -512,42 +505,53 @@ func (f *framer) parseFrame() (frame frame, err error) {
 		f.readTrace()
 	}
 
+	var err error
 	if f.header.flags&flagWarning == flagWarning {
-		f.header.warnings = f.readStringList()
+		f.header.warnings, err = f.readStringList()
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	if f.header.flags&flagCustomPayload == flagCustomPayload {
-		f.customPayload = f.readBytesMap()
+		f.customPayload, err = f.readBytesMap()
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	// assumes that the frame body has been read into rbuf
 	switch f.header.op {
 	case opError:
-		frame = f.parseErrorFrame()
+		return f.parseErrorFrame()
 	case opReady:
-		frame = f.parseReadyFrame()
+		return f.parseReadyFrame()
 	case opResult:
-		frame, err = f.parseResultFrame()
+		return f.parseResultFrame()
 	case opSupported:
-		frame = f.parseSupportedFrame()
+		return f.parseSupportedFrame()
 	case opAuthenticate:
-		frame = f.parseAuthenticateFrame()
+		return f.parseAuthenticateFrame()
 	case opAuthChallenge:
-		frame = f.parseAuthChallengeFrame()
+		return f.parseAuthChallengeFrame()
 	case opAuthSuccess:
-		frame = f.parseAuthSuccessFrame()
+		return f.parseAuthSuccessFrame()
 	case opEvent:
-		frame = f.parseEventFrame()
+		return f.parseEventFrame()
 	default:
 		return nil, NewErrProtocol("unknown op in frame header: %s", f.header.op)
 	}
-
-	return
 }
 
-func (f *framer) parseErrorFrame() frame {
-	code := f.readInt()
-	msg := f.readString()
+func (f *framer) parseErrorFrame() (frame, error) {
+	code, err := f.readInt()
+	if err != nil {
+		return nil, err
+	}
+	msg, err := f.readString()
+	if err != nil {
+		return nil, err
+	}
 
 	errD := errorFrame{
 		frameHeader: *f.header,
@@ -557,123 +561,228 @@ func (f *framer) parseErrorFrame() frame {
 
 	switch code {
 	case ErrCodeUnavailable:
-		cl := f.readConsistency()
-		required := f.readInt()
-		alive := f.readInt()
+		cl, err := f.readConsistency()
+		if err != nil {
+			return nil, err
+		}
+		required, err := f.readInt()
+		if err != nil {
+			return nil, err
+		}
+		alive, err := f.readInt()
+		if err != nil {
+			return nil, err
+		}
 		return &RequestErrUnavailable{
 			errorFrame:  errD,
 			Consistency: cl,
 			Required:    required,
 			Alive:       alive,
-		}
+		}, nil
 	case ErrCodeWriteTimeout:
-		cl := f.readConsistency()
-		received := f.readInt()
-		blockfor := f.readInt()
-		writeType := f.readString()
+		cl, err := f.readConsistency()
+		if err != nil {
+			return nil, err
+		}
+		received, err := f.readInt()
+		if err != nil {
+			return nil, err
+		}
+		blockfor, err := f.readInt()
+		if err != nil {
+			return nil, err
+		}
+		writeType, err := f.readString()
+		if err != nil {
+			return nil, err
+		}
 		return &RequestErrWriteTimeout{
 			errorFrame:  errD,
 			Consistency: cl,
 			Received:    received,
 			BlockFor:    blockfor,
 			WriteType:   writeType,
-		}
+		}, nil
 	case ErrCodeReadTimeout:
-		cl := f.readConsistency()
-		received := f.readInt()
-		blockfor := f.readInt()
-		dataPresent := f.readByte()
+		cl, err := f.readConsistency()
+		if err != nil {
+			return nil, err
+		}
+		received, err := f.readInt()
+		if err != nil {
+			return nil, err
+		}
+		blockfor, err := f.readInt()
+		if err != nil {
+			return nil, err
+		}
+		dataPresent, err := f.readByte()
+		if err != nil {
+			return nil, err
+		}
 		return &RequestErrReadTimeout{
 			errorFrame:  errD,
 			Consistency: cl,
 			Received:    received,
 			BlockFor:    blockfor,
 			DataPresent: dataPresent,
-		}
+		}, nil
 	case ErrCodeAlreadyExists:
-		ks := f.readString()
-		table := f.readString()
+		ks, err := f.readString()
+		if err != nil {
+			return nil, err
+		}
+		table, err := f.readString()
+		if err != nil {
+			return nil, err
+		}
 		return &RequestErrAlreadyExists{
 			errorFrame: errD,
 			Keyspace:   ks,
 			Table:      table,
-		}
+		}, nil
 	case ErrCodeUnprepared:
-		stmtId := f.readShortBytes()
+		stmtId, err := f.readShortBytes()
+		if err != nil {
+			return nil, err
+		}
 		return &RequestErrUnprepared{
 			errorFrame:  errD,
 			StatementId: copyBytes(stmtId), // defensively copy
-		}
+		}, nil
 	case ErrCodeReadFailure:
 		res := &RequestErrReadFailure{
 			errorFrame: errD,
 		}
-		res.Consistency = f.readConsistency()
-		res.Received = f.readInt()
-		res.BlockFor = f.readInt()
+		res.Consistency, err = f.readConsistency()
+		if err != nil {
+			return nil, err
+		}
+		res.Received, err = f.readInt()
+		if err != nil {
+			return nil, err
+		}
+		res.BlockFor, err = f.readInt()
+		if err != nil {
+			return nil, err
+		}
 		if f.proto > protoVersion4 {
-			res.ErrorMap = f.readErrorMap()
+			res.ErrorMap, err = f.readErrorMap()
+			if err != nil {
+				return nil, err
+			}
 			res.NumFailures = len(res.ErrorMap)
 		} else {
-			res.NumFailures = f.readInt()
+			res.NumFailures, err = f.readInt()
+			if err != nil {
+				return nil, err
+			}
 		}
-		res.DataPresent = f.readByte() != 0
+		b, err := f.readByte()
+		if err != nil {
+			return nil, err
+		}
+		res.DataPresent = b != 0
 
-		return res
+		return res, nil
 	case ErrCodeWriteFailure:
 		res := &RequestErrWriteFailure{
 			errorFrame: errD,
 		}
-		res.Consistency = f.readConsistency()
-		res.Received = f.readInt()
-		res.BlockFor = f.readInt()
+		res.Consistency, err = f.readConsistency()
+		if err != nil {
+			return nil, err
+		}
+		res.Received, err = f.readInt()
+		if err != nil {
+			return nil, err
+		}
+		res.BlockFor, err = f.readInt()
+		if err != nil {
+			return nil, err
+		}
 		if f.proto > protoVersion4 {
-			res.ErrorMap = f.readErrorMap()
+			res.ErrorMap, err = f.readErrorMap()
+			if err != nil {
+				return nil, err
+			}
 			res.NumFailures = len(res.ErrorMap)
 		} else {
-			res.NumFailures = f.readInt()
+			res.NumFailures, err = f.readInt()
+			if err != nil {
+				return nil, err
+			}
 		}
-		res.WriteType = f.readString()
-		return res
+		res.WriteType, err = f.readString()
+		if err != nil {
+			return nil, err
+		}
+		return res, nil
 	case ErrCodeFunctionFailure:
 		res := &RequestErrFunctionFailure{
 			errorFrame: errD,
 		}
-		res.Keyspace = f.readString()
-		res.Function = f.readString()
-		res.ArgTypes = f.readStringList()
-		return res
+		res.Keyspace, err = f.readString()
+		if err != nil {
+			return nil, err
+		}
+		res.Function, err = f.readString()
+		if err != nil {
+			return nil, err
+		}
+		res.ArgTypes, err = f.readStringList()
+		if err != nil {
+			return nil, err
+		}
+		return res, nil
 
 	case ErrCodeCDCWriteFailure:
-		res := &RequestErrCDCWriteFailure{
+		return &RequestErrCDCWriteFailure{
 			errorFrame: errD,
-		}
-		return res
+		}, nil
 	case ErrCodeCASWriteUnknown:
 		res := &RequestErrCASWriteUnknown{
 			errorFrame: errD,
 		}
-		res.Consistency = f.readConsistency()
-		res.Received = f.readInt()
-		res.BlockFor = f.readInt()
-		return res
+		res.Consistency, err = f.readConsistency()
+		if err != nil {
+			return nil, err
+		}
+		res.Received, err = f.readInt()
+		if err != nil {
+			return nil, err
+		}
+		res.BlockFor, err = f.readInt()
+		if err != nil {
+			return nil, err
+		}
+		return res, nil
 	case ErrCodeInvalid, ErrCodeBootstrapping, ErrCodeConfig, ErrCodeCredentials, ErrCodeOverloaded,
 		ErrCodeProtocol, ErrCodeServer, ErrCodeSyntax, ErrCodeTruncate, ErrCodeUnauthorized:
 		// TODO(zariel): we should have some distinct types for these errors
-		return errD
+		return errD, nil
 	default:
-		panic(fmt.Errorf("unknown error code: 0x%x", errD.code))
+		return nil, fmt.Errorf("unknown error code: 0x%x", errD.code)
 	}
 }
 
-func (f *framer) readErrorMap() (errMap ErrorMap) {
-	errMap = make(ErrorMap)
-	numErrs := f.readInt()
-	for i := 0; i < numErrs; i++ {
-		ip := f.readInetAdressOnly().String()
-		errMap[ip] = f.readShort()
+func (f *framer) readErrorMap() (ErrorMap, error) {
+	numErrs, err := f.readInt()
+	if err != nil {
+		return nil, err
 	}
-	return
+	errMap := make(ErrorMap, numErrs)
+	for i := 0; i < numErrs; i++ {
+		ip, err := f.readInetAdressOnly()
+		if err != nil {
+			return nil, err
+		}
+		errMap[ip.String()], err = f.readShort()
+		if err != nil {
+			return nil, err
+		}
+	}
+	return errMap, nil
 }
 
 func (f *framer) writeHeader(flags byte, op frameOp, stream int) {
@@ -737,10 +846,10 @@ type readyFrame struct {
 	frameHeader
 }
 
-func (f *framer) parseReadyFrame() frame {
+func (f *framer) parseReadyFrame() (frame, error) {
 	return &readyFrame{
 		frameHeader: *f.header,
-	}
+	}, nil
 }
 
 type supportedFrame struct {
@@ -751,12 +860,15 @@ type supportedFrame struct {
 
 // TODO: if we move the body buffer onto the frameHeader then we only need a single
 // framer, and can move the methods onto the header.
-func (f *framer) parseSupportedFrame() frame {
+func (f *framer) parseSupportedFrame() (frame, error) {
+	s, err := f.readStringMultiMap()
+	if err != nil {
+		return nil, err
+	}
 	return &supportedFrame{
 		frameHeader: *f.header,
-
-		supported: f.readStringMultiMap(),
-	}
+		supported:   s,
+	}, nil
 }
 
 type writeStartupFrame struct {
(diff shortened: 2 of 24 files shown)


jameshartig:
Here's an example of a JSONB type:

```go
package mypackage

import "github.com/gocql/gocql"

var typeJSONB gocql.Type = 0x0080

func init() {
	gocql.RegisterType(typeJSONB, "jsonb", gocql.SimpleCQLType{jsonbCQLType{}})
}

// jsonbCQLType implements the gocql.CQLType interface
type jsonbCQLType struct {}

// Type implements the gocql.CQLType interface
func (jsonbCQLType) Type() gocql.Type {
	return typeJSONB
}

// Zero implements the gocql.CQLType interface
func (jsonbCQLType) Zero() interface{} {
	return []byte(nil)
}

// Marshal implements the gocql.CQLType interface
func (j jsonbCQLType) Marshal(value interface{}) ([]byte, error) {
	switch v := value.(type) {
	case json.RawMessage:
		return v, nil
	case string:
		return []byte(v), nil
	case []byte:
		return v, nil
	}
	if value == nil {
		return nil, nil
	}

	rv := reflect.ValueOf(value)
	t := rv.Type()
	k := t.Kind()
	switch {
	case k == reflect.String:
		return []byte(rv.String()), nil
	case k == reflect.Slice && t.Elem().Kind() == reflect.Uint8:
		return rv.Bytes(), nil
	}
	return nil, gocql.MarshalError(fmt.Sprintf("can not marshal %T into jsonb", value))
}

// Unmarshal implements the gocql.CQLType interface
func (j jsonbCQLType) Unmarshal(data []byte, value interface{}) error {
	switch v := value.(type) {
	case *json.RawMessage:
		if data != nil {
			*v = append((*v)[:0], data...)
		} else {
			*v = nil
		}
		return nil
	case *string:
		*v = string(data)
		return nil
	case *[]byte:
		if data != nil {
			*v = append((*v)[:0], data...)
		} else {
			*v = nil
		}
		return nil
	case *interface{}:
		if data != nil {
			*v = append(json.RawMessage{}, data...)
		} else {
			*v = nil
		}
		return nil

	}

	rv := reflect.ValueOf(value)
	if rv.Kind() != reflect.Ptr {
		return gocql.UnmarshalError(fmt.Sprintf("can not unmarshal jsonb into non-pointer %T", value))
	}
	rv = rv.Elem()
	t := rv.Type()
	k := t.Kind()
	switch {
	case k == reflect.String:
		rv.SetString(string(data))
		return nil
	case k == reflect.Slice && t.Elem().Kind() == reflect.Uint8:
		var dataCopy []byte
		if data != nil {
			dataCopy = make([]byte, len(data))
			copy(dataCopy, data)
		}
		rv.SetBytes(dataCopy)
		return nil
	}
	return gocql.UnmarshalError(fmt.Sprintf("can not unmarshal jsonb into %T", value))
}
```

Or if they don't care about the `json.RawMessage` handling they can simply set an alias:

```go
var typeJSONB gocql.Type = 0x0080

func init() {
	gocql.RegisterAlias(typeJSONB, "blob")
}
```

jameshartig:
I could use some tests of custom types but given that all of the native types are using most of the same code, I feel confident in the guts of the code.

joao-r-reis:
This looks great! #1828 has been waiting to be merged for a while and it looks like the code changes might conflict with some of the changes here on this PR so I'd like to get #1828 merged before this one unless there's benefits to mergind this one first

jameshartig:
> This looks great! #1828 has been waiting to be merged for a while and it looks like the code changes might conflict with some of the changes here on this PR so I'd like to get #1828 merged before this one unless there's benefits to mergind this one first

That makes sense. Going to switch this back to draft so I can resolve those conflicts and take a stab at cleaning up the UDT case we talked about in Slack.

jameshartig:
I simplified the `TypeInfo` struct but had to keep it around so that composite types could store their children type info somewhere. Simple types can just use their `Type` as a `TypeInfo` though. This reduced the boilerplate necessary to add a new type (I updated the above jsonb example) and allowed me to remove the separate composite interface and only have a single interface that's necessary to implement.

jameshartig:
Initially, some of the unmarshal benchmarks were worse so I inlined some of the type switching which is a bit gross but massively improved the benchmarks.

jameshartig:
> Initially, some of the unmarshal benchmarks were worse so I inlined some of the type switching which is a bit gross but massively improved the benchmarks.

I do think we can roll some of those back when Go 1.24 comes out and there's improved Swiss table maps. I'll need to rerun the benchmarks but it would be nice to get rid of all of that boilerplate.

joao-r-reis:
> I do think we can roll some of those back when Go 1.24 comes out and there's improved Swiss table maps. I'll need to rerun the benchmarks but it would be nice to get rid of all of that boilerplate.

We probably want to be careful with requiring such a recent go version, if it's not crucial then we should delay it as much as possible

joao-r-reis:
Is this ready for review btw?

jameshartig:
> Is this ready for review btw?

Yes, but I would like a chance to write some more tests. I think it's in a good place though for review besides the tests.

> We probably want to be careful with requiring such a recent go version, if it's not crucial then we should delay it as much as possible

Yeah, my comment was meant for the distant future.

helpers.go
```diff
@@ -26,187 +26,51 @@ package gocql
...
-		return TypeMap
-	case "ListType":
-		return TypeList
-	case "SetType":
-		return TypeSet
-	case "TupleType":
-		return TypeTuple
-	default:
-		return TypeCustom
+// TODO: move to types.go
```
joao-r-reis: do we still want to keep these TODOs?

helpers.go
```diff
@@ -444,19 +237,41 @@ func (iter *Iter) MapScan(m map[string]interface{}) bool {
...
-			rowData.Values[i] = dest
+	cols := iter.Columns()
+	columnNames := make([]string, 0, len(cols))
+	values := make([]interface{}, 0, len(cols))
+	for _, column := range iter.Columns() {
+		if c, ok := column.TypeInfo.(TupleTypeInfo); ok {
+			for i := range c.Elems {
+				columnName := TupleColumnName(column.Name, i)
+				if dest, ok := m[columnName]; ok {
+					// TODO: check if dest is a pointer and if not, error here
```
joao-r-reis: Some TODOs here too

types.go
```diff
@@ -0,0 +1,431 @@
...
+	CQLType
+	str string
+}
+
+// RegisterType registers a new CQL data type. Type should be the CQL id for
+// the type. Name is the name of the type as returned in the metadata for the
+// column. CQLType is the implementation of the type.
+// This function is not goroutine-safe and must be called before any other usage
+// of the gocql package. Typically this should be called in an init function.
+func RegisterType(typ Type, name string, t CQLType) {
```
joao-r-reis: I wonder if we should make this a part of clusterconfig so that users can have different sessions with different type registrations within the same app

marshal.go
```diff
@@ -2456,37 +3292,29 @@ func unmarshalUDT(info TypeInfo, data []byte, value interface{}) error {
...
-	//
-	// If there is no corresponding Go type for the CQL type, NewWithError returns an error.
-	NewWithError() (interface{}, error)
 }
 
+// NativeType is a simple type that is defined by the Cassandra protocol.
+// Deprecated. Use Type instead.
+// TODO: move to types.go
 type NativeType struct {
 	proto  byte
```
joao-r-reis: I think we're using this field to basically tell the marshaller/unmarshaller what the protocol version is. Maybe we should add protocol version to the marshal interface instead? And then deprecate the current one?

types.go
```diff
@@ -0,0 +1,431 @@
...
+	return NativeType{
+		proto:  byte(proto),
+		typ:    getCassandraType(name),
+		custom: name,
+	}
+}
+
+// Marshal marshals the value for the given TypeInfo into a byte slice.
+func (t customCQLType) Marshal(info TypeInfo, value interface{}) ([]byte, error) {
+	panic("not implemented")
```
joao-r-reis: I don't understand how this new API works for custom types, typically you can have multiple custom types but all of them are of type "custom" and then the name (string) is what identifies the type (this is how vectors are implemented for example).

types.go
```diff
@@ -0,0 +1,431 @@
...
+	TypeInfoFromString(proto int, name string) TypeInfo
+
+	// Marshal should marshal the value for the given TypeInfo into a byte slice
+	Marshal(info TypeInfo, value interface{}) ([]byte, error)
+
+	// Unmarshal should unmarshal the byte slice into the value for the given
+	// TypeInfo. It must support data being nil for nullable columns and it must
+	// support pointers to empty interfaces to get the default go type for the
+	// column.
+	Unmarshal(info TypeInfo, data []byte, value interface{}) error
```
joao-r-reis: I think we need to add protocol version to this function parameters (and marshal). I've mentioned this in a previous comment above

joao-r-reis (commented):
Overall I like this rework of the type code but I think there's a couple of changes that should be made:

- Adding protocol version to the `Marshal()` and `Unmarshal` functions so the marshal/unmarshal code can provide the implementation for multiple protocol versions (e.g. `list` implementation changes according to the protocol version)
- Custom types and UDTs - for these two types we probably need a new `Register` function because the cql type doesn't identify the actual type (there's multiple UDTs and multiple custom types). This can probably be out of scope of this PR since I believe the driver doesn't handle this well anyway but maybe we should just panic when a user tries to call `RegisterType` with UDT or custom type atm.

marshal.go
```diff
@@ -1579,7 +2113,8 @@ func writeCollectionSize(info CollectionType, n int, buf *bytes.Buffer) error {
 	return nil
 }
 
-func marshalList(info TypeInfo, value interface{}) ([]byte, error) {
+// Marshal marshals the value for the given TypeInfo into a byte slice.
+func (listSetCQLType) Marshal(info TypeInfo, value interface{}) ([]byte, error) {
 	listInfo, ok := info.(CollectionType)
```
joao-r-reis: The fact that this implementation relies on accessing `listInfo.proto` which isn't exported shows that there needs to be a better (and accessible) way of knowing what protocol version is being used otherwise user defined registrations won't be able to do this.

types.go
```diff
@@ -0,0 +1,431 @@
...
+	Params(proto int) []reflect.Type
+
+	// TypeInfoFromParams should return a TypeInfo implementation for the type
+	// with the given parameters filled after the return from Params().
+	TypeInfoFromParams(proto int, params []interface{}) TypeInfo
+
+	// TypeInfoFromString should return a TypeInfo implementation for the type with
+	// the given names/classes. Only the portion within the parantheses or arrows
+	// are passed to this function. For simple types, the name passed might be empty.
+	TypeInfoFromString(proto int, name string) TypeInfo
```
joao-r-reis: I wonder if we should enable the possibility of returning an error on these methods? Otherwise the only thing that the implementation can do is `panic` which is far from ideal

types.go
joao-r-reis: Also do we not risk the driver `panic`ing here with this implementation? Or will the driver never call this if it encounters a custom type?

marshal.go
jameshartig: I've been thinking about this a lot. I'm hesitant to make the `Marshal`/`Unmarshal` interface methods have another parameter when just a few types need it, and potentially no external types need it.

I see the `TypeInfo` as a way for the types to store the information they need to marshal and unmarshal, including the list subtype, map subtypes, and also the protocol, if it's needed.

An alternative is that we could have `listSetCQLVersion2Type` and `listSetCQLType` and use a different struct for version 2 (and lower) specifically but that felt like a lot of code for one tiny difference.

types.go
jameshartig: I believe it was already the case where it would panic. I'm still a bit unclear on exactly how to handle custom types. I considered removing this completely, but we'd need to have a bit of custom code to check for `TypeCustom` and then get the name of it and look that up instead. I think that could be handled in a follow-up PR but I can try to handle it here if you want?

marshal.go
jameshartig: This is really only needed for the protocol 2 UDT error. `NativeType` and `NewNativeType` are also a publicly exposed and there might be usages that expect there to be a protocol option.

helpers.go
jameshartig: I think these can be handled in a follow-up PR. I didn't want to do it in this one because it would make the diff even worse.

types.go
jameshartig: We could do that but then it would require locking, and it couldn't be as easy as importing a package that has a side-effect of registering it's type (which is what my employer wanted to do). I don't think it would be very common for people to have different handling for different types in different clusters. Do you think so?

types.go
joao-r-reis: We could leave that for a follow up PR as long as things don't get broken by merging this PR, did this `panic` already exist before these changes?

marshal.go
joao-r-reis: It's also used for collections if I'm not mistaken

types.go
joao-r-reis: > I don't think it would be very common for people to have different handling for different types in different clusters. 

For basic native types maybe not but eventually this API will evolve to support UDTs and custom types and with these types I can definitely see a use case for having separate type registrations on the same app.

joao-r-reis:
So at this time I think the main open discussion point is whether to have these type registration API as a global API that has sideeffects on import or whether we want this to be an API tied to a cluster/session object. I have a pretty strong conviction that a global API is the wrong path but I will obviously defer to what the community agrees on so I think we need a discussion thread on the C* dev mailing list to move forward

types.go
joao-r-reis: > We could do that but then it would require locking

Why? It could be a matter of the global API returning an object that is then provided to clusterconfig or the API could be tied to the clusterconfig object, I don't see how this requires locking

marshal.go
jameshartig: The collections use their own structs and not this one.

types.go
jameshartig: I'm trying to work on supporting both but I'm still not clear how exactly you'd use this for a UDT since UDT's are already supported. Unfortunately anything outside of the package won't have access to `readBytes` and they can already make a struct implement `UnmarshalCQL` which would essentially do the same thing as registering their own type but with a lot less work.

jameshartig:
@joao-r-reis I refactored a lot of the code and even though I'm not done, I'm interested in your thoughts. I think the code is close to where something like `Vector` could be completely defined via the public API.

* A new public `RegisteredTypes` struct which holds types and could be passed to a config. I don't like that the global one is overwritable so I might change it to `GlobalTypes() *RegisteredTypes`. I think I also need to expose methods for getting a type to make it easier for custom types to reuse some existing types for marshalling/unmarshalling.
* `SimpleCQLType` can wrap a `TypeInfo` if you don't need custom parsing.
* Marshal/Unmarshal were moved off of `CQLType` and instead put on `TypeInfo`. This simplifies the protocol version handling and I added an explicit `unsupportedUDTTypeInfo`.
* I added explicit handling of Custom and removed the Marshal/Unmarshal panic, that said, I'm still trying to figure out how to handle unknown custom types. When we parse metadata we need to be able to make `TypeInfo`s for custom types that we don't necessarily know how to marshal/unmarshal so I might need to add some fallback `unknownCustomTypeInfo` type that still panics/errors in Marshal/Unmarshal.
* `Session` now has a `types` field on it. Not pulling it in yet from config. Still need to figure out the best way to "merge" the global one with the session ones. Now passing session along throughout the framer and metadata parsing.
* Removed the `NativeType` type but left `NewNativeType` which does something similar to before since there was a decent amount of usage in other repos.

Overall these changes also significantly improved the benchmarks compared to the last patch and most unmarshal benchmarks are within single-digit percentages now to trunk.

types.go
jameshartig: Yes, we should do that, and that's next on my todo list.

helpers.go
jameshartig: I did end up moving some of these over if I significantly rewrote them

types.go
```diff
@@ -0,0 +1,476 @@
...
+func (r *RegisteredTypes) registerType(typ Type, name string, t CQLType) {
+	if typ == TypeCustom {
+		panic("custom types must be registered with RegisterCustom")
+	}
+
+	if _, ok := r.byType[typ]; ok {
+		panic(fmt.Errorf("type %d already registered", typ))
+	}
+	if _, ok := r.byString[name]; ok {
+		panic(fmt.Errorf("type name %s already registered", name))
```
joao-r-reis: should probably return errors instead of panic

session.go
```diff
@@ -158,9 +159,11 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 		pageSize:        cfg.PageSize,
 		stmtsLRU:        &preparedLRU{lru: lru.New(cfg.MaxPreparedStmts)},
 		connectObserver: cfg.ConnectObserver,
-		ctx:             ctx,
-		cancel:          cancel,
-		logger:          cfg.logger(),
+		// TODO: switch to from the config
+		types:  GlobalTypes,
```
joao-r-reis: also we could add some logic here so that the user can't modify the registrations after a session has been created and therefore not requiring locking. Maybe copying the registrations to a new internal struct that only supports reading them

joao-r-reis:
I'm much more in favor of this new path. I think for the "merging" we should probably just go for a simple approach where the global registrations just aren't used at all if the user provides a struct in the cluster config.

Or you overwrite the global registrations by checking them one by one and you log a warning when an overwrite occurs.

Are you completely against adding the protocol version as a parameter to the `Marshal/Unmarshal` functions of `TypeInfo`? If the concern is how intrusive the change might be for users then we could try and explore ways to make it less intrusive but the core principal of having access to the protocol version being used while marshalling and unmarshalling is quite important in my view

types.go
jameshartig: I'm in favor of panic only because these should happen at init and there's nothing else to do but panic and the stack trace will probably be helpful for anyone calling this.

jameshartig:
> Are you completely against adding the protocol version as a parameter to the Marshal/Unmarshal functions of TypeInfo? If the concern is how intrusive the change might be for users then we could try and explore ways to make it less intrusive but the core principal of having access to the protocol version being used while marshalling and unmarshalling is quite important in my view

I split out the `TypeInfo` from the `CQLType` so that if the protocol is needed it can be set on the `TypeInfo` before it's returned from the `CQLType`. It didn't feel necessary to add it to the `Marshal`/`Unmarshal` methods because internally only the `CollectionType` needs the protocol and everything else doesn't, so why are we forcing all of them to ignore a new parameter when only 1 type needs it?

joao-r-reis:
> It didn't feel necessary to add it to the Marshal/Unmarshal methods because internally only the CollectionType needs the protocol and everything else doesn't, so why are we forcing all of them to ignore a new parameter when only 1 type needs it?

I was mostly thinking about future protocol versions that might change the format of existing types similar to what was done to collections in v3 but IIUC we will be able to do it via `TypeInfo` if it is needed correct? If this is the case then I'm ok with it. 

Also there's #1890 and I'm probably going to merge it after this one which will remove the need to check the protocol version even for collections

joao-r-reis:
I can't find the thread discussing the error vs panic but I got a notification of it so I'll reply here.

> I'm in favor of panic only because these should happen at init and there's nothing else to do but panic and the stack trace will probably be helpful for anyone calling this.

> What if we changed it to MustRegisterType? Then we could add a error-returning version later if there's an ask for it?

I think changing it to `MustRegisterType` is a good compromise. In my mind I can see some users wanting to handle that error because they don't want their app to crash in any circumstance but I'm not sure so waiting to see if there is demand for it is a good idea.

frame.go
```diff
@@ -847,84 +850,96 @@ func (w *writePrepareFrame) buildFrame(f *framer, streamID int) error {
...
-			vector := VectorType{
-				NativeType: simple,
-				SubType:    subType,
-				Dimensions: dim,
-			}
-			return vector
-		}
+	cqlt, ok := f.types.byType[typ]
+	if !ok {
+		panic(fmt.Errorf("unknown type id: %d", typ))
```
joao-r-reis: This doesn't seem like a good idea. In a "normal" scenario this shouldn't happen but if it does for some reason (bug somewhere for example) I think we can just fail the request instead of causing a crash. Before this change `readTypeInfo` would just return the `typ` without checking whether it is valid and then (if I'm reading the old code correctly) `Marshal` would just return an error because of the unknown type.

helpers.go
```diff
@@ -542,19 +154,41 @@ func (iter *Iter) MapScan(m map[string]interface{}) bool {
...
-		return false
-	}
-
-	for i, col := range rowData.Columns {
-		if dest, ok := m[col]; ok {
-			rowData.Values[i] = dest
+	cols := iter.Columns()
+	columnNames := make([]string, 0, len(cols))
+	values := make([]interface{}, 0, len(cols))
+	for _, column := range iter.Columns() {
```
joao-r-reis, line 151: Why did you choose to drop the call to `RowData()` and rewrite this?

marshal.go
```diff
@@ -131,60 +128,11 @@ func Marshal(info TypeInfo, value interface{}) ([]byte, error) {
...
-	}
-
-	// detect protocol 2 UDT
-	if strings.HasPrefix(info.Custom(), "org.apache.cassandra.db.marshal.UserType") && info.Version() < 3 {
-		return nil, ErrorUDTUnavailable
-	}
-
-	// TODO(tux21b): add the remaining types
-	return nil, fmt.Errorf("can not marshal %T into %s", value, info)
+		}*/
```
joao-r-reis: there's a commented section here

marshal.go
```diff
@@ -237,62 +185,12 @@ func Unmarshal(info TypeInfo, data []byte, value interface{}) error {
...
-	}
-
-	// detect protocol 2 UDT
-	if strings.HasPrefix(info.Custom(), "org.apache.cassandra.db.marshal.UserType") && info.Version() < 3 {
-		return ErrorUDTUnavailable
-	}
-
-	// TODO(tux21b): add the remaining types
-	return fmt.Errorf("can not unmarshal %s into %T", info, value)
+	*/
```
joao-r-reis: also a commented section here

marshal.go
```diff
@@ -1909,92 +2008,113 @@ func marshalMap(info TypeInfo, value interface{}) ([]byte, error) {
...
 	t := rv.Type()
-	if t.Kind() != reflect.Map {
-		return unmarshalErrorf("can not unmarshal %s into %T", info, value)
+	if t.Kind() == reflect.Interface {
+		if t.NumMethod() != 0 {
+			return unmarshalErrorf("can not unmarshal map into non-empty interface %T", value)
+		}
+		var key interface{}
+		// this relies on Unmarshal marshalling default values when presented with nil
+		if err := Unmarshal(c.Key, []byte(nil), &key); err != nil {
```
joao-r-reis: This looks a bit odd to me, in my mind `nil` should result in a `nil` value as output from unmarshal but we don't need to invoke the unmarshal function to do it so it could just be a part of the "contract" of this interface that you must return the zero value / default as output if `nil` is the input. On the other hand having an explicit `NewDefault` or `NewZeroValue` or something like this would make it more clear and we could just invoke that method here and in `RowData()` and other places that rely on getting zero values from types.

marshal.go
```diff
@@ -1909,92 +2008,113 @@ func marshalMap(info TypeInfo, value interface{}) ([]byte, error) {
...
+			return err
+		}
+		if key == nil {
+			panic(fmt.Errorf("key was nil after unmarshalling from %T", c.Key))
```
joao-r-reis: a lot of panics here that should just be returned errors to avoid app crashes (I understand that these particular checks are to make sure `Unmarshal` is behaving as expected but still having an app with a lot of failed requests is better than an app crashing most of the time.

joao-r-reis: I've found other panics in this file so I recommend reevaluating all of them but I can specifically call out them in separate comments if you'd like

jameshartig:
> I was mostly thinking about future protocol versions that might change the format of existing types similar to what was done to collections in v3 but IIUC we will be able to do it via `TypeInfo` if it is needed correct? If this is the case then I'm ok with it.

Yeah, I wouldn't say it's perfect but given that now nothing needs to know the protocol version I think it's a good tradeoff between breaking the existing API and it still being possible.

> I think changing it to MustRegisterType is a good compromise. In my mind I can see some users wanting to handle that error because they don't want their app to crash in any circumstance but I'm not sure so waiting to see if there is demand for it is a good idea.

Yeah I went back and forth on it but ended with giving the option of getting an error if you wanted one.

marshal.go
```diff
@@ -2623,251 +2990,23 @@ func unmarshalUDT(info TypeInfo, data []byte, value interface{}) error {
...
-			fmt.Fprint(buf, ",")
-		} else {
-			first = false
+// NewNativeType returns a TypeInfo from the global registered types.
+// Deprecated.
+// TODO: make sure this works with known public usages
+func NewNativeType(proto byte, typ Type, custom string) TypeInfo {
+	if typ == TypeCustom {
+		// TODO: do we need to parse custom? what if its a composite type?
+		rt := GlobalTypes.custom[custom]
```
joao-r-reis: Yeah custom types are weird especially now that we have composite custom types like Vector... Adding a type registration by custom type name was the way to go until Vector came along... Now I'm not so sure. 

The mechanism to provide a custom type registration in the C# and Java drivers are very similar to this, the full type name is matched so I'm pretty sure composite custom types are not supported. However, in the java driver you can provide a codec for the Vector Type and it will be recognized even if the vector type is implemented as a custom type internally so maybe it's fine to keep the custom type registration as it is here as long as it is possible for users to register a marshaler for the Vector type (also note that vector in v6 will be a native type not custom so that's something worth considering as we're designing this atm).

marshal.go
```diff
@@ -2404,12 +2660,139 @@ type UDTUnmarshaler interface {
 	UnmarshalUDT(name string, info TypeInfo, data []byte) error
 }
 
-func marshalUDT(info TypeInfo, value interface{}) ([]byte, error) {
-	udt := info.(UDTTypeInfo)
+type udtCQLType struct {
+	types *RegisteredTypes
+}
```
joao-r-reis, line 2732: I'm still not sure what the goal of the UDT registration is with this API, I don't think users can leverage this API to replace the usage of `UDTMarshaler` and `UDTUnmarshaler` right? I wonder if we should just return an error when a user tries to register a UDT at the moment and create a new ticket to add proper UDT support to this API later in a 2.x minor release.

joao-r-reis: Also, proper UDT support will have to include reading the system tables to get the information about the fields of the UDT so a UDT marshaler will need access to a connection or some kind of API that allows the marshaler to obtain that information from the session.

With that in mind it is very likely that the interface for a UDT marshaler will be different from the other ones or alternatively we only allow users to provide "go type <-> cql user type" mappings instead of a complete udt marshaler. This is another reason to just return an error if users try to do anything with UDTs when using this new API for now.

joao-r-reis (commented):
Added a few comments. Note that some comments from my previous review are still relevant and the threads are still open.

As mentioned in a separate comment, UDT support looks very basic and not really useful for now, users still need to implement the old udt marshaler interface to do anything useful with UDTs if I'm understanding correctly. Due to this I think we should limit what's exported in this API so that we don't leak stuff related to UDTs as this is very likely to change when we add proper UDT support in the future.

frame.go
jameshartig: Fair point, I'll add a new unknownType type for this and then try to write some tests around this.

marshal.go
jameshartig: I'll go through and either comment or replace all of them, thanks for the reminder!

marshal.go
jameshartig: okay for now we can just avoid composite types and get back to this later if we need to

marshal.go
jameshartig: Yeah I envision a `RegisterUDT` in the future but we can wait until post-2.0 for that.

marshal.go
jameshartig: Woops, this was a reminder for me to implement the vector code but I forgot to remove it

marshal.go
jameshartig: Same, this was a reminder and forgot to remove it.

helpers.go
jameshartig: This was using `RowData` to get the default values for every single column and then replacing them, so if you passed a map with all existing columns defined then the `RowData` call was completely pointless. I also had ambitions of getting rid of `RowData`. I did end up getting rid of `rowMap` though.

session.go
jameshartig: I added the copying. Should I add an explicit error to panic after it's set on a session? You can't access the field on the session which I thought was good enough to "protect" against it not being used to register new types.

types.go
jameshartig: Done!

session.go
joao-r-reis: Hmm I think that's good enough honestly, the only way a user would be able to do it is via reflection right?

jameshartig:
I think I addressed all outstanding comments and removed most of the panic calls. I also managed to move around some things and speed up almost every single benchmark.

I'll give it another once over tomorrow.

jameshartig:
@joao-r-reis I just realized the Unmarshal docs say:
> // If value is a pointer to pointer, it is set to nil if the CQL value is
> // null. Otherwise, nulls are unmarshalled as zero value.

We might want to clarify that because that could be read as the "zero value" for the column and not the "zero value" for the Go type. Thoughts?

marshal.go
```diff
@@ -232,95 +176,97 @@ func Unmarshal(info TypeInfo, data []byte, value interface{}) error {
...
+		switch valueElemRef.Kind() {
+		case reflect.Ptr:
+			if data == nil {
+				if valueElemRef.IsNil() {
+					return nil
+				}
+				valueRef.Elem().Set(reflect.Zero(valueElemRef.Type()))
+				return nil
+			}
+			// TODO: can we do this?
```
jameshartig: @joao-r-reis we technically don't need to be allocating a new value if the existing pointer is non-nil but this would be considered a breaking change. The [original commit](https://github.com/jackc/pgx/issues/1174) doesn't mention this as a goal and there was a [related issue filed regarding reuse of byte slices](https://github.com/apache/cassandra-gocql-driver/issues/1348) that was closed as working-as-intended. `json`'s unmarshaller does not allocate a new value. But I also found at least [one case of someone in another driver](https://github.com/jackc/pgx/issues/1174) saying this is confusing. Thoughts?

jameshartig:
I made the metadata methods not error on unknown types and instead return the unknownTypeInfo since it didn't feel right to be erroring just because some return type on a function isn't known.

I also changed `Params(proto int) []reflect.Type` to `Params(proto int) []interface{}` because making the public API handle `reflect.Type` seemed odd and unlike any other public API but it also ended up speeding up the relevant framer methods. @joao-r-reis let me know if that makes sense to you.

marshal.go
joao-r-reis: discussed offline but yeah I think we should probably not change that behavior just to be safe

conn.go
```diff
@@ -249,6 +249,14 @@ func (s *Session) dialWithoutObserver(ctx context.Context, host *HostInfo, cfg *
 		writeTimeout = cfg.WriteTimeout
 	}
 
+	logger := cfg.Logger
```
joao-r-reis, line 252: why not `cfg.logger()`?

joao-r-reis (approved):
Amazing work, this is a significant improvement of both internal and public APIs related to marshaling and CQL types.

conn.go
jameshartig: I made it fallback to the session's logger if the connection didn't have one. If I called `logger()` it would set a default logger without looking at the session.

PR #1856, opened 2025-01-03
CASSGO-44: keep nil slices in MapScan

jameshartig:
Previously if you read a NULL column as a varchar using MapScan it would store the value []byte{} in the map rather than []byte() which made it impossible to know if the value was NULL or not.

Patch by James Hartig for CASSGO-44

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bumped actions/upload-artifact and actions/cache versions to v4 in CI workflow (CASSGO-48)
 
+- Keep nil slices in MapScan (CASSGO-44)
+
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


joao-r-reis:
Have you checked https://issues.apache.org/jira/browse/CASSGO-36 ? Seems related and there's a PR for it I think

joao-r-reis:
Oh I see you have commented on it

joao-r-reis:
Is this issue specific to `varchar` and `[]byte`? Is there a way to tell whether you're reading a `NULL` value from the db on other column types?

jameshartig:
> Is this issue specific to varchar and []byte? Is there a way to tell whether you're reading a NULL value from the db on other column types?

This is just specific to slices of any kind. The way to know if you're reading NULL from other column types is to use a pointer.

joao-r-reis (approved):

Issue #1370, opened 2019-11-06
NewSession sometimes panics when it should return an error


miroswan:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?
3.0.9

### What version of Gocql are you using?
v0.0.0-20190810123941-df4b9cc33030

### What did you do?
gocql.NewSession panics when cassandra has not yet reported that it can receive connections 

```
panic: no valid connect address for host: [HostInfo hostname="0.0.0.0" connectAddress="0.0.0.0" peer="<nil>" rpc_address="<nil>" broadcast_address="<nil>" preferred_ip="<nil>" connect_addr="0.0.0.0" connect_addr_source="invalid" port=9042 data_centre="" rack="" host_id="" version="v0.0.0" state=UP num_tokens=0]. Is your cluster configured correctly?
```

### What did you expect to see?
I expect an error to be returned instead of a panic. This is where things seem to be breaking down:

https://github.com/gocql/gocql/blob/2274c8a9fab68a2be1c558b32eab48cb853163aa/session.go#L221

The ConnectAddress method does not return an error on failure. It panics. If this method returned an error in this condition, we could check for it here and bubble it up. This would allow me to keep checking until Cassandra is up instead of the thread crashing.

miroswan:
I see now from the code that using 0.0.0.0 is not allowed. I've resolved this issue for myself, but I do think it's in our best interest to bubble up errrors to the caller of NewSession instead of causing a panic.

tengu-alt:
@miroswan As I see the panic in this place is reasonable. The `0.0.0.0` should not be used, as it described here:
https://docs.datastax.com/en/cassandra-oss/2.1/cassandra/configuration/configCassandra_yaml_r.html

joao-r-reis:
I actually think there's some merit in returning an error here instead of `panic`. Specially now that we allow custom host dialers so it is possible that the contact point is not even used

tengu-alt:
> I actually think there's some merit in returning an error here instead of `panic`. Specially now that we allow custom host dialers so it is possible that the contact point is not even used

I see. I will take care of it.

OleksiienkoMykyta:
It was solved by https://github.com/apache/cassandra-gocql-driver/pull/1858, @tengu-alt looks like you just forgot to add closes for the Issue, or there are should there be some additional patches provided?

tengu-alt:
Looks like that, Issue should be closed.

[...]

PR #1858, opened 2025-01-13
CASSGO-45 ConnectAddress() should not panic and HostInfo should not be created with invalid connect address

tengu-alt:
Fix for the #1370

Panic from the `ConnectAddress()` was removed.
`HostInfo` struct creation was refactored to create via constructor to make sure the address is valid.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1690,7 +1690,11 @@ func (c *Conn) awaitSchemaAgreement(ctx context.Context) (err error) {
 		}
 
 		for _, row := range rows {
-			host, err := c.session.hostInfoFromMap(row, &HostInfo{connectAddress: c.host.ConnectAddress(), port: c.session.cfg.Port})
+			h, err := newHostInfo(c.host.ConnectAddress(), c.session.cfg.Port)
+			if err != nil {
+				goto cont
+			}
+			host, err := c.session.hostInfoFromMap(row, h)
 			if err != nil {
 				goto cont
 			}
--- a/control.go
+++ b/control.go
@@ -146,7 +146,11 @@ func hostInfo(addr string, defaultPort int) ([]*HostInfo, error) {
 
 	// Check if host is a literal IP address
 	if ip := net.ParseIP(host); ip != nil {
-		hosts = append(hosts, &HostInfo{hostname: host, connectAddress: ip, port: port})
+		h, err := newHostInfo(ip, port)
+		if err != nil {
+			return nil, err
+		}
+		hosts = append(hosts, h)
 		return hosts, nil
 	}
 
@@ -172,7 +176,12 @@ func hostInfo(addr string, defaultPort int) ([]*HostInfo, error) {
 	}
 
 	for _, ip := range ips {
-		hosts = append(hosts, &HostInfo{hostname: host, connectAddress: ip, port: port})
+		h, err := newHostInfo(ip, port)
+		if err != nil {
+			return nil, err
+		}
+
+		hosts = append(hosts, h)
 	}
 
 	return hosts, nil
--- a/host_source.go
+++ b/host_source.go
@@ -181,6 +181,18 @@ type HostInfo struct {
 	tokens           []string
 }
 
+func newHostInfo(addr net.IP, port int) (*HostInfo, error) {
+	if !validIpAddr(addr) {
+		return nil, errors.New("invalid host address")
+	}
+	host := &HostInfo{}
+	host.hostname = addr.String()
+	host.port = port
+
+	host.connectAddress = addr
+	return host, nil
+}
+
 func (h *HostInfo) Equal(host *HostInfo) bool {
 	if h == host {
 		// prevent rlock reentry
@@ -213,14 +225,12 @@ func (h *HostInfo) connectAddressLocked() (net.IP, string) {
 	} else if validIpAddr(h.rpcAddress) {
 		return h.rpcAddress, "rpc_adress"
 	} else if validIpAddr(h.preferredIP) {
-		// where does perferred_ip get set?
 		return h.preferredIP, "preferred_ip"
 	} else if validIpAddr(h.broadcastAddress) {
 		return h.broadcastAddress, "broadcast_address"
-	} else if validIpAddr(h.peer) {
-		return h.peer, "peer"
 	}
-	return net.IPv4zero, "invalid"
+	return h.peer, "peer"
+
 }
 
 // nodeToNodeAddress returns address broadcasted between node to nodes.
@@ -240,24 +250,13 @@ func (h *HostInfo) nodeToNodeAddress() net.IP {
 }
 
 // Returns the address that should be used to connect to the host.
-// If you wish to override this, use an AddressTranslator or
-// use a HostFilter to SetConnectAddress()
+// If you wish to override this, use an AddressTranslator
 func (h *HostInfo) ConnectAddress() net.IP {
 	h.mu.RLock()
 	defer h.mu.RUnlock()
 
-	if addr, _ := h.connectAddressLocked(); validIpAddr(addr) {
-		return addr
-	}
-	panic(fmt.Sprintf("no valid connect address for host: %v. Is your cluster configured correctly?", h))
-}
-
-func (h *HostInfo) SetConnectAddress(address net.IP) *HostInfo {
-	// TODO(zariel): should this not be exported?
-	h.mu.Lock()
-	defer h.mu.Unlock()
-	h.connectAddress = address
-	return h
+	addr, _ := h.connectAddressLocked()
+	return addr
 }
 
 func (h *HostInfo) BroadcastAddress() net.IP {
@@ -491,6 +490,10 @@ func checkSystemSchema(control *controlConn) (bool, error) {
 	return true, nil
 }
 
+func (s *Session) newHostInfoFromMap(addr net.IP, port int, row map[string]interface{}) (*HostInfo, error) {
+	return s.hostInfoFromMap(row, &HostInfo{connectAddress: addr, port: port})
+}
+
 // Given a map that represents a row from either system.local or system.peers
 // return as much information as we can in *HostInfo
 func (s *Session) hostInfoFromMap(row map[string]interface{}, host *HostInfo) (*HostInfo, error) {
@@ -606,6 +609,9 @@ func (s *Session) hostInfoFromMap(row map[string]interface{}, host *HostInfo) (*
 	}
 
 	ip, port := s.cfg.translateAddressPort(host.ConnectAddress(), host.port)
+	if !validIpAddr(ip) {
+		return nil, fmt.Errorf("invalid host address (before translation: %v:%v, after translation: %v:%v)", host.ConnectAddress(), host.port, ip.String(), port)
+	}
 	host.connectAddress = ip
 	host.port = port
 
@@ -623,7 +629,7 @@ func (s *Session) hostInfoFromIter(iter *Iter, connectAddress net.IP, defaultPor
 		return nil, errors.New("query returned 0 rows")
 	}
 
-	host, err := s.hostInfoFromMap(rows[0], &HostInfo{connectAddress: connectAddress, port: defaultPort})
+	host, err := s.newHostInfoFromMap(connectAddress, defaultPort, rows[0])
 	if err != nil {
 		return nil, err
 	}
@@ -674,7 +680,7 @@ func (r *ringDescriber) getClusterPeerInfo(localHost *HostInfo) ([]*HostInfo, er
 
 	for _, row := range rows {
 		// extract all available info about the peer
-		host, err := r.session.hostInfoFromMap(row, &HostInfo{port: r.session.cfg.Port})
+		host, err := r.session.newHostInfoFromMap(nil, r.session.cfg.Port, row)
 		if err != nil {
 			return nil, err
 		} else if !isValidPeer(host) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Standardized spelling of datacenter (CASSGO-35)
 
+- Refactor HostInfo creation and ConnectAddress() method (CASSGO-45)
+
 ### Fixed
 - Cassandra version unmarshal fix (CASSGO-49)
 


cassandra_test.go
```diff
@@ -3086,7 +3090,11 @@ func TestDiscoverViaProxy(t *testing.T) {
 	for _, host := range clusterHosts {
 		found := false
 		for _, hi := range session.pool.hostConnPools {
-			if hi.host.ConnectAddress().String() == host {
+			connAddr, err := hi.host.ConnectAddress()
+			if err != nil {
+				t.Error(err)
```
joao-r-reis: maybe this should be `t.Fatal` so the behavior of this test is consistent after the change?

control.go
```diff
@@ -261,16 +261,21 @@ func (c *controlConn) connect(hosts []*HostInfo) error {
 	var conn *Conn
 	var err error
 	for _, host := range hosts {
+		connAddr, err := host.ConnectAddress()
+		if err != nil {
+			c.session.logger.Printf("gocql: %v\n", err)
```
joao-r-reis: Usually we add a message for this error before we add the actual error string so something like:
`c.session.logger.Printf("gocql: unable to use host for control connection, skipping it: %v\n", err)`

control.go
```diff
@@ -423,16 +432,21 @@ func (c *controlConn) attemptReconnectToAnyOfHosts(hosts []*HostInfo) (*Conn, er
...
```
joao-r-reis: same as comment above

control_test.go
```diff
@@ -49,8 +49,13 @@ func TestHostInfo_Lookup(t *testing.T) {
 		}
 
 		host := hosts[0]
-		if !host.ConnectAddress().Equal(test.ip) {
-			t.Errorf("expected ip %v got %v for addr %q", test.ip, host.ConnectAddress(), test.addr)
+		connAddr, err := host.ConnectAddress()
+		if err != nil {
+			t.Errorf("%d: %v", i, err)
```
joao-r-reis: `t.Errorf("could not get connect address of host %q to compare with expected: %v", test.addr, err)`

filters.go
```diff
@@ -72,10 +72,13 @@ func WhiteListHostFilter(hosts ...string) HostFilter {
 
 	m := make(map[string]bool, len(hostInfos))
 	for _, host := range hostInfos {
-		m[host.ConnectAddress().String()] = true
+		connAddr, _ := host.ConnectAddress()
```
joao-r-reis: if `err != nil` I don't think it should be whitelisted, it should just `continue`

filters.go
```diff
@@ -72,10 +72,13 @@ func WhiteListHostFilter(hosts ...string) HostFilter {
...
+		m[connAddr.String()] = true
 	}
 
 	return HostFilterFunc(func(host *HostInfo) bool {
-		return m[host.ConnectAddress().String()]
+		connAddr, _ := host.ConnectAddress()
+		return m[connAddr.String()]
```
joao-r-reis: if `err != nil` it should just return `false`

host_source.go
```diff
@@ -164,8 +164,9 @@ func (h *HostInfo) Equal(host *HostInfo) bool {
 		// prevent rlock reentry
 		return true
 	}
-
-	return h.ConnectAddress().Equal(host.ConnectAddress())
+	hConnAddr, _ := h.ConnectAddress()
+	hostConnAddr, _ := host.ConnectAddress()
+	return hConnAddr.Equal(hostConnAddr)
```
joao-r-reis: maybe something like this would be better?

```
hConnAddr, err1 := h.ConnectAddress()
hostConnAddr, err2 := host.ConnectAddress()
return err1 == err2 && hConnAddr.Equal(hostConnAddr)
```

integration_test.go
```diff
@@ -139,7 +147,11 @@ func TestHostFilterInitial(t *testing.T) {
 	// we'll filter out the second host
 	filtered := clusterHosts[1]
 	cluster.HostFilter = HostFilterFunc(func(host *HostInfo) bool {
-		if host.ConnectAddress().String() == filtered {
+		connAddr, err := host.ConnectAddress()
+		if err != nil {
+			t.Error(err)
```
joao-r-reis: Maybe it makes more sense to use `t.Fatal` here? (and the 2 other similar cases in this file)

policies.go
```diff
@@ -96,7 +96,8 @@ func (c *cowHostList) remove(ip net.IP) bool {
 	found := false
 	newL := make([]*HostInfo, 0, size)
 	for i := 0; i < len(l); i++ {
-		if !l[i].ConnectAddress().Equal(ip) {
+		connAddr, _ := l[i].ConnectAddress()
+		if !connAddr.Equal(ip) {
```
joao-r-reis: `add()` should check if `.ConnectAddress()` returns an error and if it does it shouldn't add it. If we do that then yeah we can ignore the error here because it's guaranteed to always be `nil`

policies.go
```diff
@@ -353,7 +354,8 @@ func (r *roundRobinHostPolicy) AddHost(host *HostInfo) {
 }
 
 func (r *roundRobinHostPolicy) RemoveHost(host *HostInfo) {
-	r.hosts.remove(host.ConnectAddress())
+	connAddr, _ := host.ConnectAddress()
```
joao-r-reis: We can't ignore the error here, we're basically trying to remove an address that doesn't exist. Just return if there is an error

policies.go
```diff
@@ -519,7 +521,8 @@ func (t *tokenAwareHostPolicy) AddHosts(hosts []*HostInfo) {
 
 func (t *tokenAwareHostPolicy) RemoveHost(host *HostInfo) {
 	t.mu.Lock()
-	if t.hosts.remove(host.ConnectAddress()) {
+	connAddr, _ := host.ConnectAddress()
+	if t.hosts.remove(connAddr) {
```
joao-r-reis: Same as above but needs to call `RemoveHost` on the fallback policy instead of returning right away

policies.go
```diff
@@ -723,7 +726,8 @@ func (r *hostPoolHostPolicy) SetHosts(hosts []*HostInfo) {
 	hostMap := make(map[string]*HostInfo, len(hosts))
 
 	for i, host := range hosts {
-		ip := host.ConnectAddress().String()
+		connAddr, _ := host.ConnectAddress()
+		ip := connAddr.String()
```
joao-r-reis: skip the host if unable to get `connAddr`

policies.go
```diff
@@ -735,7 +739,8 @@ func (r *hostPoolHostPolicy) SetHosts(hosts []*HostInfo) {
 }
 
 func (r *hostPoolHostPolicy) AddHost(host *HostInfo) {
-	ip := host.ConnectAddress().String()
+	connAddr, _ := host.ConnectAddress()
+	ip := connAddr.String()
```
joao-r-reis: if unable to get `connAddr` just return

policies.go
```diff
@@ -756,7 +761,8 @@ func (r *hostPoolHostPolicy) AddHost(host *HostInfo) {
 }
 
 func (r *hostPoolHostPolicy) RemoveHost(host *HostInfo) {
-	ip := host.ConnectAddress().String()
+	connAddr, _ := host.ConnectAddress()
+	ip := connAddr.String()
```
joao-r-reis: same as above

policies.go
```diff
@@ -818,7 +825,8 @@ func (host selectedHostPoolHost) Info() *HostInfo {
 }
 
 func (host selectedHostPoolHost) Mark(err error) {
-	ip := host.info.ConnectAddress().String()
+	connAddr, _ := host.info.ConnectAddress()
```
joao-r-reis: if I'm reading the code correctly this should be fine if we ensure that hosts with invalid connect addr are never added to `hostPoolHostPolicy`

policies.go
```diff
@@ -862,10 +870,11 @@ func (d *dcAwareRR) AddHost(host *HostInfo) {
 }
 
 func (d *dcAwareRR) RemoveHost(host *HostInfo) {
+	connAddr, _ := host.ConnectAddress()
```
joao-r-reis: just return if unable to get conn addr but also need to add a check for this in `AddHost` so hosts with invalid connect addresses are not added

policies.go
```diff
@@ -970,7 +979,8 @@ func (d *rackAwareRR) AddHost(host *HostInfo) {
 
 func (d *rackAwareRR) RemoveHost(host *HostInfo) {
 	dist := d.HostTier(host)
-	d.hosts[dist].remove(host.ConnectAddress())
+	connAddr, _ := host.ConnectAddress()
+	d.hosts[dist].remove(connAddr)
```
joao-r-reis: same as above, check in `AddHost` so we don't add hosts with invalid addresses and then we can just return here when unable to get connAddr

session.go
```diff
@@ -394,7 +394,8 @@ func (s *Session) reconnectDownedHosts(intv time.Duration) {
 			if gocqlDebug {
 				buf := bytes.NewBufferString("Session.ring:")
 				for _, h := range hosts {
-					buf.WriteString("[" + h.ConnectAddress().String() + ":" + h.State().String() + "]")
+					connAddr, _ := h.ConnectAddress()
```
joao-r-reis: we shouldn't really ignore these errors, it's better to handle them than assuming that `connAddr` is always valid, just `continue` if unable to get `connAddr`

session.go
```diff
@@ -847,11 +848,12 @@ func (qm *queryMetrics) hostMetrics(host *HostInfo) *hostMetrics {
 // hostMetricsLocked gets or creates host metrics for given host.
 // It must be called only while holding qm.l lock.
 func (qm *queryMetrics) hostMetricsLocked(host *HostInfo) *hostMetrics {
-	metrics, exists := qm.m[host.ConnectAddress().String()]
+	connAddr, _ := host.ConnectAddress()
+	metrics, exists := qm.m[connAddr.String()]
```
joao-r-reis: We definitely should not be creating metrics objects for hosts with invalid connect addresses since they won't be used anyway.

token.go
```diff
@@ -213,14 +213,15 @@ func (t *tokenRing) String() string {
...
+		connAddr, _ := th.host.ConnectAddress()
 		buf.WriteString(sep)
 		sep = ","
 		buf.WriteString("\n\t[")
 		buf.WriteString(strconv.Itoa(i))
 		buf.WriteString("]")
 		buf.WriteString(th.token.String())
 		buf.WriteString(":")
-		buf.WriteString(th.host.ConnectAddress().String())
+		buf.WriteString(connAddr.String())
```
joao-r-reis: Will this print an empty string if `err != nil` ? Printing empty string on the address part could be fine but we could also print `nil` instead

joao-r-reis (commented):
Left some comments. The host selection / load balancing policies need some changes, we shouldn't assume the address is always valid on those implementations.

joao-r-reis:
Also can you create a JIRA for this?

tengu-alt:
> Also can you create a JIRA for this?

Of course, currently working on refactoring.

cassandra_test.go
tengu-alt: Agree, fixed

control.go
tengu-alt: Fixed

host_source.go
tengu-alt: agree, fixed

integration_test.go
tengu-alt: done

policies.go
tengu-alt: I've add the check to the `add()`

policies.go
tengu-alt: fixed

policies.go
tengu-alt: `SetHosts` and `AddHost` are checking it.

policies.go
tengu-alt: done

session.go
tengu-alt: Agree, fixed

token.go
tengu-alt: As I have checked, it will print a `nil`

joao-r-reis:
I just had a realization that we are probably going about this the wrong way. Instead of changing `ConnectAddress()` to return an `error` we should probably ensure that we never create `Host` objects with invalid addresses in the first place, I think that would result in a clearer API and less intrusive changes to users.

Sorry for not realizing this earlier, I know you already worked a bit on this approach but I really think we should change the approach here. 

First step is to find all occurences of straight `Host` struct initialization and replace them with a `newHostInfo(addr net.IP, port int) (*Host, error)` so that the ip validation occurs here in this method instead of `host.ConnectAddress()`. The only place where it will require some more changes is in `func (r *ringDescriber) getClusterPeerInfo(localHost *HostInfo) ([]*HostInfo, error)` because here the object is created without an IP address since it will be set after reading the columns from `system.peers`. A possible approach is to move the logic from `h.connectAddressLocked()` to a function outside of the `hostinfo` type, use it to compute the connect address and then use the result to create the host info object afterwards. After this change, I think `h.ConnectAddress()` can always read the value from the `h.connectAddress` field.  After this change, `h.SetConnectAddress` should be changed to return an error in case the provided address is invalid (this is only in case a user is using this function because the driver itself doesn't use it).

tengu-alt:
Understood, I will work on it.

tengu-alt:
> e only place where it will require some more changes is in func (r *ringDescriber) getClusterPeerInfo(localHost *HostInfo) ([]*HostInfo, error)

tengu-alt:
>  A possible approach is to move the logic from h.connectAddressLocked() to a function outside of the hostinfo type, use it to compute the connect address and then use the result to create the host info object afterwards. 

I don't clearly understand how we can use the logic from the `h.connectAddressLocked()` if it computes `connectAddress` according to the IP's from the `host` which needs to be filled. As I see those IP's are filled inside of the `hostInfoFromMap()` method which requires `*HostInfo`.

I think it is better to create raw structure here as it is now, and add the validation inside of the `hostInfoFromMap()`
something like this:
```
	ip, port := s.cfg.translateAddressPort(host.ConnectAddress(), host.port)
	if !validIpAddr(ip) {
		return nil, errors.New("invalid connect address")
	}
	host.connectAddress = ip
	host.port = port
```
 or validate it inside of `getClusterPeerInfo()`

joao-r-reis:
> I don't clearly understand how we can use the logic from the h.connectAddressLocked() if it computes connectAddress according to the IP's from the host which needs to be filled. As I see those IP's are filled inside of the hostInfoFromMap() method which requires *HostInfo.

You would have those IPs as parameters of the function instead for example.

> I think it is better to create raw structure here as it is now, and add the validation inside of the hostInfoFromMap()
something like this:

This can also work, as long as we guarantee that we don't have `HostInfo` objects going around with invalid IPs then I'm ok with it. I suggested the `NewHostInfo` function and replacing direct struct initialization with this function call because it would be a guaranteed way to ensure `HostInfo` objects always have valid IP addresses in the future

tengu-alt:
@joao-r-reis I've made an update. I decided to add validation into `hostInfoFromMap()` for cases when we can't provide `connectAddress` for the constructor func (I have found another place inside of `func (r *ringDescriber) getLocalHostInfo() (*HostInfo, error)`)

host_source.go
```diff
@@ -159,6 +159,18 @@ type HostInfo struct {
 	tokens           []string
 }
 
+func newHostInfo(addr net.IP, port int) (*HostInfo, error) {
+	host := &HostInfo{}
+	host.hostname = addr.String()
+	host.port = port
+	if !validIpAddr(addr) {
```
joao-r-reis, line 185: move this check to the top

host_source.go
```diff
@@ -584,6 +598,9 @@ func (s *Session) hostInfoFromMap(row map[string]interface{}, host *HostInfo) (*
 	}
 
 	ip, port := s.cfg.translateAddressPort(host.ConnectAddress(), host.port)
+	if !validIpAddr(ip) {
+		return nil, errors.New("invalid connect address")
```
joao-r-reis:
```
hostAddr := host.ConnectAddress()
return nil, fmt.Errorf("invalid connect address after translating %v:%v", hostAddr.String(), host.port)
```

host_source.go
```diff
@@ -224,18 +236,20 @@ func (h *HostInfo) ConnectAddress() net.IP {
 	h.mu.RLock()
 	defer h.mu.RUnlock()
 
-	if addr, _ := h.connectAddressLocked(); validIpAddr(addr) {
-		return addr
-	}
-	panic(fmt.Sprintf("no valid connect address for host: %v. Is your cluster configured correctly?", h))
+	addr, _ := h.connectAddressLocked()
```
joao-r-reis, line 258: There's no case where the `connectAddressLocked()` function should ever return a non nil err now so it can be changed to just return the address (basically remove the validation).

conn.go
```diff
@@ -1697,7 +1697,11 @@ func (c *Conn) awaitSchemaAgreement(ctx context.Context) (err error) {
 		}
 
 		for _, row := range rows {
-			host, err := c.session.hostInfoFromMap(row, &HostInfo{connectAddress: c.host.ConnectAddress(), port: c.session.cfg.Port})
+			h, err := newHostInfo(c.host.connectAddress, c.session.cfg.Port)
```
joao-r-reis: `c.host.ConnectAdress()`

control.go
```diff
@@ -172,7 +176,12 @@ func hostInfo(addr string, defaultPort int) ([]*HostInfo, error) {
 	}
 
 	for _, ip := range ips {
-		hosts = append(hosts, &HostInfo{hostname: host, connectAddress: ip, port: port})
+		h, err := newHostInfo(ip, port)
```
joao-r-reis, line 179: my suggestion to use `newHostInfo` was so that we could standardize every creation of a host info object to use this function so we ensure there is no host info object with an invalid connect address but as it stands in this PR we still have a few places where the host info struct is being created and initialized directly.

E.g. in `hostInfoFromIter` the host info can be created with a nil connect address but it is provided immediately to `hostInfoFromMap` which will fix it. This is fine with me but we should make it a bit more consistent.

As I see it, there's currently 3 different ways to create a host info object:

1) create an "empty" host info object with a nil connect address + default port and call `hostInfoFromMap` to "fill" it
2) create a host info object from a contact point
3) update existing host info object with new values from system.peers/system.local (also done with `hostInfoFromMap`) - this is usually done on a host info created by 2)

For 1) I think we can add a new method `session.newHostInfoFromMap(addr net.IP, port int, row map[string]interface{})` that just wraps around `hostInfoFromMap`:

```
func (s *Session) newHostInfoFromMap(addr net.IP, port int, row map[string]interface{}) (*HostInfo, error) {
	return s.hostInfoFromMap(row, &HostInfo{connectAddress: addr , port: port})
}
```

For 2. we can use the new `newHostInfo` function you created.

For 3) we can use `hostInfoFromMap` or we can use `session.newHostInfoFromMap` just like in 1)


In summary we will have `hostInfo` struct being created in 2 places only: `session.newHostInfoFromMap()` and `newHostInfo()` and in both of these 2 places we are ensuring that the object being returned has a valid address. Any host info struct initialization in the codebase should be replaced by a call to either `session.newHostInfoFromMap` or `newHostInfo()`

joao-r-reis (commented):
Added a comment about existing HostInfo struct initialization references in the codebase

control.go
tengu-alt: Souns good, as I understand we need to keep the places when we create `HostInfo` using the `newHostInfo()` as it is now(in the PR), and wrap the `hostInfoFromMap` where `HostInfo` being provided as raw struct. Is it correct?
And, of course apply the suggestions above.

host_source.go
tengu-alt: `connectAddressLocked` returns string source of an address and "invalid" if it's ivalid. 
```
if validIpAddr(h.connectAddress) {
		return h.connectAddress, "connect_address"
	} else if validIpAddr(h.rpcAddress) {
		return h.rpcAddress, "rpc_adress"
```

control.go
joao-r-reis: Yeah

host_source.go
joao-r-reis: Yes but h.connectaddress is never supposed to be invalid now

tengu-alt: Got it.

host_source.go
```diff
@@ -224,18 +234,20 @@ func (h *HostInfo) ConnectAddress() net.IP {
...
+	return addr
 }
 
-func (h *HostInfo) SetConnectAddress(address net.IP) *HostInfo {
-	// TODO(zariel): should this not be exported?
+func (h *HostInfo) SetConnectAddress(address net.IP) (*HostInfo, error) {
```
joao-r-reis: I honestly think we should just remove this method, what do you think @jameshartig ? Allowing users to just modify the connect address on a host is weird, there's already an `AddressTranslator` interface for this.

host_source.go
```diff
@@ -584,6 +600,9 @@ func (s *Session) hostInfoFromMap(row map[string]interface{}, host *HostInfo) (*
 	}
 
 	ip, port := s.cfg.translateAddressPort(host.ConnectAddress(), host.port)
+	if !validIpAddr(ip) {
+		return nil, fmt.Errorf("invalid connect address after translating %v:%v", ip.String(), host.port)
```
joao-r-reis: Just realized this will be called even if there is no address translator so maybe changing this error message to be less confusing:

"invalid host address (before translation: %v:%v, after translation: %v:%v)", host.ConnectAddress(), host.port, ip.String(), port)

host_source.go
```diff
@@ -186,19 +198,17 @@ func validIpAddr(addr net.IP) bool {
...
 		return h.connectAddress, "connect_address"
-	} else if validIpAddr(h.rpcAddress) {
+	} else if h.rpcAddress != nil {
 		return h.rpcAddress, "rpc_adress"
-	} else if validIpAddr(h.preferredIP) {
-		// where does perferred_ip get set?
+	} else if h.preferredIP != nil {
 		return h.preferredIP, "preferred_ip"
-	} else if validIpAddr(h.broadcastAddress) {
+	} else if h.broadcastAddress != nil {
```
joao-r-reis: We should keep the `validIpAddr` check on all of these (except the last) so that the "fallback" priority still works otherwise this would be a behavior change and could cause bugs.

Just skip the check on the last one (h.peer), this will be checked either way in `hostInfoFromMap` before returning the host info object so we're good.

joao-r-reis (commented):
Couple of comments

host_source.go
tengu-alt: Refactored

host_source.go
tengu-alt: Fixed

host_source.go
jameshartig: Typically I'd like to go through a version or so of it being deprecated but since we're about to cut a major release I think it makes sense to remove it and we can always add it back if there's a reason to.

jameshartig (changes requested):
Overall good, but I agree, let's remove SetConnectAddress

tengu-alt:
Done.

jameshartig (approved):

jameshartig:
@joao-r-reis want to give your approval?

joao-r-reis (approved):

tengu-alt:
@joao-r-reis I have updated the changelog and the commit message.

PR #1842, opened 2024-11-18
CASSGO-32 move lz4 compressor to lz4 package within the gocql module

OleksiienkoMykyta:
This PR should be reviewed right after https://github.com/apache/cassandra-gocql-driver/pull/1822 is merged.

PR #1861, opened 2025-01-23
CASSGO-48 Bump actions/upload-artifact and actions/cache versions to v4

worryg0d:
Bumped due to cancelation of running CI workflows with the deprecated version of actions/upload-artifact@v3 and in order to prevent future interruptions with deprecation of actions/cache@v2.

Patch by Bohdan Siryk; Reviewed by João Reis, Stanislav Bychkov for CASSGO-48

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,7 +44,7 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: ${{ matrix.go }}
-      - uses: actions/cache@v2
+      - uses: actions/cache@v4
         id: gomod-cache
         with:
           path: ~/go/pkg/mod
@@ -113,9 +113,9 @@ jobs:
           go test -v -tags "${{ matrix.tags }} gocql_debug" -timeout=5m -race ${{ env.args }}
       - name: 'Save ccm logs'
         if: 'failure()'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
-          name: ccm-cluster
+          name: ccm-cluster-cassandra-${{ matrix.cassandra_version }}-go-${{ matrix.go }}-tag-${{ matrix.tags }}
           path: /home/runner/.ccm/test
           retention-days: 5
   integration-auth-cassandra:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Remove deprecated global logger (CASSGO-24)
 
+- Bumped actions/upload-artifact and actions/cache versions to v4 in CI workflow (CASSGO-48)
+
 ### Fixed
 
 - Retry policy now takes into account query idempotency (CASSGO-27)


worryg0d:
I bumped the version of `actions/upload-artifact@v3` to v4 to unblock running workflows on other PRs.

@joao-r-reis Could you please look at this one? Also, I noticed warnings about deprecation of `actions/cache@v2`. Should I bump its version to v4 in this PR as well?

.github/workflows/main.yml
```diff
@@ -113,7 +113,7 @@ jobs:
           go test -v -tags "${{ matrix.tags }} gocql_debug" -timeout=5m -race ${{ env.args }}
       - name: 'Save ccm logs'
         if: 'failure()'
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: ccm-cluster
           path: /home/runner/.ccm/test
```
ribaraka: There are breaking changes in version 4. 
The [actions/upload-artifact](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes) documentation states that artifacts are immutable, and you can no longer upload to the same artifact name multiple times. It should be either upload to multiple artifacts with different names or upload everything at once. Otherwise, you will encounter an error.

Since this job uses a matrix, and judging by the documentation and current implementation, it could lead to errors if multiple matrix jobs fail.

worryg0d: Oh, I see the issue here. I tested it on my fork forcing the workflow to fail and it really doesn't upload the artifacts...
```
Run actions/upload-artifact@v4
With the provided path, there will be 922 files uploaded
Artifact name is valid!
Root directory input is valid!
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

I'm going to work on a fix for this. Thanks a lot!

ribaraka:
> I noticed warnings about deprecation of `actions/cache@v2`

It should be safe to bump from actions/cache@v2 to actions/cache@v4, and doing so is recommended to stay current with the latest improvements and avoid issues like the one experienced by @worryg0d. I ran the job using the current code (which still uses deprecated action versions) and did not encounter any CI cancellations. However, there are ongoing discussions about sunsetting older versions, as highlighted in [this GitHub discussion](https://github.com/actions/cache/discussions/1510) and [this deprecation notice](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/).

.github/workflows/main.yml
worryg0d: Fixed by constructing the name of the artifact using variables from the matrix:
```
ccm-cluster-cassandra-${{ matrix.cassandra_version }}-go-${{ matrix.go }}-tag-${{ matrix.tags }}
``` 
I was about to add `matrix.compressor` as well, but we have only `snappy` in the matrix and #1567 is not merged yet.
If anybody has concerns about this, please feel free to discuss it here.

worryg0d:
> I ran the job using the current code (which still uses deprecated action versions) and did not encounter any CI cancellations.

By **_the current code_** do you mean the code from this PR? If so, I didn't encounter any issues with `actions/cache@v2` too. The only thing I notice that in my workflow runs details a warning annotation appears:
```
Your workflow is using a version of actions/cache that is scheduled for deprecation, actions/cache@v2. 
Please update your workflow to use either v3 or v4 of actions/cache to avoid interruptions. 
Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-downShow
```
However, as you said, it will be sunsetted soon

ribaraka:
> By **_the current code_** do you mean the code from this PR?

I was reffering to the latest code from the trunk branch (with the recent commit ba7d5634e98df911989d0f219cd3d1cdff279a83) instead of this PR.

.github/workflows/main.yml
joao-r-reis: I'm fine with not adding `matrix.compressor` here but we have to remember to do it in #1567

joao-r-reis:
If bumping `actions/cache` is simple then I'd say we should do it in this PR as well to avoid interruptions when it does get deprecated

.github/workflows/main.yml
worryg0d: Oops, I'm sorry, I misread the issue. I meant #1842 which is CASSGO-32

worryg0d:
In the [discussion](https://github.com/actions/cache/discussions/1510) highlited:
```
The new changes to the actions/cache should be seamless and fully backward compatible. 
You are not expected to change anything in your workflows to use the new service beyond upgrading to the latest releases.
``` 

So it looks like there are no problems with bumping it too

joao-r-reis:
> In the [discussion](https://github.com/actions/cache/discussions/1510) highlited:
> 
> ```
> The new changes to the actions/cache should be seamless and fully backward compatible. 
> You are not expected to change anything in your workflows to use the new service beyond upgrading to the latest releases.
> ```
> 
> So it looks like there are no problems with bumping it too

Nice, feel free to go for it and update the JIRA, PR title and commit message and then I'll merge this

.github/workflows/main.yml
joao-r-reis: Yeah I was thinking of #1842 too my bad

worryg0d:
Bumped actions/cache to v4 as well. @joao-r-reis could you please approve the workflow run? Thanks!

Issue #1601, opened 2022-02-28
session.Closed returns false when cassandra went down or stopped after client made a successful session.


amanjain08:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?
3.11.3

### What version of Gocql are you using?
v0.0.0-20200221113847-372a19b1a852

### What version of Go are you using?
1.14

### What did you do?
I tried getting the metadata of keyspace to check the table existence while gocql client is connected to cassandra. After I stopped the cassandra, gocql wasn't throwing any error and returning cached metadata to check the table existence.
Also, when I checked session.Closed it returns false when cassandra wasn't running.

### What did you expect to see?
Error should have been thrown.
 session.Closed should have returned true, when cassandra went down or stopped after client made a successful session.

### What did you see instead?
 session.Closed it returns false when cassandra wasn't running.
---

If you are having connectivity related issues please share the following additional information

### Describe your Cassandra cluster
please provide the following information

- output of `nodetool status`
- output of `SELECT peer, rpc_address FROM system.peers`
- rebuild your application with the `gocql_debug` tag and post the output

amanjain08:
we are using gocql in a central health manager where other apps will call central health manager to check for the cassandra connectivity and the existence of tables those are required by the apps. So, if we get the cache data then there is a chance that app might start and cassandra is not running. Kind of hard dependency on cassandra. 

So, now we are executing a query with limit 1 that is returning error whenever cassandra is not connected and fulfilling the purpose. 

Reason behind using older version of gocql is that we are using old version of go 1.14

tengu-alt:
> We should update the documentation to describe what Close() does.

I'll Handle this. Also, I am currently working on doc update which covers undocumented user API.

OleksiienkoMykyta:
@joao-r-reis, could you please check the PR(https://github.com/apache/cassandra-gocql-driver/issues/1601), looks like it fixes it.

PR #1863, opened 2025-02-12
CASSGO-55 - User API doc enhancement

tengu-alt:
Some of the user-exposed API is not documented which can cause misuse. This PR provides doc update fo the public methods and interfaces. Also, some deprecated features were removed.
Fix for the #1601

### DIFF
--- a/filters.go
+++ b/filters.go
@@ -47,6 +47,7 @@ func AcceptAllFilter() HostFilter {
 	})
 }
 
+// DenyAllFilter will deny all hosts
 func DenyAllFilter() HostFilter {
 	return HostFilterFunc(func(host *HostInfo) bool {
 		return false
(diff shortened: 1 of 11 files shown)


filters.go
```diff
@@ -61,12 +62,6 @@ func DataCenterHostFilter(dataCenter string) HostFilter {
 	})
 }
 
-// Deprecated: Use DataCenterHostFilter instead.
-// DataCentreHostFilter is an alias that doesn't use the preferred spelling.
-func DataCentreHostFilter(dataCenter string) HostFilter {
```
joao-r-reis: This was deprecated very recently for the 2.0 release, we don't want to remove it now.

joao-r-reis (commented):
Added 1 comment. https://github.com/apache/cassandra-gocql-driver/pull/1868 will change the query and batch API so I'd like to get that merged before merging this one.

filters.go
tengu-alt: Got it, I have returned it.

tengu-alt:
Sorry for that, I've accidentally pushed without changes.

PR #1864, opened 2025-02-21
CASSGO-59 CASSGO-56 Test fix for hostpool package and newHostInfo() expose

tengu-alt:
hostpool package test was using recently removed `SetConnectAddress()` 
The test was refactored to create hostInfo by constructor func, and the `newHostInfo` constructor was exposed. 
Also fix for the CASSGO-56

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,9 +22,9 @@ jobs:
       - uses: actions/setup-go@v4
         with:
           go-version: ${{ matrix.go }}
-      - run: go vet
+      - run: go vet ./...
       - name: Run unit tests
-        run: go test -v -tags unit -race
+        run: go test -v -tags unit -race ./...
   integration-cassandra:
     timeout-minutes: 15
     needs:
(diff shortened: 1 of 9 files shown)


joao-r-reis (approved):

joao-r-reis (approved):

joao-r-reis:
We should fix the CI issue here on this PR too so we know we're fixing the issue and there won't be regressions

tengu-alt:
> We should fix the CI issue here on this PR

Understood, I'll do it.

joao-r-reis:
@tengu-alt I don't see `TestHostPolicy_HostPool` in the CI log still, I think the fix is to add `./..` to `go test`, we should do it for all steps in the workflow I think

tengu-alt:
> I think the fix is to add ./.. to go test, we should do it for all steps in the workflow I think

Agree, I've added the `./...` to jobs `go test`, also I've added the `unit` build tags for other packages tests(e.g. `murmur` and `lru`) to prevent double run during integration tests.

tengu-alt:
I had to remove the `./...` from integration test runs because we already have those within `env.args`

tengu-alt:
@joao-r-reis  It seems that `TestHostPolicy_HostPool` [is passed test](https://github.com/apache/cassandra-gocql-driver/actions/runs/13971227224/job/39392956606#step:5:456). Please take a look.

PR #1866, opened 2025-02-27
CASSGO-61 fix connection timeout override with custom

### LINKED ISSUE #953: Session Timeout and Context Timeout
Looking at the "exec()" method of "Conn" in "conn.go",  it seems like there's two ways we can hit a timeout condition when a Context with a deadline is provided to the Query. 

As I understand it, the "Session" ClusterConfig.Timeout value is propagated to Conn which is used in the "timeoutCh" and and the Query's context is used for "ctxDone". Coming from the Datastax drivers, this is unexpected behavior.

For example, from the Python driver docs on `session.default_timeout`

> default_timeout = 10.0
A default timeout, measured in seconds, for queries executed through execute() or execute_async(). This default may be overridden with the timeout parameter for either of those methods.

Source: https://datastax.github.io/python-driver/api/cassandra/cluster.html#cassandra.cluster.Session

We don't have to copy the same behavior, but expectations are a bit unclear. I think one of the following can be done:

* better documentation/best practices around `ClusterConfig.Timeout` and/or `Query().WithContext` and timeouts. E.g. Session timeout should be 0 or very large when using  Contexts to manage timeout limits.
* When Context is provided, ignore the Conn.timeout and/or call a separate 'Conn' method that only respects the context timeout, errors, and quits of the connection.

tengu-alt:
Resolves #953 
This PR provides the fix for the connection timeout override via query context timeout, allowing users to set the custom timeout for the query.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -44,7 +44,7 @@ import (
 	"time"
 	"unicode"
 
-	inf "gopkg.in/inf.v0"
+	"gopkg.in/inf.v0"
 
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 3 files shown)


cassandra_test.go
```diff
@@ -3955,3 +3955,46 @@ func TestRoutingKeyCacheUsesOverriddenKeyspace(t *testing.T) {
...
+	for _, conPool := range session.executor.pool.hostConnPools {
+		conPool.mu.Lock()
+		for _, conn := range conPool.conns {
+			conn.r.SetTimeout(50)
+		}
+		conPool.mu.Unlock()
+	}
+	session.executor.pool.mu.Unlock()
+	err = session.Query("INSERT INTO gocql_test.named_query(id, value) VALUES(2, 'value')").Exec()
+	if err == nil || err != ErrTimeoutNoResponse {
```
OleksiienkoMykyta: If you are expecting exactly ErrTimeoutNoResponse, this part is redundant "err == nil" wdyt?

OleksiienkoMykyta (commented):
Also, it would be nice to provide details to documentation with code examples, how to set a custom timeout, and how it works.

In general, LGTM)

cassandra_test.go
tengu-alt: Agree, fixed.

tengu-alt:
> Also, it would be nice to provide details to documentation

I've added the code example to the `query.WithContext()` method.

OleksiienkoMykyta (approved):

jameshartig:
When the context is cancelled because it had a deadline, we aren't calling `c.handleTimeout`. We probably need to check the context error and if it had a deadline when `<-ctxDone` and then call `c.handleTimeout`.

tengu-alt:
> We probably need to check the context error and if it had a deadline when <-ctxDone and then call c.handleTimeout.

Agree. I've also added this check [here](https://github.com/apache/cassandra-gocql-driver/blob/trunk/conn.go#L1181) in case if deadline hits before we create `ctxDone` chan.

PR #1867, opened 2025-03-06
CASSGO-62 Check the at least one host passed on initial host lookup 

### LINKED ISSUE #738: ensure at least one initial host passes the host filter if present

tengu-alt:
This PR provides a check to ensure that at least one host passes filter during the initial host lookup.
Resolves: #738

### DIFF
--- a/session.go
+++ b/session.go
@@ -250,6 +250,9 @@ func (s *Session) init() error {
 					filteredHosts = append(filteredHosts, host)
 				}
 			}
+			if len(filteredHosts) < 1 {
+				return errors.New("expected at least one host will pass the filter")
+			}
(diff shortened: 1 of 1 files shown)


jameshartig:
What was the behavior previously? I believe it would've errored with `ErrNoConnectionsStarted`? The #738 issue doesn't have any reasoning or explanation for why it was filed.

tengu-alt:
> I believe it would've errored with ErrNoConnectionsStarted? 

It is correct. The reason for this check is to prevent further code execution before `ErrNoConnectionsStarted` (e.g when the localHost passed the filter for the first time and fails for the second)

PR #1868, opened 2025-03-13
CASSGO-22 CASSGO-73 Changes to Query and Batch to make them safely reusable

joao-r-reis:
# API Changes

### `ExecutableQuery`

`ExecutableQuery` is currently an interface that `Query` and `Batch` implements (and is referenced by `HostSelectionPolicy`). However, it is also used in driver internals so the interface contains private methods which makes it impossible for users to "mock" the interface for testing purposes.

In this PR, `ExecutableQuery` is changed so it contains only public methods and is no longer implemented by `Query` and `Batch`. Now, `ExecutableQuery` is used exclusively as a "hook" in `HostSelectionPolicy`. This does mean that users can no longer attempt to cast an `ExecutableQuery` to `Query` or `Batch` but I've added a new method that provides this functionality (`Statement`).

```
type ExecutableQuery interface {
	GetRoutingKey() ([]byte, error)
	Keyspace() string
	Table() string
	IsIdempotent() bool
	Statement() Statement
}
```

### `Statement` 

This is the new interface that represents (and is implemented by) `Query` and `Batch`. In this PR it's only referenced by `ExecutableQuery`.

```
type Statement interface {
	Iter() *Iter
	Exec() error
}
```

### `internalRequest`

New interface that is used by driver internals to decouple the public API of Query/Batch from the internal API.
When creating an internal request object the driver now copies most if not all of the query/batch properties so that users can submit a Query/Batch for execution and then re-use immediately without having to be concerned about the object being modified by the driver execution. It also makes it less error prone because the driver is free to modify these properties (e.g. page state, consistency, query metrics) without causing a change on the objects that the users are using.

```
type internalRequest interface {
	execute(ctx context.Context, conn *Conn) *Iter
	attempt(keyspace string, end, start time.Time, iter *Iter, host *HostInfo)
	retryPolicy() RetryPolicy
	speculativeExecutionPolicy() SpeculativeExecutionPolicy
	getQueryMetrics() *queryMetrics
	RetryableQuery
	ExecutableQuery
}
```

### Query Metrics (i.e. `Query/Batch.Attempts()`, `Query/Batch.Latency()`)

This functionality makes the API a bit awkward because the user submits a query for execution and then inspects the query object to retrieve the side effects of that execution after it's done. It's a better design (imo) to have this kind of data available in the `Iter` object since this is the type that represents the return value of a query/batch.

Due to this, `AddAttempts()`, `Attempts()`, `Latency()` and `AddLatency()` have been removed from `Query` and `Batch`. `Latency()` and `Attempts()` have been added to `Iter` and a new `Batch.Iter()` method has been added so that users can obtain the `Iter` object when executing a batch (even though it doesn't make a lot of sense conceptually due to the name but `Iter` is what the driver uses to return data about a request so it's not just about "iterating").

### ```queryPool```

`Query` objects were allocated from a `sync.Pool` but I don't see the gain of keeping it especially now that `queryMetrics` have been moved to `Iter`. Also users can create these query objects once and store them if they are worried about memory/GC.

### ``` Query/Batch.GetRoutingKey() ```

Having this method as part of the public API of `Query` and `Batch` makes it difficult to manage from the driver maintainer POV, it's supposed to be an internal method that can be called by implementations of `HostSelectionPolicy` but since `ExecutableQuery` was implemented by `Query` and `Batch` this meant that these two types had to have this method on their public API as well.

With this PR, this method is deprecated on `Query` and `Batch` since these two types no longer implement `ExecutableQuery` so we can keep the `Query`/`Batch` API much simpler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for Native Protocol 5. Following protocol changes exposed new API
   Query.SetKeyspace(), Query.WithNowInSeconds(), Batch.SetKeyspace(), Batch.WithNowInSeconds() (CASSGO-1)
 - Externally-defined type registration (CASSGO-43)
+- Add Query and Batch to ObservedQuery and ObservedBatch (CASSGO-73)
 
 ### Changed
 
@@ -43,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - inet columns default to net.IP when using MapScan or SliceMap (CASSGO-43)
 - NativeType removed (CASSGO-43)
 - `New` and `NewWithError` removed and replaced with `Zero` (CASSGO-43)
+- Changes to Query and Batch to make them safely reusable (CASSGO-22)
 
 ### Fixed
 
--- a/conn.go
+++ b/conn.go
@@ -1497,32 +1497,34 @@ func marshalQueryValue(typ TypeInfo, value interface{}, dst *queryValues) error 
 	return nil
 }
 
-func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
+func (c *Conn) executeQuery(ctx context.Context, q *internalQuery) *Iter {
+	qryOpts := q.qryOpts
 	params := queryParams{
-		consistency: qry.cons,
+		consistency: q.GetConsistency(),
 	}
+	iter := newIter(q.metrics, q.Keyspace(), q.routingInfo, q.qryOpts.getKeyspace)
 
 	// frame checks that it is not 0
-	params.serialConsistency = qry.serialCons
-	params.defaultTimestamp = qry.defaultTimestamp
-	params.defaultTimestampValue = qry.defaultTimestampValue
+	params.serialConsistency = qryOpts.serialCons
+	params.defaultTimestamp = qryOpts.defaultTimestamp
+	params.defaultTimestampValue = qryOpts.defaultTimestampValue
 
-	if len(qry.pageState) > 0 {
-		params.pagingState = qry.pageState
+	if len(q.pageState) > 0 {
+		params.pagingState = q.pageState
 	}
-	if qry.pageSize > 0 {
-		params.pageSize = qry.pageSize
+	if qryOpts.pageSize > 0 {
+		params.pageSize = qryOpts.pageSize
 	}
 	if c.version > protoVersion4 {
-		params.keyspace = qry.keyspace
-		params.nowInSeconds = qry.nowInSecondsValue
+		params.keyspace = qryOpts.keyspace
+		params.nowInSeconds = qryOpts.nowInSecondsValue
 	}
 
 	// If a keyspace for the qry is overriden,
 	// then we should use it to create stmt cache key
 	usedKeyspace := c.currentKeyspace
-	if qry.keyspace != "" {
-		usedKeyspace = qry.keyspace
+	if qryOpts.keyspace != "" {
+		usedKeyspace = qryOpts.keyspace
 	}
 
 	var (
@@ -1530,17 +1532,18 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
 		info  *preparedStatment
 	)
 
-	if !qry.skipPrepare && qry.shouldPrepare() {
+	if !qryOpts.skipPrepare && shouldPrepare(qryOpts.stmt) {
 		// Prepare all DML queries. Other queries can not be prepared.
 		var err error
-		info, err = c.prepareStatement(ctx, qry.stmt, qry.trace, usedKeyspace)
+		info, err = c.prepareStatement(ctx, qryOpts.stmt, qryOpts.trace, usedKeyspace)
 		if err != nil {
-			return &Iter{err: err}
+			iter.err = err
+			return iter
 		}
 
-		values := qry.values
-		if qry.binding != nil {
-			values, err = qry.binding(&QueryInfo{
+		values := qryOpts.values
+		if qryOpts.binding != nil {
+			values, err = qryOpts.binding(&QueryInfo{
 				Id:          info.id,
 				Args:        info.request.columns,
 				Rval:        info.response.columns,
@@ -1548,12 +1551,14 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
 			})
 
 			if err != nil {
-				return &Iter{err: err}
+				iter.err = err
+				return iter
 			}
 		}
 
 		if len(values) != info.request.actualColCount {
-			return &Iter{err: fmt.Errorf("gocql: expected %d values send got %d", info.request.actualColCount, len(values))}
+			iter.err = fmt.Errorf("gocql: expected %d values send got %d", info.request.actualColCount, len(values))
+			return iter
 		}
 
 		params.values = make([]queryValues, len(values))
@@ -1562,59 +1567,63 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
 			value := values[i]
 			typ := info.request.columns[i].TypeInfo
 			if err := marshalQueryValue(typ, value, v); err != nil {
-				return &Iter{err: err}
+				iter.err = err
+				return iter
 			}
 		}
 
 		// if the metadata was not present in the response then we should not skip it
-		params.skipMeta = !(c.session.cfg.DisableSkipMetadata || qry.disableSkipMetadata) && info != nil && info.response.flags&flagNoMetaData == 0
+		params.skipMeta = !(c.session.cfg.DisableSkipMetadata || qryOpts.disableSkipMetadata) && info != nil && info.response.flags&flagNoMetaData == 0
 
 		frame = &writeExecuteFrame{
 			preparedID:       info.id,
 			params:           params,
-			customPayload:    qry.customPayload,
+			customPayload:    qryOpts.customPayload,
 			resultMetadataID: info.resultMetadataID,
 		}
 
 		// Set "keyspace" and "table" property in the query if it is present in preparedMetadata
-		qry.routingInfo.mu.Lock()
-		qry.routingInfo.keyspace = info.request.keyspace
+		q.routingInfo.mu.Lock()
+		q.routingInfo.keyspace = info.request.keyspace
 		if info.request.keyspace == "" {
-			qry.routingInfo.keyspace = usedKeyspace
+			q.routingInfo.keyspace = usedKeyspace
 		}
-		qry.routingInfo.table = info.request.table
-		qry.routingInfo.mu.Unlock()
+		q.routingInfo.table = info.request.table
+		q.routingInfo.mu.Unlock()
 	} else {
 		frame = &writeQueryFrame{
-			statement:     qry.stmt,
+			statement:     qryOpts.stmt,
 			params:        params,
-			customPayload: qry.customPayload,
+			customPayload: qryOpts.customPayload,
 		}
 	}
 
-	framer, err := c.exec(ctx, frame, qry.trace)
+	framer, err := c.exec(ctx, frame, qryOpts.trace)
 	if err != nil {
-		return &Iter{err: err}
+		iter.err = err
+		return iter
 	}
 
 	resp, err := framer.parseFrame()
 	if err != nil {
-		return &Iter{err: err}
+		iter.err = err
+		return iter
 	}
 
-	if len(framer.traceID) > 0 && qry.trace != nil {
-		qry.trace.Trace(framer.traceID)
+	if len(framer.traceID) > 0 && qryOpts.trace != nil {
+		qryOpts.trace.Trace(framer.traceID)
 	}
 
 	switch x := resp.(type) {
 	case *resultVoidFrame:
-		return &Iter{framer: framer}
+		iter.framer = framer
+		return iter
 	case *resultRowsFrame:
 		if x.meta.newMetadataID != nil {
 			// If a RESULT/Rows message reports
 			//      changed resultset metadata with the Metadata_changed flag, the reported new
 			//      resultset metadata must be used in subsequent executions
-			stmtCacheKey := c.session.stmtsLRU.keyFor(c.host.HostID(), usedKeyspace, qry.stmt)
+			stmtCacheKey := c.session.stmtsLRU.keyFor(c.host.HostID(), usedKeyspace, qryOpts.stmt)
 			oldInflight, ok := c.session.stmtsLRU.get(stmtCacheKey)
 			if ok {
 				newInflight := &inflightPrepare{
@@ -1635,33 +1644,32 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
 				info = newInflight.preparedStatment
 			}
 		}
-
-		iter := &Iter{
-			meta:    x.meta,
-			framer:  framer,
-			numRows: x.numRows,
-		}
+		iter.meta = x.meta
+		iter.framer = framer
+		iter.numRows = x.numRows
 
 		if x.meta.noMetaData() {
 			if info != nil {
 				iter.meta = info.response
 				iter.meta.pagingState = copyBytes(x.meta.pagingState)
 			} else {
-				return &Iter{framer: framer, err: errors.New("gocql: did not receive metadata but prepared info is nil")}
+				iter = newErrIter(errors.New("gocql: did not receive metadata but prepared info is nil"), q.metrics, q.Keyspace(), q.routingInfo, q.qryOpts.getKeyspace)
+				iter.framer = framer
+				return iter
 			}
 		} else {
 			iter.meta = x.meta
 		}
 
-		if x.meta.morePages() && !qry.disableAutoPage {
-			newQry := new(Query)
-			*newQry = *qry
+		if x.meta.morePages() && !qryOpts.disableAutoPage {
+			newQry := new(internalQuery)
+			*newQry = *q
 			newQry.pageState = copyBytes(x.meta.pagingState)
 			newQry.metrics = &queryMetrics{m: make(map[string]*hostMetrics)}
 
 			iter.next = &nextIter{
-				qry: newQry,
-				pos: int((1 - qry.prefetch) * float64(x.numRows)),
+				q:   newQry,
+				pos: int((1 - qryOpts.prefetch) * float64(x.numRows)),
 			}
 
 			if iter.next.pos < 1 {
@@ -1671,9 +1679,10 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
 
 		return iter
 	case *resultKeyspaceFrame:
-		return &Iter{framer: framer}
+		iter.framer = framer
+		return iter
 	case *schemaChangeKeyspace, *schemaChangeTable, *schemaChangeFunction, *schemaChangeAggregate, *schemaChangeType:
-		iter := &Iter{framer: framer}
+		iter.framer = framer
 		if err := c.awaitSchemaAgreement(ctx); err != nil {
 			// TODO: should have this behind a flag
 			c.logger.Println(err)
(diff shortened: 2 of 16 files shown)


query_executor.go
```diff
@@ -201,16 +210,385 @@ func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery, hostIter Ne
...
+
+func newBatchOptions(b *Batch) *batchOptions {
+	// make we get a new array so if user keeps appending entries on the Batch object it doesn't affect this execution
+	newEntries := make([]BatchEntry, len(b.Entries))
+	for i, e := range b.Entries {
+		newEntries[i] = e
+	}
+	return &batchOptions{
+		Type:                  b.Type,
+		Entries:               b.Entries,
```
worryg0d: Here you should use created `newEntries` instead of `b.Entries`

query_executor.go
```diff
@@ -201,16 +210,385 @@ func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery, hostIter Ne
...
+
+func (b *internalBatch) borrowForExecution() {
+	// empty, because Batch has no equivalent of Query.Release()
+	// that would race with speculative executions.
+}
+
+func (b *internalBatch) releaseAfterExecution() {
+	// empty, because Batch has no equivalent of Query.Release()
+	// that would race with speculative executions.
+}
```
worryg0d: Since both these methods are removed from the interface you can get rid of them here too

conn.go
```diff
@@ -1523,32 +1534,35 @@ func (c *Conn) UseKeyspace(keyspace string) error {
...
 
 	for i := 0; i < n; i++ {
-		entry := &batch.Entries[i]
-		b := &req.statements[i]
+		entry := &b.batchOpts.Entries[i]
+		batchStmt := &req.statements[i]
 
 		if len(entry.Args) > 0 || entry.binding != nil {
-			info, err := c.prepareStatement(batch.Context(), entry.Stmt, batch.trace)
+			info, err := c.prepareStatement(b.batchOpts.context, entry.Stmt, b.batchOpts.trace)
```
worryg0d: If I'm not missing anything here both `ctx` arg of the method and `b.batchOpts.context` should be have the same context value.
The `ctx` arg is being passed by `queryExecutor.executeQuery()` at line 102  by calling `qry.Context()` which should return the underlying `internalBatch.batchOpts.context`.

worryg0d: The only usage of `ctx` is at line 1634 and that is a recursive call to `executeBatch` on an unprepared Cassandra response so changing `b.batchOpts.context` to `ctx` will make the code less confusing and clearer

worryg0d (commented):
Hello @joao-r-reis,

I spent some time reviewing your PR. Great work! It simplifies public query and batch API which is pretty good.

Removing `queryPool` is a good idea. If anyone really needs this, it can be handled on their application side.

I left some minor comments but overall implementation is well.

I am only a bit concerned about the changes this PR provides to the driver internals, especially the conn part which overlaps with #1822.

conn.go
joao-r-reis: Agree, done 👍

query_executor.go
joao-r-reis: Nice catch, done

query_executor.go
joao-r-reis: Nice catch, done 👍

joao-r-reis:
> I am only a bit concerned about the changes this PR provides to the driver internals, especially the conn part which overlaps with https://github.com/apache/cassandra-gocql-driver/pull/1822.

Yeah I'll have to spend some time rebasing this branch and even writing some tests, for now it's good enough if reviewers can provide feedback on the current state of the PR especially the public API changes

query_executor.go
```diff
@@ -27,32 +27,43 @@ package gocql
 import (
 	"context"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
+// ExecutableQuery is an interface that represents a query or batch statement that
+// exposes the correct functions for the HostSelectionPolicy to operate correctly.
 type ExecutableQuery interface {
```
lukasz-antoniak: What do you think about renaming type to `HostPolicyQuery`?

query_executor.go
```diff
@@ -201,16 +210,375 @@ func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery, hostIter Ne
...
-func (q *queryExecutor) run(ctx context.Context, qry ExecutableQuery, hostIter NextHost, results chan<- *Iter) {
+func (q *queryExecutor) run(ctx context.Context, qry internalRequest, hostIter NextHost, results chan<- *Iter) {
 	select {
 	case results <- q.do(ctx, qry, hostIter):
 	case <-ctx.Done():
 	}
-	qry.releaseAfterExecution()
+}
+
+type queryOptions struct {
```
lukasz-antoniak, line 258: Shall we group here all relevant options into sub-types, for example:
```
Paging
    pageSize
    initialPageState
    disableAutoPage
Monitoring
    observer
    trace
Parameters
    values
    binding
WriteTimestamp
    defaultTimestamp
    defaultTimestampValue
Consistency
    initialConsistency
    serialConsistency
FrameOptions
    customPayload
    disableSkipMetadata

stmt
prefetch
rt
spec
context
idempotent
keyspace
skipPrepare
routingKey
```
Grouping is not definitive, it is just an idea.

query_executor.go
```diff
@@ -201,16 +210,375 @@ func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery, hostIter Ne
...
+		getKeyspace:           q.getKeyspace,
+	}
+}
+
+type internalQuery struct {
+	originalQuery *Query
+	qryOpts       *queryOptions
+	pageState     []byte
+	metrics       *queryMetrics
+	refCount      uint32
```
lukasz-antoniak: Unused?

session.go
```diff
@@ -727,33 +720,36 @@ func (s *Session) routingKeyInfo(ctx context.Context, stmt string) (*routingKeyI
...
 // otherwise an error is returned describing the failure.
 func (b *Batch) Exec() error {
 	iter := b.session.executeBatch(b)
 	return iter.Close()
 }
 
+// Iter executes a batch operation and returns an Iter object
+// that can be used to access properties related to the execution like Iter.Attempts and Iter.Latency
+func (b *Batch) Iter() *Iter {
+	return b.session.executeBatch(b)
```
lukasz-antoniak: I feel that people may execute sequence of `Exec()` and `Iter()`. Shall we make an assertion that only one of the is invoked?

jameshartig:
> executing a batch (even though it doesn't make a lot of sense conceptually due to the name but Iter is what the driver uses to return data about a request so it's not just about "iterating").

Note that Yugabyte lets you iterate over the [row status](https://docs.yugabyte.com/preview/api/ycql/batch/#row-status) from a batch operation.

jameshartig:
> With this PR, this method is deprecated on Query and Batch since these two types no longer implement ExecutableQuery so we can keep the Query/Batch API much simpler.

Why not just remove it? Seems like any `HostSelectionPolicy` implementation would need to make major changes already.

jameshartig:
I don't want to bikeshed about names but `ExecutableQuery` is a bit confusing since it contains `Query` and `Batch`. Did that name come from somewhere? Would `ExecutableStatement` make more sense since it has a statement?

conn.go
```diff
@@ -1371,76 +1376,80 @@ func (c *Conn) executeQuery(ctx context.Context, qry *Query) *Iter {
...
-			customPayload: qry.customPayload,
+			customPayload: qryOpts.customPayload,
 		}
 
 		// Set "keyspace" and "table" property in the query if it is present in preparedMetadata
-		qry.routingInfo.mu.Lock()
-		qry.routingInfo.keyspace = info.request.keyspace
-		qry.routingInfo.table = info.request.table
-		qry.routingInfo.mu.Unlock()
+		q.routingInfo.mu.Lock()
```
jameshartig, line 1586: Does this still need to lock? I don't think it does unless we expect `GetRoutingKey` to be called from separate goroutines?

jameshartig: Also can we just change `routingInfo` to not be a struct anymore? Seems like it only was because...
> // routingInfo is a pointer because Query can be copied and copyable struct can't hold a mutex.

Do we still allow copies of `internalQuery`? If so, maybe we should make a Clone method?

query_executor.go
```diff
@@ -201,16 +210,375 @@ func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery, hostIter Ne
...
+	disableAutoPage       bool
+	skipPrepare           bool
+	routingKey            []byte
+
+	// getKeyspace is field so that it can be overriden in tests
+	getKeyspace func() string
+}
+
+func newQueryOptions(q *Query) *queryOptions {
+	return &queryOptions{
```
jameshartig, line 309: Should we be copying `values`, `customPayload`, `initialPageState`, `routingKey`? it might not be obvious that those are not copied and can't be mutated.

query_executor.go
```diff
@@ -201,16 +210,375 @@ func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery, hostIter Ne
...
+func (q *internalQuery) Context() context.Context {
+	return q.qryOpts.context
+}
+
+func (q *internalQuery) Statement() Statement {
+	return q.originalQuery
+}
+
+type batchOptions struct {
+	Type                  BatchType
```
jameshartig: Any reason why some of these are public?

session.go
jameshartig: Hmm, I see what you mean but I don't think we can lock them into one or another if we want it to be reusable.

jameshartig:
What about making the methods `Exec(context.Context)` and `Iter(context.Context) error` so there's no need to shallow copy query in `WithContext`? This is a bigger change though that we might not want to make. Taht said, I think Query would be a bit easier to re-use then. I know this differs from `http.Request` but in that case it might be a symptom of not wanting to break the existing API when context's were introduced.

session.go
joao-r-reis: > I feel that people may execute sequence of `Exec()` and `Iter()`. Shall we make an assertion that only one of the is invoked?

They can't execute sequence (`Exec().Iter()`) unless they do it like:
```
err := b.Exec()
iter := b.Iter()
``` 

Which I believe is already something that can be done with the current API and it will trigger two requests. We can try to improve documentation if this is a concern.

+1 to what James said

joao-r-reis:
> > With this PR, this method is deprecated on Query and Batch since these two types no longer implement ExecutableQuery so we can keep the Query/Batch API much simpler.
> 
> Why not just remove it? Seems like any `HostSelectionPolicy` implementation would need to make major changes already.

Hmm I don't think `HostSelectionPolicy` implementations have to make major changes after this PR unless they rely on casting the `ExecutableQuery` object into `Query` or `Batch` (and even this is a small change, they just need to call .Statement() before doing so). 

> I don't want to bikeshed about names but ExecutableQuery is a bit confusing since it contains Query and Batch. Did that name come from somewhere? Would ExecutableStatement make more sense since it has a statement?

I kept `ExecutableQuery` name unchanged so `HostSelectionPolicy` implementations can pretty much continue to work without many changes (or even any change at all). cc @lukasz-antoniak because you also brought this up in 1 of your comments.
I'm concerned that renaming this interface will increase the chance users will have to change their code when upgrading the driver but I do agree that the current name isn't good.

joao-r-reis:
> What about making the methods `Exec(context.Context)` and `Iter(context.Context) error` so there's no need to shallow copy query in `WithContext`? This is a bigger change though that we might not want to make. Taht said, I think Query would be a bit easier to re-use then. I know this differs from `http.Request` but in that case it might be a symptom of not wanting to break the existing API when context's were introduced.

I like the idea but I think it would be too much to ask of users when upgrading since it would affect every single statement on their app... I'd be down to adding an overload and deprecating `WithContext` though... I just don't know what names would fit for the new Iter/Exec overloads...

query_executor.go
joao-r-reis: The type itself is not public and it is never exposed so it wasn't a deliberate choice, just a result of copy paste. I can make them all private for consistency

query_executor.go
joao-r-reis: I think copying `values` and `customPayload` might be a bit too much due to performance reasons. `initialPageState` should probably be copied for safety and since its length is usually pretty small. `routingKey` I'm not so sure, I think it will be `nil` the vast majority of cases anyway but I can make sure it gets copied as well

conn.go
joao-r-reis: > Does this still need to lock? I don't think it does unless we expect GetRoutingKey to be called from separate goroutines?

I think this is still the case due to speculative executions (and possibly retries I'm not sure).

> Also can we just change routingInfo to not be a struct anymore? Seems like it only was because...

>>    // routingInfo is a pointer because Query can be copied and copyable struct can't hold a mutex.

> Do we still allow copies of internalQuery? If so, maybe we should make a Clone method?

I don't think it needs to be a struct anymore, I'll change it.

query_executor.go
joao-r-reis: I posted a comment about the renaming of this interface below in a response to James

query_executor.go
joao-r-reis: yeah I'll do this

query_executor.go
joao-r-reis: Good catch, leftover from a copy paste

worryg0d:
> I'm concerned that renaming this interface will increase the chance users will have to change their code when upgrading the driver but I do agree that the current name isn't good.

We can add a type alias to leave the `ExecutableQuery` name for the migration period:
```go
// Deprecated: Will be removed in the future major release.
// Please use Statement instead.
type ExecutableQuery = Statement

type Statement interface {
...
}
```

I tested it on go 1.19.13 and it works fine

worryg0d:
> I just don't know what names would fit for the new Iter/Exec overloads...

Well, we can use the common **Context** suffix for those overloads, like it is done in sql package from standard lib - https://pkg.go.dev/database/sql#Conn.QueryContext
https://pkg.go.dev/database/sql#DB.ExecContext

conn.go
joao-r-reis: So I made some changes to the `routingInfo` part but I didn't remove the struct... I think it still serves a purpose by making it clear that the mutex is used to protect access against the two fields within the struct. Do you still prefer to remove the struct and just have the mutex and two fields in the request type itself?

query_executor.go
joao-r-reis: On a second thought I'm not sure if I like this idea that much, I'd have to create a few new types when maybe I can just group them a bit more by arranging the way they are declared... Wdyt

joao-r-reis: Something like this:

```
type queryOptions struct {
	stmt string

	// Paging
	pageSize         int
	initialPageState []byte
	disableAutoPage  bool

	// Monitoring
	trace    Tracer
	observer QueryObserver

	// Parameters
	values  []interface{}
	binding func(q *QueryInfo) ([]interface{}, error)

	// Timestamp
	defaultTimestamp      bool
	defaultTimestampValue int64

	// Consistency
	initialConsistency Consistency
	serialCons         SerialConsistency

	// Protocol flag
	disableSkipMetadata bool

	customPayload map[string][]byte
	prefetch      float64
	rt            RetryPolicy
	spec          SpeculativeExecutionPolicy
	context       context.Context
	idempotent    bool
	keyspace      string
	skipPrepare   bool
	routingKey    []byte

	// getKeyspace is field so that it can be overriden in tests
	getKeyspace func() string
}
```

joao-r-reis:
I've addressed all PR comments, I'll work on rebasing the branch now and then I'll work on the change mentioned [here](https://github.com/apache/cassandra-gocql-driver/pull/1870#issuecomment-2769579576).

joao-r-reis:
Rebase is done, it was a bit more complex than I thought it would be so I'd appreciate if you guys could take a look again @jameshartig @lukasz-antoniak @worryg0d

query_executor.go
```diff
@@ -222,16 +238,452 @@ func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery, hostIter Ne
...
+	getKeyspace func() string
+}
+
+func newQueryOptions(q *Query, ctx context.Context) *queryOptions {
+	var newPageState, newRoutingKey []byte
+	if q.initialPageState != nil {
+		pageState := q.initialPageState
+		newPageState = make([]byte, len(pageState))
+		copy(newPageState, pageState)
+	}
```
ribaraka: `newPageState` seems to be redundant here.

query_executor.go
ribaraka: still persist. but changed its position :sweat_smile:

query_executor.go
```diff
@@ -222,16 +238,452 @@ func (q *queryExecutor) do(ctx context.Context, qry ExecutableQuery, hostIter Ne
...
+
+func newInternalBatch(batch *Batch, ctx context.Context) *internalBatch {
+	return &internalBatch{
+		originalBatch: batch,
+		batchOpts:     newBatchOptions(batch, ctx),
+		metrics:       &queryMetrics{m: make(map[string]*hostMetrics)},
+		routingInfo:   &queryRoutingInfo{},
+		session:       batch.session,
+	}
+}
```
ribaraka: `consistency` is missing

session.go
```diff
@@ -743,33 +736,47 @@ func (s *Session) routingKeyInfo(ctx context.Context, stmt string, keyspace stri
...
-func (s *Session) executeBatch(batch *Batch) *Iter {
+// Iter executes a batch operation and returns an Iter object
+// that can be used to access properties related to the execution like Iter.Attempts and Iter.Latency
+func (b *Batch) Iter() *Iter { return b.session.executeBatch(b, nil) }
+
+// Iter executes a batch operation with the provided context and returns an Iter object
+// that can be used to access properties related to the execution like Iter.Attempts and Iter.Latency
+func (b *Batch) IterContext(ctx context.Context) *Iter {
+	return b.session.executeBatch(b, ctx)
+}
```
ribaraka: nitpicks: 
1. The `Iter` method should call the `IterContext`
2. The comment for `IterContext` should start with the method name

query_executor.go
```diff
@@ -27,33 +27,51 @@ package gocql
...
+	Statement() Statement
+}
 
-	withContext(context.Context) ExecutableQuery
+// Statement is an interface that represents a CQL statement that the driver can execute
+// (currently Query and Batch via Session.Query and Session.Batch)
+type Statement interface {
+	Iter() *Iter
+	Exec() error
+}
```
ribaraka: adding here `IterContext` and `ExecContetx`?

query_executor.go
joao-r-reis: Well the reason for this is to ensure that a query execution isn't affected if the user tries to modify the page byte slice after query has been submitted for execution

session.go
joao-r-reis: done

query_executor.go
joao-r-reis: nice catch, done

query_executor.go
joao-r-reis: good suggestion 👍 done

query_executor.go
joao-r-reis: I do wonder if we're just adding a performance penalty to safeguard against something that users won't attempt to do 99% of the times

query_executor.go
joao-r-reis: oops, maybe I added it back during rebase, fixed

query_executor.go
ribaraka: I mean that `newPageState` is copied but never used. And it doesn't seem the `queryOptions` struct has such a field.

joao-r-reis: oh you're right, I removed this field from this type and moved it up to `internalQuery` but I forgot to remove this part, nice catch. Done

ribaraka (approved):

joao-r-reis:
@jameshartig @worryg0d do you guys have some free time to take a look at this again during the next few days?

jameshartig (approved):

CHANGELOG.md
```diff
@@ -54,7 +54,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Move "execute batch" methods to Batch type (CASSGO-57)
 
-- Make `Session` immutable by removing setters and associated mutex (CASSGO-23)
+- Make Session immutable by removing setters and associated mutex (CASSGO-23)
+
+- Changes to Query and Batch to make them safely reusable (CASSGO-22)
```
joao-r-reis, line 47: reminder to add CASSGO-73 here

jameshartig (approved):

PR #1870, opened 2025-03-20
Add IsIdempotent to ObservedQuery and ObservedBatch

max-melentyev:
To let observers know

### DIFF
--- a/session.go
+++ b/session.go
@@ -1135,16 +1135,17 @@ func (q *Query) attempt(keyspace string, end, start time.Time, iter *Iter, host 
 
 	if q.observer != nil {
 		q.observer.ObserveQuery(q.Context(), ObservedQuery{
-			Keyspace:  keyspace,
-			Statement: q.stmt,
-			Values:    q.values,
-			Start:     start,
-			End:       end,
-			Rows:      iter.numRows,
-			Host:      host,
-			Metrics:   metricsForHost,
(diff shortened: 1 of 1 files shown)


jameshartig:
In #1868 we're making Query/Batch reusable. I wonder if we should include those in the observations instead of piecemeal adding a bunch of fields from them. @joao-r-reis thoughts?

joao-r-reis:
Yeah I agree with that, I can make that change in #1868 but my implementation doesn't make them threadsafe, just reusable so if a user tries to change the query/batch object within an implementation of these observers then bad things might happen...

max-melentyev:
Cool! What do you think about adding custom data to a query/batch that is passed to observers? We use a context.WithValue() to pass query names to observer, but this approach can lead to errors because engineers can use the same context for different queries. Would be great to have an approach when data is assigned for individual queries.

joao-r-reis:
> Cool! What do you think about adding custom data to a query/batch that is passed to observers? We use a context.WithValue() to pass query names to observer, but this approach can lead to errors because engineers can use the same context for different queries. Would be great to have an approach when data is assigned for individual queries.

This seems like a reasonable feature request but I'd recommend creating a new issue/PR for this as it seems unrelated to this one. I'm handling the current one in #1868 by adding the original query/batch objects to the `Observed...` types.

PR #1874, opened 2025-04-02
CASSGO-71 Make HostFilter interface easier to test

### LINKED ISSUE #1311: External HostFilter implementations cannot be tested against HostFilter interface

n/a

master

### What did you do?
```
// hostfilter.go
type datacenterRegexFilter struct { pattern *regexp.Regexp }
func (d *datacenterRegexFilter) Accept(host *gocql.HostInfo) bool {
    return d.pattern.MatchString(host.DataCenter())
}
var _ gocql.HostFilter = (*datacenterRegexFilter)(nil)

// hostfilter_test.go
func TestDatacenterRegexHostFilter(t *testing.T) {
    f := &datacenterRegexFilter{pattern: regexp.MustCompile("us-east-1a|us-east-1d")}
    assert.True(t, f.Accept(&gocql.HostInfo{
        // dataCenter: "us-east-1a"
    }))
}
```

Test is successful

### What did you see instead?
Because `dataCenter` is unexported and there is no setter, this test cannot be written. The [tests for other host filters](https://github.com/gocql/gocql/blob/master/filters_test.go#L20) can be written because they create new `*HostInfo` structs inline by assigning to unexported fields. But implementations that use the `HostFilter` interface to provide a custom filter cannot be tested without boilerplate code. 

Host filters that only filter on the connection address can use the public setter [`SetConnectAddress`](https://godoc.org/github.com/gocql/gocql#HostInfo.SetConnectAddress). However filters that use other fields like `dataCenter` cannot use the unexported setters like `setDataCenter`.

OleksiienkoMykyta:
Closes #1311 
Currently, external HostFilter implementations cannot be tested against HostFilter interface.
To fix that, the Accept method has an Host interface as the parameter instead of HostInfo so that it is possible to mock it.

### DIFF
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -480,6 +480,12 @@ func TestCAS(t *testing.T) {
 		t.Fatalf("insert should have not been applied: title=%v revID=%v modified=%v", titleCAS, revidCAS, modifiedCAS)
 	}
 
+	// TODO: This test failing with error due to "function dateof" on cassandra side.
+	//  It was officially removed in version 5.0.0. The recommended replacement for dateOf is the toTimestamp function.
+	//   As we are not testing against deprecated cassandra versions, it makes sense to update tests to keep them up to date
+	//	 === RUN   TestCAS
+	// 	 cassandra_test.go:487: insert: Unknown function dateof called
+	// 	 --- FAIL: TestCAS (0.97s)
 	insertBatch := session.Batch(LoggedBatch)
 	insertBatch.Query("INSERT INTO cas_table (title, revid, last_modified) VALUES ('_foo', 2c3af400-73a4-11e5-9381-29463d90c3f0, DATEOF(NOW()))")
 	insertBatch.Query("INSERT INTO cas_table (title, revid, last_modified) VALUES ('_foo', 3e4ad2f1-73a4-11e5-9381-29463d90c3f0, DATEOF(NOW()))")
@@ -951,7 +957,7 @@ func TestReconnection(t *testing.T) {
 		t.Fatal("Host should be NodeDown but not.")
 	}
 
-	time.Sleep(cluster.ReconnectInterval + h.Version().nodeUpDelay() + 1*time.Second)
+	time.Sleep(cluster.ReconnectInterval + 1*time.Second)
 
 	if h.State() != NodeUp {
 		t.Fatal("Host should be NodeUp but not. Failed to reconnect.")
--- a/events.go
+++ b/events.go
@@ -227,9 +227,6 @@ func (s *Session) handleNodeUp(eventIp net.IP, eventPort int) {
 		return
 	}
 
-	if d := host.Version().nodeUpDelay(); d > 0 {
-		time.Sleep(d)
-	}
 	s.startPoolFill(host)
 }
 
--- a/filters.go
+++ b/filters.go
@@ -24,31 +24,66 @@
 
 package gocql
 
-import "fmt"
+import (
+	"fmt"
+	"net"
+)
 
 // HostFilter interface is used when a host is discovered via server sent events.
 type HostFilter interface {
 	// Called when a new host is discovered, returning true will cause the host
 	// to be added to the pools.
-	Accept(host *HostInfo) bool
+	Accept(host Host) bool
+}
+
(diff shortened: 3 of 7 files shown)


OleksiienkoMykyta:
@joao-r-reis, could you please review the PR?

filters.go
```diff
@@ -24,39 +24,74 @@
...
+// Since cassVersion is an unexported type, the CassVersion interface is introduced
+// to allow better testability and increase test coverage.
+type CassVersion interface {
+	Set(v string) error
+	UnmarshalCQL(info TypeInfo, data []byte) error
+	AtLeast(major, minor, patch int) bool
+	String() string
 }
 
 // HostFilterFunc converts a func(host HostInfo) bool into a HostFilter
```
tengu-alt, line 71: I think it also needs to be updated to `func(host Host) bool`

tengu-alt (commented):
Left one comment. Overall LGTM

cassandra_test.go
```diff
@@ -480,6 +480,12 @@ func TestCAS(t *testing.T) {
 		t.Fatalf("insert should have not been applied: title=%v revID=%v modified=%v", titleCAS, revidCAS, modifiedCAS)
 	}
 
+	// TODO: This test failing with error due to "function dateof" on cassandra side.
+	//  It was officially removed in version 5.0.0. The recommended replacement for dateOf is the toTimestamp function.
+	//   As we are not testing against deprecated cassandra versions, it makes sense to update tests to keep them up to date
```
joao-r-reis, line 485: I believe this was done in #1828 so once that is merged and this PR is rebased it should be fixed

events.go
```diff
@@ -227,9 +227,6 @@ func (s *Session) handleNodeUp(eventIp net.IP, eventPort int) {
 		return
 	}
 
-	if d := host.Version().nodeUpDelay(); d > 0 {
```
joao-r-reis, line 230: Why remove this?

filters.go
```diff
@@ -24,39 +24,74 @@
...
+	String() string
+}
+
+// Since cassVersion is an unexported type, the CassVersion interface is introduced
+// to allow better testability and increase test coverage.
+type CassVersion interface {
+	Set(v string) error
+	UnmarshalCQL(info TypeInfo, data []byte) error
+	AtLeast(major, minor, patch int) bool
+	String() string
```
joao-r-reis, line 68: I think this needs some changes, just making an interface out of the methods of the private type feels wrong here, we shouldn't expose `Set` or `Unmarshal` method and we should definitely expose a way to get the actual version fields.

Ideally `CassVersion` would be:

```
type CassVersion interface {
	Major() int
        Minor() int
        Patch() int
        Qualifier() string
	String() string
}
```

The methods on `cassVersion` can be converted into global private functions.
```
func (c *cassVersion) Set(v string) error 
func (c *cassVersion) UnmarshalCQL(info TypeInfo, data []byte) error 
func (c *cassVersion) unmarshal(data []byte) error {
func (c cassVersion) Before(major, minor, patch int) bool 
func (c cassVersion) AtLeast(major, minor, patch int) bool
```

```
func parseCassVersion(v string) (CassVersion, error) {
        c := &cassVersion{}
	if v == "" {
		return c, nil
	}

	return c.unmarshalCQL(nil, []byte(v))
}

func (c *cassVersion) unmarshalCQL(info TypeInfo, data []byte) error {
	return c.unmarshal(data)
}

// can be kept as is
func (c *cassVersion) unmarshal(data []byte) error

// these two should just be global, private and have `CassVersion` parameters instead
func before(v1 CassVersion, v2 CassVersion) bool 
func atLeast(v1 CassVersion, v2 CassVersion) bool
```

joao-r-reis (commented):
Added some comments.

I think there's other parts of the public API that need their `HostInfo` references changed to `Host` like observers, `Iter` and policies

joao-r-reis:
@OleksiienkoMykyta do you have time to address my review comments? We're trying to resolve the pending PRs for the 2.0 release

joao-r-reis:
After discussing with @jameshartig I've opened #1892 as an alternative

joao-r-reis:
Closed in favor of #1892

PR #1875, opened 2025-04-08
CASSGO-50 Endless query execution fix proposal

tengu-alt:
I've researched possible endless query execution when `HostSelectionPolicy` returns the same downed host ([CASSGO-50](https://issues.apache.org/jira/browse/CASSGO-50)) and propose three approaches to prevent it.

- Wrap `hostIter` to prevent it from returning downed host before executing the query. 
 e.g.
```
  hostIter = func() SelectedHost {
    checkIter := hostIter()
    if !checkIter.Info().IsUp() {
      return nil
    }
    return checkIter
  }
```
But that approach can cause breaking changes for applications which already relies on user-defined policies that can return downed hosts.

- Properly documented `type NextHost func()`:
Add a comment for `NextHost` type describing that `NextHost` should return not downedhost or return nil

- Add the threshold limit based on the number of hosts inside `(q *queryExecutor) do` which will prevent the endless execution without the breaking changes

Implementations of the threshold and documentation are presented in this PR.

### DIFF
--- a/policies.go
+++ b/policies.go
@@ -323,6 +323,7 @@ func (host *selectedHost) Info() *HostInfo {
 func (host *selectedHost) Mark(err error) {}
 
 // NextHost is an iteration function over picked hosts
+// Should return nil eventually to prevent endless query execution.
 type NextHost func() SelectedHost
 
 // RoundRobinHostPolicy is a round-robin load balancing policy, where each host
--- a/query_executor.go
+++ b/query_executor.go
@@ -27,6 +27,7 @@ package gocql
 import (
 	"context"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -89,14 +90,16 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
 	// check if the host id is specified for the query,
 	// if it is, the query should be executed at the corresponding host.
 	if hostID := qry.GetHostID(); hostID != "" {
+		host, ok := q.pool.session.ring.getHost(hostID)
+		if !ok {
+			return nil, ErrNoConnections
+		}
+		var returnedHostOnce int32 = 0
 		hostIter = func() SelectedHost {
-			pool, ok := q.pool.getPoolByHostID(hostID)
-			// if the specified host is down
-			// we return nil to avoid endless query execution in queryExecutor.do()
-			if !ok || !pool.host.IsUp() {
-				return nil
+			if atomic.CompareAndSwapInt32(&returnedHostOnce, 0, 1) {
+				return (*selectedHost)(host)
 			}
-			return (*selectedHost)(pool.host)
+			return nil
 		}
 	}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Don't panic in MapExecuteBatchCAS if no `[applied]` column is returned (CASSGO-42)
 
+- Endless query execution fix (CASSGO-50)
+
 ## [1.7.0] - 2024-09-23
 
 This release is the first after the donation of gocql to the Apache Software Foundation (ASF)
--- a/events_ccm_test.go
+++ b/events_ccm_test.go
@@ -104,7 +104,7 @@ func TestEventNodeDownControl(t *testing.T) {
 	}
 	session.pool.mu.RUnlock()
 
-	host := session.ring.getHost(node.Addr)
+	host, _ := session.ring.getHost(node.Addr)
 	if host == nil {
 		t.Fatal("node not in metadata ring")
 	} else if host.IsUp() {
@@ -146,7 +146,7 @@ func TestEventNodeDown(t *testing.T) {
 		t.Fatal("node not removed after remove event")
 	}
 
-	host := session.ring.getHost(node.Addr)
+	host, _ := session.ring.getHost(node.Addr)
 	if host == nil {
 		t.Fatal("node not in metadata ring")
 	} else if host.IsUp() {
@@ -203,7 +203,7 @@ func TestEventNodeUp(t *testing.T) {
 		t.Fatal("node not added after node added event")
 	}
 
-	host := session.ring.getHost(node.Addr)
+	host, _ := session.ring.getHost(node.Addr)
 	if host == nil {
 		t.Fatal("node not in metadata ring")
 	} else if !host.IsUp() {
--- a/ring.go
+++ b/ring.go
@@ -67,11 +67,11 @@ func (r *ring) getHostByIP(ip string) (*HostInfo, bool) {
 	return r.hosts[hi], ok
 }
 
-func (r *ring) getHost(hostID string) *HostInfo {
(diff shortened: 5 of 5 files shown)


tengu-alt:
@joao-r-reis @jameshartig can you please take a look at this?

joao-r-reis:
Hmm I don't like the idea of introducing a dependency on the number of connection pools but wrapping `hostIter` won't solve the issue because whether the host is down or nil doesn't matter much, it will lead to the same behavior...

I actually think it's fine for the driver to assume that the host selection policy is behaving correctly (i.e. not returning the same host over and over) but we still have the issue that in #1793 we added an internal policy that does this (returns the same host everytime the function is called).

I've commented this before but I truly think the best option would be to make the internal host selection policy added in #1793 return the host only once even if that comes at a cost of having no retries or speculative executions when the user uses this feature.

Later down the line if we see that there is demand for retries/speculative executions when the featured added in #1793 is used then we can reevaluate.

One potential alternative if we absolutely want to keep retries when a specific host is being targeted by the query is to add a special flag so that the query executor treats this host selection policy differently (i.e. break from the loop immediately when the returned host is nil or downed).

joao-r-reis:
@tengu-alt do you have time to address [this](https://github.com/apache/cassandra-gocql-driver/pull/1875#issuecomment-2818077493)? We're trying to resolve the pending PRs for the 2.0 release

tengu-alt:
> @tengu-alt do you have time to address https://github.com/apache/cassandra-gocql-driver/pull/1875#issuecomment-2818077493? We're trying to resolve the pending PRs for the 2.0 release

Sorry for the late response. I've pushed another approach according to your suggestions.  
I propose the deletion of the pool after we got it using the `getPoolByHostID`, and keep the documentation about policies correct behavior.

connectionpool.go
```diff
@@ -246,6 +246,9 @@ func (p *policyConnPool) getPool(host *HostInfo) (pool *hostConnPool, ok bool) {
 func (p *policyConnPool) getPoolByHostID(hostID string) (pool *hostConnPool, ok bool) {
 	p.mu.RLock()
 	pool, ok = p.hostConnPools[hostID]
+	if ok {
+		p.hostConnPools[hostID] = nil
```
joao-r-reis: What's the goal here? In my mind this will make it so that only one query will ever run on a specific pool no?

tengu-alt: Good catch, fixed. I've changed the `hostIter` func so it returns the `SelectedHost` once.
.

query_executor.go
```diff
@@ -89,14 +89,18 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
...
+			return nil, ErrNoConnections
+		}
+		host := pool.host
 		hostIter = func() SelectedHost {
-			pool, ok := q.pool.getPoolByHostID(hostID)
-			// if the specified host is down
-			// we return nil to avoid endless query execution in queryExecutor.do()
-			if !ok || !pool.host.IsUp() {
-				return nil
+			if host != nil {
```
joao-r-reis: To ensure this `hostIter` can be safely called concurrently I think we should make the outer variable an `int` and then use `atomic.CompareAndSwap` in this `if`

```
		returnedHostOnce := 0
		hostIter = func() SelectedHost {
			if atomic.CompareAndSwapInt32(&returnedHostOnce, 0, 1) {
                            return return (*selectedHost)(host)
                        }
                        return nil
                }
```

query_executor.go
```diff
@@ -89,14 +89,18 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
 	// check if the host id is specified for the query,
 	// if it is, the query should be executed at the corresponding host.
 	if hostID := qry.GetHostID(); hostID != "" {
+		pool, ok := q.pool.getPoolByHostID(hostID)
+		if !ok || !pool.host.IsUp() {
```
joao-r-reis: I don't think we need this now that we return `nil` after the first execution since this check is already being done in the query executor. We needed this before because we needed a way to prevent the `hostIter` from returning the same unresponsive host over and over again.

query_executor.go
tengu-alt: Understood, refactored.

query_executor.go
tengu-alt: Got it, changed.

policies.go
```diff
@@ -323,6 +323,7 @@ func (host *selectedHost) Info() *HostInfo {
 func (host *selectedHost) Mark(err error) {}
 
 // NextHost is an iteration function over picked hosts
+// Should return nil if SelectedHost is not Up to prevent endless query execution.
```
joao-r-reis: the driver checks if the host is not up, the function should "eventually" return nil regardless of what causes it.

`// Should return nil eventually to prevent endless query execution.`

query_executor.go
```diff
@@ -89,14 +90,17 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
 	// check if the host id is specified for the query,
 	// if it is, the query should be executed at the corresponding host.
 	if hostID := qry.GetHostID(); hostID != "" {
+		pool, ok := q.pool.getPoolByHostID(hostID)
+		if !ok {
+			return nil, ErrNoConnections
+		}
```
joao-r-reis, line 96: remove lines 93-96, it's not needed as I explained before

query_executor.go
```diff
@@ -89,14 +90,17 @@ func (q *queryExecutor) executeQuery(qry ExecutableQuery) (*Iter, error) {
...
+		host := pool.host
+		var returnedHostOnce int32 = 0
 		hostIter = func() SelectedHost {
-			pool, ok := q.pool.getPoolByHostID(hostID)
-			// if the specified host is down
-			// we return nil to avoid endless query execution in queryExecutor.do()
-			if !ok || !pool.host.IsUp() {
-				return nil
+			if atomic.CompareAndSwapInt32(&returnedHostOnce, 0, 1) {
+				return (*selectedHost)(host)
```
joao-r-reis, line 100: we should add a test to ensure that if a host id is set on the query then it only gets tried once even if it fails

query_executor.go
tengu-alt: I don't clearly understand: we need the pool to get the host from it.

joao-r-reis: oh sorry hmm I think it would make more sense to just get the host from the ring here, I didn't even realize it was only getting the pool to get the host before.
Can you replace this with a call to `session.ring.getHost(id)`? Maybe change `getHost` to return `host, bool` so that we can check whether the lookup failed here and return an error

policies.go
tengu-alt: Changed.

query_executor.go
tengu-alt: Got it, refactored

query_executor.go
tengu-alt: As I checked, `TestQuery_SetHostID` in the `cassandra_test.go` already covers this case.

joao-r-reis (approved):
👍 nice, can you add to the changelog and format the commit message? Then I'll merge

tengu-alt:
> 👍 nice, can you add to the changelog and format the commit message? Then I'll merge

Done.

PR #1879, opened 2025-04-14
<framer> optimization `setLength()` method

illia-li:
Optimizes `setLength()` method of the `framer`

Usually used version more than 2, so it's better to change the code so that it handles this particular case first.


jameshartig:
We'll be dropping protocol version < 3 in #1890

Issue #1473, opened 2020-07-29
Data race when calling SetConsistency


ncabatoff:
Please answer these questions before submitting your issue. Thanks!

### What version of Cassandra are you using?

Docker image cassandra:3.11.

### What version of Gocql are you using?

```
$ grep gocql go.mod
        github.com/gocql/gocql v0.0.0-20200624222514-34081eda590e
```

### What did you do?

```
	sess, err := cluster.CreateSession()
	if err != nil {
		return nil, err
	}
	sess.SetConsistency(consistency)
```

Elsewhere:
```
        sess.Query(...)
```

### What did you expect to see?

No data race when running with `-race`.

### What did you see instead?

```
=== FAIL: physical/cassandra TestCassandraBackend (25.07s)
==================
WARNING: DATA RACE
Read at 0x00c0001d8380 by goroutine 25:
  github.com/hashicorp/vault/vendor/github.com/gocql/gocql.(*Conn).UseKeyspace()
      /go/src/github.com/hashicorp/vault/vendor/github.com/gocql/gocql/conn.go:1235 +0x186
  github.com/hashicorp/vault/vendor/github.com/gocql/gocql.(*hostConnPool).connect()
      /go/src/github.com/hashicorp/vault/vendor/github.com/gocql/gocql/connectionpool.go:534 +0x53d
  github.com/hashicorp/vault/vendor/github.com/gocql/gocql.(*hostConnPool).connectMany.func1()
      /go/src/github.com/hashicorp/vault/vendor/github.com/gocql/gocql/connectionpool.go:487 +0x79

Previous write at 0x00c0001d8380 by goroutine 7:
  github.com/hashicorp/vault/vendor/github.com/gocql/gocql.(*Session).SetConsistency()
      /go/src/github.com/hashicorp/vault/vendor/github.com/gocql/gocql/session.go:336 +0x54
  github.com/hashicorp/vault/physical/cassandra.NewCassandraBackend()
      /go/src/github.com/hashicorp/vault/physical/cassandra/cassandra.go:144 +0x917
  github.com/hashicorp/vault/physical/cassandra.TestCassandraBackend()
      /go/src/github.com/hashicorp/vault/physical/cassandra/cassandra_test.go:29 +0x2dc
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1039 +0x1eb

Goroutine 25 (running) created at:
[...]

PR #1882, opened 2025-04-18
CASSGO-23 Make Session immutable

### LINKED ISSUE #357: CASSGO-23 Make Session immutable

ribaraka:
Closes https://github.com/apache/cassandra-gocql-driver/issues/357

The previous design allowed modifying Session properties through setters. This added unnecessary overhead, as a mutex had to be acquired for every query execution. Additionally, as demonstrated in [#1473](https://github.com/apache/cassandra-gocql-driver/issues/1473) this mutability could lead to race conditions. By making Session immutable, we:  
1. Eliminate race condition.  
1. Eliminate the performance overhead of mutex acquisition  
2. Simplify the concurrency model  
3. Making cleaner configuration

This PR removing those setters and eliminating the need for the mutex. Configuration of refactored properties (i.e. consistency, trace etc.) is now handled at session creation through `ClusterConfig` and/or at execution time through **per-query** methods.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Move "execute batch" methods to Batch type (CASSGO-57)
 
+- Make `Session` immutable by removing setters and associated mutex (CASSGO-23)
+
 ### Fixed
 - Cassandra version unmarshal fix (CASSGO-49)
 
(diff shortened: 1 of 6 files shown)


joao-r-reis (approved):
Nice and simple 👍

joao-r-reis:
Hi can you fix the conflict on `CHANGELOG.md` and fix the commit message? I'll merge this afterwards

ribaraka:
@joao-r-reis, done

joao-r-reis (approved):

PR #1886, opened 2025-04-27
CASSGO-72 Fix connection issue in Amazon Keyspaces

### LINKED ISSUE #1873: CASSGO-72 Connection trouble with Amazon Keyspaces
I have a trouble connecting to AWS Amazon Keyspaces (for Apache Cassandra). My program on EC2 can connect to Amazon Keyspaces without any issues for a while, but after a few days or weeks, it loses the connection and any query causes the error below.

```
```

Go version: 1.23.4
GoCQL version: 1.7.0

I built the program with gocql_debug enabled, and I got following logs.

```
2025/03/25 03:20:29 gocql: Session.handleNodeConnected: 172.16.1.14:9142
2025/03/25 03:20:29 gocql: conns of pool after stopped "172.16.1.14": 2
2025/03/25 03:20:29 gocql: Session.handleNodeConnected: 172.16.1.28:9142
2025/03/25 03:20:29 gocql: conns of pool after stopped "172.16.1.28": 2
2025/03/25 03:21:29 Session.ring:[172.16.1.14:UP][172.16.1.28:UP]

...

2025/03/26 15:11:11 gocql: unable to dial "[HostInfo hostname=\"\" connectAddress=\"127.0.0.1\" peer=\"<nil>\" rpc_address=\"127.0.0.1\" broadcast_address=\"127.0.0.1\" preferred_ip=\"<nil>\" connect_addr=\"127.0.0.1\" connect_addr_source=\"connect_address\" port=9142 data_centre=\"ap-northeast-1\" rack=\"ap-northeast-1\" host_id=\"be0f3a14-e107-3fee-a5e5-415c10539abd\" version=\"v3.11.2\" state=UP num_tokens=0]": dial tcp 127.0.0.1:9142: connect: connection refused
2025/03/26 15:11:11 gocql: filling stopped "127.0.0.1": dial tcp 127.0.0.1:9142: connect: connection refused
2025/03/26 15:11:11 gocql: conns of pool after stopped "127.0.0.1": 0
2025/03/26 15:11:11 gocql: Session.handleNodeDown: 127.0.0.1:9142
2025/03/26 15:11:11 gocql: unable to refresh ring: get existing host=[HostInfo hostname="" connectAddress="172.16.1.14" peer="172.16.1.14" rpc_address="172.16.1.14" broadcast_address="<nil>" preferred_ip="172.16.1.14" connect_addr="172.16.1.14" connect_addr_source="connect_address" port=9142 data_centre="ap-northeast-1" rack="ap-northeast-1" host_id="be0f3a14-e107-3fee-a5e5-415c10539abd" version="v3.11.2" state=UP num_tokens=1] from prevHosts: cannot find host
2025/03/26 15:11:29 Session.ring:[127.0.0.1:DOWN][172.16.1.28:UP]

...

2025/03/26 22:43:35 gocql: unable to dial "[HostInfo hostname=\"\" connectAddress=\"127.0.0.1\" peer=\"<nil>\" rpc_address=\"127.0.0.1\" broadcast_address=\"127.0.0.1\" preferred_ip=\"<nil>\" connect_addr=\"127.0.0.1\" connect_addr_source=\"connect_address\" port=9142 data_centre=\"ap-northeast-1\" rack=\"ap-northeast-1\" host_id=\"b666465e-cb85-3efa-b3ab-f6cf139e5a39\" version=\"v3.11.2\" state=UP num_tokens=0]": dial tcp 127.0.0.1:9142: connect: connection refused
2025/03/26 22:43:35 gocql: filling stopped "127.0.0.1": dial tcp 127.0.0.1:9142: connect: connection refused
2025/03/26 22:43:35 gocql: conns of pool after stopped "127.0.0.1": 0
2025/03/26 22:43:35 gocql: Session.handleNodeDown: 127.0.0.1:9142
2025/03/26 22:43:35 gocql: unable to refresh ring: get existing host=[HostInfo hostname="" connectAddress="172.16.1.28" peer="172.16.1.28" rpc_address="172.16.1.28" broadcast_address="<nil>" preferred_ip="172.16.1.28" connect_addr="172.16.1.28" connect_addr_source="connect_address" port=9142 data_centre="ap-northeast-1" rack="ap-northeast-1" host_id="b666465e-cb85-3efa-b3ab-f6cf139e5a39" version="v3.11.2" state=UP num_tokens=1] from prevHosts: cannot find host
2025/03/26 22:44:29 Session.ring:[127.0.0.1:DOWN][127.0.0.1:DOWN]
```

On startup, It has two hosts 172.16.1.14 and 172.16.1.28. After a while, the connection to 172.16.1.14 got lost with error `cannot find host` and try to reconnect to 127.0.0.1 instead of 172.16.1.14. After another while, the other connection also got lost with the same error and also try to reconnect to 127.0.0.1 instead of 172.16.1.28. As a result, all connections got lost.

So here are my questions:

First, in what situation the error `cannot find host` occur? Is this an expected error? I read the source code, but I couldn't understand it well.
Second, what makes it reconnect to 127.0.0.1 instead of original address? Is this an expected behavior?

If anyone has any idea, please let me know.

yomipq:
This PR fixes #1873 

In Amazon Keyspaces, system.local returns the localhost address and system.peers returns a host that contains the same hostID as localhost. This causes the connection issue in refreshRing(), where host address is overwritten with the localhost address and a host with the same hostID results in the "cannot find host" error.

This commit fixes the issue by making map from a slice of hosts. This approach is also used in `func (s *Session) init()` https://github.com/apache/cassandra-gocql-driver/blob/trunk/session.go#L272-L275

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -722,7 +722,21 @@ func (r *ringDescriber) GetHosts() ([]*HostInfo, string, error) {
 		return r.prevHosts, r.prevPartitioner, err
 	}
 
-	hosts := append([]*HostInfo{localHost}, peerHosts...)
+	// if the same HostID exists in peerHosts, localHost needs to be excluded to work properly with Amazon Keyspaces (CASSGO-72, #1873)
+	var hasSameHostID bool
+	for _, h := range peerHosts {
+		if h.HostID() == localHost.HostID() {
+			hasSameHostID = true
+		}
+	}
+
+	var hosts []*HostInfo
+	if hasSameHostID {
+		hosts = peerHosts
+	} else {
+		hosts = append([]*HostInfo{localHost}, peerHosts...)
+	}
+
 	var partitioner string
 	if len(hosts) > 0 {
 		partitioner = hosts[0].Partitioner()


host_source.go
```diff
@@ -752,9 +752,14 @@ func refreshRing(r *ringDescriber) error {
 		return err
 	}
 
+	hostMap := make(map[string]*HostInfo, len(hosts))
+	for _, host := range hosts {
+		hostMap[host.HostID()] = host
+	}
+
```
dkropachev: I don't think this solution is straighforward, it relays on the fact that `r.GetHosts()` provides hosts in the certain order, if that order changes it stop working.
It would be better to update `GetHosts` to ignore hosts that are already in the list

dkropachev (changes requested):

host_source.go
yomipq: Thank you for reviewing. I will try to fix `GetHosts()` .

yomipq: I fixed `GetHosts()` to ignore localHost if the same Host ID exists.
Could you please confirm?

dkropachev: looks good, can you please add comment why this code is there, mentioning issue number and `Amazon Keyspaces`

yomipq: Yes, I added the comment.

jameshartig:
This needs a test.

Also, this addresses one of the issues but doesn't address the issue with the control connection querying the local table and getting back the wrong information. See https://github.com/apache/cassandra-gocql-driver/issues/1873#issuecomment-2778506348 for more details.

I was going to suggest that we do what that comment suggests and call `refreshRing` from within `setupConn` but that's going to end up causing double hosts lookups on initialization and it will not respect `DisableInitialHostLookup` if that's set.

The problem with `setupConn` is that it'll end up overwriting the host information with incorrect info since the local table has wrong information in it. It sounds like other drivers just do a full ring refresh but I don't think we can do that in this case. Should we just be ignoring some information in `addOrUpdate`? @joao-r-reis thoughts?

host_source.go
```diff
@@ -722,7 +722,21 @@ func (r *ringDescriber) GetHosts() ([]*HostInfo, string, error) {
...
+	var hosts []*HostInfo
+	if hasSameHostID {
+		hosts = peerHosts
+	} else {
+		hosts = append([]*HostInfo{localHost}, peerHosts...)
+	}
+
 	var partitioner string
 	if len(hosts) > 0 {
 		partitioner = hosts[0].Partitioner()
```
jameshartig, line 742: I think this needs to instead look at `localHost` instead because I don't think the peers table contains partitioner information.

joao-r-reis:
> The problem with setupConn is that it'll end up overwriting the host information with incorrect info since the local table has wrong information in it. It sounds like other drivers just do a full ring refresh but I don't think we can do that in this case. Should we just be ignoring some information in addOrUpdate? @joao-r-reis thoughts?

Yeah the fix is to make sure we do a full ring refresh instead of just updating the local host so that we can overwrite the local host info with the info from system.peers. Doing a full refresh even without any overwriting would also mitigate the issue because the current behavior of every host turning into 127.0.0.1 eventually after enough reconnects would no longer exist.

joao-r-reis:
The problem is that `(d *refreshDebouncer) refreshNow()` will attempt to get the control connection and if it's not up it will try to reconnect it so calling `refreshNow()` inside the control connection itself might not be ideal? This needs some more exploration to be honest. The easy solution would be to call the full refresh method bypassing the debouncer but at that point might as well remove the `refreshNow()` method of the debouncer and just use it for topology events.

yomipq:
Thank you for comments. I'm afraid but I don't understand the details for now. I will look into your comments and the codes.

joao-r-reis:
Closing this for now. I mentioned in https://github.com/apache/cassandra-gocql-driver/issues/1873#issuecomment-3436533650 that 2.0.0 might have fixed this already so we need someone that can verify whether CASSGO-72 can be closed or needs to stay open.

PR #1889, opened 2025-05-10
CASSGO-74: Closing a nil Session should not panic

tjons:
Encountered this while using the library this week and iterating through some unit test cases. 

The `Session` type methods are not nil-guarded, and will panic in a number of situations including nil-pointer dereferences. As a matter of practice, Go libraries customarily try to save panicking for errors where the calling program absolutely cannot continue normally. While some operations on the `Session` do deserve a panic (or ideally, an error), closing a nil Session is non-destructive to the state of the program and does not prevent normal operations for the user of the library.

### DIFF
--- a/session.go
+++ b/session.go
@@ -454,6 +454,10 @@ func (s *Session) Bind(stmt string, b func(q *QueryInfo) ([]interface{}, error))
 // Close closes all connections. The session is unusable after this
 // operation.
 func (s *Session) Close() {
+	// Closing a nil session is a non-event, return.
+	if s == nil {
+		return
+	}
 
 	s.sessionStateMu.Lock()
 	if s.isClosing {


session.go
```diff
@@ -454,6 +454,10 @@ func (s *Session) Bind(stmt string, b func(q *QueryInfo) ([]interface{}, error))
 // Close closes all connections. The session is unusable after this
 // operation.
 func (s *Session) Close() {
+	// Closing a nil session is a non-event, return.
+	if s == nil {
```
tolbertam, line 458: Agree that since there is nothing to do here, simply no-oping and returning is how I'd want this to behave too, and seems more pleasant than SIGSEV'ing.

I guess the question is how far this should go?  e.g. calling `session.Closed()` would still panic when maybe it would be ok to return `true` here as well?   Doing so would also prevent other methods `KeyspaceMetadata()` from panicking and instead return an error instead which could be nice (but other methods would still panic which creates an inconsistency).

But I think that changing the behavior for API methods broadly around nil receivers might be something hard to address without a major release.   For example `Query` returns `* Query`, and panics with a nil session, but i'm not sure how it should behave if we handled nil gracefully since we can't change the API to return an error without a major release.

Making this particular change around `Close()` seems fine to me before a major release.  I can't imagine anyone depending on this to panic.

Curious what others think about this and how the API should behave generally around nil receivers.

tjons: @tolbertam thanks for your feedback... I agree, I think there's a lot of opportunity to clean up this behavior in general. I had to resist the urge to boil the ocean and fix everything :sweat_smile:. Personally, I think I'd advocate for generally nil-guarding, returning an error to the user in most cases, and saving panics for when absolutely necessary. 

Unfortunately, that is a breaking signature change and as you said, challenging to handle without a major release. I'm new to the project but I've seen some discussion in slack about @joao-r-reis working on a v2 release... @joao-r-reis have you thought about potentially revisiting the API of the driver before committing to a v2 major version and the associated support commitments? There are a number of other things that might be interesting to consider in spirit of being more idomatic for Go developers, such as a more modular package structure. I'm using this client a lot and happy to write more about my experiences so far.

joao-r-reis: There's already a few API changes on `trunk` in preparation for the 2.0 release and we are not really looking to add any more work to 2.0 because we want to close it ASAP, it's been in progress for a long time and we initially wanted to do the release in April so we're already a bit delayed.

joao-r-reis: Also we don't want to do large scale API changes that will make it harder for users to upgrade, we're trying to make the least amount of changes possible

joao-r-reis: Irt to this change specifically, I'm not sure why a user calling `.Close()` on a `nil` session would be a gocql bug, should the driver guard against this in every single method? Seems a bit unreasonable

joao-r-reis: I'd be more open to this idea if the driver set the receiver to `nil` sometimes or if `nil` was a valid value in the context of the API but both of these things are not true

tjons: It's not a true "bug" in gocql but it's somewhat outside the accepted norm for Go libraries, particularly low-level ones. If you were reconsidering the API here, I'd encourage considering how to represent user-error in a way that's ergonomic to the user. I empathize with the need to ship the next release, and while I'd handle this differently I respect that the maintainers have their own priorities. 

I tried to find the middle ground approach here: if the library can fail silently and nondestructively I believe it is best to do so.

joao-r-reis: Do you have some kind of reference for this being considered the norm in Go libraries? Granted I don't have a lot of experience in the Go land but in all my time reading code from common Go libraries I've rarely seen libraries guard against `nil` receivers

tjons: Yep! https://go.dev/doc/effective_go#errors

> But what if the error is unrecoverable? Sometimes the program simply cannot continue.
> 
> For this purpose, there is a built-in function panic that in effect creates a run-time error that will stop the program (but see the next section). The function takes a single argument of arbitrary type—often a string—to be printed as the program dies. It's also a way to indicate that something impossible has happened, such as exiting an infinite loop.

And later: 
> This is only an example but real library functions should avoid panic. **If the problem can be masked or worked around, it's always better to let things continue to run rather than taking down the whole program.** One possible counterexample is during initialization: if the library truly cannot set itself up, it might be reasonable to panic, so to speak. 
(Emphasis mine)

joao-r-reis: I think that whole section is related to checks that the function is already performing and deciding whether to return an error or deliberately calling `panic()`. I don't see any reference to receiver `nil` checks.

Even in their example they don't check the receiver for `nil`:

```
func (e *PathError) Error() string {
    return e.Op + " " + e.Path + ": " + e.Err.Error()
}
```

jameshartig: This is a small change for such a large conversation but looking at the stdlib, I don't see any checks like this. 
* close a nil `http.Server`: panic
* close a nil `io.ReadCloser` from an `http.Response`: panic
* close a nil `sql.Conn`: panic

In general you should not be calling methods on nil pointers. I don't think we want to start some sort of precedent that we're going to be safe guarding callers who aren't checking for nil. This codebase is filled with pointers and methods that aren't guarding against this and I don't see the value in fixing one spot let alone spending the time to fix all of them.

jameshartig:
Going to close this. It would be helpful to share the code that made it seem reasonable call `Close` on a nil `Session`. If there's something that is returning nil that shouldn't we should fix that instead.

PR #1890, opened 2025-05-29
CASSGO-75: remove proto 1 and 2 support

### LINKED ISSUE #1888: Remove protoVersion2 support
Especially in frame.go, it is used in the data path and just 'pollutes' the code. Does anyone still need support for this ancient protocol? (and if they do, they can keep using older drivers).

dkropachev:
You can't find servers that support only proto 1 or 2. 
And there is no technical benefits to restrict proto to these versions. 
So, we can easily remove code that is taking care of these versions.

Fixes: https://github.com/apache/cassandra-gocql-driver/issues/1888


joao-r-reis:
C* 2.0 only supports v1 and v2 but I think that's ok. The java driver also removed support for these two versions. This should be done on a major release though so I think I'm going to get this in for 2.0. Can you create a JIRA? If you don't have an account I'll create the ticket

dkropachev:
@joao-r-reis , all tests are fixed, can you please restart cicd.

dkropachev:
> C* 2.0 only supports v1 and v2 but I think that's ok. The java driver also removed support for these two versions. This should be done on a major release though so I think I'm going to get this in for 2.0. Can you create a JIRA? If you don't have an account I'll create the ticket

Done: https://issues.apache.org/jira/browse/CASSGO-75

dkropachev:
Both failures due to the problem with java downloading:
```
Downloading: java 17.0.12-zulu

In progress...
9.1%curl: (92) HTTP/2 stream 1 was not closed cleanly: INTERNAL_ERROR (err 2)
```

That has nothing to do with the PR.
Btw, it signals that `sdkman` cache is not working.

joao-r-reis:
I asked if there are any concerns about this proposal in the slack channel, I'm going to wait a couple of days to see if anything pops up before reviewing and merging this.

marshal.go
```diff
@@ -113,10 +113,6 @@ type Unmarshaler interface {
 //
 // The marshal/unmarshal error provides a list of supported types when an unsupported type is attempted.
 func Marshal(info TypeInfo, value interface{}) ([]byte, error) {
-	if info.Version() < protoVersion1 {
-		panic("protocol version not set")
-	}
```
joao-r-reis, line 118: we probably want to keep this, IIUC it's checking if the field is set or left as empty

dkropachev: added back

jameshartig (approved):

jameshartig:
@dkropachev apologies, I merged another MR which caused a conflict in `writeHeader`. Just make that method use a single `append`.

joao-r-reis:
Can you add the entry to the changelog?

dkropachev:
> Can you add the entry to the changelog?

Done

joao-r-reis (approved):

PR #1891, opened 2025-05-31
CASSGO-77: Add golangci-lint integration

dkropachev:
`golangci-lint` is standard de facto in golang community.
Using will help with:

- Having code uniformity
- Enforcing certain linting rules
- Prevent bugs
- Prevent unused variables, increasing driver efficiency
- Easy and speedup reviewing proceccess by detecting problems early

 
`govet/fieldalignment` stands out from rest of the linters, it will help to define structures in a way that will make them occupy less memory and provide better productivty as result.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,33 @@
+version: "2"
+
+issues:
+  max-same-issues: 50
+
+formatters:
+  enable:
+    - goimports
+
+  settings:
+    goimports:
+      local-prefixes:
+        - github.com/gocql/gocql
+        - github.com/apache/cassandra-gocql-driver
+    golines:
+      max-len: 120
+
+linters:
+  default: none
+  enable:
+    - nolintlint
+    - govet
+  settings:
+    govet:
+      enable-all: true
+      disable:
+        - shadow
+        - fieldalignment
+
+    nolintlint:
+      allow-no-explanation: [ golines ]
+      require-explanation: true
+      require-specific: true
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ TEST_COMPRESSOR ?= no-compression
 TEST_INTEGRATION_TAGS ?= integration
 
 CCM_VERSION ?= 39b8222b31a6c7afe8fe845d16981088a5a735ad
+GOLANGCI_VERSION = v2.1.6
 JVM_EXTRA_OPTS ?= -Dcassandra.test.fail_writes_ks=test -Dcassandra.custom_query_handler_class=org.apache.cassandra.cql3.CustomPayloadMirroringQueryHandler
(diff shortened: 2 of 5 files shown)


joao-r-reis:
I don't think it's a good idea to say something like "X is standard de facto in golang community" when advocating for a change because a lot of the times when I search online to determine whether this is true or not I'll find a lot of discussion around it and not really any consensus. 

In this particular case from what I'm reading it looks like `golangci-lint` is just a way of running multiple linters so you could just add `golanci-lint` and run the `go vet` linter through `golanci-lint`. Instead what this PR is doing is adding a bunch of more linters instead of just `go vet` in addition to adding `golangci-lint` integration.

Personally I'm fine with using `golangci-lint` but I think the current list of linters that you have is way too large and as a result the diff is huge. This will cause all existing open PRs to have giant conflicts and personally I don't think it's worth the hassle. Also it will probably make it very frustrating to contribute to the driver because how harsh the linter checks are. Maybe you can try running `golangci-lint` with the default linters only and see if it's a good compromise.

dkropachev:
> I don't think it's a good idea to say something like "X is standard de facto in golang community" when advocating for a change because a lot of the times when I search online to determine whether this is true or not I'll find a lot of discussion around it and not really any consensus.
> 
> In this particular case from what I'm reading it looks like `golangci-lint` is just a way of running multiple linters so you could just add `golanci-lint` and run the `go vet` linter through `golanci-lint`. Instead what this PR is doing is adding a bunch of more linters instead of just `go vet` in addition to adding `golangci-lint` integration.
> 
> Personally I'm fine with using `golangci-lint` but I think the current list of linters that you have is way too large and as a result the diff is huge. This will cause all existing open PRs to have giant conflicts and personally I don't think it's worth the hassle. Also it will probably make it very frustrating to contribute to the driver because how harsh the linter checks are. Maybe you can try running `golangci-lint` with the default linters only and see if it's a good compromise.

What we can do here, is to replace `govet` with `golangci` with minimal changes, and later on add more linters in a separate PRs, advocating for each of them separately.

joao-r-reis:
> What we can do here, is to replace govet with golangci with minimal changes, and later on add more linters in a separate PRs, advocating for each of them separately.

Yeah that is a good approach 👍

dkropachev:
> > What we can do here, is to replace govet with golangci with minimal changes, and later on add more linters in a separate PRs, advocating for each of them separately.
> 
> Yeah that is a good approach 👍

Done, please take a look.

joao-r-reis (approved):
👍 can you squash commits and format the commit message like [this](https://github.com/apache/cassandra-gocql-driver/blob/trunk/CONTRIBUTING.md#commit-message)? Then I'll merge

PR #1892, opened 2025-06-13
CASSGO-71 Add way to create HostInfo objects for testing purposes

joao-r-reis:
Also renamed NewHostInfo to NewHostInfoFromContactPoint, this isn't a breaking change because this function was exported recently and no release has gone out since then.

What is a breaking change is making `host.SetHostID` private, I don't see why we would let users modify the host id

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -155,6 +155,8 @@ func (c cassVersion) nodeUpDelay() time.Duration {
 	return 10 * time.Second
 }
 
+// HostInfo represents a server Host/Node. You can create a HostInfo object with either NewHostInfoFromAddrPort or
+// NewTestHostInfoFromRow.
 type HostInfo struct {
 	// TODO(zariel): reduce locking maybe, not all values will change, but to ensure
 	// that we are thread safe use a mutex to access all fields.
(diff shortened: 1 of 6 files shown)


host_source.go
```diff
@@ -181,9 +183,12 @@ type HostInfo struct {
...
 }
 
-// NewHostInfo creates HostInfo with provided connectAddress and port.
+// NewHostInfoFromContactPoint creates HostInfo with provided connectAddress and port.
 // It returns an error if addr is invalid.
-func NewHostInfo(addr net.IP, port int) (*HostInfo, error) {
+//
+// If you're looking for a way to create a HostInfo object with more than just an address and port for
+// testing purposes then you can use NewTestHostInfoFromRow
+func NewHostInfoFromContactPoint(addr net.IP, port int) (*HostInfo, error) {
```
jameshartig: nit: What's the reason behind "ContactPoint"? What about "NewHostInfoFromAddrPort"? Not terribly opposed with contact point though

host_source.go
```diff
@@ -492,17 +497,42 @@ func checkSystemSchema(control *controlConn) (bool, error) {
...
+//		"rpc_address":       net.ParseIP("10.0.0.1"),
+//		"peer":              net.ParseIP("10.0.0.1"), // system.peers only
+//		"data_center":       "dc1",
+//		"rack":              "rack1",
+//		"host_id":           MustRandomUUID(),        // can also use ParseUUID("550e8400-e29b-41d4-a716-446655440000")
+//		"release_version":   "4.0.0",
+//		"native_port":       9042,
+//	}
+//	host, err := NewTestHostInfoFromRow(row)
+func NewTestHostInfoFromRow(row map[string]interface{}) (*HostInfo, error) {
```
jameshartig, line 525: Should we have the same method on `*Session` too so it could still call `AddressTranslator`?

jameshartig (approved):
Just a few minor comments

host_source.go
joao-r-reis: I checked and the only time we create host info objects from Addr and Port is during init when we are converting contact points into HostPoints, everywhere else we always use the row from the system table. I'm ok with changing to `NewHostInfoFromAddrPort` though

host_source.go
joao-r-reis: Since this is meant to be used for unit testing implementations of policies and other gocql public interfaces I don't think users will be using it with a `session` so I don't think it's necessary to involve the address translator. Maybe we can consider it if we get requests for it in the future?

host_source.go
joao-r-reis: Changed it

PR #1895, opened 2025-06-19
CASSGO-6 Accept peers with empty rack

### LINKED ISSUE #1706: Unable to discover cluster nodes with an empty rack name

jameshartig:
This fixes #1706.

### DIFF
--- a/control.go
+++ b/control.go
@@ -322,7 +322,12 @@ func (c *controlConn) setupConn(conn *Conn, sessionInit bool) error {
 	iter := conn.querySystemLocal(context.TODO())
 	host, err := c.session.hostInfoFromIter(iter, conn.host.connectAddress, conn.r.RemoteAddr().(*net.TCPAddr).Port)
 	if err != nil {
-		return err
+		// just cleanup
+		iter.Close()
+		return fmt.Errorf("could not retrieve control host info: %w", err)
+	}
+	if host == nil {
+		return errors.New("could not retrieve control host info: query returned 0 rows")
 	}
 
 	var exists bool
--- a/host_source.go
+++ b/host_source.go
@@ -171,6 +171,7 @@ type HostInfo struct {
 	port             int
 	dataCenter       string
 	rack             string
+	missingRack      bool
 	hostId           string
 	workload         string
 	graph            bool
@@ -413,8 +414,9 @@ func (h *HostInfo) update(from *HostInfo) {
 	if h.dataCenter == "" {
 		h.dataCenter = from.dataCenter
 	}
-	if h.rack == "" {
+	if h.missingRack {
 		h.rack = from.rack
+		h.missingRack = from.missingRack
 	}
 	if h.hostId == "" {
 		h.hostId = from.hostId
@@ -530,7 +532,7 @@ func newHostInfoFromRow(s *Session, defaultAddr net.IP, defaultPort int, row map
 	const assertErrorMsg = "Assertion failed for %s, type was %T"
 	var ok bool
 
-	host := &HostInfo{connectAddress: defaultAddr, port: defaultPort}
+	host := &HostInfo{connectAddress: defaultAddr, port: defaultPort, missingRack: true}
 
 	// Process all fields from the row
 	for key, value := range row {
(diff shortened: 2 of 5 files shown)


host_source.go
```diff
@@ -720,18 +770,25 @@ func (r *ringDescriber) getClusterPeerInfo(localHost *HostInfo) ([]*HostInfo, er
...
 		if err != nil {
-			return nil, err
-		} else if !isValidPeer(host) {
+			// if the error came from the iterator then return it, otherwise ignore
+			// and warn
+			if iterErr := iter.Close(); iterErr != nil {
+				return nil, fmt.Errorf("unable to fetch peer host info: %s", iterErr)
+			}
+			// skip over peers that we couldn't parse
+			r.session.logger.Warning("Failed to parse peer this host will be ignored.", newLogFieldError("err", err))
```
jameshartig, line 785: This technically changes the behavior but this seems reasonable? @joao-r-reis thoughts?

control.go
```diff
@@ -322,7 +322,11 @@ func (c *controlConn) setupConn(conn *Conn, sessionInit bool) error {
 	iter := conn.querySystemLocal(context.TODO())
 	host, err := c.session.hostInfoFromIter(iter, conn.host.connectAddress, conn.r.RemoteAddr().(*net.TCPAddr).Port)
 	if err != nil {
-		return err
+		iter.Close()
```
joao-r-reis, line 326: Should we handle the error returned by Close()?

host_source.go
```diff
@@ -413,8 +414,9 @@ func (h *HostInfo) update(from *HostInfo) {
 	if h.dataCenter == "" {
 		h.dataCenter = from.dataCenter
 	}
-	if h.rack == "" {
+	if h.missingRack || h.rack == "" {
```
joao-r-reis: Maybe we should remove the `h.rack == ""` from here? If now we treat `""` as a valid rack.

host_source.go
```diff
@@ -700,8 +747,12 @@ func (r *ringDescriber) getLocalHostInfo() (*HostInfo, error) {
 
 	host, err := r.session.hostInfoFromIter(iter, nil, r.session.cfg.Port)
 	if err != nil {
+		iter.Close()
```
joao-r-reis, line 751: Same as above, should we handle the error returned by this call?

host_source.go
joao-r-reis: Yeah seems reasonable to me

host_source.go
```diff
@@ -720,18 +770,25 @@ func (r *ringDescriber) getClusterPeerInfo(localHost *HostInfo) ([]*HostInfo, er
...
+			if iterErr := iter.Close(); iterErr != nil {
+				return nil, fmt.Errorf("unable to fetch peer host info: %s", iterErr)
+			}
+			// skip over peers that we couldn't parse
+			r.session.logger.Warning("Failed to parse peer this host will be ignored.", newLogFieldError("err", err))
+			continue
+		}
+		// if nil then none left
+		if host == nil {
+			break
```
joao-r-reis, line 790: Is this ever a scenario? I'm reading the code and it doesn't look like it. It doesn't look like `hostInfoFromMap` can ever return a nil host. It should return either `nil, err` or `host, nil` where host is never nil.

host_source.go
```diff
@@ -700,8 +747,12 @@ func (r *ringDescriber) getLocalHostInfo() (*HostInfo, error) {
 
 	host, err := r.session.hostInfoFromIter(iter, nil, r.session.cfg.Port)
 	if err != nil {
+		iter.Close()
 		return nil, fmt.Errorf("could not retrieve local host info: %w", err)
 	}
+	if host == nil {
+		return nil, errors.New("could not retrieve local host info: query returned 0 rows")
+	}
```
joao-r-reis, line 756: Same as my other comment, doesn't look like `host` can ever be `nil`

joao-r-reis (commented):
Looks good, left a few comments

control.go
jameshartig: This was mostly to just cleanup the iterator. We should probably return the error from `hostInfoFromIter`.

host_source.go
jameshartig: Yeah, I could see that, I was trying to handle filling in the rack if it was empty but since "" is valid I should remove it.

host_source.go
jameshartig: This was just to cleanup the iterator.

host_source.go
jameshartig: It can now return `nil, nil`. I documented that:
> // this will return nil, nil if there were no rows left in the Iter

joao-r-reis (approved):

joao-r-reis:
Looks good, don't forget to update commit message before merging

PR #1897, opened 2025-07-01
CASSGO-80 Update example and test code for 2.0

joao-r-reis:
Updates example and test code according to changes in 2.0.0 release (mostly deprecations).

Also adds examples for new 2.0 features.

### DIFF
--- a/example_batch_test.go
+++ b/example_batch_test.go
@@ -49,31 +49,46 @@ func Example_batch() {
 
 	ctx := context.Background()
 
-	b := session.Batch(gocql.UnloggedBatch).WithContext(ctx)
-	b.Entries = append(b.Entries, gocql.BatchEntry{
-		Stmt:       "INSERT INTO example.batches (pk, ck, description) VALUES (?, ?, ?)",
-		Args:       []interface{}{1, 2, "1.2"},
-		Idempotent: true,
-	})
-	b.Entries = append(b.Entries, gocql.BatchEntry{
-		Stmt:       "INSERT INTO example.batches (pk, ck, description) VALUES (?, ?, ?)",
-		Args:       []interface{}{1, 3, "1.3"},
-		Idempotent: true,
-	})
+	// Example 1: Simple batch using the Query() method - recommended approach
+	batch := session.Batch(gocql.LoggedBatch)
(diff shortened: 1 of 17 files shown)


example_batch_test.go
lukasz-antoniak, line 1: What do you think about removing `ProtoVersion` assignment from examples to prevent users hardcoding it in their applications?

lukasz-antoniak:
Shall we also add an example for types registration (like [here](https://github.com/apache/cassandra-gocql-driver/pull/1855))?

example_batch_test.go
joao-r-reis: Honestly I think it might be better to keep that for now until we have a chance to properly rework the protocol version negotiation to what other drivers do...

joao-r-reis:
> Shall we also add an example for types registration (like [here](https://github.com/apache/cassandra-gocql-driver/pull/1855))?

The current state of that feature is probably only useful for users that are using a different DB other than Cassandra like yugabyte. That example is for a yugabyte type that Cassandra itself doesn't support. Eventually we should add proper custom and udt type support to this API and at that point I think it will become useful to the common user and at that point I think it definitely deserves an example. As it stands I don't even have a way to verify that the jsonb example works because it requires a yugabyte db so I'd say we don't add one for now.

lukasz-antoniak:
LGTM

lukasz-antoniak (approved):

example_lwt_batch_test.go
```diff
@@ -72,7 +72,7 @@ func ExampleSession_MapExecuteBatchCAS() {
 			Args: []interface{}{"B", "pk1", "ck2", ck2Version},
 		})
 		m := make(map[string]interface{})
-		applied, iter, err := b.WithContext(ctx).MapExecCAS(m)
+		applied, iter, err := b.MapExecCAS(m)
```
worryg0d: nit: I think here should be shown an example context api as in other examples.

`applied, iter, err := b.MapExecCASContext(ctx, m)`

worryg0d (approved):
LGTM. Left a little nit.

example_lwt_batch_test.go
joao-r-reis: done

PR #1898, opened 2025-07-02
CASSGO-81 Add missing Context variant methods

joao-r-reis:
In #1868  we deprecated Query.WithContext and Batch.WithContext. We added ExecContext and IterContext as alternatives but we didn't add these alternatives for other ways to execute queries and batches like Query.Scan or Batch.ExecCAS.

We also didn't deprecate Query.Context() and we should have

### DIFF
--- a/session.go
+++ b/session.go
@@ -719,7 +719,7 @@ func (s *Session) routingKeyInfo(ctx context.Context, stmt string, keyspace stri
 // Exec executes a batch operation and returns nil if successful
 // otherwise an error is returned describing the failure.
 func (b *Batch) Exec() error {
-	iter := b.session.executeBatch(b, nil)
+	iter := b.session.executeBatch(b, b.context)
 	return iter.Close()
 }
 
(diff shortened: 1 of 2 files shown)


joao-r-reis:
~~TODO: add to changelog and rewrite commit msg~~

session.go
```diff
@@ -786,7 +786,16 @@ func (s *Session) ExecuteBatchCAS(batch *Batch, dest ...interface{}) (applied bo
 // Further scans on the interator must also remember to include
 // the applied boolean as the first argument to *Iter.Scan
 func (b *Batch) ExecCAS(dest ...interface{}) (applied bool, iter *Iter, err error) {
-	iter = b.session.executeBatch(b, nil)
+	return b.ExecCASContext(nil, dest...)
```
jameshartig: The context docs say:
> Do not pass a nil [Context](https://pkg.go.dev/context#Context), even if a function permits it. Pass [context.TODO](https://pkg.go.dev/context#TODO) if you are unsure about which Context to use.

We should either pass a TODO or pass context.Background()

joao-r-reis: yeah I thought about that but then the internal code that defaults to the old context field that is deprecated will not work...

joao-r-reis: maybe I can use context.TODO and treat context.TODO as if the user didn't provide it

jameshartig: Oh I didn't make the connection with `newQueryOptions` and it checking for nil specifically. Could we just do `q.Context()` in the query ones and `b.Context()` in the batch ones here instead off `nil`?

joao-r-reis: oh I didn't think about that, yeah I'll do that

jameshartig (approved):

PR #1899, opened 2025-07-02
CASSGO-82 CASSGO-83 fix tinyint decode bug and vector decode bug with SliceMap

joao-r-reis:
CASSGO-82

Tinyint unmarshal silently errors and unmarshals tinyint values as 0 in some situations.

This PR also removes other cases of silent errors in decode functions.

CASSGO-83

fix issue where reading vectors with SliceMap is not possible

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -432,15 +432,19 @@ func (smallIntTypeInfo) Marshal(value interface{}) ([]byte, error) {
 
 // Unmarshal unmarshals the byte slice into the value.
 func (s smallIntTypeInfo) Unmarshal(data []byte, value interface{}) error {
+	decodedData, err := decShort(data)
+	if err != nil {
+		return unmarshalErrorf("%s", err.Error())
+	}
 	if iptr, ok := value.(*interface{}); ok && iptr != nil {
 		var v int16
-		if err := unmarshalIntlike(TypeSmallInt, int64(decShort(data)), data, &v); err != nil {
+		if err := unmarshalIntlike(TypeSmallInt, int64(decodedData), data, &v); err != nil {
 			return err
 		}
 		*iptr = v
 		return nil
 	}
-	return unmarshalIntlike(TypeSmallInt, int64(decShort(data)), data, value)
+	return unmarshalIntlike(TypeSmallInt, int64(decodedData), data, value)
 }
 
 type tinyIntTypeInfo struct{}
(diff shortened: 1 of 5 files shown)


marshal.go
```diff
@@ -2352,7 +2370,7 @@ func (inetType) Type() Type {
 
 // Zero returns the zero value for the inet CQL type.
 func (inetType) Zero() interface{} {
-	return net.IP(nil)
+	return net.IP{}
```
jameshartig: What's this change for? net.IP is a slice and the zero value for a slice is nil.

vector.go
```diff
@@ -173,7 +183,7 @@ func (v VectorType) Unmarshal(data []byte, value interface{}) error {
 		}
 		return nil
 	}
-	return unmarshalErrorf("can not unmarshal %s into %T. Accepted types: slice, array.", v, value)
+	return unmarshalErrorf("can not unmarshal %s into %T. Accepted types: slice, array, *interface{}.", v, value)
```
jameshartig: this should just be `interface{}` or we should make the others be `*slice`, `*array`

marshal.go
```diff
@@ -636,26 +644,30 @@ func (intTypeInfo) Marshal(value interface{}) ([]byte, error) {
...
 
 func encInt(x int32) []byte {
 	return []byte{byte(x >> 24), byte(x >> 16), byte(x >> 8), byte(x)}
 }
 
-func decInt(x []byte) int32 {
+func decInt(x []byte) (int32, error) {
 	if len(x) != 4 {
-		return 0
+		return 0, fmt.Errorf("expected 4 bytes decoding int but got %v", len(x))
```
jameshartig, line 672: I'm confused, isn't this going to make `NULL` now return an error instead of `0` like it did before?

vector.go
```diff
@@ -143,6 +150,9 @@ func (v VectorType) Unmarshal(data []byte, value interface{}) error {
 			}
 		} else {
 			rv.Set(reflect.MakeSlice(t, v.Dimensions, v.Dimensions))
+			if rv.Kind() == reflect.Interface {
```
jameshartig, line 153: I don't think this can ever be true because you just set `rv` to a slice.

vector.go
```diff
@@ -124,6 +124,13 @@ func (v VectorType) Unmarshal(data []byte, value interface{}) error {
 	}
 	rv = rv.Elem()
 	t := rv.Type()
+	if t.Kind() == reflect.Interface {
+		if t.NumMethod() != 0 {
+			return unmarshalErrorf("can not unmarshal into non-empty interface %T", value)
+		}
+		t = reflect.TypeOf(v.Zero())
```
jameshartig, line 131: This is assuming that the interface holds that type but it might not, I think you need to instead check if it's nil and then if its not nil, then do `rv = rv.Elem() t = rv.Type()`

marshal.go
jameshartig: For MapScan/SliceMap they end up never leaving the Unmarshal method because we handle `*interface{}` and `nil` data directly in there. But I'm worried now if you Unmarshal with a `*int8` and a NULL it's going to error rather than set the value to `0`.

marshal.go
joao-r-reis: I went back and forth on this and I forgot to revert back to `nil` good catch

marshal.go
joao-r-reis: > I'm confused, isn't this going to make NULL now return an error instead of 0 like it did before?

Yeah I started seeing a couple of test failures and I noticed that, was just working on it now. Should be fixed now.

vector.go
joao-r-reis: It can be a variable of type `interface{}` that just got assigned a slice no?

marshal.go
joao-r-reis: should be fixed now

vector.go
joao-r-reis: done

jameshartig (approved):

jameshartig (approved):

PR #1901, opened 2025-07-05
CASSGO-79 Update doc.go for 2.0.0 release

joao-r-reis:
**You can preview the doc.go changes [by running `pkgsite` ](https://github.com/golang/pkgsite?tab=readme-ov-file#pkgsite-a-documentation-server)**

This patch updates the existing pkg.go.dev documentation and adds documentation for the new packages.

This patch also updates the README ands adds a new UPGRADE_GUIDE.md with some documentation for users that will be upgrading from 1.x to 2.x.

### DIFF
--- a/UPGRADE_GUIDE.md
+++ b/UPGRADE_GUIDE.md
@@ -1,0 +1,821 @@
+<!--
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+# GoCQL Major Version Upgrade Guide
+
+This guide helps you migrate between major versions of the GoCQL driver. Each major version introduces significant changes that may require code modifications.
+
+
+
+## Available Upgrade Paths
+
+- [v1.x → v2.x](#upgrading-from-v1x-to-v2x)
+- [Future version upgrades will be documented here as they become available]
+
+---
+
+## Upgrading from v1.x to v2.x
+
+Version 2.0.0 represents a major overhaul of the GoCQL driver with significant API changes, new features, and improvements. This migration requires careful planning and testing.
+
+**Important Prerequisites:**
+- **Minimum Go version**: Go 1.19+
+- **Minimum Cassandra version**: 2.1+ (recommended 4.1+ for full feature support)
+- **Supported protocol versions**: 3, 4, 5 (Cassandra 2.1+, versions 1 and 2 are no longer supported)
+
+### Table of Contents
+
+- [Protocol Version / Cassandra Version Support](#protocol-version--cassandra-version-support)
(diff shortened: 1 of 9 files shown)


UPGRADE_GUIDE.md
```diff
@@ -0,0 +1,748 @@
...
+- **Memory efficiency**: `nil` slices use less memory than empty slices
+- **Consistency**: Direct `Scan()` calls already behaved this way
+
+**Impact:** This affects code using `MapScan()` and `SliceMap()` with collection columns (lists, sets, maps) that explicitly checks for `nil` or performs operations that don't handle `nil` slices/maps safely.
+
+---
+
+### Deprecation Notices
+
+The following features are deprecated but still functional in v2.x. **These should not be used in new code.** Plan to migrate away from these before v3.x:
```
worryg0d, line 739: Nit: What do you think about adding deprecated errors from #1903 to this list for consistency?

doc.go
```diff
@@ -50,25 +54,230 @@
...
+//		WithNowInSeconds(specificTimestamp).
+//		Bind(id, data).Exec()
+//
+// These features are also available on batch operations:
+//
+//	err := session.Batch(LoggedBatch).
+//		Query("INSERT INTO table (id, data) VALUES (?, ?)", id, data).
+//		SetKeyspace("tenant1").
+//		WithNowInSeconds(specificTimestamp).
+//		Exec()
```
worryg0d: Nit: It may cause a misleading impression that batches also support a specific host targeting feature. Should we explicitly say that `SetHostID()` is not supported for batches?

`These features are also available on batch operations, excluding SetHostID()`

worryg0d (approved):
Great work! I left some little nits, but everything is good

doc.go
```diff
@@ -326,24 +559,129 @@
...
+//   - Only use when you specifically need linearizable consistency
+//   - If a SERIAL read finds an uncommitted transaction, it will commit that transaction
+//   - Most applications should use regular consistency levels (ONE, QUORUM, etc.)
+//
+// # Immutable Execution
+//
+// [Query] and [Batch] objects follow an immutable execution model that enables safe reuse and concurrent
+// execution without object mutation.
+//
+// Query and Batch Object Reusability:
```
worryg0d, line 624: Just wondering, should we mention here that the underlying `sync.Pool` for Query objects has been removed. So if the codebase relies on this fact (for example, creating many Query objects), it may experience a performance regression

UPGRADE_GUIDE.md
```diff
@@ -0,0 +1,748 @@
...
+cluster := gocql.NewCluster("127.0.0.1")
+// ProtoVersion will be auto-negotiated to the highest supported version
+// This is recommended as it works with any Cassandra 2.1+ version
+```
+
+**Note:** Since gocql v2.x requires Cassandra 2.1+ anyway, most users should use auto-negotiation for the best compatibility unless you have a cluster with nodes that have different Cassandra versions.
+
+---
+
+### Breaking Changes
```
worryg0d: As changes in CASSGO-90 (#1912) are breaking one, we should mention them in this section

doc.go
joao-r-reis: Hmm according to the tests I ran I don't think users will notice it but I can add a note to mention that the driver no longer uses sync.Pool for Query objects

UPGRADE_GUIDE.md
joao-r-reis: done

doc.go
joao-r-reis: added a section for this

doc.go
joao-r-reis: done

UPGRADE_GUIDE.md
joao-r-reis: done

doc.go
worryg0d: Cool. Thanks! Anyway, I think it is better to explicitly note that query pooling is no longer handled by the driver side, even if the impact is not significant

worryg0d (approved):
LGTM

PR #1902, opened 2025-07-15
CASSGO-86 Remove release date from new releases and add 2.0.0-rc1

joao-r-reis:
With the ASF donation we no longer have the ability to specify an exact release date due to the voting process. For this reason we will stop including a release date on the changelog.

This patch also makes it clear what changes were released in 2.0.0-rc1 so that users can know what is different between rc1 and GA.

Patch by João Reis; reviewed by TBD for CASSGO-86

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [2.0.0] - Unreleased
+## [2.0.0]
 
 ### Removed
+
+#### 2.0.0-rc1
 
 - Drop support for old CQL protocol versions: 1 and 2 (CASSGO-75)
 - Cleanup of deprecated elements (CASSGO-12)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -5,10 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [2.0.0] - Unreleased
+## [2.0.0]
```
emerkle826, line 8: :suggestion:
Would it be better to just change this to:
```suggestion
## [2.0.0-rc1]
```
for now and remove both of the `[2.0.0]` references? Add them back when the 2.0.0 release is made?

Looking at the changes now, it looks like you want all of the RC changes nested under the 2.0.0 proper release notes, as opposed to having a complete 2.0.0-rc1 section separate from the 2.0.0 section that will come later. If you want the notes nested, than you can ignore this comment and consider it :+1: from me.

emerkle826 (approved):
Approving as-is with just a suggestion on the 2.0.0-rc1 notes alignment.

CHANGELOG.md
joao-r-reis: Yeah my goal was to keep all the rc changes nested under 2.0.0, similar to what the main project does in [CHANGES.txt](https://github.com/apache/cassandra/blob/8289bb4cdee20c59add34858b817640e7f88518e/CHANGES.txt#L323-L331).

I'm concerned that if we completely separate 2.0.0-rc1 and 2.0.0, users that are only interested in GA versions will miss the actual 2.0 changelog which is mostly in the rc1 notes.

emerkle826: Makes sense. :+1:

jfleming-ic (approved):

Jacko161 (approved):

PR #1906, opened 2025-08-26
CASSGO-88 Do not set beta protocol flag when using v5

joao-r-reis:
Driver was automatically setting beta flag when using v5 protocol which causes issues when trying to connect to Cassandra versions that supported beta v5 like 3.11.

Patch by João Reis; reviewed by TBD for CASSGO-88

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1317,7 +1317,21 @@ func (c *Conn) execInternal(ctx context.Context, req frameBuilder, tracer Tracer
 		defer c.releaseStream(call)
 
 		if v := resp.framer.header.version.version(); v != c.version {
-			return nil, NewErrProtocol("unexpected protocol version in response: got %d expected %d", v, c.version)
+			errProtocol := NewErrProtocol("unexpected protocol version in response: got %d expected %d", v, c.version)
+			responseFrame, err := resp.framer.parseFrame()
+			if err != nil {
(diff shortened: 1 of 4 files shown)


conn.go
```diff
@@ -1317,7 +1317,24 @@ func (c *Conn) execInternal(ctx context.Context, req frameBuilder, tracer Tracer
...
+			errProtocol := NewErrProtocol("unexpected protocol version in response: got %d expected %d", v, c.version)
+			responseFrame, err := resp.framer.parseFrame()
+			if err != nil {
+				c.logger.Warning("Framer error while attempting to parse potential protocol error.",
+					newLogFieldError("err", err))
+				return nil, errProtocol
+			}
+			//goland:noinspection GoTypeAssertionOnErrors
+			errFrame, isErrFrame := responseFrame.(errorFrame)
+			if !isErrFrame {
```
lukasz-antoniak: NIT: How about merging two below conditions together?

lukasz-antoniak (commented):
Tested locally with 3.11.7 and connection can be established without configuring the protocol version.

conn.go
joao-r-reis: done

lukasz-antoniak (approved):

PR #1907, opened 2025-08-27
Replace golang/snappy with klauspost/compress/snappy

gberg0:
`golang/snappy` was archived recently which most security scanner tools deem EOL and a no-no. `klauspost/compress/snappy` is a [drop-in replacement](https://github.com/klauspost/compress/tree/master/snappy).


gberg0:
Nevermind. Found a replacement in https://github.com/scylladb/gocql

Issue #845, opened 2016-12-01
gocql prepared all queries


GQHero:
Hi,

It seems that gocql will prepare all the queries like CRUD.
But we have lots of random read and write, which may cause the cache soon run out.
Is there some ways to avoid using prepared query?

Thanks.

vrischmann:
[This](https://sourcegraph.com/github.com/gocql/gocql@db6f35eb602dd0770a7162e0e23e5cfb1cc05b91/-/blob/session.go#L826:17) is always called when executing a query, and looks like the behavior is hard coded.

Right now it looks like your best bet would be to modify gocql directly and change `shouldPrepare` to your liking.

Zariel:
The issue arises from the fact that the driver wont know how to serialise your types into the columns Cassandra expects without preparing, all prepare really does is cache the type information which is used when serialising the request.

You shouldn't really be thrashing the cache, if you're using queries without bind params then thats a simple fix.

rockxsj:
the java driver support raw query, we need too

martin-sucha:
@rockxsj what is your use case?

rockxsj:
> @rockxsj what is your use case?

I have a data gateway, which for accept data sink request and persistent to Cassandra.  Most sink request can use prepared statement bind with params, but some complex data type I want direct execute the statement send by request.

martin-sucha:
I've re-read this issue and it is still not clear to me when it is an issue using prepared statements. @rockxsj could you please provide a more concrete example of "some complex data type" that requires directly executing the query without preparing? Or what exactly is the issue with using the prepared statements?

OleksiienkoMykyta:
I don't see the issue using prepared queries, it's a little overhead if it's a single-use query, but in other cases, it provides more efficiency, security, and performance. Not sure that there is something to fix.

The issue reporter hasn't responded since Dec 1, 2016.
Candidate to be closed.

utking:
@OleksiienkoMykyta,

We have a Cassandra v4 cluster with 3 nodes carrying ~3TB of data, and switching from Java version with raw queries to Go version where everything is prepared has dramatically increased using (and flushing) the prep statements cache. The Go version significantly increased CPU usage (~90% across all the cores) read/write and resulted in OS LoadAverage from 10 to 30, while with the Java version it was 2-5 for the same workloads (switching back to the Java version, confirmed that the LoadAverage goes down immediately).
[...]

PR #1908, opened 2025-09-01
feat: replace custom LRU cache with otter cache library

rhymf14:
Replace internal LRU cache implementation for prepared statements and routing key info with the more efficient otter cache library.

🤖 Generated with [Claude Code](https://claude.ai/code)


worryg0d:
Hi @rhymf14. 

I had a quick look to understand the changes better, and here’s a short breakdown:

1. I see that it replaces gocql’s built-in LRU cache with the external caching library [maypok86/otter](maypok86/otterhttps://github.com/maypok86/otter). The motivation behind this change isn’t entirely clear—could you share the main goal? Adding some benchmarks to compare both implementations would also be good.
2. There’s no GitHub issue or Jira ticket linked. Having one would be great for tracking purposes.
3. This patch bumps the required Go version to 1.24.6, which is a big jump from the current 1.19 used on the trunk branch.
4. It looks like the patch doesn’t include recent changes from trunk, such as CASSGO-1 and CASSGO-22. Since they overlap with this implementation, it may be worth reconciling them.

joao-r-reis:
Related: https://github.com/apache/cassandra-gocql-driver/issues/845

joao-r-reis:
We really want to avoid adding 3rd party dependencies unless you can provide benchmark data that shows this would significantly improve the performance of the driver

PR #1910, opened 2025-09-23
Small performance optimization in lz4 compressor implementation

joao-r-reis:
Ran some benchmarks to compare 2.0.0-rc1 with 1.7.0 and noticed a 2-3% degradation when compression was enabled. I made these quick optimizations, re-ran the tests and the degradation went away (0-1%).

Benchmark code [here](https://github.com/joao-r-reis/gocql-benchmarks) (`first` branch, will eventually merge it)

Results [here](https://docs.google.com/spreadsheets/d/18VB_W5H5kInuyerwWdK_3Z4cwZ4H3n353IbJDoEkLJk/edit?usp=sharing) in the "compression fix test" tab.

### DIFF
--- a/lz4/lz4.go
+++ b/lz4/lz4.go
@@ -53,16 +53,17 @@ func (s LZ4Compressor) AppendCompressedWithLength(dst, src []byte) ([]byte, erro
 	maxLength := lz4.CompressBlockBound(len(src))
 	oldDstLen := len(dst)
 	dst = grow(dst, maxLength+dataLengthSize)
+	slicedDst := sliceDest(dst, oldDstLen)
 
 	var compressor lz4.Compressor
-	n, err := compressor.CompressBlock(src, dst[oldDstLen+dataLengthSize:])
+	n, err := compressor.CompressBlock(src, slicedDst[dataLengthSize:])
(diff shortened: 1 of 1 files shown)


worryg0d (approved):
Honestly, I didn't expect the **copy** function call on a zero-length source to cost that much... But well, it is still a call, so it makes sense. I assume the same applies to the zero value slicing thing...

Also, this is why I don't like to make such big patches with a lot of different areas involved. It takes a lot of time to identify such things.

Anyway, thanks @joao-r-reis! You've done a really great job with this one. If it were me, I would spend much more time trying to identify this. On the other hand, this is a good lesson for me

joao-r-reis:
Yeah honestly I microbenchmarked the `grow` function alone and the `copy` call didn't seem to cause much impact if any so I'm not 100% confident that this change actually helped, the 1-2% difference I saw was probably just normal variance but I'm rerunning tests and will try to profile later today to see if I understand exactly why compression looks to be slower now

joao-r-reis:
Ok after some profiling I decided to go with a different approach to mitigate the performance regression that we are seeing. I'm not convinced the changes on this PR even helped that much anyway so I'm closing this in favor of #1912 (and repurposed the JIRA ticket)

PR #1912, opened 2025-09-24
CASSGO-90 Don't collect host metrics if a query/batch observer is not provided

joao-r-reis:
~~In 2.0.0-rc1 we got rid of the query object pooling and now the extra allocs associated with the code that is collecting QueryMetrics is 1 of the factors that are leading to a small performance regression.~~

~~Since the majority of the users don't use Query Metrics, it makes more sense to make these opt-in.~~

~~Benchmarks [here](https://docs.google.com/spreadsheets/d/18VB_W5H5kInuyerwWdK_3Z4cwZ4H3n353IbJDoEkLJk/edit?usp=sharing) (this branch vs v1.7.0).  With this change, the results are much more acceptable. I still think the v5 implementation has room for improvement but this is the first version of our v5 implementation and we can iterate on it in follow up releases. The most important part is that now v4 (in 1.7.0) vs v4 (in 2.0 - this branch) looks ok.~~

Don't collect host metrics if a query/batch observer is not provided, this is a small optimization.

### DIFF
--- a/session.go
+++ b/session.go
@@ -389,7 +389,7 @@ func (s *Session) AwaitSchemaAgreement(ctx context.Context) error {
 		return errNoControl
 	}
 	return s.control.withConn(func(conn *Conn) *Iter {
-		return newErrIter(conn.awaitSchemaAgreement(ctx), newQueryMetrics(), "", nil, nil)
+		return newErrIter(conn.awaitSchemaAgreement(ctx), &queryMetrics{}, "", nil, nil)
 	}).err
 }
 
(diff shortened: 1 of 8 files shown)


session.go
```diff
@@ -944,14 +949,26 @@ func (qm *queryMetrics) attempt(addAttempts int, addLatency time.Duration,
...
-		hostMetricsCopy = new(hostMetrics)
-		*hostMetricsCopy = *updateHostMetrics
-	}
-
 	qm.l.Unlock()
-	return totalAttempts, hostMetricsCopy
+	return totalAttempts, updateHostMetrics
+}
+
+var nilHostMetrics = &hostMetrics{}
```
worryg0d: I think the `nil` prefix is misleading, because value of this var is not actually nil. Probably the `empty` prefix is better?

The same applies to `nilQueryMetricsImpl`

session.go
```diff
@@ -1500,11 +1530,17 @@ func (iter *Iter) Columns() []ColumnInfo {
 }
 
 // Attempts returns the number of times the statement was executed.
+//
+// Set Query.WithQueryMetrics to true
+// so that this Attempts metric is available, otherwise this method will always return 0.
```
worryg0d: From what I understand, attempts are always available, so this comment is not true

worryg0d (approved):
It's a good idea to make query metrics collectors optional. If someone doesn't care about them or has their own mechanisms to collect such metrics, the driver should allow disabling them. Also, it reduces the amount of mem allocations when we don't really need them.

I left some comments regarding naming and godoc.



Now it makes more sense to me than a `copy` call on a zero-length source to impact performance that much. Ofc it has some place, but believing that this causes 10% performance degradation is really hard

session.go
joao-r-reis: Yeah my first implementation of this was a truly nil object but I realized later that `attempts` is used for retry policies so can't get rid of that one. Also I called it `nil` because of the [Null Object Pattern](https://en.wikipedia.org/wiki/Null_object_pattern) but I'm not even sure if using this name in the context of Go is a good idea so I'll just rename it to `empty` anyway.

session.go
joao-r-reis: Good catch

joao-r-reis:
After discussing with @jameshartig we decided to go with another approach. Instead of doing a breaking change and adding a new API we'll just stop collecting host metrics (therefore removing the map allocation) when an observer is not provided.

jameshartig (approved):
LGTM

worryg0d (approved):
LGTM

PR #1913, opened 2025-10-10
CASSGO-91 Prefer connection address over system local addresses

joao-r-reis:
There is an unintended change in 2.0-rc1 causing the driver to prefer system table addresses over the connection address which breaks deployments that rely on this behavior from 1.x.

This patch fixes this and keeps the behavior the same as it was in 1.x.

### DIFF
--- a/host_source.go
+++ b/host_source.go
@@ -267,6 +267,13 @@ func (h *HostInfo) ConnectAddress() net.IP {
 	return addr
 }
 
+// actualConnectAddress can be used to access the connectAddress field with the lock (mu).
+func (h *HostInfo) actualConnectAddress() net.IP {
+	h.mu.RLock()
+	defer h.mu.RUnlock()
+	return h.connectAddress
+}
+
 func (h *HostInfo) BroadcastAddress() net.IP {
 	h.mu.RLock()
 	defer h.mu.RUnlock()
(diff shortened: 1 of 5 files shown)


host_source.go
```diff
@@ -680,16 +680,7 @@ func newHostInfoFromRow(s *Session, defaultAddr net.IP, defaultPort int, row map
...
-	if validIpAddr(host.rpcAddress) {
-		host.connectAddress = host.rpcAddress
-	} else if validIpAddr(host.preferredIP) {
-		host.connectAddress = host.preferredIP
-	} else if validIpAddr(host.broadcastAddress) {
-		host.connectAddress = host.broadcastAddress
-	} else if validIpAddr(host.peer) {
-		host.connectAddress = host.peer
-	}
+	host.connectAddress, _ = host.connectAddressLocked()
```
jameshartig, line 693: This means it'll never change from `defaultAddr` and that address is the initial connect address. That seems wrong because now every host will have the same address and you can never connect to another one.

jameshartig: Actually it seems like we almost always pass in `nil` for connectAddress in `hostInfoFromIter` so how is this working?

joao-r-reis: I don't think this is the case, the initial connect address is retrieved from the current connection OR `nil`, if it's nil then this will make it so that the IP comes from system tables because `connectAddressLocked()` goes down a list of 4 ip addresses until it finds a valid one

joao-r-reis:
Pushed a change [to this line here](https://github.com/apache/cassandra-gocql-driver/pull/1913/files#diff-b3a920dda70b3d7b468688c05344b329a3395949e2c50c9203fef3b92485da4dR739), this might fix #1873

host_source.go
joao-r-reis: For peers we set the argument as `nil` so the connect address here will be nil and `connectAddressLocked` will look at the other fields

host_source.go
```diff
@@ -745,7 +736,7 @@ func (r *ringDescriber) getLocalHostInfo() (*HostInfo, error) {
 		return nil, errNoControl
 	}
 
-	host, err := r.session.hostInfoFromIter(iter, nil, r.session.cfg.Port)
+	host, err := r.session.hostInfoFromIter(iter, iter.host.connectAddress, r.session.cfg.Port)
```
jameshartig: Doesn't this need to lock mu? Do we need a new function that's like `ActualConnectAddress` that does the locking?

joao-r-reis: oh you're right I think, I should probably use `host.ConnectAddress()` instead

joao-r-reis: actually `ConnectAddress()` will do all the logic to falback to other columns so I probably need to add a new method

joao-r-reis: done

jameshartig (approved):
Looks good but I'd like to see some more comments around the changed lines so we understand the nuances of why they're doing something specific because it's not obvious when looking at the code in places.

joao-r-reis:
> Looks good but I'd like to see some more comments around the changed lines so we understand the nuances of why they're doing something specific because it's not obvious when looking at the code in places.

good point, done

PR #1915, opened 2025-10-23
CASSGO-92: Add a public method to retrieve StatementMetadata and LogField methods

jameshartig:
StatementMetadata can be called on a session to get the bind, result, and pk information for a given query. Previously this wasn't publicly exposed but is necessary for some implementations of HostSelectionPolicy like token-aware. This might also be useful for CI tooling or runtime analysis of queries and the types of columns.

NewLogField* are methods to to return a LogField with name and a specific type.

Finally, session init was cleaned up to prevent a HostSelectionPolicy from causing a panic if it tried to make a query during init. The interface was documented that queries should not be attempted.

Patch by James Hartig for CASSGO-92

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.1.0]
+
+### Added
+
+- Session.StatementMetadata (CASSGO-92)
+- NewLogFieldIP, NewLogFieldError, NewLogFieldStringer, NewLogFieldString, NewLogFieldInt, NewLogFieldBool (CASSGO-92)
+
+### Fixed
+
+- Prevent panic with queries during session init (CASSGO-92)
+
 ## [2.0.0]
 
 ### Removed
--- a/logger.go
+++ b/logger.go
@@ -48,7 +48,7 @@ func logHelper(logger StructuredLogger, level LogLevel, msg string, fields ...Lo
 	case LogLevelError:
 		logger.Error(msg, fields...)
 	default:
-		logger.Error("Unknown log level", newLogFieldInt("level", int(level)), newLogFieldString("msg", msg))
+		logger.Error("Unknown log level", NewLogFieldInt("level", int(level)), NewLogFieldString("msg", msg))
 	}
 }
 
@@ -229,7 +229,8 @@ func newLogField(name string, value LogFieldValue) LogField {
 	}
 }
 
-func newLogFieldIp(name string, value net.IP) LogField {
+// NewLogFieldIP creates a new LogField with the given name and net.IP.
+func NewLogFieldIP(name string, value net.IP) LogField {
 	var str string
 	if value == nil {
 		str = "<nil>"
@@ -239,7 +240,8 @@ func newLogFieldIp(name string, value net.IP) LogField {
 	return newLogField(name, logFieldValueString(str))
 }
 
-func newLogFieldError(name string, value error) LogField {
+// NewLogFieldError creates a new LogField with the given name and error.
+func NewLogFieldError(name string, value error) LogField {
 	var str string
 	if value != nil {
 		str = value.Error()
@@ -247,7 +249,8 @@ func newLogFieldError(name string, value error) LogField {
 	return newLogField(name, logFieldValueString(str))
 }
 
-func newLogFieldStringer(name string, value fmt.Stringer) LogField {
+// NewLogFieldStringer creates a new LogField with the given name and fmt.Stringer.
+func NewLogFieldStringer(name string, value fmt.Stringer) LogField {
 	var str string
 	if value != nil {
 		str = value.String()
@@ -255,15 +258,18 @@ func newLogFieldStringer(name string, value fmt.Stringer) LogField {
 	return newLogField(name, logFieldValueString(str))
 }
 
-func newLogFieldString(name string, value string) LogField {
+// NewLogFieldString creates a new LogField with the given name and string.
+func NewLogFieldString(name string, value string) LogField {
 	return newLogField(name, logFieldValueString(value))
 }
 
-func newLogFieldInt(name string, value int) LogField {
+// NewLogFieldInt creates a new LogField with the given name and int.
+func NewLogFieldInt(name string, value int) LogField {
 	return newLogField(name, logFieldValueInt64(int64(value)))
 }
 
-func newLogFieldBool(name string, value bool) LogField {
+// NewLogFieldBool creates a new LogField with the given name and bool.
+func NewLogFieldBool(name string, value bool) LogField {
 	return newLogField(name, logFieldValueBool(value))
 }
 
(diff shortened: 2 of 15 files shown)


logger.go
```diff
@@ -222,6 +222,13 @@ type LogField struct {
 	Value LogFieldValue
 }
 
+// NewLogField creates a new LogField with the given name and value. Currently
+// supports the following types: string, int, bool, fmt.Stringer, error, net.IP.
+// If the type is not supported, it'll use the fmt package to print the value.
+func NewLogField(name string, value interface{}) LogField {
```
joao-r-reis: I believe the boxing will result in extra allocs, We should just make the private methods public (newLogField, newLogFieldIp, etc.)

joao-r-reis: We could add a newLogFieldAny method but it should have a warning in the docs to say that it isn't as efficient as using the specific type method variants

session.go
```diff
@@ -199,6 +186,20 @@ func NewSession(cfg ClusterConfig) (*Session, error) {
 	}
 	s.connCfg = connCfg
 
+	if cfg.PoolConfig.HostSelectionPolicy == nil {
+		cfg.PoolConfig.HostSelectionPolicy = RoundRobinHostPolicy()
+	}
+	s.pool = cfg.PoolConfig.buildPool(s)
+	s.policy = cfg.PoolConfig.HostSelectionPolicy
+
+	// set the executor here in case the policy needs to execute queries in Init
```
joao-r-reis, line 195: Allowing queries to be executed before the session is even initialized is so confusing to me, the session doesn't even have a control connection until it is initialized. I understand the motivation behind letting policies query something in the DB to initialize but I feel like if such use case does exist it should be better to add a new method to the policy called `OnSessionInit` or something that will be called by the session after its initialization has completed. However adding a new method to the interface is a breaking change so we'd have to add a new interface and type check if the policy implements the new interface

joao-r-reis: I feel like this use case is so niche that it would be better to tell those users to implement a policy with "two states" and the user is responsible for changing the state of that policy when they create a session

CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
```
joao-r-reis: I think we can set this to 2.1.0

session.go
```diff
@@ -652,114 +653,113 @@ func (s *Session) routingKeyInfo(ctx context.Context, stmt string, keyspace stri
...
-	keyspaceMetadata, inflight.err = s.KeyspaceMetadata(info.request.columns[0].Keyspace)
-	if inflight.err != nil {
-		// don't cache this error
-		s.routingKeyInfoCache.Remove(stmt)
-		return nil, inflight.err
+	// ResultColumns are the columns that are returned by the statement.
+	ResultColumns []ColumnInfo
+}
+
+// StatementMetadata returns metadata for a statement. If keyspace is empty,
```
joao-r-reis, line 692: the docs here should mention that the statement will be prepared here. Also if a user uses this method it won't use the cache at all which is a bit odd

joao-r-reis:
the description of CASSGO-92 has a few more things than were included here in this PR, maybe we should make CASSGO-92 an Epic and create a JIRA for what's included here and possibly other JIRAs to address the remaining points of CASSGO-92. Or maybe keep CASSGO-92 as is and just create sub tasks?

session.go
jameshartig: Its not obvious right now because it's called `Init` and it's passed a session but it doesn't say the session isn't ready or isn't usable. If the policy is ready there isn't a reason why the policy should have assumed the session isn't.

I'm not arguing that we change things but we should at least document the intention of init and the state of the session.

session.go
jameshartig: I went back and forth on this. I documented in the commit message the reasoning but essentially I didn't want to impose a cache that the user couldn't invalidate. It seems simple enough for the user to do their own caching if they need to.

I'm not opposed to adding a cache here but then we'd need a way to invalidate it.

logger.go
jameshartig: I'll expose the rest of them. I wasn't sure if we wanted to commit to those but I think they make sense.

jameshartig:
> the description of [CASSGO-92](https://issues.apache.org/jira/browse/CASSGO-92) has a few more things than were included here in this PR, maybe we should make [CASSGO-92](https://issues.apache.org/jira/browse/CASSGO-92) an Epic and create a JIRA for what's included here and possibly other JIRAs to address the remaining points of [CASSGO-92](https://issues.apache.org/jira/browse/CASSGO-92). Or maybe keep [CASSGO-92](https://issues.apache.org/jira/browse/CASSGO-92) as is and just create sub tasks?

I wasn't necessarily intending on it being realistic to fix all of the things I brought up especially the ones with workarounds. Could I just make another issue for the `Context()` method on `ExecutableStatement`? Not sure if it's worth breaking the interface but it might be? We can decide separately but as-is it's not insurmountable.

session.go
joao-r-reis: yeah honestly it doesn't make a lot of sense that we're passing the Session in the Init method because it's not clear exactly which methods of the Session type API can actually be used by the policy at that point. I think the best we can do at this point is to document this in the the Init method docs because any other way of addressing this probably requires a breaking change

session.go
joao-r-reis: Ok we can keep it like this, we can always discuss caching later if anyone requests it

joao-r-reis:
> I wasn't necessarily intending on it being realistic to fix all of the things I brought up especially the ones with workarounds. 

That makes sense, I'd say we rename the issue to `Add a public method to retrieve StatementMetadata` or something like that to make it more specific as to what this PR is doing but you can keep the description that talks about the usability of the policy because that's good context that we can refer to later.

> Could I just make another issue for the `Context()` method on `ExecutableStatement`? Not sure if it's worth breaking the interface but it might be? We can decide separately but as-is it's not insurmountable.

Yeah go for it. I'm also not sure about whether it's worth breaking the interface, I think I'd rather document how users can achieve that by type casting the object returned by `Statement()`. I added the `Statement()` method so that users could keep the functionality of type casting to `Query` or `Batch` which might be useful at times.

I do think we should create JIRAs for these things that require breaking changes and tag them with an appropriate label so that we can address them when we work on a 3.0. I'd say creating one to review the HostSelectionPolicy API (especially the Init method) and one to review `ExecutableStatement`.

jameshartig:
@joao-r-reis updated the JIRA issue and updated this PR according to the comments.

session.go
jameshartig: I updated the interface to clarify `Init`. That said, I want to keep the changes I made since they're small and prevent the panic. In a host policy I was debugging, it starts a background goroutine to refresh the partitions and if that runs too soon it could still cause a panic without these changes.

joao-r-reis: 👍 I'm actually curious to know what happens now with these changes, there is an executor but there is no hosts or connection pools, the ideal fix would be for the policy to get notified somehow that the session has finished initializing. Ideally the driver would do that on its own but without this I guess the user creating the policy has to write that code on the app side

In any case we can keep these changes sure

joao-r-reis (approved):
👍

PR #1916, opened 2025-10-29
CASSGO-95: Fix Iter.RowData Values to point to concrete types

jameshartig:
There were multiple issues with the previous RowData method. First, it was returning invalid values because it called c.Zero() instead of Zero on column. Second, it would've returned an interface to a pointer to an interface which is frustrating to work with.

Now it restores the previous v1 behavior and returns pointers to the underlying types wrapped in an interface.

Added tests to ensure the behavior is as expected.

Patch by James Hartig for CASSGO-95


joao-r-reis (approved):
good catch

PR #1920, opened 2025-11-28
CASSGO-97 Protocol version negotiation doesn't work if server replies with stream id different than 0

worryg0d:
Protocol version negotiation doesn't work if server replies with stream id different than 0

Previously, protocol negotiation didn't work properly when C* was responding with stream id different from 0.
 
This patch changes the way protocol negotiation works. Instead of parsing a supported protocol version from C* error response, the driver tries to connect with each supported protocol starting from the latest.

Patch by Bohdan Siryk; Reviewed by <> for CASSGO-98

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -378,6 +378,13 @@ func (s *startupCoordinator) setupConn(ctx context.Context) error {
 	select {
 	case err := <-startupErr:
 		if err != nil {
+			if s.checkProtocolRelatedError(err) {
+				return &unsupportedProtocolVersionError{
+					err:      err,
+					hostInfo: s.conn.host,
+					version:  protoVersion(s.conn.version),
+				}
+			}
 			return err
 		}
 	case <-ctx.Done():
@@ -385,6 +392,38 @@ func (s *startupCoordinator) setupConn(ctx context.Context) error {
 	}
 
 	return nil
+}
+
+// Checks if the error is protocol related and should be retried during startup.
+// It returns the frame that caused the error and whether the error should be retried.
+func (s *startupCoordinator) checkProtocolRelatedError(err error) bool {
+	var unwrappedFrame frame
+
+	var protocolErr *protocolError
+	if !errors.As(err, &protocolErr) {
+		var errFrame errorFrame
+		if !errors.As(err, &errFrame) {
+			return false
+		} else {
+			unwrappedFrame = errFrame
+		}
+	} else {
+		unwrappedFrame = protocolErr.frame
+	}
+
+	switch frame := unwrappedFrame.(type) {
+	case *supportedFrame:
+		// We can receive a supportedFrame wrapped in protocolError from Conn.recv if the host responds to a 0 stream id.
+		// If we receive a supportedFrame then we know that the host is not compatible with the protocol version, but it is reachable, so we can retry
+		return true
+	case errorFrame:
+		// If we receive an errorFrame with codes ErrCodeProtocol or ErrCodeServer,
+		// then we should try to downgrade a protocol version, so do not skip the host
+		return frame.code == ErrCodeProtocol || frame.code == ErrCodeServer
+	default:
+		// In any other case we should not retry as it means the host is not reachable or some other error happened
+		return false
+	}
 }
 
 func (s *startupCoordinator) write(ctx context.Context, frame frameBuilder, startupCompleted *atomic.Bool) (frame, error) {
@@ -408,12 +447,14 @@ func (s *startupCoordinator) options(ctx context.Context, startupCompleted *atom
 		return err
 	}
 
-	supported, ok := frame.(*supportedFrame)
-	if !ok {
-		return NewErrProtocol("Unknown type of response to startup frame: %T", frame)
+	switch frame := frame.(type) {
+	case *supportedFrame:
+		return s.startup(ctx, frame.supported, startupCompleted)
+	case error:
+		return frame
+	default:
+		return NewErrProtocol("Unknown type of response to startup frame: %T (frame=%s)", frame, frame.String())
 	}
-
-	return s.startup(ctx, supported.supported, startupCompleted)
 }
 
 func (s *startupCoordinator) startup(ctx context.Context, supported map[string][]string, startupCompleted *atomic.Bool) error {
--- a/control.go
+++ b/control.go
@@ -32,7 +32,6 @@ import (
 	"math/rand"
 	"net"
 	"os"
-	"regexp"
 	"strconv"
 	"sync"
 	"sync/atomic"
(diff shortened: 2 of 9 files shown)


worryg0d:
WIP: currently working on tests for the whole protocol negotiation logic....

worryg0d:
Ready for review

control.go
```diff
@@ -262,26 +217,27 @@ func (c *controlConn) discoverProtocol(hosts []*HostInfo) (int, error) {
...
+			}
 
-		if err == nil {
-			c.session.logger.Debug("Discovered protocol version using host.",
-				NewLogFieldInt("protocol_version", connCfg.ProtoVersion), NewLogFieldIP("host_addr", host.ConnectAddress()), NewLogFieldString("host_id", host.HostID()))
-			return connCfg.ProtoVersion, nil
-		}
+			if err == nil {
+				c.session.logger.Debug("Discovered protocol version using host.",
+					NewLogFieldInt("protocol_version", connCfg.ProtoVersion), NewLogFieldIP("host_addr", host.ConnectAddress()))
```
joao-r-reis: any particular reason you removed the host_id log field from both calls? Is it because it's not populated?

protocol_negotiation_test.go
```diff
@@ -0,0 +1,179 @@
...
+	if r.supportsBetaProtocol(reqFrame.header.version) && !r.hasBetaFlag(reqFrame.header) {
+		respFrame.writeHeader(0, opError, reqFrame.header.stream)
+		respFrame.writeInt(ErrCodeProtocol)
+		respFrame.writeString(r.createBetaFlagUnsetProtocolErrorMessage(reqFrame.header.version))
+		return nil
+	}
+
+	// if a client uses an unsupported protocol version, we respond with an error
+	if !r.supportsProtocol(reqFrame.header.version) {
+		respFrame.writeHeader(0, opError, reqFrame.header.stream)
```
joao-r-reis: can we have a test case that replies with the stream id of the request in addition to this one? This behavior is not defined in the spec and I think some C* versions might behave differently

control.go
```diff
@@ -262,26 +217,27 @@ func (c *controlConn) discoverProtocol(hosts []*HostInfo) (int, error) {
...
+				return connCfg.ProtoVersion, nil
+			}
 
-		if proto := parseProtocolFromError(err); proto > 0 {
-			c.session.logger.Debug("Discovered protocol version using host after parsing protocol error.",
-				NewLogFieldInt("protocol_version", proto), NewLogFieldIP("host_addr", host.ConnectAddress()), NewLogFieldString("host_id", host.HostID()))
-			return proto, nil
+			c.session.logger.Debug("Failed to connect to the host using protocol version.",
```
joao-r-reis: We should only attempt to reconnect to the same host with a lower protocol version if the error is assumed to be related to unsupported protocol version: 
- t's an error to the first request (OPTIONS)
- the error type is PROTOCOL_ERROR or SERVER_ERROR - SERVER_ERROR is for old C* versions that reported this error as SERVER_ERROR instead of PROTOCOL_ERROR

I propose we use a custom internal error type (unsupportedProtocolVersionErr for example) just for this case that we can return inside the dial method and we test for it here in discoverProtocol(). Reference code [in the java driver here](https://github.com/apache/cassandra-java-driver/blob/62eade21bfeb16a12ce71013fbaebf1c19b5ae96/core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java#L238-L259).

[Technically the java driver does check for the error string](https://github.com/apache/cassandra-java-driver/blob/62eade21bfeb16a12ce71013fbaebf1c19b5ae96/core/src/main/java/com/datastax/oss/driver/internal/core/channel/ProtocolInitHandler.java#L327-L334) which I didn't know but I still believe we shouldn't do that, checking the above 2 conditions is enough.

protocol_negotiation_test.go
```diff
@@ -0,0 +1,179 @@
+//go:build all || unit
+// +build all unit
+
+package gocql
```
joao-r-reis, line 22: add license header, we should add a check for this in CI at some point

protocol_negotiation_test.go
```diff
@@ -0,0 +1,179 @@
...
+		respFrame.writeInt(0) // <flags>
+		respFrame.writeInt(0) // <columns_count>
+		// <rows_count>
+		respFrame.writeInt(0)
+	}
+
+	return nil
+}
+
+func TestProtocolNegotiation(t *testing.T) {
```
joao-r-reis, line 171: Very good job with this test, I feel like it was a weak spot in our test suite and now it's much better. Can you add `-v` to the `go test` command in the GH workflow step that runs the unit tests so we can double check which tests are running?

joao-r-reis (commented):
Looking good

control.go
worryg0d: Yes, it is not populated, so there is really no useful information

protocol_negotiation_test.go
worryg0d: Could you please clarify what you exactly mean? The test server already replies with the stream id of the request

control.go
worryg0d: That makes sense, we don't really have to re-try each time if the error is not protocol-related

protocol_negotiation_test.go
worryg0d: Good catch!

protocol_negotiation_test.go
worryg0d: Thanks. I'll add verbose for unit tests to ci

worryg0d:
Thanks, @joao-r-reis, for your feedback. I'm going to address this next week

protocol_negotiation_test.go
joao-r-reis: Oh I'm sorry I thought it was replying with stream id 0, in that case then can you add a test case where the response stream id is 0 and the request stream id is not 0?

worryg0d:
Hi @joao-r-reis

Sorry for the late reply — I’ve been a bit tied up with work lately.

I’ve now addressed your comments, so please take a look whenever you have a moment.

Thanks.

protocol_negotiation_test.go
worryg0d: Added test cases for 0 stream id

conn.go
```diff
@@ -659,6 +700,8 @@ func (c *Conn) processFrame(ctx context.Context, r io.Reader) error {
 		return err
 	}
 
+	c.logger.Debug("Received frame header", NewLogFieldString("header", head.String()))
```
joao-r-reis: Is this `processFrame` function called on every frame or just handshake frames? Running this log call on every frame post handshake seems a bit unnecessary no?

logger.go
```diff
@@ -162,7 +162,8 @@ func (l *defaultLogger) Debug(msg string, fields ...LogField) {
 func (l *defaultLogger) write(prefix string, msg string, fields []LogField) {
 	buf := bytes.Buffer{}
 	writeLogMsg(&buf, prefix, msg, fields)
-	log.Println(buf.String())
+	// During testing it allows seeing where exactly logging methods are called instead of this method
+	log.Output(3, buf.String())
```
joao-r-reis: This is used in production code as well, I'd prefer to keep this behavior unchanged for users

joao-r-reis (commented):
Looks good, left 2 comments only

conn.go
worryg0d: Nice catch. I needed this when I was implementing it, but it is not meant to be left here. Will remove it

logger.go
worryg0d: Oh, I assumed that this for testing purpose only... Well if this is the case I'm going to revert this change. But it would actually useful to see the exact caller...

joao-r-reis: Yeah I agree that the caller info is very useful and I'd +1 that if it was only used in tests

conn.go
worryg0d: Removed

logger.go
worryg0d: Reverted this change

worryg0d:
Addressed review comments, squashed commits, and added @joao-r-reis as a reviewer to the commit message

joao-r-reis (approved):
👍

worryg0d:
Thanks for merging it

PR #1923, opened 2026-01-12
CASSGO-106: shuffle replicas for token-aware-policy

tejaslodayadd:
Description:

- Keep the non-shuffling default and emit a warning when the shuffle decision isn't explicit.
- AddDoNotShuffleReplicas, and ShuffledTokenAwareHostPolicy to toggle replica shuffling.
- Improves replica load distribution while preserving correctness for LWTs.

Why shuffling:
Shuffling replicas results in a proper load balancing, almost in all cases, except LWT, which is voided even if shuffling is enabled.

This fix is ported over from scylla/gocql: 
https://github.com/scylladb/gocql/issues/480
https://github.com/scylladb/gocql/pull/485

PR #1924, opened 2026-01-12
CASSGO-103: Add GetSerialConsistency method to Query and Batch

tejaslodayadd:
Closes https://issues.apache.org/jira/browse/CASSGO-103

### DIFF
--- a/session.go
+++ b/session.go
@@ -1272,6 +1272,15 @@ func (q *Query) SerialConsistency(cons Consistency) *Query {
 	return q
 }
 
+// GetSerialConsistency returns the currently configured serial consistency level
+// for the query. If the serial consistency level has not been set, it returns nil.
+func (q *Query) GetSerialConsistency() *Consistency {
+	if !q.serialCons.isSerial() {
+		return nil
+	}
+	return &q.serialCons
+}
+
 // PageState sets the paging state for the query to resume paging from a specific
 // point in time. Setting this will disable to query paging for this query, and
 // must be used for all subsequent pages.
@@ -2077,6 +2086,15 @@ func (b *Batch) SerialConsistency(cons Consistency) *Batch {
 	}
 	b.serialCons = cons
 	return b
+}
+
+// GetSerialConsistency returns the currently configured serial consistency level
+// for the batch. If the serial consistency level has not been set, it returns nil.
+func (b *Batch) GetSerialConsistency() *Consistency {
+	if !b.serialCons.isSerial() {
+		return nil
+	}
+	return &b.serialCons
 }
 
 // DefaultTimestamp will enable the with default timestamp flag on the query.
--- a/session_test.go
+++ b/session_test.go
@@ -108,6 +108,15 @@ func TestQueryBasicAPI(t *testing.T) {
 		t.Fatalf("expected Query.GetConsistency to return 'All', got '%s'", qry.GetConsistency())
 	}
 
+	if sc := qry.GetSerialConsistency(); sc != nil {
+		t.Fatalf("expected Query.GetSerialConsistency to return nil when not set, got '%s'", sc)
+	}
+
+	qry.SerialConsistency(Serial)
+	if sc := qry.GetSerialConsistency(); sc == nil || *sc != Serial {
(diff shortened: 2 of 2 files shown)


millerjp (approved):

session.go
```diff
@@ -2079,6 +2085,12 @@ func (b *Batch) SerialConsistency(cons Consistency) *Batch {
 	return b
 }
 
+// GetSerialConsistency returns the currently configured serial consistency level for
+// the batch.
+func (b *Batch) GetSerialConsistency() Consistency {
+	return b.serialCons
```
joao-r-reis: If serial consistency was never set then this will actually be an invalid value (ANY) and if a user tries to use this value to set the serial consistency of another query the driver will panic. We can return *Consistency so that we can return nil in case it wasn't set and document it on the function's godoc

session.go
```diff
@@ -1272,6 +1272,12 @@ func (q *Query) SerialConsistency(cons Consistency) *Query {
 	return q
 }
 
+// GetSerialConsistency returns the currently configured serial consistency level for
+// the query.
+func (q *Query) GetSerialConsistency() Consistency {
+	return q.serialCons
+}
```
tejaslodayadd: If serial consistency was never set, `serialCons` will have its zero value which is `ANY` — an invalid serial consistency level. If a user then takes this return value and tries to set it on another query via `SerialConsistency()`, the driver will panic (since `SerialConsistency` panics on values other than `Serial` or `LocalSerial`).

Consider returning `*Consistency` instead so we can return `nil` when serial consistency was never explicitly set, and document this behavior in the godoc. Something like:

```go
// GetSerialConsistency returns the currently configured serial consistency level for
// the query. It returns nil if no serial consistency has been set.
func (q *Query) GetSerialConsistency() *Consistency {
	if q.serialCons == 0 {
		return nil
	}
	return &q.serialCons
}
```

Same applies to the `Batch.GetSerialConsistency()` below.

tejaslodayadd (commented):
The `GetSerialConsistency` methods on both `Query` and `Batch` should return `*Consistency` instead of `Consistency` to handle the case where serial consistency was never explicitly set. Currently, the zero value of `Consistency` is `ANY`, which is not a valid serial consistency level and will cause a panic if passed to `SerialConsistency()`.

session.go
tejaslodayadd: done! thanks.

joao-r-reis:
Sorry I know I suggested the `*Consistency` change but I should have considered a `(Consistency, bool)` alternative which might be better. I've started a thread on Slack to see if I can get more opinions from others on this before merging. It might seem like a small change and that I might be a bit pedantic but since it's a public API addition we can't break it once it's merged so I want to make sure we get it right.

joao-r-reis:
I'll take over this and make the change but I'll keep you as an author on the commit. We want to release v2.1.0 soon and I want to get this and #1923 in

joao-r-reis:
https://github.com/apache/cassandra-gocql-driver/pull/1934

session.go
```diff
@@ -1272,6 +1272,15 @@ func (q *Query) SerialConsistency(cons Consistency) *Query {
 	return q
 }
 
+// GetSerialConsistency returns the currently configured serial consistency level
+// for the query. If the serial consistency level has not been set, it returns nil.
+func (q *Query) GetSerialConsistency() *Consistency {
```
worryg0d, line 1277: Hmm, instead of returning a pointer maybe you should try follow go-ideomatic approach with bool value which indicates if value is presented, wdyt? I'm unsure which is better, but this one is common in go, so...

```go
func (q *Query) GetSerialConsistency() (Consistency, bool)
```

joao-r-reis: Yeah I made that change in #1934

joao-r-reis: This PR will be closed, let's discuss #1934 instead

worryg0d: Ah, sorry, miss-opened this one

joao-r-reis:
Made the changes in #1934 and merged in https://github.com/apache/cassandra-gocql-driver/commit/343007bf5fb06156e723bf17af45663de679fae3

PR #1926, opened 2026-01-16
CASSGO-104 CASSGO-107: TokenAwareHostPolicy should populate replica maps for non-default keyspaces

### LINKED ISSUE #1925: TokenAwareHostPolicy ignores replica shuffling for non-default keyspaces
This is a follow-up for the issue reported in #1621.

PR #1714 fixes the problem only partially. The keyspace is read from the prepared statement, but the replica map contains information only for the default keyspace configured for the session. As a result, the code at [policies.go:L615](https://github.com/apache/cassandra-gocql-driver/blob/f1e31a58f7e0c25e58e2e2a0a0c6de358e643e8b/policies.go#L615) returns nil, causing the query plan to fall back to primary-replica selection.

Impact: requests are routed only to the primary replica, even when replica shuffling is enabled, leading to reduced load distribution.

raj14243:
When a session is created with one keyspace but queries are executed against a different keyspace (using keyspace.table syntax), the TokenAwareHostPolicy routes queries only to the primary replica and ignores ShuffleReplicas() setting because replica maps are only populated for the session's default keyspace.

This causes hotspotting and reduced load distribution across the cluster for multi-keyspace workloads.

This change implements lazy loading of replica maps in the Pick() method. When replicas are not found for a keyspace, the ensureReplicasForKeyspace() method is called to populate them on-demand using the session's existing KeyspaceMetadata capability.

Follow the discussion: [ASF](https://the-asf.slack.com/archives/C05LPRVNZV1/p1768514415652969)
patch by Raj Ummadisetty for CASSGO-104

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Session.StatementMetadata (CASSGO-92)
 - NewLogFieldIP, NewLogFieldError, NewLogFieldStringer, NewLogFieldString, NewLogFieldInt, NewLogFieldBool (CASSGO-92)
+- Introduced configurable schema metadata caching modes to control what metadata is cached (CASSGO-107)
 
 ### Fixed
 
@@ -20,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Use protocol downgrading approach during protocol negotiation (CASSGO-97)
 - Prevent panic iin compileMetadata() when final func is not defined for an aggregate (CASSGO-105)
 - Framer drops error silently (CASSGO-108)
+- TokenAwareHostPolicy now populates replica maps for non-default keyspaces (CASSGO-104)
 
 ## [2.0.0]
 
--- a/cassandra_test.go
+++ b/cassandra_test.go
@@ -2230,12 +2230,21 @@ func TestEmptyTimestamp(t *testing.T) {
 	}
 }
 
-// Integration test of just querying for data from the system.schema_keyspace table where the keyspace DOES exist.
+// Integration test of querying for data from the system.schema_keyspace table for single and all the keyspaces.
 func TestGetKeyspaceMetadata(t *testing.T) {
 	session := createSession(t)
 	defer session.Close()
+	t.Run("SingleKeyspace", func(t *testing.T) {
+		keyspaceMetadata, err := getKeyspaceMetadata(session, "gocql_test")
+		assertGetKeyspaceMetadata(t, keyspaceMetadata, err)
+	})
+	t.Run("AllKeyspaces", func(t *testing.T) {
+		keyspacesMetadata, err := getAllKeyspaceMetadata(session)
+		assertGetKeyspaceMetadata(t, keyspacesMetadata["gocql_test"], err)
+	})
+}
 
-	keyspaceMetadata, err := getKeyspaceMetadata(session, "gocql_test")
+func assertGetKeyspaceMetadata(t *testing.T, keyspaceMetadata *KeyspaceMetadata, err error) {
 	if err != nil {
 		t.Fatalf("failed to query the keyspace metadata with err: %v", err)
 	}
@@ -2276,16 +2285,25 @@ func TestGetKeyspaceMetadataFails(t *testing.T) {
 	}
 }
 
-// Integration test of just querying for data from the system.schema_columnfamilies table
-func TestGetTableMetadata(t *testing.T) {
+// Integration test of querying for table data from the system.schema_columnfamilies table for single keyspace and all keyspaces
+func TestGetAllTableMetadata(t *testing.T) {
 	session := createSession(t)
 	defer session.Close()
-
 	if err := createTable(session, "CREATE TABLE gocql_test.test_table_metadata (first_id int, second_id int, third_id int, PRIMARY KEY (first_id, second_id))"); err != nil {
 		t.Fatalf("failed to create table with error '%v'", err)
 	}
+	t.Run("SingleKeyspace", func(t *testing.T) {
+		tables, err := getTableMetadata(session, "gocql_test")
+		assertGetTableMetadata(t, session, tables, err)
+	})
+	t.Run("AllKeyspaces", func(t *testing.T) {
+		tables, err := getAllTablesMetadata(session)
+		assertGetTableMetadata(t, session, tables["gocql_test"], err)
+	})
+}
 
-	tables, err := getTableMetadata(session, "gocql_test")
+func assertGetTableMetadata(t *testing.T, session *Session, tables []TableMetadata, err error) {
+
 	if err != nil {
 		t.Fatalf("failed to query the table metadata with err: %v", err)
 	}
@@ -2360,7 +2378,18 @@ func TestGetColumnMetadata(t *testing.T) {
 		t.Fatalf("failed to create index with err: %v", err)
 	}
 
-	columns, err := getColumnMetadata(session, "gocql_test")
+	t.Run("SingleKeyspace", func(t *testing.T) {
+		columns, err := getColumnMetadata(session, "gocql_test")
+		assertGetColumnMetadata(t, session, columns, err)
+	})
+
+	t.Run("AllKeyspaces", func(t *testing.T) {
+		columns, err := getAllColumnMetadata(session)
+		assertGetColumnMetadata(t, session, columns["gocql_test"], err)
+	})
+}
+
+func assertGetColumnMetadata(t *testing.T, session *Session, columns []ColumnMetadata, err error) {
 	if err != nil {
 		t.Fatalf("failed to query column metadata with err: %v", err)
 	}
@@ -2457,8 +2486,17 @@ func TestMaterializedViewMetadata(t *testing.T) {
 	session := createSession(t)
 	defer session.Close()
 	createMaterializedViews(t, session)
+	t.Run("SingleKeyspace", func(t *testing.T) {
+		materializedViews, err := getMaterializedViewsMetadata(session, "gocql_test")
+		assertMaterializedViewMetadata(t, materializedViews, err)
+	})
+	t.Run("AllKeyspaces", func(t *testing.T) {
+		materializedViews, err := getAllMaterializedViewsMetadata(session)
+		assertMaterializedViewMetadata(t, materializedViews["gocql_test"], err)
+	})
+}
 
-	materializedViews, err := getMaterializedViewsMetadata(session, "gocql_test")
+func assertMaterializedViewMetadata(t *testing.T, materializedViews []MaterializedViewMetadata, err error) {
 	if err != nil {
 		t.Fatalf("failed to query view metadata with err: %v", err)
 	}
@@ -2542,8 +2580,17 @@ func TestAggregateMetadata(t *testing.T) {
 	session := createSession(t)
 	defer session.Close()
 	createAggregate(t, session)
+	t.Run("SingleKeyspace", func(t *testing.T) {
+		aggregates, err := getAggregatesMetadata(session, "gocql_test")
+		assertAggregateMetadata(t, aggregates, err)
+	})
+	t.Run("AllKeyspaces", func(t *testing.T) {
+		aggregates, err := getAllAggregatesMetadata(session)
+		assertAggregateMetadata(t, aggregates["gocql_test"], err)
+	})
+}
 
-	aggregates, err := getAggregatesMetadata(session, "gocql_test")
+func assertAggregateMetadata(t *testing.T, aggregates []AggregateMetadata, err error) {
 	if err != nil {
 		t.Fatalf("failed to query aggregate metadata with err: %v", err)
 	}
@@ -2590,8 +2637,18 @@ func TestFunctionMetadata(t *testing.T) {
 	session := createSession(t)
 	defer session.Close()
 	createFunctions(t, session)
+	t.Run("SingleKeyspace", func(t *testing.T) {
+		functions, err := getFunctionsMetadata(session, "gocql_test")
+		assertFunctionMetadata(t, functions, err)
+	})
+	t.Run("AllKeyspaces", func(t *testing.T) {
+		functions, err := getAllFunctionsMetadata(session)
+		assertFunctionMetadata(t, functions["gocql_test"], err)
+	})
 
-	functions, err := getFunctionsMetadata(session, "gocql_test")
+}
+
+func assertFunctionMetadata(t *testing.T, functions []FunctionMetadata, err error) {
 	if err != nil {
 		t.Fatalf("failed to query function metadata with err: %v", err)
 	}
@@ -2661,8 +2718,31 @@ func TestFunctionMetadata(t *testing.T) {
 	}
 }
 
-// Integration test of querying and composition the keyspace metadata
+// Integration test of querying keyspace metadata with different MetadataCacheMode settings
 func TestKeyspaceMetadata(t *testing.T) {
+	testCases := []struct {
+		name      string
+		cacheMode MetadataCacheMode
+		// When true, expect full metadata (tables, aggregates, views, types)
+		// When false, only expect keyspace-level metadata
+		expectFullMetadata bool
+	}{
+		{
+			name:               "Full",
+			cacheMode:          Full,
+			expectFullMetadata: true,
+		},
+		{
+			name:               "KeyspaceOnly",
+			cacheMode:          KeyspaceOnly,
+			expectFullMetadata: false,
+		},
+		{
+			name:               "Disabled",
+			cacheMode:          Disabled,
+			expectFullMetadata: true,
+		},
+	}
 	session := createSession(t)
 	defer session.Close()
 
@@ -2677,119 +2757,156 @@ func TestKeyspaceMetadata(t *testing.T) {
 		t.Fatalf("failed to create index with err: %v", err)
 	}
 
-	keyspaceMetadata, err := session.KeyspaceMetadata("gocql_test")
-	if err != nil {
-		t.Fatalf("failed to query keyspace metadata with err: %v", err)
-	}
-	if keyspaceMetadata == nil {
-		t.Fatal("expected the keyspace metadata to not be nil, but it was nil")
-	}
-	if keyspaceMetadata.Name != session.cfg.Keyspace {
-		t.Fatalf("Expected the keyspace name to be %s but was %s", session.cfg.Keyspace, keyspaceMetadata.Name)
-	}
-	if len(keyspaceMetadata.Tables) == 0 {
-		t.Errorf("Expected tables but there were none")
-	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			session := createSession(t, func(config *ClusterConfig) {
+				config.MetadataCacheMode = tc.cacheMode
+			})
+			defer session.Close()
+			// Query keyspace metadata
+			keyspaceMetadata, err := session.KeyspaceMetadata("gocql_test")
+			if err != nil {
+				t.Fatalf("failed to query keyspace metadata with err: %v", err)
+			}
+			if keyspaceMetadata == nil {
+				t.Fatal("expected the keyspace metadata to not be nil, but it was nil")
+			}
+			if keyspaceMetadata.Name != session.cfg.Keyspace {
+				t.Fatalf("Expected the keyspace name to be %s but was %s", session.cfg.Keyspace, keyspaceMetadata.Name)
+			}
+			// When cache mode is Disabled, verify that the cache is empty
+			if tc.cacheMode == Disabled {
+				cachedMeta := session.schemaDescriber.getSchemaMetaForRead()
+				if cachedMeta != nil && len(cachedMeta.keyspaceMeta) > 0 {
+					t.Errorf("Expected empty cache in Disabled mode, but found %d keyspaces cached", len(cachedMeta.keyspaceMeta))
+				}
+			}
 
-	tableMetadata, found := keyspaceMetadata.Tables["test_metadata"]
-	if !found {
-		t.Fatalf("failed to find the test_metadata table metadata")
-	}
+			if tc.expectFullMetadata {
+				if len(keyspaceMetadata.Tables) == 0 {
+					t.Errorf("Expected tables but there were none")
+				}
 
-	if len(tableMetadata.PartitionKey) != 1 {
-		t.Errorf("expected partition key length of 1, but was %d", len(tableMetadata.PartitionKey))
-	}
-	for i, column := range tableMetadata.PartitionKey {
-		if column == nil {
-			t.Errorf("partition key column metadata at index %d was nil", i)
-		}
-	}
-	if tableMetadata.PartitionKey[0].Name != "first_id" {
-		t.Errorf("Expected the first partition key column to be 'first_id' but was '%s'", tableMetadata.PartitionKey[0].Name)
-	}
-	if len(tableMetadata.ClusteringColumns) != 1 {
-		t.Fatalf("expected clustering columns length of 1, but was %d", len(tableMetadata.ClusteringColumns))
-	}
-	for i, column := range tableMetadata.ClusteringColumns {
-		if column == nil {
-			t.Fatalf("clustering column metadata at index %d was nil", i)
-		}
-	}
-	if tableMetadata.ClusteringColumns[0].Name != "second_id" {
-		t.Errorf("Expected the first clustering column to be 'second_id' but was '%s'", tableMetadata.ClusteringColumns[0].Name)
-	}
-	thirdColumn, found := tableMetadata.Columns["third_id"]
-	if !found {
-		t.Fatalf("Expected a column definition for 'third_id'")
-	}
-	if !session.useSystemSchema && thirdColumn.Index.Name != "index_metadata" {
-		// TODO(zariel): scan index info from system_schema
-		t.Errorf("Expected column index named 'index_metadata' but was '%s'", thirdColumn.Index.Name)
-	}
+				tableMetadata, found := keyspaceMetadata.Tables["test_metadata"]
+				if !found {
+					t.Fatalf("failed to find the test_metadata table metadata")
+				}
 
-	aggregate, found := keyspaceMetadata.Aggregates["average"]
-	if !found {
-		t.Fatal("failed to find the aggregate 'average' in metadata")
-	}
-	if aggregate.FinalFunc.Name != "avgfinal" {
-		t.Fatalf("expected final function %s, but got %s", "avgFinal", aggregate.FinalFunc.Name)
-	}
-	if aggregate.StateFunc.Name != "avgstate" {
-		t.Fatalf("expected state function %s, but got %s", "avgstate", aggregate.StateFunc.Name)
-	}
-	aggregate, found = keyspaceMetadata.Aggregates["average2"]
-	if !found {
-		t.Fatal("failed to find the aggregate 'average2' in metadata")
-	}
-	if aggregate.FinalFunc.Name != "avgfinal" {
-		t.Fatalf("expected final function %s, but got %s", "avgFinal", aggregate.FinalFunc.Name)
-	}
-	if aggregate.StateFunc.Name != "avgstate" {
-		t.Fatalf("expected state function %s, but got %s", "avgstate", aggregate.StateFunc.Name)
-	}
-	_, found = keyspaceMetadata.UserTypes["basicview"]
-	if !found {
-		t.Fatal("failed to find the types in metadata")
-	}
-	textType := TypeText
-	if flagCassVersion.Before(3, 0, 0) {
-		textType = TypeVarchar
-	}
-	expectedType := UserTypeMetadata{
-		Keyspace:   "gocql_test",
-		Name:       "basicview",
-		FieldNames: []string{"birthday", "nationality", "weight", "height"},
-		FieldTypes: []TypeInfo{
-			timestampTypeInfo{},
-			varcharLikeTypeInfo{
-				typ: textType,
-			},
-			varcharLikeTypeInfo{
-				typ: textType,
-			},
-			varcharLikeTypeInfo{
-				typ: textType,
-			},
-		},
-	}
-	if !reflect.DeepEqual(*keyspaceMetadata.UserTypes["basicview"], expectedType) {
-		t.Fatalf("type is %#v, but expected %#v", keyspaceMetadata.UserTypes["basicview"], expectedType)
-	}
-	if flagCassVersion.Major >= 3 {
-		materializedView, found := keyspaceMetadata.MaterializedViews["view_view"]
-		if !found {
-			t.Fatal("failed to find materialized view view_view in metadata")
-		}
-		if materializedView.BaseTable.Name != "view_table" {
-			t.Fatalf("expected name: %s, materialized view base table name: %s", "view_table", materializedView.BaseTable.Name)
-		}
-		materializedView, found = keyspaceMetadata.MaterializedViews["view_view2"]
-		if !found {
-			t.Fatal("failed to find materialized view view_view2 in metadata")
-		}
-		if materializedView.BaseTable.Name != "view_table2" {
-			t.Fatalf("expected name: %s, materialized view base table name: %s", "view_table2", materializedView.BaseTable.Name)
-		}
+				if len(tableMetadata.PartitionKey) != 1 {
+					t.Errorf("expected partition key length of 1, but was %d", len(tableMetadata.PartitionKey))
+				}
+				for i, column := range tableMetadata.PartitionKey {
+					if column == nil {
+						t.Errorf("partition key column metadata at index %d was nil", i)
+					}
+				}
+				if tableMetadata.PartitionKey[0].Name != "first_id" {
+					t.Errorf("Expected the first partition key column to be 'first_id' but was '%s'", tableMetadata.PartitionKey[0].Name)
+				}
+				if len(tableMetadata.ClusteringColumns) != 1 {
+					t.Fatalf("expected clustering columns length of 1, but was %d", len(tableMetadata.ClusteringColumns))
+				}
+				for i, column := range tableMetadata.ClusteringColumns {
+					if column == nil {
+						t.Fatalf("clustering column metadata at index %d was nil", i)
+					}
+				}
+				if tableMetadata.ClusteringColumns[0].Name != "second_id" {
+					t.Errorf("Expected the first clustering column to be 'second_id' but was '%s'", tableMetadata.ClusteringColumns[0].Name)
+				}
+				thirdColumn, found := tableMetadata.Columns["third_id"]
+				if !found {
+					t.Fatalf("Expected a column definition for 'third_id'")
+				}
+				if !session.useSystemSchema && thirdColumn.Index.Name != "index_metadata" {
+					// TODO(zariel): scan index info from system_schema
+					t.Errorf("Expected column index named 'index_metadata' but was '%s'", thirdColumn.Index.Name)
+				}
+
+				aggregate, found := keyspaceMetadata.Aggregates["average"]
+				if !found {
+					t.Fatal("failed to find the aggregate 'average' in metadata")
+				}
+				if aggregate.FinalFunc.Name != "avgfinal" {
+					t.Fatalf("expected final function %s, but got %s", "avgFinal", aggregate.FinalFunc.Name)
+				}
+				if aggregate.StateFunc.Name != "avgstate" {
+					t.Fatalf("expected state function %s, but got %s", "avgstate", aggregate.StateFunc.Name)
+				}
+				aggregate, found = keyspaceMetadata.Aggregates["average2"]
+				if !found {
+					t.Fatal("failed to find the aggregate 'average2' in metadata")
+				}
+				if aggregate.FinalFunc.Name != "avgfinal" {
+					t.Fatalf("expected final function %s, but got %s", "avgFinal", aggregate.FinalFunc.Name)
+				}
+				if aggregate.StateFunc.Name != "avgstate" {
+					t.Fatalf("expected state function %s, but got %s", "avgstate", aggregate.StateFunc.Name)
+				}
+				_, found = keyspaceMetadata.UserTypes["basicview"]
+				if !found {
+					t.Fatal("failed to find the types in metadata")
+				}
+				textType := TypeText
+				if flagCassVersion.Before(3, 0, 0) {
+					textType = TypeVarchar
+				}
+				expectedType := UserTypeMetadata{
+					Keyspace:   "gocql_test",
+					Name:       "basicview",
+					FieldNames: []string{"birthday", "nationality", "weight", "height"},
+					FieldTypes: []TypeInfo{
+						timestampTypeInfo{},
+						varcharLikeTypeInfo{
+							typ: textType,
+						},
+						varcharLikeTypeInfo{
+							typ: textType,
+						},
+						varcharLikeTypeInfo{
+							typ: textType,
+						},
+					},
+				}
+				if !reflect.DeepEqual(*keyspaceMetadata.UserTypes["basicview"], expectedType) {
+					t.Fatalf("type is %#v, but expected %#v", keyspaceMetadata.UserTypes["basicview"], expectedType)
+				}
+				if flagCassVersion.Major >= 3 {
+					materializedView, found := keyspaceMetadata.MaterializedViews["view_view"]
+					if !found {
+						t.Fatal("failed to find materialized view view_view in metadata")
+					}
+					if materializedView.BaseTable.Name != "view_table" {
+						t.Fatalf("expected name: %s, materialized view base table name: %s", "view_table", materializedView.BaseTable.Name)
+					}
+					materializedView, found = keyspaceMetadata.MaterializedViews["view_view2"]
+					if !found {
+						t.Fatal("failed to find materialized view view_view2 in metadata")
+					}
+					if materializedView.BaseTable.Name != "view_table2" {
+						t.Fatalf("expected name: %s, materialized view base table name: %s", "view_table2", materializedView.BaseTable.Name)
+					}
+				}
+			} else {
+				// KeyspaceOnly mode should only return keyspace metadata
+				// Tables, Functions, Aggregates, MaterializedViews, UserTypes should be nil
+				if keyspaceMetadata.Tables != nil {
+					t.Errorf("Expected no tables in KeyspaceOnly mode, but got %d tables", len(keyspaceMetadata.Tables))
+				}
+				if keyspaceMetadata.Aggregates != nil {
+					t.Errorf("Expected no aggregates in KeyspaceOnly mode, but got %d aggregates", len(keyspaceMetadata.Aggregates))
+				}
+				if keyspaceMetadata.Functions != nil {
+					t.Errorf("Expected no functions in KeyspaceOnly mode, but got %d functions", len(keyspaceMetadata.Functions))
+				}
+				if keyspaceMetadata.UserTypes != nil {
+					t.Errorf("Expected no user types in KeyspaceOnly mode, but got %d types", len(keyspaceMetadata.UserTypes))
+				}
+				if keyspaceMetadata.MaterializedViews != nil {
+					t.Errorf("Expected no materialized views in KeyspaceOnly mode, but got %d views", len(keyspaceMetadata.MaterializedViews))
+				}
+			}
+		})
 	}
 }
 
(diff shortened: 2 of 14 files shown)


joao-r-reis:
Most if not all other drivers have a tokenmap structure in the driver itself that can be leveraged by the tokenaware policy. This tokenmap has the replica metadata for every keyspace (found via a system table query) and it is computed every time there is a control connection reconnection, a topology refresh or a schema changed event. 

In GoCQL this logic seems to be built into the tokenaware policy itself which is fine but it looks like it is only computing replicas for one keyspace which is far from ideal. This patch does fix the issue as described but it adds a round trip for the first query of each keyspace and at a first glance it seems like if a lot of queries go out at the same time for a keyspace that doesn't have replicas yet then all of those queries will trigger these extra round trips.

I believe the correct solution would be to fetch the list of keyspaces and update the replicas for all keyspaces where currently this is being done just for a single keyspace instead of trying to compute replicas on the fly when a request has already been submitted to the driver.

References:

https://github.com/apache/cassandra-java-driver/blob/16260261d3df50fcf24fac1fc2d37896c4a111bf/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/DefaultTokenMap.java#L200

https://github.com/apache/cassandra-java-driver/blob/35bae4f72dea82bb0001d97e7ab1981a6ff68edd/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultMetadata.java#L104

metadata.go
```diff
@@ -519,6 +551,39 @@ func getKeyspaceMetadata(session *Session, keyspaceName string) (*KeyspaceMetada
 	return keyspace, nil
 }
 
+// getAllKeyspaceNames queries system tables for all keyspace names.
+func getAllKeyspaceNames(session *Session) ([]string, error) {
```
worryg0d: Basically, only the query statements are different in both branches, so it is possible to simplify the whole function to:
```go
var queryStmt string 
if session.useSystemSchema { // Cassandra 3.x+
	queryStmt = `SELECT keyspace_name FROM system_schema.keyspaces`
} else { // Cassandra 2.x
	queryStmt = `SELECT keyspace_name FROM system.schema_keyspaces`
}

iter := session.control.query(stmt)
var keyspaces []string
var keyspaceName string
for iter.Scan(&keyspaceName) {
	keyspaces = append(keyspaces, keyspaceName)
}

	return nil, fmt.Errorf("error querying keyspace names: %v", err)
}

...
```

Also, I'm unsure if we have to support Cassandra < 3, as we don't even have it in our CI. I don't say you should remove this, we can leave this for now to be consistent with other schema related functions

metadata.go
```diff
@@ -278,6 +278,38 @@ func (s *schemaDescriber) getSchema(keyspaceName string) (*KeyspaceMetadata, err
...
 
+// returns schema metadata of all the cached keyspaces
+func (s *schemaDescriber) getSchemas() ([]*KeyspaceMetadata, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	schemas := make([]*KeyspaceMetadata, 0, len(s.cache))
+	for _, metadata := range s.cache {
+		schemas = append(schemas, metadata)
+	}
+	return schemas, nil
```
worryg0d: error return value seems to be redundant; it is never returned anything except `nil.`

session.go
```diff
@@ -598,6 +606,23 @@ func (s *Session) KeyspaceMetadata(keyspace string) (*KeyspaceMetadata, error) {
...
 
+// GetAllKeyspacesMetadata returns metadata for all keyspaces in the cluster.
+// Uses cached metadata where available.
+func (s *Session) GetAllKeyspacesMetadata() ([]*KeyspaceMetadata, error) {
+	if s.Closed() {
+		return nil, ErrSessionClosed
+	}
+	return s.schemaDescriber.getSchemas()
+}
+
```
worryg0d: Looks like `shemaDescriber.getSchemas()` method doesn't query C* to retrieve schema information, but returns only the cached one, so the comment `Uses cached metadata where available` is not entirely true.

I think we should update cached metadata when `GetAllKeyspacesMetadata()` is called, as it will return outdated schema metadata to the token-aware policy itself.

raj14243: > so the comment Uses cached metadata where available is not entirely true.

I will update it to "Uses cached metadata"

> I think we should update cached metadata when GetAllKeyspacesMetadata() is called, as it will return outdated schema metadata to the token-aware policy itself.

The current implementation intentionally reads only from the cache because schema updates are expected to be applied via [handleSchemaEvent](https://github.com/apache/cassandra-gocql-driver/blob/f1e31a58f7e0c25e58e2e2a0a0c6de358e643e8b/events.go#L135) . I noticed we currently refresh the cache only for the schemaChangeKeyspace event; I’ll change this so the cache is refreshed for all relevant schema events. That should keep the schema metadata up to date for token-aware routing. Let me know if I missed any cases.

events.go
```diff
@@ -133,27 +133,42 @@ func (s *Session) handleEvent(framer *framer) {
 }
 
 func (s *Session) handleSchemaEvent(frames []frame) {
```
worryg0d: I'm unsure if this is the right way to handle this. My understanding is that `schemaChangeKeyspace` event is sent by Cassandra only in case when some DDL is running for a keyspace. From the protocol doc:

https://cassandra.apache.org/doc/latest/cassandra/reference/native-protocol.html

```
An event pushed by the server. A client will only receive events for the
  types it has REGISTERed to. The body of an EVENT message will start with a
  [string] representing the event type. The rest of the message depends on the
  event type. The valid event types are:
...
 "SCHEMA_CHANGE": events related to schema change. After the event type,
      the rest of the message will be <change_type><target><options> where:
        - <change_type> is a [string] representing the type of changed involved.
          It will be one of "CREATED", "UPDATED" or "DROPPED".
        - <target> is a [string] that can be one of "KEYSPACE", "TABLE", "TYPE",
          "FUNCTION" or "AGGREGATE" and describes what has been modified
          ("TYPE" stands for modifications related to user types, "FUNCTION"
          for modifications related to user defined functions, "AGGREGATE"
          for modifications related to user defined aggregates).
        - <options> depends on the preceding <target>:
          - If <target> is "KEYSPACE", then <options> will be a single [string]
            representing the keyspace changed.
          - If <target> is "TABLE" or "TYPE", then
            <options> will be 2 [string]: the first one will be the keyspace
            containing the affected object, and the second one will be the name
            of said affected object (either the table, user type, function, or
            aggregate name).
          - If <target> is "FUNCTION" or "AGGREGATE", multiple arguments follow:
            - [string] keyspace containing the user defined function / aggregate
            - [string] the function/aggregate name
            - [string list] one string for each argument type (as CQL type)
```

So, basically we're loosing informating about the **target**. Let's say we dropped a table in a keyspace, then Cassandra will send an event to the clients about that change. The problem here is that on each change, we call `policy.KeyspacesChanged()`, which consists of two fields:

```
type KeyspaceUpdateEvent struct {
	Keyspace string
	Change   string
}
```

So, we lost a target of the change, and this actually changes the behavior of the API - previously gocql was notified on DDL run for individual keyspace, and the current implementation notifies on each DDL runs including tables, aggregates, etc.

Also, I don't think that just adding `Target` field to the `KeyspaceUpdateEvent` is a good idea, because there could be custom implementations of `HostSelectionPolicy` which could rely on old behavior.

raj14243: Good catch! thanks for sharing the protocol. I had assumed `change` would capture the target as well, but it only captures the action. With this new information I see two approaches:

1. In `handleSchemaEvent`, refresh the metadata for the keyspaces that were cleared.
2. Add an explicit callback to the `HostSelectionPolicy` for targets other than keyspace.

I'll go with option 1 for now, since refreshing the cache where it’s removed feels more logical. I'm open to other suggestions.

worryg0d: I think we can go with the first option. The second one would require us to break api or to introduce internal hidden interfaces with type assertions just for this case, which I would prefer to avoid.

events.go
```diff
@@ -133,40 +133,40 @@ func (s *Session) handleEvent(framer *framer) {
...
-	type keyspaceChange struct {
-		keyspace string
-		change   string
+
+	if len(frames) > 0 {
+		s.control.awaitSchemaAgreement()
 	}
```
worryg0d: nit: Just thinking if it really makes sense to return from this if len(frame) < 1, so we prevent allocation of a keyspacesToRefresh map. I'm unsure if go compiler optimizes this

worryg0d (approved):
To me, everything looks good 👍, but I would ask the community / other contributors to review this to ensure we didn't miss something

metadata.go
joao-r-reis: Hmm honestly we shouldn't break support of C* 2.x if we don't have a good reason because C* 2.2 still supports v3 and v4 which the driver supports but it is true that since we don't have it on our CI it's hard to tell if we even still support it right now. We removed v2 protocol support because to find a C* version that didn't support at least v3 we had to go back to 2.0.x which is very very old but 2.1.x and 2.2.x support v3 (and v4 in the 2.2.x case) so it was a bit different in my view. 

If someone opens a bug mentioning that C* 2.2 or 2.1 support is broken on the driver I'd be in favor of fixing the bug even if we don't provide full support by including it in our CI.

raj14243: Agree to this.

events.go
```diff
@@ -133,27 +133,42 @@ func (s *Session) handleEvent(framer *framer) {
...
+			keyspacesToRefresh[f.keyspace] = struct{}{}
 		}
 	}
-}
 
-func (s *Session) handleKeyspaceChange(keyspace, change string) {
-	s.control.awaitSchemaAgreement()
-	s.policy.KeyspaceChanged(KeyspaceUpdateEvent{Keyspace: keyspace, Change: change})
+	for keyspace := range keyspacesToRefresh {
+		if _, err := s.schemaDescriber.getSchema(keyspace); err != nil {
```
joao-r-reis: Just to reiterate, we should try to borrow from the java driver implementation as much as possible for stuff that we know works well (this is one of those cases). The java driver has a debouncer that does a full refresh of the schema. It does a `SELECT *` on everything and just computes the entire schema for every keyspace every time but this is OK because it's behind a debouncer so even if a lot of events come in the event handling is a basic "schedule" request on the debouncer and at some point in time the debouncer will make sure to run the refresh schema operation on all keyspaces. This is a simpler but effective approach that I think we should take here. 

Technically there is already a `eventdebouncer` (different from the one I added) but that one doesn't support requesting a "flush now" operation which is part of the reason why I added mine (`refreshDebouncer`) for the node events. The `eventdebouncer` for node/topology events basically just schedules things on the refresh debouncer. I think we can basically just do the exact same for the schema events with a new refresh debouncer instance (I think the same type can be reused, you provide a custom refresh function that is called when flushing).

We could improve on it by collecting the keyspaces on the events and only processing the keyspaces that were changed (even though we SELECT all, it would save some computation time) but that is not needed for the first implementation of this.

joao-r-reis: The refresh schema operation should also log the time it took to do the full refresh at debug level so that it is easier for us to benchmark its performance and troubleshoot potential issues.

raj14243: Aligned on next steps https://the-asf.slack.com/archives/C05LPRVNZV1/p1768514415652969

events.go
```diff
@@ -137,16 +137,16 @@ func (s *Session) handleSchemaEvent(frames []frame) {
 	for _, frame := range frames {
 		switch f := frame.(type) {
 		case *schemaChangeKeyspace:
-			s.schemaDescriber.clearSchema(f.keyspace)
+			s.schemaDescriber.debounceRefreshSchemaMetadata()
 			s.handleKeyspaceChange(f.keyspace, f.change)
```
joao-r-reis: Make this a part of the schema refresh call (I go into more detail in my comment on that function)

events.go
joao-r-reis: this function is calling the debouncer already

metadata.go
```diff
@@ -243,51 +245,66 @@ const (
...
-	delete(s.cache, keyspaceName)
-}
-
 // forcibly updates the current KeyspaceMetadata held by the schema describer
 // for a given named keyspace.
+//
+// Deprecated: Schema refreshes should be triggered through the schemaRefresher
+// refreshDebouncer (debounceRefreshSchemaMetadata) which batches and debounces
+// refresh requests for all keyspaces via refreshSchemas.
 func (s *schemaDescriber) refreshSchema(keyspaceName string) error {
```
joao-r-reis: this isn't exported so it can just be removed (along with the functions that fetch data by keyspace that become dead code)

metadata.go
```diff
@@ -243,51 +245,66 @@ const (
...
 
 // creates a session bound schema describer which will query and cache
 // keyspace metadata
 func newSchemaDescriber(session *Session) *schemaDescriber {
-	return &schemaDescriber{
+	meta := new(schemaMeta)
+	describer := &schemaDescriber{
 		session: session,
-		cache:   map[string]*KeyspaceMetadata{},
 	}
```
joao-r-reis, line 278: missing initialization of the debouncer

metadata.go
```diff
@@ -327,11 +344,99 @@ func (s *schemaDescriber) refreshSchema(keyspaceName string) error {
...
+			columns[keyspaceName],
+			functions[keyspaceName],
+			aggregates[keyspaceName],
+			userTypes[keyspaceName],
+			materializedViews[keyspaceName])
+		// update the cache
+		keyspaceMeta[keyspaceName] = keyspace
+	}
+	sd := session.schemaDescriber
+	sd.mu.Lock()
```
joao-r-reis: do we still need the lock if the refresh only happens in the debouncer?

metadata.go
```diff
@@ -327,11 +344,99 @@ func (s *schemaDescriber) refreshSchema(keyspaceName string) error {
...
+	meta.keyspaceMeta = keyspaceMeta
+	sd.schemaMeta.Store(meta)
+	sd.mu.Unlock()
+	// Notify policy if it supports schema refresh notifications
+	if notifier, ok := session.policy.(SchemaRefreshNotifier); ok {
+		notifier.SchemaRefreshed(sd.getSchemaMetaForRead())
+	}
 
 	return nil
 }
```
joao-r-reis, line 487: log at debug level the amount it took to run the full refresh

policies.go
```diff
@@ -324,6 +324,10 @@ type HostSelectionPolicy interface {
 	Pick(statement ExecutableStatement) NextHost
 }
 
+type SchemaRefreshNotifier interface {
```
joao-r-reis: make the interface and the method unexported

metadata.go
```diff
@@ -327,11 +344,99 @@ func (s *schemaDescriber) refreshSchema(keyspaceName string) error {
...
+	}
+	sd := session.schemaDescriber
+	sd.mu.Lock()
+	meta := sd.getSchemaMetaForUpdate()
+	meta.keyspaceMeta = keyspaceMeta
+	sd.schemaMeta.Store(meta)
+	sd.mu.Unlock()
+	// Notify policy if it supports schema refresh notifications
+	if notifier, ok := session.policy.(SchemaRefreshNotifier); ok {
+		notifier.SchemaRefreshed(sd.getSchemaMetaForRead())
```
joao-r-reis: we only need to refresh the replicas if keyspace metadata changed, we should do that computation here to determine whether we need to notify. And we probably should change the interface name (and method) to something more specific to just keyspaces not schema in general.

metadata.go
```diff
@@ -327,11 +344,99 @@ func (s *schemaDescriber) refreshSchema(keyspaceName string) error {
...
+}
+
+// forcibly updates the current KeyspaceMetadata held by the schema describer
+// for all the keyspaces.
+// This function is called via schemaRefresher refreshDebouncer to batch and
+// debounce schema refresh requests.
+func refreshSchemas(session *Session) error {
+	var err error
+
+	// query the system keyspace for schema data
```
joao-r-reis, line 383: wait for schema agreement here

metadata.go
```diff
@@ -327,11 +344,99 @@ func (s *schemaDescriber) refreshSchema(keyspaceName string) error {
...
+	sd := session.schemaDescriber
+	sd.mu.Lock()
+	meta := sd.getSchemaMetaForUpdate()
+	meta.keyspaceMeta = keyspaceMeta
+	sd.schemaMeta.Store(meta)
+	sd.mu.Unlock()
+	// Notify policy if it supports schema refresh notifications
+	if notifier, ok := session.policy.(SchemaRefreshNotifier); ok {
+		notifier.SchemaRefreshed(sd.getSchemaMetaForRead())
+	}
```
joao-r-reis, line 485: we should change this to something like:

```
// rough code, just typing it now
notifier, notifierExists := session.policy.(SchemaRefreshNotifier)
keyspacesChanged := false
for keyspaceName, keyspaceMetadata := range newKeyspaceMetadata {
    var changeType string
    oldKeyspaceMetadata, exists := oldKeyspaces[keyspaceName]
    if !exists {
          changeType = "CREATED"
          keyspacesChanged = true
    } else {
           // compare keyspace replication
           // if keyspace replication is different then changeType = "UPDATED" and keyspacesChanged = true
    }

    if !notifierExists && changeType != "" {
        session.policy.KeyspaceChanged(KeyspaceUpdateEvent{Keyspace: keyspaceName, Change:changeType}
    }
}

// also need to do the same process for keyspaces that were dropped with changetype "DROPPED", ideally without having to loop the entire old keyspace keyset

// in the end call notifier
if keyspacesChanged && notifierExists {
    notifier.SchemaRefreshed(sd.getSchemaMetaForRead())
}
```

With this process we can remove `func (s *Session) handleKeyspaceChange(keyspace, change string)` completely.

I'd probably implement the interface on the other policies that the driver has so we can avoid the unecessary `session.policy.KeyspaceChanged()` calls (empty implementations basically)

metadata.go
```diff
@@ -243,51 +245,66 @@ const (
...
-	s.mu.Lock()
-	defer s.mu.Unlock()
-
-	metadata, found := s.cache[keyspaceName]
+	metadata, found := s.getSchemaMetaForRead().keyspaceMeta[keyspaceName]
 	if !found {
-		// refresh the cache for this keyspace
-		err := s.refreshSchema(keyspaceName)
+		// refresh the cache
+		err := s.refreshSchemaMetadata()
```
joao-r-reis: With this new implementation we expect the driver to always have the cache up to date so if we don't find the metadata then just return `ErrKeyspaceDoesNotExist` right away.

For this to work properly though we need to add a call to `s.refreshSchemaMetadata()` as a replacement to this block in `session.init()`:

```
	// Invoke KeyspaceChanged to let the policy cache the session keyspace
	// parameters. This is used by tokenAwareHostPolicy to discover replicas.
	if !s.cfg.disableControlConn && s.cfg.Keyspace != "" {
		s.policy.KeyspaceChanged(KeyspaceUpdateEvent{Keyspace: s.cfg.Keyspace})
	}
```

And we need to add a `s.refreshSchemaMetadata()` call to `(c *controlConn) setupConn` inside the `if c.session.initialized() {` block

policies.go
```diff
@@ -447,43 +452,39 @@ func (t *tokenAwareHostPolicy) Init(s *Session) {
...
-	if err == nil {
-		strat := getStrategy(ks, t.logger)
+func (t *tokenAwareHostPolicy) updateReplicas(meta *clusterMeta, schemaMeta *schemaMeta) {
+	schema := schemaMeta.keyspaceMeta
+	newReplicas := make(map[string]tokenRingReplicas, len(schema))
+	for keyspace, metadata := range schema {
+		strat := getStrategy(metadata, t.logger)
 		if strat != nil {
 			if meta != nil && meta.tokenRing != nil {
 				newReplicas[keyspace] = strat.replicaMap(meta.tokenRing)
```
joao-r-reis: We need to add some kind of caching here to avoid computing the tokenmap every time for every keyspace. Most of the times a lot of the keyspaces share the same replication configuration. 

The java driver converts the replication configuration to a `Map<string,string>` and stores this in a `Map<string, Map<string, string>>` so that it can lookup the replication configuration per keyspace name. Then it adds all of these replication configuration map objects to a `Set` and then iterates on this set to compute the replicas and stores the replicas per replication configuration instead of per keyspace. [Reference here](https://github.com/apache/cassandra-java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/DefaultTokenMap.java#L78-L95).

Then on the `getReplicas` method it looks up the replication configuration by keyspace name and then looks up the replicas by replication configuration. [Reference here](https://github.com/apache/cassandra-java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/DefaultTokenMap.java#L176-L178).

To accomplish this in Go we can use a comparable struct instead of `Map<String,String>` as a "key" representing the replication configuration which allows us to basically create a cache of replicas by replication configuration. Or we can convert the replication configuration to a string but personally I think that's more error prone.

joao-r-reis (commented):
Good stuff! There's still some work left but it's looking really good and the overall approach looks solid.

I am not able to review with high confidence the code that handles the schema tables for older C* versions but I will help testing it out locally when the PR is ready. 

I think the goal was to attempt to keep supporting older C* versions as best as we can but not providing "full support" which means we don't have them on CI but it doesn't work that well in my opinion. Either we have them in CI or we remove code that is specific to those versions. For now I'll help out with some manual testing on this PR to try and avoid breaking users that might be on those older C* versions that still use the old system schema tables.

metadata.go
raj-ummadisetty: not once I remove the unused single keyspace functions

metadata.go
raj-ummadisetty: I thought I did this, maybe got removed during rebase

policies.go
raj-ummadisetty: I think we did discuss this previously and felt this can be handled as a follow up. I'll take a stab

metadata.go
raj-ummadisetty: As discussed in the thread, we will need it for `KeyspaceMetadata(keyspace string)` when caching is disabled.

policies.go
joao-r-reis: Oh sorry if I misled you but I feel like this is necessary especially if we make this behavior enabled by default to avoid increasing the session startup time by a lot in clusters with a ton of keyspaces

metadata.go
```diff
@@ -243,95 +245,226 @@ const (
...
+	schemaRefresher *refreshDebouncer
+	schemaMeta      atomic.Value // *schemaMeta
+}
 
-	cache map[string]*KeyspaceMetadata
+// Constants for keyspace change events
+const (
+	KeyspaceCreated = "CREATED"
+	KeyspaceUpdated = "UPDATED"
+	KeyspaceDropped = "DROPPED"
```
joao-r-reis: I'd probably rename these 3 constants slightly because they are on the main `gocql` package so there's no context clues to indicate that these refer to the [change type of the schema events](https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v5.spec#L997-L998). Something like `SchemaChangeTypeCreated`, maybe?

metadata.go
joao-r-reis: This is still missing I believe

metadata.go
```diff
@@ -243,95 +245,226 @@ const (
...
+
+	for _, createdKeyspace := range newKeyspaces {
+		session.policy.KeyspaceChanged(KeyspaceUpdateEvent{Keyspace: createdKeyspace, Change: KeyspaceCreated})
+	}
+	for _, droppedKeyspace := range droppedKeyspaces {
+		session.policy.KeyspaceChanged(KeyspaceUpdateEvent{Keyspace: droppedKeyspace, Change: KeyspaceDropped})
+	}
+	for _, updatedKeyspace := range updatedKeyspaces {
+		session.policy.KeyspaceChanged(KeyspaceUpdateEvent{Keyspace: updatedKeyspace, Change: KeyspaceUpdated})
+	}
```
joao-r-reis, line 485: I would type check the policy to see if it implements the notifier interface and if it does it only calls the notifier method. If it doesn't then call the regular `KeyspaceChanged` method like it's doing here. 

If the policy implements notifier technically all we care about is to check if there was any kind of change just so we know that we need to call the notify method but this computation of what actually changed (created,dropped,updated) is necessary to maintain compatibility with the old interface AND to later implement proper schema events on the public API which is being worked on in parallel on another PR (I've asked the author of that PR to wait until your implementation here is merged)

policies.go
```diff
@@ -455,35 +457,57 @@ func (t *tokenAwareHostPolicy) IsLocal(host *HostInfo) bool {
...
 }
 
-// updateReplicas updates replicas in clusterMeta.
-// It must be called with t.mu mutex locked.
+// updateReplicas updates replicas in clusterMeta for keyspace schema changes.
 // meta must not be nil and it's replicas field will be updated.
-func (t *tokenAwareHostPolicy) updateReplicas(meta *clusterMeta, keyspace string) {
-	newReplicas := make(map[string]tokenRingReplicas, len(meta.replicas))
-
+func (t *tokenAwareHostPolicy) updateReplicas(keyspace string) {
```
joao-r-reis, line 483: let's remove this method and make `KeyspaceChanged(update KeyspaceUpdateEvent)` NOOP on this policy, basically bringing back the internal notifier interface that you had before. 

This `updateReplicas` is basically the implementation that an external policy will have to implement if they want to replicate what our token aware policy is doing but we can cheat a little bit and make it simpler (and more performant in init) for our policy by just doing a full update everytime so the `updateAllReplicas` method that you added is enough.

policies.go
```diff
@@ -612,7 +636,13 @@ func (t *tokenAwareHostPolicy) Pick(qry ExecutableStatement) NextHost {
 	}
 
 	token := meta.tokenRing.partitioner.Hash(routingKey)
-	ht := meta.replicas[qry.Keyspace()].replicasFor(token)
+	var ht *hostTokens
+	if ksMeta, ok := t.getSchemaMeta().keyspaceMeta[qry.Keyspace()]; ok {
+		strategy := ksMeta.placementStrategy
+		if strategy != nil {
+			ht = meta.replicas[strategy.strategyKey()].replicasFor(token)
+		}
```
joao-r-reis, line 678: perfect 👌

topology.go
```diff
@@ -163,6 +179,33 @@ type networkTopology struct {
...
+	}
+	sort.Strings(dcs)
+
+	// Build key with sorted dc=rf pairs using strings.Builder for efficiency
+	var b strings.Builder
+	b.WriteString(n.strategyClass())
+	b.WriteString(":")
+	for i, dc := range dcs {
+		if i > 0 {
+			b.WriteString(",")
```
joao-r-reis: I'm a bit concerned about the use of `,` and `=` here because I'm not actually sure if these characters are forbidden in datacenter names so it could potentially lead to "collisions" between different strategies.

I would prepend the length of the datacenter name and use `:` and `|` as delimiters:

`{Strategy}|{LenDC}:{DC}:{RF}|...`
example: `NetworkTopologyStrategy|3:dc1:3|3:dc2:1`

I think this will prevent issues where the dc name is weird for example let's say there is only one dc with name `dc1:3|3:dc2`, the prefix length would make it different from the above case `NetworkTopologyStrategy|11:dc1:3|3:dc2:1`

topology.go
```diff
@@ -163,6 +179,33 @@ type networkTopology struct {
 	dcs map[string]int
 }
 
+func (n *networkTopology) strategyClass() string {
+	return networkTopologyStrategyClass
+}
+
+func (n *networkTopology) strategyKey() string {
```
joao-r-reis, line 206: Let's put the computation of the key at init time of the strategy so it happens only once and this method only returns the string. This method will be called multiple times and it will be called once per query at least (on the `Pick` method.

topology.go
```diff
@@ -163,6 +179,33 @@ type networkTopology struct {
...
+	// Sort datacenter names for deterministic output
+	dcs := make([]string, 0, len(n.dcs))
+	for dc := range n.dcs {
+		dcs = append(dcs, dc)
+	}
+	sort.Strings(dcs)
+
+	// Build key with sorted dc=rf pairs using strings.Builder for efficiency
+	var b strings.Builder
+	b.WriteString(n.strategyClass())
```
joao-r-reis: let's call `b.Grow(64)` right after creating the builder to pre-allocate 64 bytes, it should reduce the number of allocations quite a bit without much effort.

topology.go
```diff
@@ -127,10 +130,23 @@ func getStrategy(ks *KeyspaceMetadata, logger StructuredLogger) placementStrateg
...
 type simpleStrategy struct {
 	rf int
 }
 
+func (s *simpleStrategy) strategyClass() string {
+	return simpleStrategyClass
+}
+
+func (s *simpleStrategy) strategyKey() string {
+	return fmt.Sprintf("%s:rf=%d", s.strategyClass(), s.rf)
```
joao-r-reis: This looks good, I would only move the computation like I mentioned on my comment for the NT strategy.

session.go
```diff
@@ -367,8 +367,12 @@ func (s *Session) init() error {
 
 	// Invoke KeyspaceChanged to let the policy cache the session keyspace
 	// parameters. This is used by tokenAwareHostPolicy to discover replicas.
-	if !s.cfg.disableControlConn && s.cfg.Keyspace != "" {
-		s.policy.KeyspaceChanged(KeyspaceUpdateEvent{Keyspace: s.cfg.Keyspace})
+	if !s.cfg.disableControlConn && s.schemaDescriber != nil {
+		err := s.schemaDescriber.refreshSchemaMetadata()
+		if err != nil {
+			s.logger.Warning("Unable to refresh schema metadata.",
```
joao-r-reis: Add some info to this log message: `Unable to initialize schema metadata. Token aware host selection policies might fall back to round robin and attempts to retrieve keyspace metadata might fail with ErrKeyspaceDoesNotExist.`

cluster.go
```diff
@@ -286,6 +294,8 @@ type ClusterConfig struct {
 
 	// internal config for testing
 	disableControlConn bool
+
+	MetadataCacheMode MetadataCacheMode
```
joao-r-reis, line 320: add `//` api docs explaining what each mode does.

```
// MetadataCacheMode specifies whether the driver should read and parse the system tables to store schema metadata information. The behavior of Session.KeyspaceMetadata and tokenAwareHostPolicy will depend on this mode.
//
// Full is the default and specifies that the driver will read all the system tables and store all the schema metadata related to a keyspace including tables, functions, aggregates, UDTs and materialized views. In this mode token aware policies will work normally (note that token aware policy is not the default HostSelectionPolicy). Session.KeyspaceMetadata will only look up the metadata cache, it will not issue any system table query.
//
// KeyspaceOnly is a mode that allows token aware policies to work normally but prevents the driver from querying and storing all the other metadata that is mentioned in the Full mode's description above. Session.KeyspaceMetadata will continue to only look up the metadata cache but the returned object will have some of its fields set as `nil` (`Tables`, `Functions`, `Aggregates`, `MaterializedViews`, `UserTypes`).
//
// Disabled will completely disable this feature. Token aware policies will fall back to round robin and Session.KeyspaceMetadata will change its behavior to always query the system table (caching is completely disabled).
```

Maybe we move the specific docs for each mode to the actual constants and reference them in the docs for this config field.

CHANGELOG.md
joao-r-reis, line 13: can you add https://issues.apache.org/jira/browse/CASSGO-107 here as well? I just created it I think it makes sense to document this new metadata cache system as a separate ticket. I updated the title of the PR as well.

joao-r-reis (commented):
Looks really good. The main comments that I added are related to the strategy class cache (change the key computation a bit and move it) and the notifier internal interface that you removed. I think the internal interface still makes sense so that we can optimize the token aware policy to only run the update once per schema refresh.

Can't stress enough how good this looks 💯

metadata.go
raj14243: i'll bring back the notifier interface

policies.go
raj14243: I chose to optimize for schema changes rather than startup time. We pay the initialization cost once, but for future schema updates the number of keyspaces that need changes will probably be low. In that case `updateReplicas` will be called only a few times, whereas `updateAllReplicas` loops through all keyspaces. I’m not married to either approach, since we compute once per strategy, both should be fast, so I’ll follow your suggestion.

metadata.go
joao-r-reis: > And we need to add a s.refreshSchemaMetadata() call to (c *controlConn) setupConn inside the if c.session.initialized() { block

this is still missing I believe

policies.go
joao-r-reis: That's a fair point, how about we call the `updateAllReplicas` method (with the internal interface) when it's the first refresh? That would be the best of both worlds I think. Either we check session.initialized() or we just store a simple boolean flag in the refresh debouncer

metadata.go
raj14243: Is this to handle reconnects ?

joao-r-reis: Yeah, the driver can't receive schema events while the control connection is disconnected so we may lose updates. Refreshing when the control connection reconnects ensures it is always up to date and we don't miss events

control.go
```diff
@@ -364,6 +364,7 @@ func (c *controlConn) setupConn(conn *Conn, sessionInit bool) error {
 		NewLogFieldIP("host_addr", host.ConnectAddress()), NewLogFieldString("host_id", host.HostID()))
 
 	if c.session.initialized() {
+		c.session.schemaDescriber.debounceRefreshSchemaMetadata()
```
joao-r-reis: I think we should call refresh here instead of debounce so that the reconnection is considered complete when the refresh ends

metadata.go
```diff
@@ -243,95 +246,255 @@ const (
...
 // queries the cluster for schema information for a specific keyspace
 type schemaDescriber struct {
-	session *Session
-	mu      sync.Mutex
+	session         *Session
+	schemaRefresher *refreshDebouncer
+	schemaMeta      atomic.Value // *schemaMeta
+}
+
+// Schema change type constants as defined in the Cassandra Native Protocol v5 specification.
```
joao-r-reis: Technically it has been defined in the specs of all versions not just v5, going back to v1 https://github.com/apache/cassandra/blob/cassandra-2.2/doc/native_protocol_v1.spec

metadata.go
```diff
@@ -243,95 +246,255 @@ const (
...
+				updatedKeyspaces = append(updatedKeyspaces, keyspaceName)
+			}
+		}
+	}
+	droppedKeyspaces = sd.getDroppedKeyspaces(oldKeyspaceMeta, keyspaces)
+	meta.keyspaceMeta = keyspaceMeta
+	sd.schemaMeta.Store(meta)
+
+	// Notify policy if it supports schema refresh notifications
+	if notifier, ok := session.policy.(schemaRefreshNotifier); ok {
```
joao-r-reis, line 473: We can make this special case happen only at the first refresh so subsequent refreshes are more efficient like you suggested

metadata.go
```diff
@@ -243,95 +246,255 @@ const (
...
+		userTypes, err = getAllUserTypeMetadata(session)
+		if err != nil {
+			return err
+		}
+		materializedViews, err = getAllMaterializedViewsMetadata(session)
+		if err != nil {
+			return err
+		}
+		fullMetadataElapsed := time.Since(fullMetadataStart)
+		session.logger.Debug("Full metadata fetch completed",
```
joao-r-reis: let's move this log message to just after `sd.schemaMeta.Store(meta)` and change the message to `Schema Metadata refresh completed.`

joao-r-reis (commented):
Tests are failing, apparently in some cases the `awaitSchemaAgreement` call is hanging which causes the `refreshDebouncer.stop()` call to hang as well. I can try to help troubleshoot this on monday.

On a separate note I've just noticed that the default timeout for schema agreement is 60 seconds, let's have a different timeout for the call that happens inside the schema refresher, let's set that one to 10 seconds and ignore the config setting.

metadata.go
raj14243: I do not think its worth the complexity and confusion it will create with the internal interface.

metadata.go
raj14243: There is already a defer `Schema refresh completed` I'll call this `Schema metadata cache refresh completed`

control.go
```diff
@@ -364,6 +364,13 @@ func (c *controlConn) setupConn(conn *Conn, sessionInit bool) error {
 		NewLogFieldIP("host_addr", host.ConnectAddress()), NewLogFieldString("host_id", host.HostID()))
 
 	if c.session.initialized() {
+		refreshErr := c.session.schemaDescriber.refreshSchemaMetadata()
+		if refreshErr != nil {
+			c.session.logger.Warning("Failed to refresh schema metadata after reconnecting. "+
+				"Schema might be stale or missing, causing token-aware routing to fall back to the configured fallback policy. "+
+				"Keyspace metadata queries will fail with ErrKeyspaceDoesNotExist until schema refresh succeeds.",
```
joao-r-reis: `"Keyspace metadata queries will fail with ErrKeyspaceDoesNotExist until schema refresh succeeds."` 

might fail but not guaranteed because it will depend on whether any prior refreshes were successful.

```
"Failed to refresh schema metadata after reconnecting. Schema metadata might be stale or missing. If it's missing then it will cause token-aware policies to fall back to the configured fallback policy and keyspace metadata queries will fail with ErrKeyspaceDoesNotExist until schema refresh succeeds."
```

joao-r-reis:
Ok only added 1 last comment now but it's a NIT regarding a log message, everything else has been resolved 👍 I didn't have time yet to troubleshoot the test issue but working on it. After we fix the issue causing test failures we're good to merge

metadata.go
```diff
@@ -243,95 +246,257 @@ const (
...
 
+// forcibly updates the current KeyspaceMetadata held by the schema describer
+// for all the keyspaces.
+// This function is called via schemaRefresher refreshDebouncer to batch and
+// debounce schema refresh requests.
+func refreshSchemas(session *Session) error {
+	start := time.Now()
+	defer func() {
+		elapsed := time.Since(start)
+		session.logger.Debug("Schema refresh completed",
```
joao-r-reis: I just noticed while troubleshooting that this log message will be misleading when the refresh fails because it's always printed the same way regardless of whether this functions fails or not

metadata.go
```diff
@@ -243,95 +246,257 @@ const (
...
+	var tables map[string][]TableMetadata
+	var columns map[string][]ColumnMetadata
+	var functions map[string][]FunctionMetadata
+	var aggregates map[string][]AggregateMetadata
+	var userTypes map[string][]UserTypeMetadata
+	var materializedViews map[string][]MaterializedViewMetadata
+	if session.cfg.MetadataCacheMode == Full {
+		tables, err = getAllTablesMetadata(session)
+		if err != nil {
+			return err
```
joao-r-reis: while troubleshooting the test issue it would have helped if this was something like `fmt.Errorf("failed to retrieve table metadata: %w", err)` for all of these error conditions

metadata.go
```diff
@@ -1061,14 +1561,25 @@ func getFunctionsMetadata(session *Session, keyspaceName string) ([]FunctionMeta
 		FROM %s
 		WHERE keyspace_name = ?`, tableName)
```
joao-r-reis: test is failing because you forgot to remove the `WHERE keyspace_name = ?` clause here

raj-ummadisetty: 🤦‍♂️

joao-r-reis:
So one thing that is a bit concerning is that even though the schema refresh was failing we didn't have any test failures related to token awareness which means we are lacking a good integration test that verifies that the token aware policy is working correctly...

joao-r-reis:
There's a `TestTokenAwareConnPool` test that should do it but it's missing an "assertion" section to actually verify that the queries are going to the correct hosts

raj14243:
> So one thing that is a bit concerning is that even though the schema refresh was failing we didn't have any test failures related to token awareness which means we are lacking a good integration test that verifies that the token aware policy is working correctly...

I added bunch of unit tests but an integration test failing would be useful. I am also adding additional integration tests to test the `MetadataCacheMode` as well

joao-r-reis:
Good news is that my local test runs with C* 2.2 and 3.11 look good and my manual tests with a cluster that has a lot of keyspaces and tables also shows there's no significant slow down of the session init time with this new change.

raj-ummadisetty:
> Good news is that my local test runs with C* 2.2 and 3.11 look good and my manual tests with a cluster that has a lot of keyspaces and tables also shows there's no significant slow down of the session init time with this new change.

🏆 that's awesome. I was able to get integrations tests to run locally. I'll add more tests and send out a commit

cassandra_test.go
```diff
@@ -2677,119 +2757,149 @@ func TestKeyspaceMetadata(t *testing.T) {
...
+			if keyspaceMetadata.Name != session.cfg.Keyspace {
+				t.Fatalf("Expected the keyspace name to be %s but was %s", session.cfg.Keyspace, keyspaceMetadata.Name)
+			}
 
-	tableMetadata, found := keyspaceMetadata.Tables["test_metadata"]
-	if !found {
-		t.Fatalf("failed to find the test_metadata table metadata")
-	}
+			if tc.expectFullMetadata {
+				if len(keyspaceMetadata.Tables) == 0 {
```
joao-r-reis, line 2786: can we add a quick check for when metadata cache is disabled to make sure the cache is empty?

cassandra_test.go
```diff
@@ -2955,13 +3065,40 @@ func TestTokenAwareConnPool(t *testing.T) {
 		t.Fatalf("failed to insert with err: %v", err)
 	}
 
-	query = session.Query("SELECT data FROM test_token_aware where id = ?", 42).Consistency(One)
+	// Verify token-aware routing: queries with the same partition key should
+	// consistently go to the same host (proving token awareness)
+	var queriedHosts []*HostInfo
+
+	observer := funcQueryObserver(func(ctx context.Context, o ObservedQuery) {
```
joao-r-reis: I think in addition to this we should also use `Query.Trace()`, if I'm not mistaken the trace events (each event is a row from the table I think) obtained from C* would have the `Source` be the same as `Coordinator` which means every event originated from the coordinator so there were no hops to nodes that weren't the coordinator (maybe need to set consistency level ONE?) if token awareness is working correctly. 

I think the default implementation `NewTraceWriter` already reads the trace data from the system tables and it writes to the provided `io.Writer`. We could create a custom test tracer that stores or outputs this data in a different format but the existing tracer is probably good enough for what we want to test but it will require some string parsing/matching. See [C# driver example here](https://github.com/datastax/csharp-driver/blob/f68365ba64117d57b3c26763612dcfff35dab64f/src/Cassandra.IntegrationTests/Policies/Tests/LoadBalancingPolicyShortTests.cs#L84-L88).

cluster.go
```diff
@@ -42,6 +42,36 @@ type PoolConfig struct {
...
+const (
+	// Full mode reads and caches all schema metadata including keyspaces, tables, columns,
+	// functions, aggregates, user-defined types, and materialized views.
+	//
+	// Token-aware routing works normally with full replica information.
+	// Session.KeyspaceMetadata returns cached metadata without querying system tables.
+	Full MetadataCacheMode = iota
+
+	// KeyspaceOnly mode reads and caches only keyspace metadata (replication strategy and options).
+	// This enables token-aware routing without the overhead of caching detailed schema information.
```
joao-r-reis: `This enables token-aware routing (if TokenAwareHostPolicy is used)`

cluster.go
```diff
@@ -42,6 +42,36 @@ type PoolConfig struct {
...
+// This affects the behavior of Session.KeyspaceMetadata and token-aware host selection policies.
+//
+// See the individual mode constants (Full, KeyspaceOnly, Disabled) for detailed behavior of each mode.
+type MetadataCacheMode int
+
+const (
+	// Full mode reads and caches all schema metadata including keyspaces, tables, columns,
+	// functions, aggregates, user-defined types, and materialized views.
+	//
+	// Token-aware routing works normally with full replica information.
```
joao-r-reis: `Token-aware routing works normally (if TokenAwareHostPolicy is used) with full replica information.`

cluster.go
```diff
@@ -42,6 +42,36 @@ type PoolConfig struct {
...
+	// Session.KeyspaceMetadata returns cached metadata without querying system tables.
+	Full MetadataCacheMode = iota
+
+	// KeyspaceOnly mode reads and caches only keyspace metadata (replication strategy and options).
+	// This enables token-aware routing without the overhead of caching detailed schema information.
+	//
+	// Token-aware routing works normally with full replica information.
```
joao-r-reis: `Token-aware routing works normally (if TokenAwareHostPolicy is used) with full replica information.`

metadata.go
joao-r-reis: I still think this is misleading, I really think this message should say "failed" or just not exist if the refresh failed.

joao-r-reis (commented):
Tests are passing 👍 , to verify token awareness I think we should use query traces in addition to the observer (check my comment for more details).

cassandra_test.go
raj14243: Nice, in that case I can just use tracer and not require observer.

joao-r-reis (approved):
💯 can you squash all the commits into a single one and format the commit message [according to our guidelines](https://github.com/apache/cassandra-gocql-driver/blob/trunk/CONTRIBUTING.md#commit-message)? Or check an example commit on trunk.

PR #1927, opened 2026-01-29
CASSGO-105 Prevent panic iin compileMetadata() when final func is not defined for an aggregate

worryg0d:
Previously, when there was a user defined aggregate without FINAL_FUNC defined the compileMetadata() function would panic due to a nil pointer dereference. Gocql should properly handle this case since FINAL_FUNC is optional for user defined aggregates.

This patch fixes the described problem by adding nil-checking before dereferencing.

Patch by Bohdan Siryk; Reviewed by João Reis for CASSGO-105


joao-r-reis (approved):
👍

joao-r-reis:
add my name as a reviewer to the commit msg and I'll merge it afterwards, thanks!

worryg0d:
Updated the commit message, thanks!

PR #1932, opened 2026-03-17
Add options to shuffle replicas for token-aware policy and log warning when default is used

joao-r-reis:
Shuffling replicas produces much better load distribution for token-aware host selection, while leaving the previous (non-shuffled) behavior available for workloads that need it
(notably LWTs, which are voided if replicas are shuffled).

Also logs a warning if the default behavior is used since the default (non shuffled) is not optimal for the majority of applications. If non shuffled is the intended behavior then explicitly setting it will disable the log message.

patch by tejaslodayadd; reviewed by worryg0d, joao-r-reis for CASSGO-106


joao-r-reis:
Meant to open this on my fork just to verify everything with the commit is correct (with the correct author, etc.)

Ignore this

PR #1933, opened 2026-03-17
CASSGO-110 Bump Go version support from 1.22 and 1.23 to 1.25 and 1.26

joao-r-reis:


worryg0d (approved):
CI looks good, nothing fails and no extra warnings👍

One thing, we should upgrade versions of **action** dependencies the CI since deprecation of nodejs 20 in GH actions, see warning annotations under the job. We need to raise a ticket to bump versions of those dependencies

https://github.com/apache/cassandra-gocql-driver/actions/runs/23212260657/job/67464349942?pr=1933

https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

joao-r-reis:
Good catch

PR #1936, opened 2026-03-18
CASSGO-113: Add dedicated error types for more error codes

paoloelefante:
Add dedicated error types for all generic server error codes
Introduce distinct concrete types for all server error codes that were
previously returned as generic errorFrame values:

RequestErrOverloaded, RequestErrBootstrapping, RequestErrInvalid,
RequestErrConfig, RequestErrCredentials, RequestErrProtocol,
RequestErrServer, RequestErrSyntax, RequestErrTruncate,
RequestErrUnauthorized

Each type embeds errorFrame and follows the existing pattern used for
RequestErrUnavailable, RequestErrWriteTimeout, RequestErrReadTimeout.

The catch-all case and the TODO comment in parseErrorFrame are removed.

Add tests covering parsing of all ten error codes.

Patch by Paolo Elefante; reviewed by João Reis, Bohdan Siryk for [CASSGO-113](https://issues.apache.org/jira/browse/CASSGO-113)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for session ready, host state, topology change and schema changes custom listeners (CASSGO-101)
 - Add Session.AllKeyspaceMetadata() (CASSGO-109)
 - Add GetSerialConsistency method to Query and Batch (CASSGO-103)
+- Add RequestErrOverloaded, RequestErrBootstrapping, RequestErrInvalid, RequestErrConfig, RequestErrCredentials, RequestErrSyntax, RequestErrTruncate, RequestErrUnauthorized for dedicated error handling (CASSGO-113)
 
 ### Changed
 
--- a/frame.go
+++ b/frame.go
@@ -759,10 +759,24 @@ func (f *framer) parseErrorFrame() (frame, error) {
 			return nil, err
 		}
 		return res, nil
-	case ErrCodeInvalid, ErrCodeBootstrapping, ErrCodeConfig, ErrCodeCredentials, ErrCodeOverloaded,
-		ErrCodeProtocol, ErrCodeServer, ErrCodeSyntax, ErrCodeTruncate, ErrCodeUnauthorized:
-		// TODO(zariel): we should have some distinct types for these errors
+	case ErrCodeOverloaded:
+		return &RequestErrOverloaded{errorFrame: errD}, nil
+	case ErrCodeBootstrapping:
+		return &RequestErrBootstrapping{errorFrame: errD}, nil
+	case ErrCodeInvalid:
+		return &RequestErrInvalid{errorFrame: errD}, nil
+	case ErrCodeConfig:
+		return &RequestErrConfig{errorFrame: errD}, nil
+	case ErrCodeCredentials:
+		return &RequestErrCredentials{errorFrame: errD}, nil
+	case ErrCodeServer, ErrCodeProtocol:
 		return errD, nil
+	case ErrCodeSyntax:
+		return &RequestErrSyntax{errorFrame: errD}, nil
+	case ErrCodeTruncate:
+		return &RequestErrTruncate{errorFrame: errD}, nil
+	case ErrCodeUnauthorized:
+		return &RequestErrUnauthorized{errorFrame: errD}, nil
 	default:
 		return nil, fmt.Errorf("unknown error code: 0x%x", errD.code)
 	}
--- a/errors.go
+++ b/errors.go
@@ -154,6 +154,46 @@ func (e *RequestErrUnavailable) String() string {
 	return fmt.Sprintf("[request_error_unavailable consistency=%s required=%d alive=%d]", e.Consistency, e.Required, e.Alive)
 }
 
+// RequestErrOverloaded represents an overloaded error returned by Cassandra.
+type RequestErrOverloaded struct {
+	errorFrame
+}
+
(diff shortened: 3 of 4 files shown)


dmarro89 (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add RequestErrOverloaded and RequestErrBootstrapping for dedicated error handling (CASSGO-113)
```
joao-r-reis: Very minor thing but can you move this to the bottom of the section just to keep the order in which these patches were committed?

frame.go
```diff
@@ -759,7 +759,11 @@ func (f *framer) parseErrorFrame() (frame, error) {
...
 		}
 		return res, nil
-	case ErrCodeInvalid, ErrCodeBootstrapping, ErrCodeConfig, ErrCodeCredentials, ErrCodeOverloaded,
+	case ErrCodeOverloaded:
+		return &RequestErrOverloaded{errorFrame: errD}, nil
+	case ErrCodeBootstrapping:
+		return &RequestErrBootstrapping{errorFrame: errD}, nil
+	case ErrCodeInvalid, ErrCodeConfig, ErrCodeCredentials,
 		ErrCodeProtocol, ErrCodeServer, ErrCodeSyntax, ErrCodeTruncate, ErrCodeUnauthorized:
 		// TODO(zariel): we should have some distinct types for these errors
```
joao-r-reis: We want all of these to have their own types too but it's up to you if you want to add these or if you want to keep this patch focused on overloaded and bootstrapping, let me know

joao-r-reis (commented):
Left a couple of comments but they are both nitpicks. If you want to keep this patch and CASSGO-113 focused on just these 2 error types instead of all the ones that are missing a dedicated type then I can just go ahead and merge it as is (after you add my name as reviewer to the commit message)

joao-r-reis:
Also, we're really close to starting the process for v2.1.0 release. If I don't hear back from you until EOD today I might just merge this as is in any case, the patch is good to be merged regardless and this way it can be included in v2.1.0 (although I will have to commit it manually to update the commit message and this way GH won't recognize the PR as merged, I'm not sure if this is something that you want to avoid)

worryg0d (approved):
👍 

It would be good to have integration tests for such things, but there was nothing in place, and such errors are hard to trigger, so you can leave it as is

frame.go
paoloelefante: Thanks for the feedback.
Agreed: introducing dedicated types for all currently generic server
error codes is the right approach, both for consistency and to eliminate
the TODO that has been there since the original implementation.
I have extended the PR to cover the remaining codes: ErrCodeInvalid,
ErrCodeConfig, ErrCodeCredentials, ErrCodeProtocol, ErrCodeServer,
ErrCodeSyntax, ErrCodeTruncate, ErrCodeUnauthorized.
The catch-all case and the TODO comment are now removed.

CHANGELOG.md
paoloelefante: Done.

joao-r-reis (approved):
👍 can you add the reviewers to the commit message? Then I'll merge, thanks

frame.go
```diff
@@ -759,10 +759,26 @@ func (f *framer) parseErrorFrame() (frame, error) {
...
+	case ErrCodeProtocol:
+		return &RequestErrProtocol{errorFrame: errD}, nil
+	case ErrCodeServer:
+		return &RequestErrServer{errorFrame: errD}, nil
+	case ErrCodeSyntax:
+		return &RequestErrSyntax{errorFrame: errD}, nil
+	case ErrCodeTruncate:
+		return &RequestErrTruncate{errorFrame: errD}, nil
+	case ErrCodeUnauthorized:
+		return &RequestErrUnauthorized{errorFrame: errD}, nil
```
joao-r-reis, line 779: After checking the test failures I realize that the internal protocol negotiation mechanism depends on these errorframes so let's keep the `Server` and `Protocol` errors as they were to keep the behavior so we don't have to go and change the protocol negotiation code in this PR. After removing the new types for `Server` and `Protocol` errors tests are passing again locally

worryg0d: I spent some time reworking the protocol negotiation mechanism and ended up with this:
```go
// Checks if the error is protocol related and should be retried during startup.
// It returns the frame that caused the error and whether the error should be retried.
func (s *startupCoordinator) checkProtocolRelatedError(err error) bool {
	var protocolErr *protocolError
	if errors.As(err, &protocolErr) {
		if _, supportedFrame := protocolErr.frame.(*supportedFrame); supportedFrame {
			// We can receive a supportedFrame wrapped in protocolError from Conn.recv if the host responds to a 0 stream id.
			// If we receive a supportedFrame then we know that the host is not compatible with the protocol version, but it is reachable, so we can retry
			return true
		} else if requestErr, ok := protocolErr.frame.(RequestError); ok {
			// Unwrapping underlying error frame to check if it is protocol related
			err = requestErr
		}
	}

	switch err.(type) {
	case *RequestErrServer, *RequestErrProtocol:
		// If we receive an errorFrame with codes ErrCodeProtocol or ErrCodeServer,
		// then we should try to downgrade a protocol version, so do not skip the host
		return true
	default:
		// In any other case we should not retry as it means the host is not reachable or some other error happened
		return false
	}
}
```

But I would agree that it should be handled outside of this PR

paoloelefante: Fixed. Removed RequestErrServer and RequestErrProtocol to preserve the existing protocol negotiation behavior. The remaining eight types are intact. I added the reviewes in the commit message. Thanks!

PR #1939, opened 2026-03-31
CASSGO-115 Iter.MapScan is unable to scan data of user-defined types fix

worryg0d:
Previously, Iter.MapScan was unable to scan columns of user-defined types due to a bug which leaded to never updated destination variable. With this patch it is handled correctly.

Patch by Bohdan Siryk; reviewed by TBD for CASSGO-115

### DIFF
--- a/marshal.go
+++ b/marshal.go
@@ -2921,12 +2921,6 @@ func (udt UDTTypeInfo) Marshal(value interface{}) ([]byte, error) {
 
 // Unmarshal unmarshals the byte slice into the value.
 func (udt UDTTypeInfo) Unmarshal(data []byte, value interface{}) error {
-	// do this up here so we don't need to duplicate all of the map logic below
-	if iptr, ok := value.(*interface{}); ok && iptr != nil {
-		v := map[string]interface{}{}
-		*iptr = v
-		value = &v
-	}
 	switch v := value.(type) {
 	case UDTUnmarshaler:
 		for id, e := range udt.Elements {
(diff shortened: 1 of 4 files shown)


worryg0d:
I'm unsure what caused a couple of jobs to be canceled, probably quota...

marshal.go
```diff
@@ -2923,9 +2923,13 @@ func (udt UDTTypeInfo) Marshal(value interface{}) ([]byte, error) {
...
-		v := map[string]interface{}{}
-		*iptr = v
+		var v map[string]interface{}
 		value = &v
+		// Deffering this to the end of the method so dst var will hold updated map value
+		// https://issues.apache.org/jira/projects/CASSGO/issues/CASSGO-115
+		defer func() {
+			*iptr = v
+		}()
 	}
```
joao-r-reis: Hmm I think I'd rather just extract the code in `case *map[string]interface{}:` to a private method and add a `case *interface{}:` that calls this new method instead of doing this. It would be easier to read and less error prone imo

joao-r-reis: something roughly like this:

```
func (udt UDTTypeInfo) Unmarshal(data []byte, value interface{}) error {
    switch v := value.(type) {
    case *interface{}:
        if v == nil {
            return unmarshalErrorf("can not unmarshal into nil *interface{}")
        }
        // Unmarshal into a temporary map, then assign to *interface{}
        m := map[string]interface{}{}
        if err := udt.unmarshalIntoMap(data, &m); err != nil {
            return err
        }
        *v = m
        return nil
        
    case UDTUnmarshaler:
        // ... existing code ...
        
    case *map[string]interface{}:
        return udt.unmarshalIntoMap(data, v)
        
    // ... rest of cases ...
    }
}

// Extract common map unmarshaling logic
func (udt UDTTypeInfo) unmarshalIntoMap(data []byte, v *map[string]interface{}) error {
    if data == nil {
        *v = nil
        return nil
    }
    
    m := map[string]interface{}{}
    for id, e := range udt.Elements {
        if len(data) == 0 {
            break
        }
        if len(data) < 4 {
            return unmarshalErrorf("can not unmarshal UDT: field [%d]%s: unexpected eof", id, e.Name)
        }
        
        var p []byte
        p, data = readBytes(data)
        
        elemValue := reflect.New(reflect.TypeOf(e.Type.Zero()))
        if err := Unmarshal(e.Type, p, elemValue.Interface()); err != nil {
            return err
        }
        m[e.Name] = elemValue.Elem().Interface()
    }
    *v = m
    return nil
}



```

worryg0d: yeah, thinking about this now I would agree having a separate method is much better, at least from maintanability point of view

this `defer func { *iptr = v }` magic is definetly something that potential contributor could really miss...

worryg0d: Updated patch - now there is an unmarshalIntoMap method, which is called for both *map[string]interface{} and *interface{} cases

jameshartig (approved):

jameshartig:
Need to update the commit sign off but lgtm

joao-r-reis (approved):
nice

worryg0d:
I've updated the commit message and added 2.1.1 section to the changelog

joao-r-reis:
I'll merge this as soon as 2.1.0 is released, working on it now

worryg0d:
Thanks @joao-r-reis, btw is change log entry fine?

joao-r-reis:
> Thanks @joao-r-reis, btw is change log entry fine?

yep

PR #1941, opened 2026-04-06
CASSGO-119 Add Table field to ObservedQuery and ObservedBatch for table-level observability

raj14243:
ObservedQuery and ObservedBatch did not expose the target table name, requiring consumers (metrics collectors, loggers, tracers) to parse the CQL statement string to determine which table a query targets.

This change adds a Table field to ObservedQuery and a Tables field to ObservedBatch, populated from prepared statement metadata returned by Cassandra. The table information was already available internally via preparedMetadata (used for routing), so no additional round-trips or CQL parsing are needed.

Patch by Raj Ummadisetty for CASSGO-119

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Security-model discoverability (CASSANDRA-21464)
 - Improve host_source locking and ring refresh concurrency (CASSGO-121)
+- Add PreparedMetadata (Keyspace, Table) and IsPrepared fields to ObservedQuery, and parallel PreparedMetadata / IsPrepared slices to ObservedBatch, for statement-level observability without CQL parsing (CASSGO-119)
 
 ## [2.1.2]
 
--- a/query_executor.go
+++ b/query_executor.go
@@ -385,15 +385,20 @@ func (q *internalQuery) attempt(keyspace string, end, start time.Time, iter *Ite
 		q.qryOpts.observer.ObserveQuery(q.qryOpts.context, ObservedQuery{
 			Keyspace:  keyspace,
 			Statement: q.qryOpts.stmt,
-			Values:    q.qryOpts.values,
-			Start:     start,
-			End:       end,
-			Rows:      iter.numRows,
-			Host:      host,
-			Metrics:   metricsForHost,
-			Err:       iter.err,
-			Attempt:   attempt,
-			Query:     q.originalQuery,
+			PreparedMetadata: PreparedMetadata{
+				Keyspace: q.routingInfo.getKeyspace(),
+				Table:    q.routingInfo.getTable(),
+			},
+			IsPrepared: q.routingInfo.isPrepared(),
+			Values:     q.qryOpts.values,
+			Start:      start,
+			End:        end,
+			Rows:       iter.numRows,
+			Host:       host,
+			Metrics:    metricsForHost,
+			Err:        iter.err,
+			Attempt:    attempt,
+			Query:      q.originalQuery,
 		})
 	}
 }
@@ -432,6 +437,7 @@ func (q *internalQuery) GetRoutingKey() ([]byte, error) {
 		q.routingInfo.mu.Lock()
 		q.routingInfo.keyspace = meta.Keyspace
 		q.routingInfo.table = meta.Table
+		q.routingInfo.prepared = true
 		q.routingInfo.mu.Unlock()
 	}
 	return createRoutingKey(meta, q.qryOpts.values)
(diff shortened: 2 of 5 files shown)


worryg0d (commented):
Hi @raj14243,

I have spent some time reviewing this PR - LGTM.

One thing I noticed is that PR lacks integration testing for this change. This is a small change actually, but updating existing tests would be good anyway. There are `TestObserve` and `TestBatchObserve` integration tests for both query and batch observe respectively, that you can update to reflect this API change.

session.go
```diff
@@ -2374,6 +2374,14 @@ type ObservedQuery struct {
 	Keyspace  string
 	Statement string
 
+	// StatementKeyspace is the keyspace of the table targeted by the query.
+	// It is populated from prepared statement metadata returned by Cassandra.
+	StatementKeyspace string
```
worryg0d, line 2379: I think it is fine to populate the existing `Keyspace` field here - it is unclear to see the difference between `Keyspace` and `StatementKeyspace`

query_executor.go
```diff
@@ -383,17 +383,19 @@ func (q *internalQuery) attempt(keyspace string, end, start time.Time, iter *Ite
...
-			End:       end,
-			Rows:      iter.numRows,
-			Host:      host,
-			Metrics:   metricsForHost,
-			Err:       iter.err,
-			Attempt:   attempt,
-			Query:     q.originalQuery,
+			Keyspace:          keyspace,
+			Statement:         q.qryOpts.stmt,
+			StatementKeyspace: q.routingInfo.getKeyspace(),
```
worryg0d, line 388: `routingInfo` could not have any information about keyspaces/tables for DDL queries that are unprepareable, so you probably should use `internalQuery.Keyspace()` here. It has some predefined defaulting behavior.

We should probably think about adding a kind of parsing of DDL queries to determine keyspace and table names from it before defaulting to session-level keyspace. 

Also, there is an alternative for `internalBatch.Keyspace()` as well, but it just calls the underlying routingInfo, which is fine because it is impossible to batch DDL queries.

@joao-r-reis, could you please take a look at this?

joao-r-reis: Hmm I don't think we should be adding CQL parsing for something like this, most if not all drivers avoid doing any kind of CQL parsing. What some drivers do is expose a statement builder / mapper API that will create statements and internally set fields so the driver is aware of some metadata like keyspace, table, idempotency without having to parse it.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [2.1.1]
 
+### Added
```
joao-r-reis, line 10: Add a new 2.2.0 section and move this under there

joao-r-reis (commented):
I think the current design is not intuitive for cases where no prepared statement metadata is available.

I explored a bit and came up with something like this:

```
type PreparedMetadata struct {
    Keyspace string
    Table    string
}

type ObservedQuery struct {
    Keyspace          string
    Statement         string
    PreparedMetadata  PreparedMetadata  // zero value when unprepared
    IsPrepared        bool              // flag at query level
    // ...
}

type ObservedBatch struct {
    Keyspace          string
    Statements        []string
    PreparedMetadata  []PreparedMetadata  // zero values for unprepared
    IsPrepared        []bool              // parallel array of flags
    // ...
}
```

We could also just get rid of the `IsPrepared` flag and just make prepared metadata a pointer (so `nil` would mean it's not prepared) but that would lead to extra allocs.

Wdyt?

worryg0d:
> Wdyt?

Yeah I guess it would work, but API looks a bit gross. As alternative to exposing those fields to ObservedXY structs, you may call this method: https://github.com/apache/cassandra-gocql-driver/blob/c65c762b83eccdb6a6a083c123a5935c4302745c/session.go#L756

It basically allows to retrieve internally cached statement metadata. But this requires observers to be able to call session methods

session.go
raj14243: The existing Keyspace field is populated with Keyspace the session is initialized with. I am concerned changing that might break if someone is using it with that assumption.

query_executor.go
raj14243: I will check the `internalQuery.Keyspace()` I agree with @joao-r-reis on not having CQL parsing. I started this PR as we were internally parsing statement to extract keyspace and wanted to avoid that :)

session.go
worryg0d: Yeah, this is why I filed CASSGO-120, I'm unsure if this is the correct behavior. At least I was expecting it to be based on the query, if it is presented there

worryg0d:
Besides CASSGO-120, this PR looks good.

I guess you can get rid of CASSGO-120 in the commit message, and we will handle it separately if it is worth fixing

joao-r-reis (approved):

joao-r-reis:
Let's hold off on merging this until 2.1.2 is released

worryg0d (approved):
+1

joao-r-reis:
@raj14243 can you add both of us as reviewers to the commit message and fix the git conflict? Will merge afterwards

joao-r-reis:
Or if you don't mind github not recognizing the PR as "merged" then I can do that for you and set you as author of the commit

raj14243:
> @raj14243 can you add both of us as reviewers to the commit message and fix the git conflict? Will merge afterwards

@joao-r-reis done

PR #1929, opened 2026-02-11
CASSGO-101 CASSGO-109 Support for Node Status, Topology Change and Schema Change custom listeners

worryg0d:
Patch provides support for custom listeners of node status, topology, and schema change events

ASF Slack thread: https://the-asf.slack.com/archives/C05LPRVNZV1/p1767548766531679

PR #1943, opened 2026-04-09
CASSGO-39 Add exec attempt interceptor (round 3)

### LINKED ISSUE #1786: CASSGO-39 Add query attempt interceptor

abenn135:
This supersedes https://github.com/apache/cassandra-gocql-driver/pull/1820 rebasing on top of, and taking into account, changes in v2.

Notably, this removes the ability to mutate the Statement in the Interceptor -- it is now embedded in the `internalRequest`, making it inaccessible to a public API without invasive refactoring.

Additionally, I modified `queryMetrics` in a few ways:

* Made it threadsafe, so that latency and attempt count are updated under lock, preventing a potential race when reading latency
* Separated out "attempts started" from "attempts completed". Speculative execution means that several attempts could happen concurrently, and this guarantees that (a) each attempt/interception gets its own attempt count, and (b) the retry policy respects attempts started, to prevent over-retry when earlier retries haven't completed yet.

Fixes #1786.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -258,9 +258,9 @@ type ClusterConfig struct {
 	// See https://issues.apache.org/jira/browse/CASSANDRA-10786
 	DisableSkipMetadata bool
 
-	// QueryAttemptInterceptor will set the provided query interceptor on all queries created from this session.
-	// Use it to intercept and modify queries by providing an implementation of QueryAttemptInterceptor.
-	QueryAttemptInterceptor QueryAttemptInterceptor
+	// ExecAttemptInterceptor will set the provided interceptor on all queries/batches created from this session.
+	// Use it to intercept queries by providing an implementation of ExecAttemptInterceptor.
+	ExecAttemptInterceptor ExecAttemptInterceptor
 
 	// QueryObserver will set the provided query observer on all queries created from this session.
 	// Use it to collect metrics / stats from queries by providing an implementation of QueryObserver.
--- a/cluster.go
+++ b/cluster.go
@@ -258,6 +258,10 @@ type ClusterConfig struct {
 	// See https://issues.apache.org/jira/browse/CASSANDRA-10786
 	DisableSkipMetadata bool
 
+	// QueryAttemptInterceptor will set the provided query interceptor on all queries created from this session.
+	// Use it to intercept and modify queries by providing an implementation of QueryAttemptInterceptor.
+	QueryAttemptInterceptor QueryAttemptInterceptor
+
 	// QueryObserver will set the provided query observer on all queries created from this session.
 	// Use it to collect metrics / stats from queries by providing an implementation of QueryObserver.
 	QueryObserver QueryObserver
--- a/example_interceptor_test.go
+++ b/example_interceptor_test.go
@@ -55,7 +55,7 @@ func (q MyQueryAttemptInterceptor) Intercept(
 
 	// Optionally bypass the handler and return an error to prevent query execution.
 	// For example, to simulate query timeouts.
-	if q.injectFault && attempt.Number == 1 {
+	if q.injectFault && attempt.Attempts == 0 {
 		<-time.After(1 * time.Second)
 		return gocql.NewIterWithErr(gocql.RequestErrWriteTimeout{})
 	}
--- a/example_interceptor_test.go
+++ b/example_interceptor_test.go
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/*
+ * Content before git sha 34fdeebefcbf183ed7f916f931aa0586fdaa1b40
+ * Copyright (c) 2016, The Gocql authors,
+ * provided under the BSD-3-Clause License.
+ * See the NOTICE file distributed with this work for additional information.
+ */
+
+package gocql_test
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	gocql "github.com/gocql/gocql"
+)
+
+type MyQueryAttemptInterceptor struct {
+	injectFault bool
+}
+
+func (q MyQueryAttemptInterceptor) Intercept(
+	ctx context.Context,
+	query gocql.ExecutableQuery,
+	conn *gocql.Conn,
+	handler gocql.QueryAttemptHandler,
+) *gocql.Iter {
+	switch q := query.(type) {
+	case *gocql.Query:
+		// Inspect or modify query
+		query = q
+	case *gocql.Batch:
+		// Inspect or modify batch
+		query = q
+	}
+
+	// Inspect or modify context
+	ctx = context.WithValue(ctx, "trace-id", "123")
+
+	// Optionally bypass the handler and return an error to prevent query execution.
+	// For example, to simulate query timeouts.
+	if q.injectFault && query.Attempts() == 0 {
+		<-time.After(1 * time.Second)
+		return gocql.NewIterWithErr(gocql.RequestErrWriteTimeout{})
+	}
+
+	// The interceptor *must* invoke the handler to execute the query.
+	return handler(ctx, query, conn)
+}
+
+// Example_interceptor demonstrates how to implement a QueryAttemptInterceptor.
+func Example_interceptor() {
+	cluster := gocql.NewCluster("localhost:9042")
+	cluster.QueryAttemptInterceptor = MyQueryAttemptInterceptor{injectFault: true}
+
+	session, err := cluster.CreateSession()
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer session.Close()
+
+	ctx := context.Background()
+
+	var stringValue string
+	err = session.Query("select now() from system.local").
+		WithContext(ctx).
+		RetryPolicy(&gocql.SimpleRetryPolicy{NumRetries: 2}).
+		Scan(&stringValue)
+	if err != nil {
+		log.Fatalf("query failed %T", err)
+	}
+	fmt.Println(stringValue)
+	// Output: MOOOO!
+}
--- a/example_interceptor_test.go
+++ b/example_interceptor_test.go
@@ -26,7 +26,6 @@ package gocql_test
 
 import (
 	"context"
-	"fmt"
 	"log"
 	"time"
 
@@ -39,17 +38,16 @@ type MyQueryAttemptInterceptor struct {
 
 func (q MyQueryAttemptInterceptor) Intercept(
 	ctx context.Context,
-	query gocql.ExecutableQuery,
-	conn *gocql.Conn,
+	attempt gocql.QueryAttempt,
 	handler gocql.QueryAttemptHandler,
 ) *gocql.Iter {
-	switch q := query.(type) {
+	switch q := attempt.Query.(type) {
 	case *gocql.Query:
 		// Inspect or modify query
-		query = q
+		attempt.Query = q
 	case *gocql.Batch:
 		// Inspect or modify batch
-		query = q
+		attempt.Query = q
 	}
 
 	// Inspect or modify context
@@ -57,13 +55,13 @@ func (q MyQueryAttemptInterceptor) Intercept(
 
 	// Optionally bypass the handler and return an error to prevent query execution.
 	// For example, to simulate query timeouts.
-	if q.injectFault && query.Attempts() == 0 {
+	if q.injectFault && attempt.Number == 1 {
 		<-time.After(1 * time.Second)
 		return gocql.NewIterWithErr(gocql.RequestErrWriteTimeout{})
 	}
 
 	// The interceptor *must* invoke the handler to execute the query.
-	return handler(ctx, query, conn)
+	return handler(ctx, attempt)
 }
 
 // Example_interceptor demonstrates how to implement a QueryAttemptInterceptor.
@@ -87,6 +85,4 @@ func Example_interceptor() {
 	if err != nil {
 		log.Fatalf("query failed %T", err)
 	}
-	fmt.Println(stringValue)
-	// Output: MOOOO!
 }
--- a/example_interceptor_test.go
+++ b/example_interceptor_test.go
@@ -61,13 +61,64 @@ func (q MyQueryAttemptInterceptor) Intercept(
 	}
 
 	// The interceptor *must* invoke the handler to execute the query.
-	return handler(ctx, attempt), nil
+	return handler(ctx, attempt)
 }
 
 // Example_interceptor demonstrates how to implement a QueryAttemptInterceptor.
 func Example_interceptor() {
 	cluster := gocql.NewCluster("localhost:9042")
 	cluster.QueryAttemptInterceptor = MyQueryAttemptInterceptor{injectFault: true}
+
+	session, err := cluster.CreateSession()
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer session.Close()
+
+	ctx := context.Background()
+
+	var stringValue string
+	err = session.Query("select now() from system.local").
+		WithContext(ctx).
+		RetryPolicy(&gocql.SimpleRetryPolicy{NumRetries: 2}).
+		Scan(&stringValue)
+	if err != nil {
+		log.Fatalf("query failed %T", err)
+	}
+}
+
+type QueryAttemptInterceptorChain struct {
+	interceptors []gocql.QueryAttemptInterceptor
+}
+
+func (c QueryAttemptInterceptorChain) Intercept(
+	ctx context.Context,
+	attempt gocql.QueryAttempt,
+	handler gocql.QueryAttemptHandler,
+) (*gocql.Iter, error) {
+	return c.interceptors[0].Intercept(ctx, attempt, c.getNextHandler(0, handler))
+}
+
+func (c QueryAttemptInterceptorChain) getNextHandler(curr int, final gocql.QueryAttemptHandler) gocql.QueryAttemptHandler {
+	if curr == len(c.interceptors)-1 {
+		return final
+	}
+
+	return func(ctx context.Context, attempt gocql.QueryAttempt) (*gocql.Iter, error) {
+		return c.interceptors[curr+1].Intercept(ctx, attempt, c.getNextHandler(curr+1, final))
+	}
+}
+
+// Example_interceptor_chain demonstrates how to chain QueryAttemptInterceptors.
+func Example_interceptor_chain() {
+	cluster := gocql.NewCluster("localhost:9042")
+	cluster.QueryAttemptInterceptor = QueryAttemptInterceptorChain{
+		[]gocql.QueryAttemptInterceptor{
+			MyQueryAttemptInterceptor{},
+			MyQueryAttemptInterceptor{},
+			MyQueryAttemptInterceptor{},
+		},
+	}
 
 	session, err := cluster.CreateSession()
 	if err != nil {
--- a/example_interceptor_test.go
+++ b/example_interceptor_test.go
@@ -40,7 +40,7 @@ func (q MyQueryAttemptInterceptor) Intercept(
 	ctx context.Context,
 	attempt gocql.QueryAttempt,
 	handler gocql.QueryAttemptHandler,
-) *gocql.Iter {
+) (*gocql.Iter, error) {
 	switch q := attempt.Query.(type) {
 	case *gocql.Query:
 		// Inspect or modify query
@@ -57,11 +57,11 @@ func (q MyQueryAttemptInterceptor) Intercept(
 	// For example, to simulate query timeouts.
 	if q.injectFault && attempt.Attempts == 0 {
 		<-time.After(1 * time.Second)
-		return gocql.NewIterWithErr(gocql.RequestErrWriteTimeout{})
+		return nil, gocql.RequestErrWriteTimeout{}
 	}
 
 	// The interceptor *must* invoke the handler to execute the query.
-	return handler(ctx, attempt)
+	return handler(ctx, attempt), nil
 }
 
 // Example_interceptor demonstrates how to implement a QueryAttemptInterceptor.
--- a/example_interceptor_test.go
+++ b/example_interceptor_test.go
@@ -29,7 +29,7 @@ import (
 	"log"
 	"time"
 
-	gocql "github.com/gocql/gocql"
+	gocql "github.com/apache/cassandra-gocql-driver/v2"
 )
 
 type MyQueryAttemptInterceptor struct {
@@ -41,13 +41,14 @@ func (q MyQueryAttemptInterceptor) Intercept(
 	attempt gocql.QueryAttempt,
 	handler gocql.QueryAttemptHandler,
 ) (*gocql.Iter, error) {
-	switch q := attempt.Query.(type) {
+	switch q := attempt.Statement.Statement().(type) {
 	case *gocql.Query:
-		// Inspect or modify query
-		attempt.Query = q
+		// Inspect query
+		log.Println(q.Statement())
 	case *gocql.Batch:
-		// Inspect or modify batch
-		attempt.Query = q
+		// Inspect batch
+
+		log.Println(q.Entries[0].Stmt)
 	}
 
 	// Inspect or modify context
@@ -61,7 +62,7 @@ func (q MyQueryAttemptInterceptor) Intercept(
 	}
 
 	// The interceptor *must* invoke the handler to execute the query.
-	return handler(ctx, attempt)
+	return handler(ctx)
 }
 
 // Example_interceptor demonstrates how to implement a QueryAttemptInterceptor.
@@ -79,9 +80,8 @@ func Example_interceptor() {
 
 	var stringValue string
 	err = session.Query("select now() from system.local").
-		WithContext(ctx).
 		RetryPolicy(&gocql.SimpleRetryPolicy{NumRetries: 2}).
-		Scan(&stringValue)
+		ScanContext(ctx, &stringValue)
 	if err != nil {
 		log.Fatalf("query failed %T", err)
 	}
@@ -96,16 +96,16 @@ func (c QueryAttemptInterceptorChain) Intercept(
 	attempt gocql.QueryAttempt,
 	handler gocql.QueryAttemptHandler,
 ) (*gocql.Iter, error) {
-	return c.interceptors[0].Intercept(ctx, attempt, c.getNextHandler(0, handler))
+	return c.interceptors[0].Intercept(ctx, attempt, c.getNextHandler(0, attempt, handler))
 }
 
-func (c QueryAttemptInterceptorChain) getNextHandler(curr int, final gocql.QueryAttemptHandler) gocql.QueryAttemptHandler {
+func (c QueryAttemptInterceptorChain) getNextHandler(curr int, attempt gocql.QueryAttempt, final gocql.QueryAttemptHandler) gocql.QueryAttemptHandler {
 	if curr == len(c.interceptors)-1 {
 		return final
 	}
 
-	return func(ctx context.Context, attempt gocql.QueryAttempt) (*gocql.Iter, error) {
-		return c.interceptors[curr+1].Intercept(ctx, attempt, c.getNextHandler(curr+1, final))
+	return func(ctx context.Context) (*gocql.Iter, error) {
+		return c.interceptors[curr+1].Intercept(ctx, attempt, c.getNextHandler(curr+1, attempt, final))
 	}
 }
 
@@ -130,9 +130,8 @@ func Example_interceptor_chain() {
 
 	var stringValue string
 	err = session.Query("select now() from system.local").
-		WithContext(ctx).
 		RetryPolicy(&gocql.SimpleRetryPolicy{NumRetries: 2}).
-		Scan(&stringValue)
+		ScanContext(ctx, &stringValue)
 	if err != nil {
 		log.Fatalf("query failed %T", err)
 	}
--- a/example_interceptor_test.go
+++ b/example_interceptor_test.go
@@ -32,11 +32,11 @@ import (
 	gocql "github.com/apache/cassandra-gocql-driver/v2"
 )
 
-type MyQueryAttemptInterceptor struct {
+type MyExecAttemptInterceptor struct {
 	injectFault bool
 }
 
-func (q MyQueryAttemptInterceptor) Intercept(
+func (q MyExecAttemptInterceptor) Intercept(
 	ctx context.Context,
 	attempt gocql.QueryAttempt,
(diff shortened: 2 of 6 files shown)


cluster.go
```diff
@@ -258,6 +258,10 @@ type ClusterConfig struct {
 	// See https://issues.apache.org/jira/browse/CASSANDRA-10786
 	DisableSkipMetadata bool
 
+	// QueryAttemptInterceptor will set the provided query interceptor on all queries created from this session.
+	// Use it to intercept and modify queries by providing an implementation of QueryAttemptInterceptor.
+	QueryAttemptInterceptor QueryAttemptInterceptor
```
worryg0d, line 263: nit: The **Query** prefix might create a wrong impression that this specific option is related to `gocql.Query` request types only. I'm unsure which name is actually better here, but we can mention both `gocql.Batch` and `gocql.Query` in the doc at least.

example_interceptor_test.go
```diff
@@ -0,0 +1,138 @@
...
+	var stringValue string
+	err = session.Query("select now() from system.local").
+		RetryPolicy(&gocql.SimpleRetryPolicy{NumRetries: 2}).
+		ScanContext(ctx, &stringValue)
+	if err != nil {
+		log.Fatalf("query failed %T", err)
+	}
+}
+
+type QueryAttemptInterceptorChain struct {
```
worryg0d, line 90: I think we can make this a part of gocql api - we already expose a couple of similar types for event listeners: `SessionReadyListenersMux `, `SchemaListenersMux`, etc.

example_interceptor_test.go
```diff
@@ -0,0 +1,138 @@
...
+) (*gocql.Iter, error) {
+	switch q := attempt.Statement.Statement().(type) {
+	case *gocql.Query:
+		// Inspect query
+		log.Println(q.Statement())
+	case *gocql.Batch:
+		// Inspect batch
+
+		log.Println(q.Entries[0].Stmt)
+	}
```
worryg0d, line 52: Basically, this allows in-place modification of the submitted Query / Batch, and it violates the immutable nature that the driver follows, so changes made by interceptors might make Query / Batch objects potentially unreusable.

Is there any particular use case when you want to use the ability of interceptors to modify queries over their configuration API?

worryg0d (commented):
Hey,

I'm unsure if this PR is ready for review, but I just looked at it. I like the overall idea of introducing a generic enough solution that enables us to implement request monitors, rate limiting, etc., but I'm concerned about the ability of in-place mutation of request objects. What are the actual use cases when you need to modify requests in interceptors instead of using the request configuration API?

I didn't dive deep into the implementation details, though, and mostly reviewed the public API.

coxley:
It would be nice if there was a way to determine whether an observed query/attempt was itself triggered by speculative execution vs. retry policy. There's no reliable way to determine that today which makes it hard to observe behavior changes at scale in relation to tweaking one or the other.

cluster.go
abenn135: Changed this to `ExecAttemptInterceptor`, which is more generic.

example_interceptor_test.go
abenn135: We could... this was originally intended as sample code, and I don't know how many customers will actually need to compose multiple interceptors at a time. Do you want me to move this to `query_executor.go`?

example_interceptor_test.go
abenn135: I have to apologize here -- the comments in cluster.go and doc.go were out of date. Due to refactors in the v2 API, creation of the `internalRequest` is upstream of exec invocation, and this is limits our ability to mutate the query/batch in the interceptor without re-exposing that class or more dramatic refactor. The query/batch are not mutable in the interceptor.

The interceptor still provides unique value over an Observer because it is invoked before exec() and can fail, providing an opportunity for e.g. load shedding.

abenn135:
> It would be nice if there was a way to determine whether an observed query/attempt was itself triggered by speculative execution vs. retry policy. There's no reliable way to determine that today which makes it hard to observe behavior changes at scale in relation to tweaking one or the other.

Sure, we could modify `QueryAttempt` to pass that information in. What do you think?

example_interceptor_test.go
worryg0d: I've just written a basic test to confirm that requests are immutable, and I'm able to mutate them in an interceptor:
```go
type simpleInterceptor struct {
	logger StructuredLogger
}

func (s *simpleInterceptor) Intercept(ctx context.Context, attempt QueryAttempt, handler QueryAttemptHandler) (*Iter, error) {
	switch attempt.Statement.Statement().(type) {
	case *Query:
		query := attempt.Statement.Statement().(*Query)
		s.logger.Info("Intercepting query", NewLogFieldString("statement", query.Statement()))
		query.Consistency(One)
	default:
	}
	return handler(ctx)
}

func Test_QueryObjectImmutabilityInInterceptor(t *testing.T) {
	interceptor := &simpleInterceptor{logger: NewLogger(LogLevelInfo)}
	session := createSession(t, func(config *ClusterConfig) {
		config.ExecAttemptInterceptor = interceptor
	})
	defer session.Close()

	expectedConsistency := Quorum
	query := session.Query("SELECT host_id FROM system.local").Consistency(expectedConsistency)

	var hostID string
	err := query.Scan(&hostID)
	if err != nil {
		t.Fatalf("Failed to scan host ID: %v", err)
	}

	require.Equal(t, expectedConsistency.String(), query.GetConsistency().String(), "Query object should not be mutated by the interceptor")
}
```

Output:
```
2026/05/05 11:28:35 logger.go:165: INF gocql: Intercepting query statement=DROP KEYSPACE IF EXISTS gocql_test
2026/05/05 11:28:37 logger.go:165: INF gocql: Intercepting query statement=CREATE KEYSPACE gocql_test
	WITH replication = {
		'class' : 'SimpleStrategy',
		'replication_factor' : 1
	}
2026/05/05 11:28:39 logger.go:165: INF gocql: Intercepting query statement=SELECT host_id FROM system.local
--- FAIL: Test_QueryObjectImmutabilityInInterceptor (3.73s)
    /home/worry/projects/go/github.com/worryg0d/gocql/integration_test.go:1013: 
        	Error Trace:	/home/worry/projects/go/github.com/worryg0d/gocql/integration_test.go:1013
        	Error:      	Not equal: 
        	            	expected: "QUORUM"
        	            	actual  : "ONE"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-QUORUM
        	            	+ONE
        	Test:       	Test_QueryObjectImmutabilityInInterceptor
        	Messages:   	Query object should not be mutated by the interceptor
FAIL
FAIL	github.com/apache/cassandra-gocql-driver/v2	3.736s
```

example_interceptor_test.go
worryg0d: I'm actually neutral on this, but we already have two use-cases where we could use interceptors: monitoring and rate-limiting, so having chain out of the box looks reasonable to me. If nobody has objections, you may make it part of gocql public api.

I think `query_executor.go` is fine.

abenn135: Done, thanks.

example_interceptor_test.go
abenn135: Yikes, I'm sorry, I didn't realize until now that Go abstract type members keep a pointer (not a copy) of the underlying type.

Looking at how we would use this internally, we're hoping to have the following at intercept time:

* the ctx that is being used to invoke `Exec()`
* the statement body itself (or an exemplar for a `Batch`)
* the host ID it will be sent to (for testing fault injection)
* an ability to set the trace per attempt

It might also be useful for other customers to learn other values on the `Query`, such as current consistency setting, idempotency setting, and keyspace. Perhaps the play is to do the following:

* provide a way to optionally set the trace (we just need the context to do this, though others might want other information)
* provide a read-only deep copy of just those fields of the Query or Batch that might be useful for interception

What do you think?

abenn135:
> It would be nice if there was a way to determine whether an observed query/attempt was itself triggered by speculative execution vs. retry policy.

Done in latest change.

example_interceptor_test.go
worryg0d: I see the value of having such feature - I'm just concerned that having an ability to modify underlying request objects would lead to kinda api misuse... Providing a deep copy for read would resolve this, but I'm unsure about performance impact because it is on a request hot path.

I think we need another pair of eyes on this specific case. 

@joao-r-reis hey, would you mind take a look at this if you have free capacity? This PR overlaps with query immutabillity you have been working on.

coxley:
@abenn135 Thank you very much! :)

```go
// The index of the speculative execution attempt this attempt is associated 
// with, starting with index 1. -1 indicates this is the "main" execution.
SpeculativeExecutionCount int
```

Forgive me if this is me not following well, but what does "index" mean in this case? For example:

- First attempt failed and `RetryPolicy` said to re-attempt
- Second attempt taking longer than `(gocql.SpeculativeExecutionPolicy).Delay()`
    - Assume `(gocql.SpeculativeExecutionPolicy).Attempts()` returns 2
- Third attempt is a hedge and is taking longer than the delay
- Fourth attempt is a hedge

What does `gocql.QueryAttempt.SpeculativeExecutionCount` look like at each stage? The field name makes me think that it's the cumulative count of speculative attempts so far, but the doc comment made me question.

example_interceptor_test.go
abenn135: I think I have an idea of how you could do this...

1. Add read-only QueryView/BatchView interfaces that only offer the getters on Query/Batch.
2. Move all members on Query and Batch to new privateQuery/privateBatch structs. These structs offer the public getters and therefore implement QueryView/BatchView APIs, but have package-private setters.
3. Update Query and Batch to just hold a ref to an underlying privateQuery/privateBatch, and have their public getters and setters proxy to the public getters and package-private setters on their private* members.
4. At Exec/Intercept time, you offer only the privateQuery/privateBatch, which implement the QueryView/BatchView public getters, but do not implement the setters on the Query/Batch interfaces.

This seems like it would work to provide an immutable view of the Query/Batch without a deep copy. We'd also need an affordance to set the trace, which might be done through a specific callback func or something.

Let me know what you think of this. I could do it in a separate PR if you prefer.

abenn135:
> Forgive me if this is me not following well, but what does "index" mean in this case? For example:
>
> * First attempt failed and RetryPolicy said to re-attempt

First attempt:
`Attempts: 0 SpeculativeExecutionCount: -1`

> Second attempt taking longer than (gocql.SpeculativeExecutionPolicy).Delay()

Second attempt:
`Attempts: 1 SpeculativeExecutionCount: -1`

> Assume (gocql.SpeculativeExecutionPolicy).Attempts() returns 2
> Third attempt is a hedge and is taking longer than the delay

Well, here you have two goroutines that might be executing concurrently. But the first hedge call will see `Attempts: 2 SpeculativeExecutionCount: 1`.

> Fourth attempt is a hedge

Again, each speculation spawns its own goroutine, so based on your description I think there may now be as many as three concurrent attempts. Either way, if the fourth attempt is another hedge, it would see `Attempts: 3 SpeculativeExecutionCount: 2`

Attempts is incremented each time `attemptQuery()` is called. It may increment due to any of:

* per-host timeout or something (`GetRetryType()` returns `Retry`)
* host says "go away" and we try the next eligible host (`GetRetryType()` returns `RetryNextHost`)
* speculation attempts

Maybe there is a better word than "index" here, but "index" pretty much describes "which speculation this is", i.e. is this the first speculative attempt/goroutine, the second, or so forth.

I avoided zero to minimize confusion (customers might test on <0 or <=0 to test whether this is a non-speculative attempt) but if you'd like to start counting at 0 we can do that instead.

abenn135:
Alright, after noodling on this a LONG time, I figured out a simple way to make query/batch access immutable. Please take a look when you have a moment. Thank you for your patience!

example_interceptor_test.go
joao-r-reis: > @joao-r-reis hey, would you mind take a look at this if you have free capacity? This PR overlaps with query immutabillity you have been working on.

Sorry haven't been able to take a look at this yet, between the gocql patch releases and the other project I'm working on I haven't been able to look at this yet.

toptobes:
Hi folks, I'm just doing an initial pass-through on behalf of Joao to provide some initial feedback before he gets a chance to look at it.

I'm not super experienced with the gocql driver itself, but from a purely Go point-of-view, it looks pretty solid. There's just a few relatively minor things I'll note in separate, dedicated comments.

One of my major takeaways from this though, was the potential for exec intercepters to effectively replace `[Batch|Query]Observer`s with a little work, for the sake of keeping the API simple for users, rather than having two close/competing APIs
- To be clear, I'm not talking about deleting `[Batch|Query]Observer` immedietly, I'm just suggesting we could somewhat deprecate them or otherwise put them on the back burner, especially as it's marked w/ an `// Experimental, this interface and use may change` comment. We could fairly easily just keep them and actually reimplement them under the hood as exec interceptors so we don't need to duplicate related code/mechanisms while keeping them entirely backwards compatable

**Now to be clear, I'm just throwing this out there as an idea, we can discuss further specifics if we like it; we can trash it instantly if we don't.**

One way of doing so would be to make `Intercept` return a `(ExecAttemptResult, error)` instead of `(*Iter, error)` where there's a few different ways you can implement `ExecAttemptResult`, but a simple version could be something like:
```go
type ExecAttemptResult struct {
  Iter *Iter
  q    *internalQuery
  b    *internalBatch
}

// lazy constructed for performance since they're not always needed
// using the existing ObservedQuery & ObservedBatch for easy migration & compatability
// these could also take a `e *ExecAttemptResult` to cache the results
func (e ExecAttemptResult) GetObservedQuery() ObservedQuery { ... }
func (e ExecAttemptResult) GetObservedBatch() ObservedBatch { ... }
```

Or you could just have the info directly in the `ExecAttemptResult` instead of using the `Observed[Query|Batch]` types. Whatever works. The main idea is to stay lazy so a price isn't paid on the hot path if it's not needed.

Then you'd just need to 
- Allow setting exec interceptors on a per-query basis (potentially auto-chaining them? though that could be unexpected)
- Figure out a heuristic for when to use `newHostMetricsManager` (maybe check if the interceptor satisfies an interface like `interface { RequiresHostMetrics() bool }` + returns `true`) (but also it doesn't look very performance intensive to use `newHostMetricsManager` so maybe it should just be enabled on any interceptor being set :shrug:)
- Migrate observers to be interceptors under the hood for simplicity

Maybe there's a few other small things you'd need to do (like additional setters/getters), but I don't think I'm missing anything major.

query_executor.go
```diff
@@ -69,16 +71,151 @@ type internalRequest interface {
...
 
-func (q *queryExecutor) attemptQuery(ctx context.Context, qry internalRequest, conn *Conn) *Iter {
+type OpType int
+
+const (
+	OpQuery OpType = iota
+	OpBatch
+)
+
+type ImmutableQuery interface {
```
toptobes, line 86: Very subjective, but I honestly preferred the original `[Query|Batch]View` names you suggested, they do a better job of representing exactly what's happening here (+ it __technically__ isn't entirely immutable as the returned slices can still be mutated, but I don't think it's worth creating defensive copies)

query_executor.go
```diff
@@ -69,16 +71,151 @@ type internalRequest interface {
...
+
+func (b *immutableBatch) Entries() []BatchEntry {
+	return b.batch.Entries
+}
+
+func (b *immutableBatch) Cons() Consistency {
+	return b.batch.Cons
+}
+
+type QueryAttempt struct {
```
toptobes, line 125: Should probably be called `ExecAttempt` to stay consistent & generic (otherwise it's easy to mistake as purely query-only w/ there being another batch-only variant)

query_executor.go
```diff
@@ -69,16 +71,151 @@ type internalRequest interface {
...
+	// The remote address of the connection used to execute the query.
+	RemoteAddr net.Addr
+	// The number of this query attempt. 0 for the initial attempt, 1 for the first retry, etc. Note that there may be multiple serial Attempts to different hosts within a single execution, whether main execution or speculative.
+	Attempts int
+	// The index of the speculative execution attempt this attempt is associated with, starting with index 1. -1 indicates this is the "main" execution.
+	SpeculativeExecutionCount int
+}
+
+// QueryAttemptHandler is a function that attempts query execution. The interceptor must call this once if it does not return an error.
+type QueryAttemptHandler = func(context.Context) (*Iter, error)
```
toptobes, line 144: Similarly should be called `ExecAttemptHandler`

query_executor.go
```diff
@@ -69,16 +71,151 @@ type internalRequest interface {
...
+type ExecAttemptInterceptorChain struct {
+	Interceptors []ExecAttemptInterceptor
+}
+
+func (c ExecAttemptInterceptorChain) Intercept(
+	ctx context.Context,
+	attempt QueryAttempt,
+	handler QueryAttemptHandler,
+) (*Iter, error) {
+	return c.Interceptors[0].Intercept(ctx, attempt, c.getNextHandler(0, attempt, handler))
```
toptobes, line 171: I believe this will panic without mercy if you create an empty chain (didn't see any checks which would stop `Intercept` being called here in case of an empty chain)

toptobes: I think just doing `return c.getNextHandler(-1, attempt, handler)(ctx)` should work?

query_executor.go
```diff
@@ -69,16 +71,151 @@ type internalRequest interface {
...
+		if iBatch, ok := iRequest.(*internalBatch); ok {
+			iq.Type = OpBatch
+			iq.Batch = &immutableBatch{batch: iBatch.originalBatch}
+		}
+		iter, err = q.interceptor.Intercept(ctx, iq, func(cxCtx context.Context) (*Iter, error) {
+			it := iRequest.execute(cxCtx, conn)
+			return it, it.err
+		})
+		if err != nil {
+			iter = &Iter{err: err}
```
toptobes, line 211: Should this use `newErrIter` instead? There's nowhere else where in `Iter` is created directly. I believe `iRequest` has everything you need to construct it

toptobes (changes requested):

example_interceptor_test.go
joao-r-reis: I think it's fine if the original Batch and Query objects are exposed here, the observers already do this and they have a comment on the field to let users know they should only use these as read only (`// Batch object associated with this request. Should be used as read only.`). It's not ideal but it's the best we can do until we have query and batch types that are **actually** immutable without creating an overcomplicated API.

In regards to allowing interceptors to do things like setting the trace I think that as long as we have the `QueryAttempt` struct we would have a way to add this type of functionality in the future without having to add more ways to mutate Query/Batch (e.g. add a field to the struct that interceptor implementations are expected to set).

query_executor.go
joao-r-reis: Even with this API you can still modify the slices that are being returned so I don't think there's a way to prevent the interceptors from modifying the query/batch completely unless we deep clone these but this would have to be opt-in so that applications that don't need to modify them don't have to suffer the performance penalty that comes with copying these slices on every request. And to make this opt-in the API would have to be much more complicated and I don't think it's worth going there.

The observers already expose `Query` and `Batch` directly so I think we should do the same here with a comment mentioning that they should be treated as read only.

In the future the goal is to have proper immutable Query and Batch types and this issue will go away then.

query_executor.go
```diff
@@ -69,16 +71,151 @@ type internalRequest interface {
...
+	policy      HostSelectionPolicy
+	interceptor ExecAttemptInterceptor
 }
 
-func (q *queryExecutor) attemptQuery(ctx context.Context, qry internalRequest, conn *Conn) *Iter {
+type OpType int
+
+const (
+	OpQuery OpType = iota
+	OpBatch
```
joao-r-reis, line 83: We already have a `Statement` interface that is implemented by both `Query` and `Batch`, maybe we should rename `OpType` to `StatementType`

query_executor.go
```diff
@@ -69,16 +71,151 @@ type internalRequest interface {
...
+// Interceptors are well-suited to logic that is not specific to a single query or batch, such as flow control.
+type ExecAttemptInterceptor interface {
+	// Intercept is invoked once immediately before a query or batch execution attempt, including retry attempts and
+	// speculative execution attempts.
+	//
+	// The interceptor is responsible for calling the `handler` function and returning the handler result. If the
+	// interceptor wants to bypass the handler and skip query execution, it should return an error. Failure to
+	// return either the handler result or an error will panic.
+	//
+	// Note that there is no affordance to mutate the query or batch at this stage -- the handler already encapsulates the original query/batch and cannot be modified.
```
joao-r-reis, line 157: This comment is not 100% accurate but it will have to be modified regardless once my feedback on this topic is addressed

query_executor.go
```diff
@@ -69,16 +71,151 @@ type internalRequest interface {
...
+	Intercept(ctx context.Context, attempt QueryAttempt, handler QueryAttemptHandler) (*Iter, error)
```
joao-r-reis, line 158: I'm not sure I like having the interceptor being responsible for calling the driver code that executes the request. I know this is following the "middleware" way of doing things but I'd prefer having two methods on this interface like `BeforeAttempt` and `AfterAttempt` so the driver is in full control of the request execution and just "checks in" with the interceptor at the relevant points in time.

query_executor.go
```diff
@@ -69,16 +71,151 @@ type internalRequest interface {
...
+}
+
+// Use an ExecAttemptInterceptorChain to apply multiple Interceptors at Intercept time.
+type ExecAttemptInterceptorChain struct {
```
joao-r-reis, line 162: If we want to follow the https://github.com/apache/cassandra-gocql-driver/pull/1929 precedent of adding multiplexer types for this type of feature then we should probably use similar names. In this case it would probably be `ExecAttemptInterceptorMux`

query_executor.go
```diff
@@ -69,16 +71,151 @@ type internalRequest interface {
...
+	SpeculativeExecutionCount int
+}
+
+// QueryAttemptHandler is a function that attempts query execution. The interceptor must call this once if it does not return an error.
+type QueryAttemptHandler = func(context.Context) (*Iter, error)
+
+// ExecAttemptInterceptor is the interface implemented by interceptors / middleware.
+//
+// Interceptors are well-suited to logic that is not specific to a single query or batch, such as flow control.
+type ExecAttemptInterceptor interface {
```
joao-r-reis, line 149: We could probably get away with naming this `RequestInterceptor` no? And renaming the related types as well

joao-r-reis:
> One of my major takeaways from this though, was the potential for exec intercepters to effectively replace `[Batch|Query]Observer`s with a little work, for the sake of keeping the API simple for users, rather than having two close/competing APIs
> 
>   * To be clear, I'm not talking about deleting `[Batch|Query]Observer` immedietly, I'm just suggesting we could somewhat deprecate them or otherwise put them on the back burner, especially as it's marked w/ an `// Experimental, this interface and use may change` comment. We could fairly easily just keep them and actually reimplement them under the hood as exec interceptors so we don't need to duplicate related code/mechanisms while keeping them entirely backwards compatable

I think this is something we should keep in mind going forward but we can work on this new Interceptor feature without having to worry too much about the Observers right now. Conceptually they are two different things and they can exist independently of each other but in the future we might try to just reimplement observers so they are interceptors under the hood. Something to consider in the future but out of scope for this current PR

session.go
```diff
@@ -968,26 +969,49 @@ type hostMetrics struct {
 	TotalLatency int64
 }
 
+// Query attempts could be started and finished out of order when using speculative execution. Therefore,
+// we issue attempt indexes at query start time (so that Interceptors can use them) then record those attempts
+// at completion time (so that metrics can correctly express average latency).
 type queryMetrics struct {
-	totalAttempts int64
-	totalLatency  int64
+	l                 sync.RWMutex
```
joao-r-reis, line 976: Do we need this mutex? Can't we keep using the atomic operations?

session.go
```diff
@@ -968,26 +969,49 @@ type hostMetrics struct {
...
+	return attempt
 }
 
+func (qm *queryMetrics) recordAttempt(attempt int, addLatency time.Duration, s *Session) {
+	qm.l.Lock()
+	defer qm.l.Unlock()
+	qm.totalLatency += addLatency.Nanoseconds()
+	qm.attemptsCompleted++
+	if attempt > qm.attemptsCompleted && s != nil {
+		s.logger.Debug("attempt number is greater than total attempts completed, this should not happen", NewLogFieldInt("attempt", attempt), NewLogFieldInt("attemptsCompleted", qm.attemptsCompleted))
```
joao-r-reis, line 996: If multiple speculative executions are triggered before responses come back won't this happen? I think that's expected no?

session.go
```diff
@@ -968,26 +969,49 @@ type hostMetrics struct {
...
+	}
+}
+
+// Returns total number of attempts started.
 func (qm *queryMetrics) attempts() int {
-	return int(atomic.LoadInt64(&qm.totalAttempts))
+	qm.l.RLock()
+	defer qm.l.RUnlock()
+	return qm.attemptsStarted
```
joao-r-reis, line 1004: I know you mention that changing this from "attemptsCompleted" to "attemptsStarted" was a deliberate choice but I would prefer not changing the behavior of this for existing users with custom retry policies. We can discuss this on a new issue/PR dedicated to this topic if needed.

session.go
joao-r-reis: If you remove this log statement then the session argument can also be removed

joao-r-reis (commented):
The "attempts" system needs a complete rework so that retries are completely independent from speculative executions but I don't want to stop any changes made in this context if they are needed for a particular feature which seems to be the case here?

I see that you added a "speculative request index" variable called `speculativeExecutionCount` and this is only used to populate the `SpeculativeExecutionCount` field of `QueryAttempt` correct? This seems good to me and I don't think there's any potential side effects that could be problematic. However, I would probably rename `SpeculativeExecutionCount` to `SpeculativeExecutionIndex`, I think it's clearer.

The changes to query metrics attempts were a bit more confusing to me but I think I eventually understood the reason behind them. If I understand correctly the motive behind these changes is the fact that the interceptor needs an "attempt" count BEFORE the request execution which wasn't possible to get before. I think this mixed with the fact that the PR also changes the semantics so that retry policies now look at "attempts started" instead of "attempts completed" made it more confusing to me. I think changing these semantics should be out of scope of this PR and we can discuss this particular issue on a different issue/PR dedicated to it if needed.

joao-r-reis:
> > It would be nice if there was a way to determine whether an observed query/attempt was itself triggered by speculative execution vs. retry policy.
> 
> Done in latest change.

If I'm understanding correctly it's still not possible to know what triggered a particular query attempt right? Now it's possible to know which "execution" the attempt belongs to (main execution vs speculative execution) but the attempt counter is still shared between executions if I'm not mistaken (and it should stay that way for now until we do a proper rework of this).

query_executor.go
toptobes: Tbh I'm for `Chain` or `Stack` or something here. I could be way off, but it looks like the `Mux` types in #1929 (e.g. `SessionReadyListenersMux`) seem to be a collection of independent listeners being treated as one by the driver itself, whereas with `ExecAttemptInterceptorChain`, 1) each interceptor affects the next, and 2) it operates in a stack-like fashion where the first interceptor acts both first and last due to the nature of recursion which IMO makes it fundimentally a bit different from the Muxes

query_executor.go
toptobes: I mean it _technically_ intercepts both the request and the response so it's _technically_ more correct as is–though can't deny `RequestInterceptor` sounds cleaner if you prefer that tradeoff

If we do the `BeforeAttempt`/`AfterAttempt` things and separate out the interfaces it could be a different story

session.go
toptobes: No, atomics would be worse here
- Assuming you're talking about using an `atomic.[Value|Pointer]` over the whole struct, that would cause more allocations than necessary (where the necessary amount is zero), especially under contention (unlikely as that may be)
- Assuming you're talking about stores/loads on individual fields, then that sort of defeats the point of treating `queryMetrics` as one atomic unit, no? (since you can still have mismatched reads/writes then)

Mutexes under no contention would be as fast as/faster than atomic pointers under no contention anyways (and again it looks like contention would be highly unlikely here)

And if you're worried about someone forgetting to access the mutex before read/write then the actual queryMetrics could always be encapsulated behind higher order methods/functions which handle the mutex for you

toptobes:
> I think this is something we should keep in mind going forward but we can work on this new Interceptor feature without having to worry too much about the Observers right now. Conceptually they are two different things and they can exist independently of each other but in the future we might try to just reimplement observers so they are interceptors under the hood. Something to consider in the future but out of scope for this current PR

Perfectly fine with me but just to be blunt, I think it'd be hard to do this in the future without causing breaking changes unless the appropriate initial steps are taken in this PR (or at least in this release!), even if the full marriage isn't performed just yet. 

Regardless, I still don't think it'd be the worst idea to have per-query exec interceptor overrides (similar to observers), and think we should probably have additional setters/getters so you can access the interceptors from the cluster to reuse them in a query/batch-specific interceptor chain (if that made any sense)

query_executor.go
joao-r-reis: Yeah that's fair but I also suggested on a different comment that we should probably abandon the "middleware" pattern in favor of what the driver already uses for this sort of interface. In this scenario the "chain" aspect would be lost and so `Mux` type might make sense again.

query_executor.go
abenn135: Done.

query_executor.go
abenn135: Done.

query_executor.go
abenn135: Went with `RequestInterceptor`, sounds clean, not too confusing I don't think. Could also just do `ExecInterceptor` if you prefer.

query_executor.go
abenn135: Removed.

query_executor.go
abenn135: That seems sensible to me. For our purposes we only need pre-attempt; if an attempt fails or is retried we don't need to know post-attempt; and if it succeeds we can use the Observer pattern, though I recognize there might be an effort to merge the two.

query_executor.go
abenn135: I've changed the pattern to a simple for loop given how Intercept works after changing it. Let me know if you still feel like `Mux` is the right name here; I think it's still more of a chain since the output of one interceptor is passed to the input of the next one, so there is an intended order there.

query_executor.go
abenn135: Now that Intercept isn't expected to invoke the handler, this shouldn't panic. A for loop over zero items simply skips.

query_executor.go
abenn135: Done.

session.go
abenn135: +1 to what @toptobes said :)

session.go
abenn135: Removed.

session.go
abenn135: OK, reverted.

query_executor.go
abenn135: Done.

query_executor.go
abenn135: OK, removed entirely and went back to Query/Batch, with an admonition that they should be treated immutable.

PR #1946, opened 2026-04-24
CASSGO-100 Support frames batching into segment

worryg0d:
TBD


worryg0d:
I ended up with a duplicate of the current `writeCoalescer` behavior, but with segment-encoding specifics. Changes to the framing format are rare, so I think it is reasonable not to spend time implementing a flexible solution that would allow the application of encoders in flight. But I'm open to discussion here.

Also, I introduced `segmentReader` wrapper around the `ConnReader` interface, which bundles frame-related logic in `recv()` and segment-related logic.

Also, I added a `segmentCodec` that encapsulates segment encoding/decoding, but I do not really insist on merging it, so I am open to discussion.

conn.go
```diff
@@ -795,91 +774,14 @@ func (c *Conn) releaseStream(call *callReq) {
...
-
-	// Computing how many bytes of message left to read
-	bytesToRead := head.length - len(frame) + frameHeadSize
-
-	err = c.recvPartialFrames(buf, bytesToRead)
-	if err != nil {
-		return err
+func (c *Conn) maybeSwitchToSegments() {
+	if c.version >= protoVersion5 {
+		// Use segments writter which basically batches multiple frames into a single segment before flushing them to the connection.
```
joao-r-reis: writer*

segment_codec.go
```diff
@@ -0,0 +1,277 @@
+// segment_codec.go
```
joao-r-reis: license header

conn.go
```diff
@@ -2047,6 +1937,343 @@ func (c *Conn) awaitSchemaAgreementWithTimeout(ctx context.Context, timeout time
...
+	running := false
+
+	for {
+		select {
+		case <-sw.quit:
+			return
+		case req := <-sw.writeCh:
+			frame := req.data
+			if len(frame) > maxSegmentPayloadSize {
+				sw.flushBigFrameImmediately(req)
```
joao-r-reis, line 2044: If I'm reading this correctly this will ignore the order in which the frames come in. If there's already write requests that were appended then we should flush first.

conn.go
```diff
@@ -2047,6 +1937,343 @@ func (c *Conn) awaitSchemaAgreementWithTimeout(ctx context.Context, timeout time
 	return fmt.Errorf("gocql: cluster schema versions not consistent: %+v", schemas)
 }
 
+// segmentWriter allows batching multiple frames into a signle segment before flushing them to the connection.
```
joao-r-reis: single*

conn.go
```diff
@@ -2047,6 +1937,343 @@ func (c *Conn) awaitSchemaAgreementWithTimeout(ctx context.Context, timeout time
...
+}
+
+func (sw *segmentWriter) fitsSegment(frame []byte) bool {
+	return sw.totalFramesLength+len(frame) <= maxSegmentPayloadSize
+}
+
+// Flushes the current segment and writes the results to the result listeners.
+// Should be called before resetting the segment writer.
+func (sw *segmentWriter) flushCurrentSegment() {
+	framesBuf := make([]byte, 0, sw.totalFramesLength)
```
joao-r-reis: Can we not just reuse a single instance of `bytes.Buffer`? If I'm not mistaken this slice is only needed until `encodeAndWrite` below

conn.go
```diff
@@ -2047,6 +1937,343 @@ func (c *Conn) awaitSchemaAgreementWithTimeout(ctx context.Context, timeout time
...
+	for _, req := range sw.writeRequests {
```
joao-r-reis, line 2095: do we need `writeRequests`? can we not just append to the buffer directly when the request comes?

conn.go
```diff
@@ -2047,6 +1937,343 @@ func (c *Conn) awaitSchemaAgreementWithTimeout(ctx context.Context, timeout time
...
+}
+
+func newSegmentReader(r ConnReader, segmentCodec segmentCodec) *segmentReader {
+	return &segmentReader{
+		r:            r,
+		segmentCodec: segmentCodec,
+	}
+}
+
+// why do we have a write method for reader lol
```
joao-r-reis: `ConnReader` embeds `net.Conn` unfortunately which doesn't make sense. I actually tried to find whether `gocql.Conn.Write` is used anywhere and it's only used in test code I believe but that's slightly unrelated to this. 

This could definitely be improved as it makes the code harder to read but oh well. Maybe you can replace this comment with a TODO about refactoring the code to remove the need to embed `net.Conn` in `ConnReader` and remove unused methods like `gocql.Conn.Write()`

conn.go
```diff
@@ -2047,6 +1937,343 @@ func (c *Conn) awaitSchemaAgreementWithTimeout(ctx context.Context, timeout time
...
+}
+
+func (sr *segmentReader) readSegment() error {
+	segment, isSelfContained, err := sr.segmentCodec.decode(sr.r)
+	if err != nil {
+		// TODO: does only network related errors should result in connection closure?
+		// var verr net.Error
+		// if errors.As(err, &verr) {
+		// 	return nil, false, verr
+		// }
```
joao-r-reis: This TODO needs to be addressed. Personally I think it's fine to close the connection when the driver fails to encode/decode something here because it's either an error on the decode/encode code and most likely a driver bug or it's a connection error. A connection error should lead to connection closure and the encode/decode error should never happen so I don't see a lot of benefit in trying to handle it.

conn.go
```diff
@@ -2047,6 +1937,343 @@ func (c *Conn) awaitSchemaAgreementWithTimeout(ctx context.Context, timeout time
...
+		return err
+	}
+
+	if isSelfContained {
+		// Reset the buffer to the new segment
+		// It might contain multiple frames so Read should be called mutiple times to read all of them
```
joao-r-reis: multiple*

conn.go
```diff
@@ -2047,6 +1937,343 @@ func (c *Conn) awaitSchemaAgreementWithTimeout(ctx context.Context, timeout time
...
+// Called by readNonSelfContainedSegment.
+func (sr *segmentReader) readPartialFrames(dstBuf *bytes.Buffer, bytesToRead int) error {
+	for bytesToRead > 0 {
+		frame, isSelfContained, err := sr.segmentCodec.decode(sr.r)
+		if err != nil {
+			return err
+		}
+		// Expected to receive only non self-contained segments
+		if isSelfContained {
+			return errUnexpectedSelfcontainedSegment
```
joao-r-reis: errUnexpectedSelfContainedSegment*
also why does this one deserve an actual error type and not the other errors?

conn.go
```diff
@@ -2047,6 +1937,343 @@ func (c *Conn) awaitSchemaAgreementWithTimeout(ctx context.Context, timeout time
...
+		if totalLength := dstBuf.Len() + len(frame); totalLength > dstBuf.Cap() {
+			return fmt.Errorf("gocql: expected partial frame of length %d, got %d", dstBuf.Cap(), totalLength)
+		}
+		n, _ := dstBuf.Write(frame)
+		bytesToRead -= n
+	}
+
+	if bytesToRead < 0 {
+		// This should never happen actually
+		panic("gocql: something went wrong while reading partial frames")
```
joao-r-reis: We can return an error here just in case. It would lead to the connection closing instead of crashing the application which is better imo.

segment_codec.go
```diff
@@ -0,0 +1,277 @@
...
+	computed := Crc32(data)
+	if computed != expected {
+		return fmt.Errorf("gocql: payload crc32 mismatch in segment payload: expected %d, got %d", expected, computed)
+	}
+	return nil
+}
+
+// decodeCompressedSegmentHeader reads and verifies the header of a compressed segment from the given reader.
+func (sc *segmentCodec) decodeCompressedSegmentHeader(r io.Reader) (*segmentHeader, error) {
+	var headerBuf [8]byte // TODO: potentially optimize allocation, could be stored in segmentCodec and reused if the codec is a specific for each Conn
```
joao-r-reis: I agree with the TODO, it's the same with the read and write buffer we can just reuse them if each connection has their own.

joao-r-reis (commented):
Looking good overall but left a few comments. It would be nice to run benchmarks to compare the performance vs trunk before merging this.

I have some code I used to do a basic comparison between gocql v1 and v2 maybe it can be useful to you: https://github.com/joao-r-reis/gocql-benchmarks/tree/testv3

worryg0d:
Thanks for the review. I'm going to address it in a couple of days

conn.go
worryg0d: ConnReader is public, but it is very unlikely anybody relies on this, imo, as there is nothing in the driver that implements this interface and is exported. We probably just have to make it private

conn.go
worryg0d: The other errors rely on format

conn.go
worryg0d: If such a thing happens, there is a bug in the driver. Wondering if this message is good enough here `"gocql: driver encountered unexpected state while reading partial frames of the segment, read more bytes than expected: %d. This is a bug."`

conn.go
worryg0d: It indeed ignores the order. The Java driver also does the same thing here if I got it right: // https://github.com/datastax/native-protocol/blob/6b9bfb05c3fb1e29e74eec288dd54bd78232c2b7/src/main/java/com/datastax/oss/protocol/internal/SegmentBuilder.java#L74

I implemented this on top of what the Java driver does. Hmmm, might it be problematic in any case? In the worst case of having a large number of large frames and a small number of small frames, the worst that would happen is a flush of small frames on the CoalesceWait interval tick...

worryg0d: Yeah I guess it's better to flush small frames before flushing the big one:
```
				if running {
					running = false
					sw.flushCurrentSegment()
					sw.reset()
				}
				sw.flushBigFrameImmediately(req)
```

conn.go
worryg0d: You mean make part of the segment writer? I'm concerned about having a big backed slice for each conn. It's around 128 kb

conn.go
worryg0d: I'm thinking about timeouts, actually. Currently, the driver ignores the timeout when trying to read the frame header from the conn. With this patch, the driver will also ignore reading both the segment header and body, which differs from the previous driver behavior.

I'm unsure if this is good

conn.go
worryg0d: writeRequests holds response channels that it has to notify

conn.go
joao-r-reis: Sure but you could also just use `errors.New` instead. If this error is not being checked anywhere there doesn't seem to be any utility in having the global variable.

conn.go
joao-r-reis: Are you sure it ignores the timeout? Doesn't the Read loop on `connReader` return the error immediately instead of retrying when it's a read timeout?

conn.go
joao-r-reis: `"gocql: driver encountered unexpected state while reading partial frames of the segment, read more bytes than expected: %d; please report this bug to gocql maintainers"`

conn.go
joao-r-reis: The code in `writeFlusherImpl` has fewer loops for essentially the same thing because as soon as the writerequest comes in the buffer is appended to a `net.Buffers` object and the channel is appended to a `[]chan<-` object. Then the only loop that happens is to send the result to the channels after the write succeeds.

conn.go
worryg0d: https://github.com/apache/cassandra-gocql-driver/blob/c65c762b83eccdb6a6a083c123a5935c4302745c/conn.go#L904-L925

In case of net.Error it do not break if I'm not missing anything.

https://pkg.go.dev/net#Error tells us whether it is timeout or not. Actually, probably worth updating the code to rely on net.Error.Timeout() intstead of Temporary() as suggested

conn.go
joao-r-reis: It will potentially grow up to that size but if you're that concerned about it you can just resize it after a write forces the buffer to grow that large.

Since v4 code doesn't reuse a buffer either let's keep it this way and let's create a low priority ticket to prototype and analyze the performance impact of making both write coalescers reuse a buffer.

conn.go
joao-r-reis: Interesting that the java driver behaves that way but I personally think it's better to preserve the order. In a regular application you would probably not notice it ever but if an app is sending requests in a sequential order (without waiting for the response) and the server receives the requests in a different order I'd consider that odd even if it's technically not against the protocol spec.

conn.go
joao-r-reis: Yeah I misread it you're right. I think this will sort itself out when we implement the timeout fix which should go in (and get released) before this PR since this is not a bug fix. After the timeout fix is merged then the timeout will not even exist here.

segment_codec.go
worryg0d: Added reusable buffers for headers and payload checksums

conn.go
worryg0d: Yeah, if we get rid of the read socket timeouts it won't be a problem at all

conn.go
worryg0d: `writeFlusherImpl` relies on the `net.Buffers`, which supports writev operation, which `segmentWriter` never uses.

Probably, I should try to implement something like a composition buffer from Java, which allows having multiple underlying byte buffers but treats them as a single byte buffer. It prevents copying buffers to a bigger one in this specific place

worryg0d: Actually, `net.Buffers` probably fits for this case as it implements `io.Reader`

worryg0d: I'll think about this, but it requires me to refactor codecs to work with io.Reader instead of raw byte slices

conn.go
worryg0d: I've added unit tests for segment reader and writer, where this err is checked

worryg0d:
I have run a couple of benchmarks both with and without compression for proto 5. I'm genially surprised how significant improvement for compression path. Base is trunk. Candidate is changes in this PR. 
Benchmark setup:
- Ubuntu 24.04.3 LTS x86_64 CPU i7-13700H 32 Gb RAM
- Options:
warmup cycles: 10000
cycles: 1000000
rows: 10000000
goroutines: 64
protocol version: 5


### INSERT (base vs candidate, %)
percentile | base ms | candidate ms | delta
---|---:|---:|---:
p50 | 0.659 | 0.679 | +3.03%
p90 | 2.234 | 2.298 | +2.86%
p95 | 3.145 | 3.180 | +1.11%
p99 | 5.226 | 5.194 | -0.61%

### SELECT (base vs candidate, %)
percentile | base ms | candidate ms | delta
---|---:|---:|---:
p50 | 0.652 | 0.670 | +2.76%
p90 | 2.232 | 2.307 | +3.36%
p95 | 3.197 | 3.225 | +0.88%
p99 | 5.410 | 5.328 | -1.52%

### INSERT (compression) (base vs candidate, %)
percentile | base ms | candidate ms | delta
---|---:|---:|---:
p50 | 1.933 | 0.922 | -52.30%
p90 | 5.205 | 2.485 | -52.26%
p95 | 6.836 | 3.294 | -51.82%
p99 | 10.165 | 5.216 | -48.68%

### SELECT (compression) (base vs candidate, %)
percentile | base ms | candidate ms | delta
---|---:|---:|---:
p50 | 2.017 | 0.902 | -55.28%
p90 | 5.474 | 2.467 | -54.93%
p95 | 7.246 | 3.282 | -54.71%
p99 | 10.684 | 5.204 | -51.29%

worryg0d:
Added also v4 benchmarks:
## Non-compression

### INSERT
percentile | proto v4 trunk ms | proto v5 trunk ms | proto v5 cassgo100 ms | cassgo100 vs v5 trunk | cassgo100 vs v4 trunk
---|---:|---:|---:|---:|---:
p50 | 0.628 | 0.659 | 0.679 | +3.03% | +8.12%
p90 | 2.113 | 2.234 | 2.298 | +2.86% | +8.75%
p95 | 3.052 | 3.145 | 3.180 | +1.11% | +4.19%
p99 | 5.147 | 5.226 | 5.194 | -0.61% | +0.91%

### SELECT
percentile | proto v4 trunk ms | proto v5 trunk ms | proto v5 cassgo100 ms | cassgo100 vs v5 trunk | cassgo100 vs v4 trunk
---|---:|---:|---:|---:|---:
p50 | 0.623 | 0.652 | 0.670 | +2.76% | +7.54%
p90 | 2.099 | 2.232 | 2.307 | +3.36% | +9.91%
p95 | 3.092 | 3.197 | 3.225 | +0.88% | +4.30%
p99 | 5.308 | 5.410 | 5.328 | -1.52% | +0.38%

---

## Compression

### INSERT
percentile | proto v4 trunk ms | proto v5 trunk ms | proto v5 cassgo100 ms | cassgo100 vs v5 trunk | cassgo100 vs v4 trunk
---|---:|---:|---:|---:|---:
p50 | 1.887 | 1.933 | 0.922 | -52.30% | -51.14%
p90 | 4.904 | 5.205 | 2.485 | -52.26% | -49.33%
p95 | 6.490 | 6.836 | 3.294 | -51.82% | -49.24%
p99 | 9.624 | 10.165 | 5.216 | -48.68% | -45.80%

### SELECT
percentile | proto v4 trunk ms | proto v5 trunk ms | proto v5 cassgo100 ms | cassgo100 vs v5 trunk | cassgo100 vs v4 trunk
---|---:|---:|---:|---:|---:
p50 | 1.913 | 2.017 | 0.902 | -55.28% | -52.85%
p90 | 5.178 | 5.474 | 2.467 | -54.93% | -52.36%
p95 | 6.902 | 7.246 | 3.282 | -54.71% | -52.45%
p99 | 10.164 | 10.684 | 5.204 | -51.29% | -48.80%


Memory:
## Non-compression (proto v4 trunk vs proto v5 trunk vs proto v5 cassgo100)

metric | proto v4 trunk | proto v5 trunk | proto v5 cassgo100 | cassgo100 vs v5 trunk | cassgo100 vs v4 trunk
---|---:|---:|---:|---:|---:
avg MB | 8.710 | 10.205 | 9.314 | -8.73% | +6.94%
min MB | 5.488 | 6.093 | 5.714 | -6.22% | +4.11%
max MB | 13.054 | 19.078 | 15.520 | -18.65% | +18.89%

## Compression (proto v4 trunk vs proto v5 trunk vs proto v5 cassgo100)

metric | proto v4 trunk | proto v5 trunk | proto v5 cassgo100 | cassgo100 vs v5 trunk | cassgo100 vs v4 trunk
---|---:|---:|---:|---:|---:
avg MB | 18.889 | 20.235 | 10.761 | -46.82% | -43.05%
min MB | 7.369 | 7.738 | 6.033 | -22.03% | -18.14%
max MB | 78.335 | 57.707 | 19.479 | -66.24% | -75.13%

worryg0d:
Re-run benchmarks one more time

**vs columns:** percent change for **cassgo-100 proto5** vs each v2.1.1 baseline. Latency: positive = slower, negative = faster. Memory: positive = higher usage, negative = lower usage.

## No compression (LZ4 off)


#### Latency


### INSERT

| percentile | v2.1.1 proto4 | v2.1.1 proto5 | cassgo-100 proto5 | vs v2.1.1 proto4 | vs v2.1.1 proto5 |
|---|---:|---:|---:|---:|---:|
| p50 | 0.601ms | 0.680ms | 0.705ms | +17.3% | +3.7% |
| p90 | 2.115ms | 2.365ms | 2.482ms | +17.4% | +4.9% |
| p95 | 3.067ms | 3.382ms | 3.432ms | +11.9% | +1.5% |
| p99 | 5.297ms | 5.745ms | 5.617ms | +6.0% | -2.2% |

### SELECT

| percentile | v2.1.1 proto4 | v2.1.1 proto5 | cassgo-100 proto5 | vs v2.1.1 proto4 | vs v2.1.1 proto5 |
|---|---:|---:|---:|---:|---:|
| p50 | 0.597ms | 0.674ms | 0.694ms | +16.2% | +3.0% |
| p90 | 2.086ms | 2.365ms | 2.495ms | +19.6% | +5.5% |
| p95 | 3.078ms | 3.467ms | 3.480ms | +13.1% | +0.4% |
| p99 | 5.395ms | 5.943ms | 5.772ms | +7.0% | -2.9% |

#### Memory (Go heap)


### Heap allocation

| metric | v2.1.1 proto4 | v2.1.1 proto5 | cassgo-100 proto5 | vs v2.1.1 proto4 | vs v2.1.1 proto5 |
|---|---:|---:|---:|---:|---:|
| avg | 9.795 MB | 9.318 MB | 9.950 MB | +1.6% | +6.8% |
| min | 5.462 MB | 5.927 MB | 5.504 MB | +0.8% | -7.1% |
| max | 16.638 MB | 15.660 MB | 14.863 MB | -10.7% | -5.1% |

## With compression (LZ4 on)


#### Latency


### INSERT

| percentile | v2.1.1 proto4 | v2.1.1 proto5 | cassgo-100 proto5 | vs v2.1.1 proto4 | vs v2.1.1 proto5 |
|---|---:|---:|---:|---:|---:|
| p50 | 1.941ms | 2.032ms | 0.971ms | -50.0% | -52.2% |
| p90 | 5.063ms | 5.527ms | 2.749ms | -45.7% | -50.3% |
| p95 | 6.653ms | 7.269ms | 3.717ms | -44.1% | -48.9% |
| p99 | 9.891ms | 10.735ms | 5.974ms | -39.6% | -44.4% |

### SELECT

| percentile | v2.1.1 proto4 | v2.1.1 proto5 | cassgo-100 proto5 | vs v2.1.1 proto4 | vs v2.1.1 proto5 |
|---|---:|---:|---:|---:|---:|
| p50 | 1.968ms | 2.113ms | 0.950ms | -51.7% | -55.0% |
| p90 | 5.303ms | 5.841ms | 2.732ms | -48.5% | -53.2% |
| p95 | 7.055ms | 7.721ms | 3.687ms | -47.7% | -52.2% |
| p99 | 10.454ms | 11.273ms | 5.970ms | -42.9% | -47.0% |

#### Memory (Go heap)


### Heap allocation

| metric | v2.1.1 proto4 | v2.1.1 proto5 | cassgo-100 proto5 | vs v2.1.1 proto4 | vs v2.1.1 proto5 |
|---|---:|---:|---:|---:|---:|
| avg | 20.133 MB | 16.577 MB | 11.954 MB | -40.6% | -27.9% |
| min | 7.948 MB | 7.670 MB | 6.362 MB | -20.0% | -17.1% |
| max | 62.891 MB | 56.450 MB | 20.852 MB | -66.8% | -63.1% |

joao-r-reis:
Yeah I remember running some benchmarks to compare v2 with v1 and being a bit suprised that compression was so slow but I'm still surprised to see performance gains of that order now.

Do you have any idea why there seems to be a hit on v4 non compression scenarios?

worryg0d:
> Do you have any idea why there seems to be a hit on v4 non compression scenarios

I'm suspecting vectored IO in `writeCoalscer` which `segmentWriter` doesn't use. The segment writer allocates more memory when concatenating frames before passing them to the encoder, and encoder also allocates memory during payload encoding/decoding, but it is also true for the uncompressed path...

What actually changed is that the driver now does not compress each frame, but instead each segment, and each segment could have multiple frames

worryg0d:
I added memory usage for comparison to the last benchmark report

joao-r-reis:
is this still WIP or is it ready?

worryg0d:
I did a lot of testing of different approaches to reduce performance impact on the uncompressed path. Right now, the best thing is to change the segmentCodec.encode API to take [][]bytes instead of a temporarily allocated buffer that holds all frames concatenated -1 allocations. But I didn't test it yet through the benchmark as I had done before. I'm going to bench it, and if the impact is good enough, I'll update the PR

worryg0d:
Alright, I ended up adding reusable buffers to segmentWriter and segmentCodec.

What changed:
- segmentCodec.encode now takes [][]byte payload to avoid the use of a temporary intermediate buffer; also, it takes as an argument the destination buffer in which segments should be encoded so that it can be reused afterwards
- added scratch buffers to segmentCodec to reduce the amount of memory allocation on the compressed path.

Benchmark results are here: https://docs.google.com/spreadsheets/d/1XKU0Fkw5tPQjhZNtBG86s5n0UumZ1qJUdoViThzzdQ8/edit?usp=sharing

Results are slightly better compared to the previous diff:
 - p99 SELECT workload around -9% vs -11%
 - p99 INSERT workload around -9% vs -13%

segment_codec.go
```diff
@@ -0,0 +1,363 @@
...
+
+	return &segmentHeader{
+		payloadLength:   payloadLen,
+		isSelfContained: isSelfContained,
+	}, nil
+}
+
+// decodePayload reads and verifies the payload of a segment from the given reader.
+func (sc *segmentCodec) decodePayload(r io.Reader, header *segmentHeader) ([]byte, error) {
+	payload := make([]byte, header.payloadLength)
```
joao-r-reis, line 346: we could improve even further by reusing a buffer here too but I'm ok with merging what we already have since it's already a very good improvement

conn.go
```diff
@@ -804,103 +774,15 @@ func (c *Conn) releaseStream(call *callReq) {
...
-	}
-
-	// Restore timeout for subsequent segment reads in multi-segment frames
-	if readTimeout > 0 {
-		c.r.SetTimeout(readTimeout)
+func (c *Conn) maybeSwitchToSegments(host *DialedHost) {
+	if c.version >= protoVersion5 {
+		c.logger.Debug("Switching to segments for connection", NewLogFieldStringer("write_timeout", c.session.cfg.WriteTimeout), NewLogFieldStringer("write_coalesce_wait_time", c.session.cfg.WriteCoalesceWaitTime))
+		// Use segments writer which basically batches multiple frames into a single segment before flushing them to the connection.
+		segmentWriter := newSegmentWriter(host.Conn, c.writeTimeout, c.session.cfg.WriteCoalesceWaitTime, c.ctx.Done(), c.compressor)
```
joao-r-reis, line 781: I believe the proto v4 path checks if writecoalescetime is `> 0` before running the coalescer so we should probably guard against this here too

worryg0d: I actually can't see a value of having such feature as it goes against the batching nature of v5 frame formats. However, I'm unsure if it should be tied to the generic `WriteCoalesceWaitTime` option, probably it's better to introduce a separete `SegmentsWriteTimeout`

segment_codec.go
worryg0d: Those reused buffers actually makes the solution more complicated to me imo, so we probably should think about introducing something like buffer pools at some point

conn_test.go
```diff
@@ -1597,3 +1606,492 @@ func TestConnProcessAllFramesInSingleSegment(t *testing.T) {
...
+	sw := newSegmentWriter(client, time.Second*10, time.Microsecond*400, make(chan struct{}), nil)
+	go func() {
+		_, err := sw.writeContext(context.Background(), []byte("one"))
+		require.NoError(t, err)
+	}()
+
+	go func() {
+		_, err := sw.writeContext(context.Background(), []byte("two"))
+		require.NoError(t, err)
+	}()
```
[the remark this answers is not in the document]
worryg0d: I don't think this is problematic in this context - we already have time-dependent tests across the codebase

PR #1948, opened 2026-05-07
CASSGO-122 Fix panic when using a HostFilter and keyspace is not replicated to every DC

### LINKED ISSUE #1947: CASSGO-122 Panic when using a HostFilter and keyspace is not replicated to every DC
Hey, we see this panic

```
panic: token map different size to token ring: got 0 expected 120
goroutine 415 [running]:
github.com/apache/cassandra-gocql-driver/v2.(*networkTopology).replicaMap(0x25c6b9516810, 0x25c6b96629c0)
	/go/pkg/mod/github.com/apache/cassandra-gocql-driver/v2@v2.1.0/topology.go:391 +0x1274
github.com/apache/cassandra-gocql-driver/v2.(*tokenAwareHostPolicy).updateAllReplicas(0x25c6b8c53ac8?, 0x25c6b87a1de0, 0xd326c0?)
	/go/pkg/mod/github.com/apache/cassandra-gocql-driver/v2@v2.1.0/policies.go:559 +0x112
github.com/apache/cassandra-gocql-driver/v2.(*tokenAwareHostPolicy).schemaRefreshed(0x25c6b8b95800, 0x25c6b8782dc0)
	/go/pkg/mod/github.com/apache/cassandra-gocql-driver/v2@v2.1.0/policies.go:543 +0x90
github.com/apache/cassandra-gocql-driver/v2.refreshSchemas(0x25c6b8694f08)
	/go/pkg/mod/github.com/apache/cassandra-gocql-driver/v2@v2.1.0/metadata.go:865 +0xd71
github.com/apache/cassandra-gocql-driver/v2.NewSession.func1()
	/go/pkg/mod/github.com/apache/cassandra-gocql-driver/v2@v2.1.0/session.go:174 +0x17
github.com/apache/cassandra-gocql-driver/v2.(*refreshDebouncer).flusher(0x25c6b85b3590)
	/go/pkg/mod/github.com/apache/cassandra-gocql-driver/v2@v2.1.0/host_source.go:1009 +0x17d
created by github.com/apache/cassandra-gocql-driver/v2.newRefreshDebouncer in goroutine 414
	/go/pkg/mod/github.com/apache/cassandra-gocql-driver/v2@v2.1.0/host_source.go:945 +0x145
```

It happens when session is configured with host filter to connect only to hosts in a local dc:
```go
cluster.HostFilter = gocql.HostFilterFunc(func(host *gocql.HostInfo) bool {
	return host.DataCenter() == cfg.LocalDC
})
```

And  there is a keyspace that is not replicated to all DCs:
```
CREATE KEYSPACE ks1 WITH replication = {'class': 'NetworkTopologyStrategy', 'dc1': '3'}  AND durable_writes = true;
CREATE KEYSPACE ks2 WITH replication = {'class': 'NetworkTopologyStrategy', 'dc1': '3', 'dc2': '3'}  AND durable_writes = true;
```

UPD. It happens for clients with LocalDC: dc2

worryg0d:
Patch by Bohdan Siryk; reviewed by João Reis for CASSGO-122

### DIFF
--- a/topology.go
+++ b/topology.go
@@ -66,7 +66,7 @@ func (h tokenRingReplicas) replicasFor(t token) *hostTokens {
 }
 
 type placementStrategy interface {
-	replicaMap(tokenRing *tokenRing) tokenRingReplicas
+	replicaMap(tokenRing *tokenRing, logger StructuredLogger) tokenRingReplicas
 	replicationFactor(dc string) int
 	// strategyKey returns a unique identifier string for this strategy instance.
 	// Two strategy instances with identical configuration should return the same key.
@@ -162,7 +162,7 @@ func (s *simpleStrategy) replicationFactor(dc string) int {
 	return s.rf
 }
 
-func (s *simpleStrategy) replicaMap(tokenRing *tokenRing) tokenRingReplicas {
+func (s *simpleStrategy) replicaMap(tokenRing *tokenRing, _ StructuredLogger) tokenRingReplicas {
 	tokens := tokenRing.tokens
 	ring := make(tokenRingReplicas, len(tokens))
 
@@ -255,7 +255,7 @@ func (n *networkTopology) haveRF(replicaCounts map[string]int) bool {
 	return true
 }
 
-func (n *networkTopology) replicaMap(tokenRing *tokenRing) tokenRingReplicas {
+func (n *networkTopology) replicaMap(tokenRing *tokenRing, logger StructuredLogger) tokenRingReplicas {
 	dcRacks := make(map[string]map[string]struct{}, len(n.dcs))
 	// skipped hosts in a dc
 	skipped := make(map[string][]*HostInfo, len(n.dcs))
@@ -324,7 +324,11 @@ func (n *networkTopology) replicaMap(tokenRing *tokenRing) tokenRingReplicas {
 				continue
 			} else if replicasInDC[dc] >= rf {
 				if replicasInDC[dc] > rf {
-					panic(fmt.Sprintf("replica overflow. rf=%d have=%d in dc %q", rf, replicasInDC[dc], dc))
+					logger.Warning("Replica overflow. Returning empty map.",
+						NewLogFieldInt("rf", rf),
+						NewLogFieldInt("have", replicasInDC[dc]),
+						NewLogFieldString("dc", dc))
+					return tokenRingReplicas{}
 				}
 
 				// have enough replicas in this DC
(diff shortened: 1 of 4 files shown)


worryg0d:
Looks like CI jobs are being canceled due to the time limit. Some of the jobs took around 13 minutes to setup and start C* cluster - we should try to bump C* versions in C* so it won't download them from the slow archive

joao-r-reis:
> Looks like CI jobs are being canceled due to the time limit. Some of the jobs took around 13 minutes to setup and start C* cluster - we should try to bump C* versions in C* so it won't download them from the slow archive

Yeah I've noticed that before. I'm used to retrying the failed jobs and I thought that I would eventually find time to improve the workflow so we don't have as many parallel jobs for a simple matrix of 2 Go versions and 2 C* versions.

topology_test.go
```diff
@@ -224,3 +226,33 @@ func TestPlacementStrategy_NetworkStrategy(t *testing.T) {
...
+func TestPlacementStrategy_NetworkStrategy_DoNotPanicWhenNoReplicasInRing(t *testing.T) {
+	strat := newNetworkTopology(map[string]int{
+		"dc1": 3, // replicated only in dc1
+	})
+
+	// Hosts in ring only from dc2, so no replicas should be returned.
+	hosts := []*HostInfo{
+		{hostId: "dc2:rack1:0", dataCenter: "dc2", rack: "rack1"},
+		{hostId: "dc2:rack2:1", dataCenter: "dc2", rack: "rack2"},
+		{hostId: "dc2:rack3:2", dataCenter: "dc2", rack: "rack3"},
```
joao-r-reis, line 244: We should add a couple more test cases, at least one with hosts from both dcs. Also I'd generate some test UUIDs on any online uuid generator and use them as literals here, it's a bit confusing to see the host id strings like that even if they don't break anything in this specific test

topology_test.go
```diff
@@ -224,3 +226,33 @@ func TestPlacementStrategy_NetworkStrategy(t *testing.T) {
...
+			token: orderedToken(h.hostId),
+			host:  h,
+		})
+	}
+	sort.Sort(&tokenRing{tokens: tokens})
+
+	require.NotPanics(t, func() {
+		replicas := strat.replicaMap(&tokenRing{hosts: hosts, tokens: tokens})
+		require.Empty(t, replicas, "expected no replicas, got %d", len(replicas))
+	})
```
joao-r-reis: I would also assert that the replica map is correct, not just that it doesn't panic

worryg0d: It is expected to have an empty replica map if hosts are filtered out which this test does

topology_test.go
worryg0d: We already have tests for this actually:

https://github.com/apache/cassandra-gocql-driver/blob/c65c762b83eccdb6a6a083c123a5935c4302745c/topology_test.go#L78

> it's a bit confusing to see the host id strings like that even if they don't break anything in this specific test

I use this `hostId` to construct a token in tokenRing on my own. The value of such approach is that it makes easier for devs to understand in which order tokens will be ordered. I can update it if you insist, but it won't matter because ring is manually constructed:
```
tokens := make([]hostToken, 0, len(hosts))
	for _, h := range hosts {
		tokens = append(tokens, hostToken{
			token: orderedToken(h.hostId),
			host:  h,
		})
	}
	sort.Sort(&tokenRing{tokens: tokens})
```

joao-r-reis: In that case can you add a comment explaining that then? I think it's reasonable to keep it that way

topology_test.go
joao-r-reis: Yeah but there's nothing asserting that the replica map is empty no? I know the issue is the panic but adding that assert is still useful I think

worryg0d: There is require.Empty assertion:
```
require.Empty(t, replicas, "expected no replicas, got %d", len(replicas))
```

I'm not following you probably

topology.go
```diff
@@ -381,14 +381,20 @@ func (n *networkTopology) replicaMap(tokenRing *tokenRing) tokenRingReplicas {
...
+		// We should count only DCs that driver is aware of and have a replication factor > 0
+		if _, knownDc := dcRacks[dc]; knownDc && rf > 0 {
 			dcsWithReplicas++
 		}
 	}
 
 	if dcsWithReplicas == len(dcRacks) && len(replicaRing) != len(tokens) {
-		panic(fmt.Sprintf("token map different size to token ring: got %d expected %d", len(replicaRing), len(tokens)))
+		// Fallback to empty replica map if we have all the DCs with replicas but no replicas in the ring.
+		logger.Warning("Not enough replicas in the ring for the strategy key. Falling back to empty replica map.",
```
joao-r-reis: Hmm having this as a warning that will be repeatedly logged is not great because this is essentially a shortcoming of the driver (the driver should be able to compute replicas regardless of whether the user filters a host or not). 

I propose two things:
1 - create a ticket to implement a way for the driver to have access to all host metadata even if the application has a hostfilter in place
2 - until that ticket is addressed we keep this log message but we either make sure to log it only once (per session?) or we change the log level to debug after the first time it is logged

topology.go
joao-r-reis: I just realized there's other panics here in this method, we should remove them as well...

topology.go
worryg0d: It should not be really loudy if I understand things right - replica map computation only when there is a schema refresh or `AddHost` is called. 

Schema refresh happens when there is a node / topology events incoming or control conn reconnection schedules it so that should not be that loud.

topology_test.go
joao-r-reis: 🤦 sorry I completely missed the require empty inside the `NotPanics` call

I was expecting to read something like:

```
var replicas tokenRingReplicas
require.NotPanics(t, func() { replicas = strat.replicaMap(&tokenRing{hosts: hosts, tokens: tokens}) })
require.Empty(t, replicas, "expected no replicas, got %d", len(replicas))
```

so it completely threw me off 🤣 nvm then

topology.go
joao-r-reis: as we discussed offline it looks like this condition will only be hit in some kind of corrupted state or something so we don't have to worry about it being annoying

joao-r-reis (approved):

joao-r-reis:
Don't forget to squash and edit commit message then feel free to merge 👍

worryg0d:
Thanks for reviews

Issue #1919, opened 2025-11-27
Many "Pool connection error" with small Session.Timeout


moguchev:
Version of driver: **v2.0.0** (latest)
Go version: **1.24**
Cassandra version: **5.0.4**

If set global `Session` `Timeout` equal `1*time.Second` or less in logs we can see a lot of messages:

```
"message":"Pool connection error.","addr":"10.239.171.143:9042","err":"read tcp 10.221.55.21:53299->10.229.171.143:9042: i/o timeout"
```

The same thing happens with long timeouts, but less often.

I tracked the full error flow:
`Conn.serve` -> `Conn.recv` -> `Conn.processFrame` -> `readHeader` -> `connReader.Read` -> `Conn.serve` -> `Conn.closeWithError` ->`hostConnPool.HandleError `

So the driver keeps getting read/timeout errors and closes the connection with error and establishes a new one (and so on all the time).

The most surprising thing is that Cassandra's response time doesn't exceed 40ms at 99.9p. The go client isn't under any load.

Also I've found, the Heartbeat connection is established with a general timeout, not a Connect timeout (which isn't entirely accurate). However, debugging has shown that this isn't the reason the connection is closed.

Constant client-side reconnects put a strain on Cassandra.

joao-r-reis:
> The most surprising thing is that Cassandra's response time doesn't exceed 40ms at 99.9p.

Where are you measuring this? Are you sure that is the latency that gocql itself is seeing?

joao-r-reis:
Did you start observing this behavior only after you upgraded to gocql 2.0.0?

moguchev:
> Did you start observing this behavior only after you upgraded to gocql 2.0.0?

No, this can also be observed on `v2.0.0-rc1`.  I was sure that this PR https://github.com/apache/cassandra-gocql-driver/pull/1903 would fix this error but unfortantly  it didn't

> Where are you measuring this? Are you sure that is the latency that gocql itself is seeing?

Response time is determined by Cassandra cluster monitoring data. The latency seen by gocql itself is unknown. I'll have to check.

moguchev:
What worries me most is that the driver has a hertbeat mechanism that removes connections from the pool due to errors (including timeouts). This mechanism should be the only one to close the connection in such cases. But it turns out that somewhere at the Conn level, an I/O timeout occurs, causing it to be closed...

moguchev:
May be problem somewhere here:
```go
[...]

PR #1953, opened 2026-05-20
Separate init and runtime timeout budgets, add per-request timeouts

moguchev:
Related Issue: https://github.com/apache/cassandra-gocql-driver/issues/1919

Splits previously conflated timeout budgets so a low user-supplied Timeout no longer breaks session establishment, and adds dedicated client-side budgets for system queries and per-request overrides. Also fixes a potential controlConn reconnect deadlock.

* `ClusterConfig.ReadTimeout`: socket-level read deadline, defaults to `11s`, falls back to `ConnectTimeout` if unset. Decoupled from `Timeout`, which is now purely a per-request client-side budget.
* `ClusterConfig.Metadata.SystemRequestTimeout` (default `60s`): per-request budget for `system`/`system_schema` queries and other runtime control-conn frames. Applied via `Conn.querySystem`, `controlConn.query` and `controlConn.writeFrame`.
* `Query/Batch.WithRequestTimeout` / `SetRequestTimeout` / `GetRequestTimeout`: per-request override of `ClusterConfig.Timeout`, enforced via `callReq.timer` in `Conn.exec` independently of the socket deadline. 0 disables the client-side timer.
* `Conn.finalizeConnection`: switches an initialized connection from ConnectTimeout-bound init mode to operational `ReadTimeout` / `WriteTimeout` / `Metadata.SystemRequestTimeout`. Called by `hostConnPool.connect`, `controlConn.connect` and `controlConn.attemptReconnectToAnyOfHosts`. Until then, all socket reads/writes and internal init-phase requests (`STARTUP`/`AUTH`/`OPTIONS`/`system.local`/`USE keyspace`) are pinned to `ConnectTimeout`. Fixes "no connections were made when creating the session" when Timeout is low.
* `controlConn.HandleError` now launches reconnect in a goroutine to avoid a potential deadlock during a connection storm where attemptReconnect holds the controlConn mutex while doing DNS resolution / TCP dial.
* `connReader.timeout`, `deadlineContextWriter.timeout` and `writeCoalescer.timeout` are now atomics to allow `finalizeConnection` to update them concurrently with reader/writer goroutines.
* Regression tests: `TestNewConnectWithLowTimeout`, `TestInitSchemaQueryRespectsConnectTimeout`, `TestControlConnQueryRespectsSystemRequestTimeout`, `TestQueryRequestTimeout`, `TestControlConn_HandleError_LaunchesReconnectInGoroutine`.
* `README`: new "Connection Timeouts" section documenting the layers and picking guidance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,51 @@ Some tips for getting more performance from the driver:
 * Reading data from the network to unmarshal will incur a large amount of allocations, this can adversely affect the garbage collector, tune `GOGC`
 * Close iterators after use to recycle byte buffers
 
+Connection Timeouts
+-------------------
+
+The driver exposes several independent timeouts. They live on different layers
+(per-request budget vs. socket-level deadline) and serve different purposes — set
+them together, not "the smallest wins":
+
+| Field | Default | Layer | Purpose |
+|---|---|---|---|
+| `ClusterConfig.ConnectTimeout` | `11s` | Per-request, **init only** | Covers the entire connection setup: TCP dial, `STARTUP`/`AUTH`/`OPTIONS` handshake, initial `system.local` read, `USE keyspace`. Independent of `Timeout` — a low user `Timeout` will not break session creation. |
+| `ClusterConfig.Timeout` | `11s` | Per-request, **runtime** | Default client-side timeout for user `Query`/`Batch` (inherited at construction time, see below). Should be **greater** than server-side timeouts to avoid retry storms. |
+| `ClusterConfig.WriteTimeout` | `Timeout` | Socket write deadline | Operational write deadline after `finalizeConnection`. Defaults to `Timeout` if not set. Should be `<= Timeout`. |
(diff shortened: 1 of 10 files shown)


joao-r-reis:
We want to prioritize a fix for a quick patch release so we need a PR that focuses on removing the connection read deadline and therefore making `cfg.Timeout` apply to just the request timeout (`callReq` timer).

After that we can work on a more comprehensive improvement of the existing timeout configuration fields such as this PR.

cc @worryg0d

moguchev:
> We want to prioritize a fix for a quick patch release so we need a PR that focuses on removing the connection read deadline and therefore making `cfg.Timeout` apply to just the request timeout (`callReq` timer).
> 
> After that we can work on a more comprehensive improvement of the existing timeout configuration fields such as this PR.


Of course, this PR is more for inspiration for the future works with timeouts.

PR #1954, opened 2026-05-29
CASSGO-126 Fix system.peers_v2 fallback

joao-r-reis:
The fallback to system.peers if system.peers_v2 doesn't exist doesn't work. This patch addresses this.


worryg0d (approved):
+1

PR #1955, opened 2026-06-02
CASSGO-125 Fix repeated "Pool connection error" and reconnections with small Session.Timeout

### LINKED ISSUE #1919: Many "Pool connection error" with small Session.Timeout

joao-r-reis:
In GoCQL v1.x, the read deadline is set to 0 when reading a new response frame to prevent the connection from closing when it idles.

In v2.0.0 a regression was introduced causing the read deadline to be set to clusterCfg.Timeout for every read operation which leads to connections constantly erroring and reconnecting if they idle.

This patch fixes this regression by implementing the behavior of 1.x. A follow up ticket (CASSGO-127) will rework the timeout configuration so users can tune the read and write timeouts independently from the request timeout.

Patch by João Reis; reviewed by Bohdan Siryk for CASSGO-125

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -171,6 +171,7 @@ type Conn struct {
 	w contextWriter
 
 	writeTimeout   time.Duration
+	requestTimeout time.Duration // Request timeout, used for setting up request timers
 	cfg            *ConnConfig
 	frameObserver  FrameHeaderObserver
 	streamObserver StreamObserver
@@ -279,6 +280,7 @@ func (s *Session) dialWithoutObserver(ctx context.Context, host *HostInfo, cfg *
 		logger:         logger,
 		streamObserver: s.streamObserver,
 		writeTimeout:   writeTimeout,
+		requestTimeout: cfg.ConnectTimeout,
 	}
 
 	if err := c.init(ctx, dialedHost); err != nil {
@@ -312,6 +314,7 @@ func (c *Conn) init(ctx context.Context, dialedHost *DialedHost) error {
 	}
 
 	c.r.SetTimeout(c.cfg.Timeout)
+	c.requestTimeout = c.cfg.Timeout
 
 	// dont coalesce startup frames
 	if c.session.cfg.WriteCoalesceWaitTime > 0 && !c.cfg.disableCoalesce && !dialedHost.DisableCoalesce {
@@ -335,8 +338,8 @@ type startupCoordinator struct {
 
 func (s *startupCoordinator) setupConn(ctx context.Context) error {
 	var cancel context.CancelFunc
-	if s.conn.r.GetTimeout() > 0 {
-		ctx, cancel = context.WithTimeout(ctx, s.conn.r.GetTimeout())
+	if s.conn.requestTimeout > 0 {
+		ctx, cancel = context.WithTimeout(ctx, s.conn.requestTimeout)
 	} else {
 		ctx, cancel = context.WithCancel(ctx)
 	}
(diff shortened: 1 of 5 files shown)


integration_test.go
```diff
@@ -977,3 +977,54 @@ func TestSliceMapMapScanCollectionTypes(t *testing.T) {
...
+	time.Sleep(5 * time.Second)
+
+	// Get log output for analysis
+	logOutput := strings.ToLower(logger.String())
+
+	// Count successful connection messages - should be exactly NumConns * number of nodes
+	connectedCount := strings.Count(logOutput, "pool connected to node")
+	if connectedCount != *clusterSize*cluster.NumConns {
+		t.Fatalf("Expected exactly 2 'Pool connected to node' messages, got %d:\n%s",
+			connectedCount, logOutput)
```
worryg0d: nit: expected exactly *clusterSize *cluster.NumConns - there are 3 node clusters in our CI, so it would assert for 3 conns

conn.go
```diff
@@ -908,6 +929,8 @@ func (c *connReader) Read(p []byte) (n int, err error) {
 		var nn int
 		if c.timeout > 0 {
 			c.conn.SetReadDeadline(time.Now().Add(c.timeout))
+		} else if c.timeout == 0 {
+			c.conn.SetReadDeadline(time.Time{})
 		}
```
worryg0d, line 934: It fails `TestSteam0` because the underlying conn is nil

conn_test.go
```diff
@@ -306,12 +308,18 @@ func TestTimeout(t *testing.T) {
 		}
 	}()
 
-	if err := db.Query("kill").WithContext(ctx).Exec(); err == nil {
+	now := time.Now()
+	err = db.Query("timeout").WithContext(ctx).Exec()
```
worryg0d: At some point, we would have to get rid of WithContext usage as it is deprected

joao-r-reis: Fixed it for this test since I modified it, I won't fix it on other places since it would make this commit a bit confusing, fixing it on other places should probably be a separate ticket and commit

conn.go
joao-r-reis: Fixed it by creating a conn object using `net.Pipe`

integration_test.go
joao-r-reis: fixed

integration_test.go
```diff
@@ -977,3 +977,54 @@ func TestSliceMapMapScanCollectionTypes(t *testing.T) {
...
+	// and "Pool connection error" messages will be logged repeatedly
+	time.Sleep(5 * time.Second)
+
+	// Get log output for analysis
+	logOutput := strings.ToLower(logger.String())
+
+	// Count successful connection messages - should be exactly NumConns * number of nodes
+	connectedCount := strings.Count(logOutput, "pool connected to node")
+	if connectedCount != *clusterSize*cluster.NumConns {
+		t.Fatalf("Expected exactly &d 'Pool connected to node' messages, got %d:\n%s",
```
worryg0d: Mistyped `&d` instead of `%`. Interesting that the linter didn't spot it before

worryg0d (approved):
+1

I have run integration tests locally to verify if everything is good there.

Please fix the typo, and feel free to merge.

integration_test.go
joao-r-reis: oops

Issue #956, opened 2017-08-11
On using DCAwareRoundRobinPolicy, read also searching on remote hosts


nupurbansal8:
I am using DCAwareRoundRobinPolicy for my cassandra connection. But the time taken for querying by my driver is comparatively larger than when i am using the same connection parameters in php.

The datastax driver in php uses the function to the the same:
withDatacenterAwareRoundRobinLoadBalancingPolicy( string $localDatacenter, int $hostPerRemoteDatacenter, bool $useRemoteDatacenterForLocalConsistencies ) 

When the last parameter is set as false, only local hosts are used for querying.
But this parameter setting is not available in gocql driver. And I am guessing that this may be causing the issue I am facing. Correct me if I am wrong.

**Please help in this regard what can be done.**
I was not facing this issue when I had only 1 node in cluster.

Zariel:
Can you post the output of `nodetool status`, the gocql cluster config and what version of cassandra are you using.

There is nothing in this issue which is directly actionable, comparing the driver to the PHP one is not helpful, they are not related.

robusto:
gocql handles this differently from the other drivers. See [policies.DCAwareRR](https://github.com/gocql/gocql/blob/master/policies.go#L589) compared to:

[Python Example](https://github.com/datastax/python-driver/blob/fad8764b365e76f2c2955500f2956c4396eec9b4/cassandra/policies.py#L257)
[CPP Example](https://github.com/datastax/cpp-driver/blob/7ebeac1532609e29c1f4ad8563a1a634fb151e6d/src/dc_aware_policy.cpp) (PHP driver wraps the C/C++ driver)

We could add something similar to the DCAwareRR policy. However, in the near term, I'm wondering if the desired "local or nothing" results that @nupurbansal8 wants could be achieved with a HostFilter.

```
cluster := gocql.NewConfig(hosts...)
dc := "datacenter1"
cluster.PoolConfig.HostSelectionPolicy = gocql.DCAwareRoundRobinPolicy(dc)
cluster.HostFilter = gocql.DataCentreHostFilter(dc)
```

Otherwise, as it is today, it seems like Remote nodes will eventually be selected "round-robin" (as the name implies) even if Local nodes are available.

nupurbansal8:
Hi,

**For nodetool status :** 
Datacenter: DC1
===============
Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address          Load       Tokens  Owns    Host ID                               Rack
[...]

PR #1956, opened 2026-06-08
add DCLocalRoundRobin host selection policy

isopov:
As requested in #956 we need an option to not fallback to remote datacenter in DCAware host selection policy.
Unfortunately we first try to connect to remote datacenter host and receive local hosts there. So the way with HostFilter does not work for us. It seems that such host selection policy should work for us...

### DIFF
--- a/policies.go
+++ b/policies.go
@@ -840,6 +840,52 @@ func (d *dcAwareRR) RemoveHost(host *HostInfo) {
 func (d *dcAwareRR) HostUp(host *HostInfo)   { d.AddHost(host) }
 func (d *dcAwareRR) HostDown(host *HostInfo) { d.RemoveHost(host) }
 
+func (d *dcAwareRR) Pick(q ExecutableStatement) NextHost {
+	nextStartOffset := atomic.AddUint64(&d.lastUsedHostIdx, 1)
+	return roundRobbin(int(nextStartOffset), d.localHosts.get(), d.remoteHosts.get())
+}
+
+type dcLocalRR struct {
+	local           string
+	localHosts      cowHostList
+	lastUsedHostIdx uint64
+}
+
+// DCLocalRoundRobinPolicy is a host selection policies which will allow and
+// return hosts which are in the local datacenter and will ignore hosts
+// from other datacenters
+func DCLocalRoundRobinPolicy(localDC string) HostSelectionPolicy {
+	return &dcLocalRR{local: localDC}
+}
+
+func (d *dcLocalRR) Init(*Session)                       {}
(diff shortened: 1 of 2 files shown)


joao-r-reis:
Adding a whole new type for this seems unnecessary, we just need a new boolean field on the existing policy like `localOnly`. 

Also, there's nothing in the integration test suite that verifies that the driver will be able to successfully connect and initialize with the scenario you're describing. Unfortunately this would require a cluster with two datacenters and the cluster we create for CI only has one datacenter. Maybe we could change the CI cluster to have two datacenters though.

isopov:
In my app I finished not with fork of this driver, but with this custom selection policy:
```
func wrapForceDC(internal gocql.HostSelectionPolicy, dcLocal string) gocql.HostSelectionPolicy {
	return &forceDCPolicy{
		dc:       dcLocal,
		internal: internal,
	}
}

type forceDCPolicy struct {
	dc       string
	internal gocql.HostSelectionPolicy
}

func (f *forceDCPolicy) AddHost(host *gocql.HostInfo) {
	if f.IsLocal(host) {
		f.internal.AddHost(host)
	}
}

func (f *forceDCPolicy) RemoveHost(host *gocql.HostInfo) {
	if f.IsLocal(host) {
		f.internal.RemoveHost(host)
	}
}

func (f *forceDCPolicy) HostUp(host *gocql.HostInfo) {
	if f.IsLocal(host) {
		f.internal.HostUp(host)
	}
}

func (f *forceDCPolicy) HostDown(host *gocql.HostInfo) {
	if f.IsLocal(host) {
		f.internal.HostDown(host)
	}
}

func (f *forceDCPolicy) Init(s *gocql.Session) {
	f.internal.Init(s)
}
func (f *forceDCPolicy) KeyspaceChanged(e gocql.KeyspaceUpdateEvent) {
	f.internal.KeyspaceChanged(e)
}
func (f *forceDCPolicy) SetPartitioner(p string) {
	f.internal.SetPartitioner(p)
}

func (f *forceDCPolicy) IsLocal(host *gocql.HostInfo) bool {
	return host.DataCenter() == f.dc
}

func (f *forceDCPolicy) Pick(query gocql.ExecutableQuery) gocql.NextHost {
	return f.internal.Pick(query)
}

```
And use it like this:
```
wrapForceDC(gocql.TokenAwareHostPolicy(gocql.RoundRobinHostPolicy(), gocql.ShuffleReplicas()), dcLocal)
```

isopov:
>Also, there's nothing in the integration test suite that verifies that the driver will be able to successfully connect and initialize with the scenario you're describing

We already have code dealing with several datacenters and fallback to not preferred one and no tests for it

PR #1960, opened 2026-06-22
fix protocol negotiation error handling

jameshartig:
In CASSGO-97 and #1920 the handling was changed to better support non-zero stream id error responses. But because of hardcoding of version 5 in the test harness there was a regression where older protocol responses were not supported anymore.

By adding an Unwrap method to ErrProtocol now the existing checking in checkProtocolRelatedError will correctly catch the protocolError.


worryg0d (approved):
+1

Thanks for catching this

PR #1961, opened 2026-06-30
CASSGO-132 Prevent syntetic CREATED events on HostSelectionPolicy.KeyspaceChanged

worryg0d:
This patch fixes a regression introduced in CASSGO-104 that caused the driver to fire synthetic CREATED events on HostSelectionPolicy.KeyspaceChanged for each existing keyspace during session initialization.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -405,9 +405,26 @@ func (cfg *ClusterConfig) filterHost(host *HostInfo) bool {
 	return !(cfg.HostFilter == nil || cfg.HostFilter.Accept(host))
 }
 
+func (cfg *ClusterConfig) validate() error {
+	logger := cfg.newLogger()
+
+	if cfg.Metadata.CacheMode == Disabled {
+		mrp, ok := cfg.PoolConfig.HostSelectionPolicy.(MetadataRequiredPolicy)
+		if !ok {
+			logger.Warning(metadataCacheDisabledWarningMsg)
+		} else if mrp.MetadataRequired() {
+			return ErrMetadataCacheRequired
+		}
+	}
+
+	return nil
+}
+
 // MetadataConfig configures driver's internal metadata caching and event listening.
 type MetadataConfig struct {
 	// CacheMode controls how the driver reads and caches schema metadata from Cassandra system tables.
+	//
+	// It is required to be enabled for [TokenAwareHostPolicy] and custom host selection policies.
 	//
 	// Also, it affects the behavior of schema change listeners.
 	//
@@ -465,4 +482,10 @@ var (
 	ErrNoConnectionsStarted = errors.New("no connections were made when creating the session")
 	// Deprecated: Never used or returned by the driver.
 	ErrHostQueryFailed = errors.New("unable to populate Hosts")
+)
+
+const (
+	metadataCacheDisabledWarningMsg = "Metadata cache is disabled and the host selection policy does not implement MetadataRequiredPolicy interface. " +
+		"This might result in unexpected behavior if your custom policy requires keyspace metadata. " +
+		"To avoid this warning, please implement the MetadataRequiredPolicy interface."
 )
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -30,6 +30,7 @@ package gocql
 import (
 	"net"
 	"reflect"
+	"strings"
 	"testing"
 	"time"
 )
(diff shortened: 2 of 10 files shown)


worryg0d:
The implementation in this patch fixes the regression by guarding from synthetic CREATED events the same way as schema keyspace listeners are guarded.

Also, it adds a requirement that the internal metadata cache should be enabled at least at the KeyspaceOnly level for Token Aware and Custom host policies. But this is technically a breaking change for users who already have custom policies and disabled metadata, so it's probably better to just log a warning about it.

We can introduce an optional interface that would indicate if a policy requires a metadata cache, and if a custom policy implements the interface, omit the warning:
```
// MetadataRequiredPolicy is an interface that can be implemented by host selection policies to indicate if they require the metadata cache to be enabled.
type MetadataRequiredPolicy interface {
	RequiresMetadata() bool
}
```

policies_test.go
```diff
@@ -1099,3 +1101,14 @@ func TestHostPolicy_TokenAware_TopologyChangeUpdatesAllKeyspaces(t *testing.T) {
...
+
+func Test_hostSelectionPolicyRequiresMetadataCache(t *testing.T) {
+	// Doesn't require metadata cache
+	assert.False(t, hostSelectionPolicyRequiresMetadata(RoundRobinHostPolicy()))
+	assert.False(t, hostSelectionPolicyRequiresMetadata(DCAwareRoundRobinPolicy("dc1")))
+	assert.False(t, hostSelectionPolicyRequiresMetadata(RackAwareRoundRobinPolicy("dc1", "rack1")))
+	assert.False(t, hostSelectionPolicyRequiresMetadata(SingleHostReadyPolicy(RoundRobinHostPolicy())))
+
+	// Requires metadata cache
+	assert.True(t, hostSelectionPolicyRequiresMetadata(TokenAwareHostPolicy(RoundRobinHostPolicy())))
```
joao-r-reis, line 1113: Should we have something here that ensures that a new implementation passes the assertion? Otherwise we might forget to add a new implementation to the `hostSelectionPolicyRequiresMetadata` function.

Gemini suggests `go/parser` and `go/types` to parse the source and find every struct that implements the interface. It looks to be quite a bit of code to accomplish this but it might be worth it, wdyt?

worryg0d: Yeah it's a valid point. But I'm unsure how to achieve this. Do you propose to implement kind of a code gen tool that would lookup for interface implementations and add them to the hostSelectionPolicyRequiresMetadata()?

joao-r-reis: Provided a example snippet offline

worryg0d: AST-based approach sounds rather interesting, but it comes with some limitations:
- go compiler version dependant
- we can't test policies that are not in gocql package, for example hostpool package cannot be tested due to circular dependency. Workaround is to create a separate package for testing purpose, but then we have expose `hostSelectionPolicyRequiresMetadata()` method or put it in `internal/policies` (whatever name) so it won't be part of public api.

We decided to leave it out of scope of this PR

joao-r-reis (approved):
Don't forget the changelog 👍

worryg0d:
So, we decided to introduce an optional interface `MetadataRequiredPolicy` that basically allows to bypass the check. If it is implemented on a host policy it will be queried and if the policy requires keyspace metadata while metadata cache mode is disabled, it will fail during session creation

cluster.go
```diff
@@ -466,3 +483,9 @@ var (
 	// Deprecated: Never used or returned by the driver.
 	ErrHostQueryFailed = errors.New("unable to populate Hosts")
 )
+
+const (
+	metadataCacheDisabledWarningMsg = "Metadata cache is disabled and the host selection policy does not implement MetadataRequiredPolicy interface. " +
+		"This might result in unexpected behavior in case if your custom policy requires keyspace metadata. " +
```
joao-r-reis: NIT grammar: "in case" or "if", not "in case if"

cluster_test.go
```diff
@@ -83,3 +84,70 @@ func TestClusterConfig_translateAddressAndPort_Success(t *testing.T) {
...
+
+	testCases := []testCase{
+		{
+			name:                "Metadata cache is disabled and policy does not require metadata",
+			cacheMode:           Disabled,
+			hostSelectionPolicy: RoundRobinHostPolicy(),
+			expectedError:       nil,
+		},
+		{
+			name:                "Metadata cache is disbled and policy requires metadata",
```
joao-r-reis: NIT: "disabled"

policies.go
```diff
@@ -340,6 +348,15 @@ type schemaRefreshNotifier interface {
 	schemaRefreshed(meta *schemaMeta)
 }
 
+// MetadataRequiredPolicy is an optional interface that can be implemented by HostSelectionPolicy.
+// If implemented, the policy will be queried to determine if metadata is required for the policy to work.
+// If metadata is required and MetadataCacheMode is not set to at least KeyspaceOnly, it will result in an error during session creation.
+//
+// All built-in policies implement this interface.
+type MetadataRequiredPolicy interface {
```
joao-r-reis, line 356: What about "MetadataAwarePolicy" or "MetadataDependentPolicy"? Just a suggestion though

joao-r-reis (approved):
Looking good, my comments are all NITs

policies.go
worryg0d: I was thinking about better name, I like the first option, let me update it

worryg0d: Should I also change the method name to MetadataAware()?

worryg0d: However, thinking about this more "Aware" is like considering if it exists? Probably not the best semantic...

